PR #1, opened 2017-08-07
root is a FQ path so relative to __FILE__ is unneeded

jrafanie:
First!

### DIFF
--- a/lib/manageiq/api/engine.rb
+++ b/lib/manageiq/api/engine.rb
@@ -3,9 +3,9 @@ module ManageIQ
   module Api
     class Engine < ::Rails::Engine
       isolate_namespace ManageIQ::Api
-      config.autoload_paths << File.expand_path(root.join('app', 'controllers', 'api'), __FILE__)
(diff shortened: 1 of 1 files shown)


lib/manageiq/api/engine.rb
```diff
@@ -3,9 +3,9 @@ module ManageIQ
   module Api
     class Engine < ::Rails::Engine
       isolate_namespace ManageIQ::Api
-      config.autoload_paths << File.expand_path(root.join('app', 'controllers', 'api'), __FILE__)
-      config.autoload_paths << File.expand_path(root.join('lib', 'services'), __FILE__)
-      config.autoload_paths << File.expand_path(root.join('lib'), __FILE__)
+      config.autoload_paths << File.expand_path(root.join('app', 'controllers', 'api'))
```
Fryguy: You can call expand_path directly on a pathname: `root.join(...).expand_path`

jrafanie: lol, yup

abellotti:
Nice @jrafanie   however, Mongo here only pawn in game of life  (cannot merge yet :) )

PR #3, opened 2017-08-08
 OPTIONS for /api/providers

enoodle:
This is a migration of https://github.com/ManageIQ/manageiq/pull/15660

### DIFF
--- a/app/controllers/api/providers_controller.rb
+++ b/app/controllers/api/providers_controller.rb
@@ -86,6 +86,13 @@ module Api
       end
     end
 
+    def options
+      providers_options = ManageIQ::Providers::BaseManager.leaf_subclasses.inject({}) do |po, ems|
+        po.merge(ems.ems_type => ems.options_description)
+      end
+      render_options(:providers, {"provider_settings" => providers_options})
+    end
+
     private
 
     def format_provider_custom_attributes(attribute)
--- a/spec/requests/providers_spec.rb
+++ b/spec/requests/providers_spec.rb
@@ -1134,4 +1134,14 @@ describe "Providers API" do
       end
     end
   end
+
+  context "OPTIONS /api/providers" do
+    it "returns options for all providers when no query" do
+      run_options(providers_url)
+      expect(response.parsed_body["data"]["provider_settings"].keys.count).to eq(
+        ManageIQ::Providers::BaseManager.leaf_subclasses.count
+      )
+      expect(response.parsed_body["data"]["provider_settings"]["kubernetes"]["proxy_settings"]["settings"]["http_proxy"]["label"]).to eq(N_('HTTP Proxy'))
+    end
+  end
 end


enoodle:
@jntullo @abellotti PTAL

app/controllers/api/providers_controller.rb
```diff
@@ -85,6 +85,15 @@ def import_vm_resource(type, id = nil, data = {})
       end
     end
 
+    def options
+      req_class_name = @req.to_hash[:fullpath]
+                           .split('?').last.split('&')                        # '?' for query, '&' connects parameters
+                           .select { |param| param.starts_with?("ems_type") } # 'parameters with "ems_type" name
+                           .first.split('=').last                             # choosing only the first occuracne.
+      ems_class = ExtManagementSystem.descendants.select { |ems| ems.ems_type == req_class_name }.first
+      render :json => ems_class.respond_to?(:provider_settings) ? ems_class.provider_settings : {}
```
jntullo: I think you will want to use `render_options(:providers, <your provider settings hash)` here. This will return all of the default options that are passed back from `OPTIONS /providers`, and append your provider settings hash to it.

enoodle: @jntullo I changed to `render_options`

app/controllers/api/providers_controller.rb
```diff
@@ -85,6 +85,15 @@ def import_vm_resource(type, id = nil, data = {})
       end
     end
 
+    def options
+      req_class_name = @req.to_hash[:fullpath]
+                           .split('?').last.split('&')                        # '?' for query, '&' connects parameters
+                           .select { |param| param.starts_with?("ems_type") } # 'parameters with "ems_type" name
+                           .first.split('=').last                             # choosing only the first occuracne.
+      ems_class = ExtManagementSystem.descendants.select { |ems| ems.ems_type == req_class_name }.first
```
jntullo: missed this earlier, but you should be able to get the ems type from just params[:ems_type]. Then you can actually use `ExtManagementSystem.descendant_get(params[:ems_type])` here. That will return the descendant. An example of its usage can be found https://github.com/ManageIQ/manageiq-api/blob/master/lib/services/api/authentication_service.rb#L4

enoodle: @jntullo I didn't know of `descendant_get` before, but I am not sure it will help here because it will require sending the full class name (ie `ManageIQ::Providers::Openshift::ContainerManager`) instead of the `ems_type` reported by the providers. I will change to use `params[:ems_type]`

jntullo: @enoodle ah, gotcha! yeah, then `descendant_get` won't work.

this is just nit but you could do `ExtManagementSystem.descendants.find { |ems| ems.ems_type == params[:ems_type] }`

spec/requests/api/providers_spec.rb
```diff
@@ -1116,4 +1116,14 @@ def gen_import_request
       end
     end
   end
+
+  context "#options" do
+    it "returns options for queried provider class" do
+      api_basic_authorize
+      ExtManagementSystem.descendants.select { |ems| ems.respond_to?(:provider_settings) }.each do |ems_type|
```
jntullo: one last thing 😄  you can simplify the spec by using `expect_options_results`, example here: https://github.com/ManageIQ/manageiq-api/blob/master/spec/requests/api/authentications_spec.rb#L480

spec/requests/api/providers_spec.rb
```diff
@@ -1116,4 +1116,14 @@ def gen_import_request
       end
     end
   end
+
+  context "#options" do
+    it "returns options for queried provider class" do
+      api_basic_authorize
```
imtayadeway: Doing an OPTIONS request shouldn't require authorization

app/controllers/api/providers_controller.rb
```diff
@@ -85,6 +85,11 @@ def import_vm_resource(type, id = nil, data = {})
       end
     end
 
+    def options
+      ems_class = ExtManagementSystem.descendants.find { |ems| ems.ems_type == params[:ems_type] }
+      render_options(:providers, ems_class.respond_to?(:provider_settings) ? ems_class.provider_settings : {})
```
imtayadeway: It looks like you don't render anything when you don't provide the `ems_type` param. Did you you consider rendering all the EMSs and their provider settings and leaving it to the client to figure out?

spec/requests/api/providers_spec.rb
imtayadeway: Your test logic here mirrors that of the implementation, so I'm wondering what value this test is providing? IMO, using literals to represent expected values is much clearer and will provide useful feedback in the case of breakages or false assumptions about the implementations.

spec/requests/api/providers_spec.rb
enoodle: @imtayadeway I am not sure how to remove this requirement. I also see this in many other examples: cluster, hosts, authentications. How do you suggest solving this?

spec/requests/api/providers_spec.rb
enoodle: @imtayadeway doing so will force me to use specific providers names as this is only implemented for "ManageIQ::Providers::Kubernetes::ContainerManager" currently. I am afraid of breaking tests when things changes in another repo (ie someone will change the way options are saved there and tests here will be red).  I also agree that the test logic is not super helpful but it helped me.
I will try to add a test with this provider and options that I don't think will change anytime soon.

enoodle:
@jntullo I changed to use `expect_options_results` and `descendants.find`.

spec/requests/api/providers_spec.rb
imtayadeway: I didn't realize this was still an issue.....I created #8 to resolve this

app/controllers/api/providers_controller.rb
```diff
@@ -85,6 +85,21 @@ def import_vm_resource(type, id = nil, data = {})
       end
     end
 
+    def options
+      if params[:ems_type].blank?
```
imtayadeway: I still have concerns over this param and its discoverability. At present AFAICT it is not discoverable, meaning it will have to be documented somewhere, and I'm not totally comfortable with adding complexity to OPTIONS calls. Is it needed?

app/controllers/api/providers_controller.rb
```diff
@@ -85,6 +85,21 @@ def import_vm_resource(type, id = nil, data = {})
...
+        all_options = Hash[
+          ExtManagementSystem.descendants.collect do |ems|
+            [ems.ems_type, ems.respond_to?(:provider_settings) ? ems.provider_settings : {}]
```
imtayadeway: Do provider settings come from the `Settings` object? Could this information be got at through the settings API?

spec/requests/providers_spec.rb
```diff
@@ -1116,4 +1116,29 @@ def gen_import_request
       end
     end
   end
+
+  context "OPTIONS /api/providers" do
+    it "returns options for all providers when no query" do
+      api_basic_authorize
```
imtayadeway: Now that #8 has been merged you can remove this and subsequent calls to `api_basic_authorize`

app/controllers/api/providers_controller.rb
enoodle: No, they come from "per-provider" options, for example: https://github.com/ManageIQ/manageiq-providers-kubernetes/pull/45

app/controllers/api/providers_controller.rb
enoodle: It is not a must, but I though it will be nice. I don't mind documenting it (where?) but I can also remove this.

app/controllers/api/providers_controller.rb
imtayadeway: Is there a reason that's not in [here](https://github.com/ManageIQ/manageiq-providers-kubernetes/blob/master/config/settings.yml#L3) ?

enoodle: Yes, those options themselves are per provider instance.
What I am serving here are only their descriptions but belong there as they are static and not configurable.

imtayadeway: I'm not sure I follow? I'd expect per-instance settings to be revealed via GETs to the collection member URI. We augment OPTIONS responses only as a limited way to reveal metadata - i.e. data about the resource itself in the abstract.

app/controllers/api/providers_controller.rb
imtayadeway: See also https://github.com/ManageIQ/manageiq-api/pull/3#discussion_r132243534 for my explanation there - I think I'd prefer not including this unless there is a distinct need for it. If that is the case, I suspect OPTIONS might not be the right solution

app/controllers/api/providers_controller.rb
enoodle: Right. But here we are not serving the values of those options but the answer to the question "what per-provider options are for this provider type?". This is the static description of those options. please look at https://github.com/ManageIQ/manageiq-providers-kubernetes/pull/45/files#diff-47711b761083b6ce40dacb7d3ee68f8b for example

app/controllers/api/providers_controller.rb
```diff
@@ -85,6 +85,16 @@ def import_vm_resource(type, id = nil, data = {})
       end
     end
 
+    def options
+      all_options = Hash[
+        ExtManagementSystem.descendants.collect do |ems|
```
imtayadeway: Prefer `inject` or `each_with_object` here

app/controllers/api/providers_controller.rb
```diff
@@ -85,6 +85,16 @@ def import_vm_resource(type, id = nil, data = {})
...
+          [ems.ems_type, ems.respond_to?(:provider_settings) ? ems.provider_settings : {}]
+        end
+      ]
+      all_options.delete(nil)
```
imtayadeway: Which EMSs have an `ems_type` of `nil`?

enoodle:
```
ExtManagementSystem.descendants.select { |ems| ems.ems_type == nil }.map { |ems| ems.name }
=> ["ManageIQ::Providers::BaseManager",
 "ManageIQ::Providers::DatawarehouseManager",
 "ManageIQ::Providers::MiddlewareManager",
 "ManageIQ::Providers::NetworkManager",
 "ManageIQ::Providers::ProvisioningManager",
 "ManageIQ::Providers::AutomationManager",
 "ManageIQ::Providers::CloudManager",
 "ManageIQ::Providers::ContainerManager",
 "ManageIQ::Providers::ConfigurationManager",
 "ManageIQ::Providers::StorageManager",
 "ManageIQ::Providers::InfraManager",
 "ManageIQ::Providers::EmbeddedAutomationManager",
 "ManageIQ::Providers::ExternalAutomationManager"]
```

imtayadeway: Will `.leaf_subclasses` give you what you need?

imtayadeway:
@enoodle just one more suggestion in https://github.com/ManageIQ/manageiq-api/pull/3#discussion_r132301114. Thanks!

enoodle:
@imtayadeway PTAL

app/controllers/api/providers_controller.rb
```diff
@@ -85,6 +85,13 @@ def import_vm_resource(type, id = nil, data = {})
       end
     end
 
+    def options
+      all_options = ExtManagementSystem.leaf_subclasses.inject({}) do |ao, ems|
+        ao.merge(ems.ems_type => ems.respond_to?(:provider_settings) ? ems.provider_settings : {})
```
imtayadeway: It's unfortunate that not all EMSs respond to this message. Though I think this if fine for now - do you think you could follow this up by trying to push this default down into the base class?

spec/requests/providers_spec.rb
```diff
@@ -1116,4 +1116,14 @@ def gen_import_request
       end
     end
   end
+
+  context "OPTIONS /api/providers" do
+    it "returns options for all providers when no query" do
+      run_options(providers_url)
+      expect(response.parsed_body["data"].keys.count).to eq(
+        ExtManagementSystem.leaf_subclasses.select { |ems| !ems.ems_type.blank? }.count
```
imtayadeway: Are there any leaf subclasses for which this is not true?

enoodle: No, but if a new one will be created I think it will break this test. I will remove the `select`.

app/controllers/api/providers_controller.rb
enoodle: Yes. I changed this to address `leaf_subclasses` of `ManageIQ::Providers::BaseManager` instead of `ExtManagementSystem` and it will be easy to add 
```
def self.provider_settings()
  {}
end
```
to this class.

enoodle: https://github.com/ManageIQ/manageiq/pull/15799

enoodle:
@imtayadeway PTAL

enoodle:
@miq-bot add_label wip
We will change the `provider_settings` function name to `options_description`

enoodle:
@imtayadeway @abellotti with https://github.com/ManageIQ/manageiq/pull/15799 being approved, can we merge this one?

app/controllers/api/providers_controller.rb
```diff
@@ -86,6 +86,13 @@ def import_vm_resource(type, id = nil, data = {})
       end
     end
 
+    def options
+      all_options = ManageIQ::Providers::BaseManager.leaf_subclasses.inject({}) do |ao, ems|
+        ao.merge(ems.ems_type => ems.respond_to?(:options_description) ? ems.options_description : {})
```
imtayadeway: @enoodle if https://github.com/ManageIQ/manageiq/pull/15799 is merged, can you remove this `respond_to?`?

enoodle: Changed, tests here will only be green after https://github.com/ManageIQ/manageiq/pull/15799 is merged

abellotti:
@enoodle can you attach sample output you're getting with OPTIONS /api/providers ? just the data part of the output is sufficient. Thanks.

enoodle:
Output example from current masters:

[api_provider_to_delete.txt](https://github.com/ManageIQ/manageiq-api/files/1249341/api_provider_to_delete.txt)

abellotti:
I wonder if there might be some common options we'd want to provide in the future for all providers, in which case we may want to nest the output one level down,

i.e.

```
"data" => {
   "option_descriptions" => {
       "ansible_tower_automation" => ...,
       "azure" => ...,
       ...
   }
}
```

enoodle:
@abellotti Then you suggest to have the common options description under "provider_settings"? It will make for a more compact output, but will be more awkward to consume so I don't like it too much.

abellotti:
@enoodle true, it is an extra level to fetch, but would keep the return signature future-proof.

enoodle:
@abellotti :+1: I added the "provider_settings" step.

enoodle:
Travis for this seems to be broken as it is testing the wrong commit. I will close this one to open a new PR.

PR #5, opened 2017-08-08
Move request specs into the requests folder

imtayadeway:
Now that things are separated there's no need to house these files in
a separate folder. Doing this also allows us to simplify the spec
helper a bit too.

@miq-bot assign @abellotti


abellotti:
spec/spec_helper.rb still references spec/equests/api path.

imtayadeway:
@abellotti fixed!

abellotti:
Thanks @imtayadeway for the update.  🎵

PR #6, opened 2017-08-08
[WIP] Trying to improve the code-climate rating

abellotti:
- playing first with the VM resource actions

### DIFF
--- a/app/controllers/api/vms_controller.rb
+++ b/app/controllers/api/vms_controller.rb
@@ -12,81 +12,27 @@ module Api
     RELATIONSHIP_COLLECTIONS = [:vms, :templates].freeze
 
     def start_resource(type, id = nil, _data = nil)
-      raise BadRequestError, "Must specify an id for starting a #{type} resource" unless id
-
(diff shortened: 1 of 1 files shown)


app/controllers/api/vms_controller.rb
```diff
@@ -263,6 +173,19 @@ def set_miq_server_resource(type, id, data)
 
     private
 
+    def exec_vm_resource_action(verb, action, method, type, id = nil)
```
imtayadeway: Does there need to be a default argument for id?

Also, I'd prefer no abbreviations in these method names.

5 arguments is a lot of arguments. Did you consider using kwargs (or a hash) to better convey what they do from the calling end?

Fryguy: I agree with @imtayadeway but let's handle in a followup

Fryguy:
@abellotti Bump 🐢  (just remove the WIP...JUST DO IT!!!)

abellotti:
will revisit later.

PR #8, opened 2017-08-09
Don't require authorization for OPTIONS

imtayadeway:
CORS states that no authorization should be done in pre-flight
requests. For the problematic Container Deployments collection, we
will now render the sensitive data if authorized, but don't require
authorization, rendering no data if none is stipulated.

@miq-bot add-label bug
@miq-bot assign @abellotti 

/cc @jntullo

### DIFF
--- a/app/controllers/api/container_deployments_controller.rb
+++ b/app/controllers/api/container_deployments_controller.rb
@@ -9,11 +9,19 @@ module Api
       # TODO: this service is rendering resources which (a) require
       # authentication (problematic for CORS compatibility), and (b)
       # are not being properly filtered by RBAC
-      if [HttpHeaders::MIQ_TOKEN, HttpHeaders::AUTH_TOKEN, "HTTP_AUTHORIZATION"].any? { |header| request.headers.include?(header) }
+      if authentication_requested?
         require_api_user_or_token
         render_options(:container_deployments, ContainerDeploymentService.new.all_data)
       else
         super
+      end
+    end
+
+    private
+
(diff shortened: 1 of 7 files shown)


app/controllers/api/container_deployments_controller.rb
```diff
@@ -9,7 +9,12 @@ def options
       # TODO: this service is rendering resources which (a) require
       # authentication (problematic for CORS compatibility), and (b)
       # are not being properly filtered by RBAC
-      render_options(:container_deployments, ContainerDeploymentService.new.all_data)
+      if [HttpHeaders::MIQ_TOKEN, HttpHeaders::AUTH_TOKEN, "HTTP_AUTHORIZATION"].any? { |header| request.headers.include?(header) }
+        require_api_user_or_token
+        render_options(:container_deployments, ContainerDeploymentService.new.all_data)
```
abellotti: Nice @imtayadeway !!

Can we move the logic in line 12 to a method in base_controller/authentication.rb ? maybe something like request_includes_authentication.  Thanks.

imtayadeway: We _could_, although personally I think this is an anti-pattern and I don't want to promote re-use of it. I could extract a private method in this controller if you think this would be better?

abellotti: if you don't mind, private method for now would be good.  Thanks!!

abellotti:
Thanks @imtayadeway for the update 🎵 

will merge when 🍏

PR #9, opened 2017-08-09
missing_user_features now lives in the User model

jrafanie:
- [x] depends on https://github.com/ManageIQ/manageiq/pull/15779 [MERGED]

Note, it doesn't look like the actual user validation service was needed
other than to call this method, so the instantiation of it and the
caching in an instance variable were removed.

### DIFF
--- a/app/controllers/api/base_controller/authentication.rb
+++ b/app/controllers/api/base_controller/authentication.rb
@@ -49,8 +49,7 @@ module Api
       end
 
       def validate_user_identity(user_obj)
-        @user_validation_service ||= UserValidationService.new(self)
-        missing_feature = @user_validation_service.missing_user_features(user_obj)
+        missing_feature = User.missing_user_features(user_obj)
         if missing_feature
           raise AuthenticationError, "Invalid User #{user_obj.userid} specified, User's #{missing_feature} is missing"
         end


app/controllers/api/base_controller/authentication.rb
```diff
@@ -49,10 +49,11 @@ def authorize_user_group(user_obj)
       end
 
       def validate_user_identity(user_obj)
-        @user_validation_service ||= UserValidationService.new(self)
-        missing_feature = @user_validation_service.missing_user_features(user_obj)
+        missing_feature = User.missing_user_features(user_obj)
         if missing_feature
-          raise AuthenticationError, "Invalid User #{user_obj.userid} specified, User's #{missing_feature} is missing"
+          msg = "Invalid User #{user_obj.userid} specified, User's #{missing_feature} is missing"
+          $log.warn(msg)
```
abellotti: why the extra $log.warn,  all API raised errors, including authentication are in api.log.

jrafanie: Thanks, that was debugging I forgot to remove 😮

jrafanie:
Kicking the build now that https://github.com/ManageIQ/manageiq/pull/15779 is merged

abellotti:
LGTM!! Thanks @jrafanie for updating this.  👍

PR #7, opened 2017-08-08
Add guest_devices support to the API

skovic:
This PR adds guest_devices support to the API.

PR #11, opened 2017-08-09
Add API support for additional power operations

skovic:
Updates the physical_servers REST API to add support for additional power operations.

### DIFF
--- a/spec/requests/physical_servers_spec.rb
+++ b/spec/requests/physical_servers_spec.rb
@@ -32,11 +32,47 @@ RSpec.describe "physical_servers API" do
         expect(response).to have_http_status(:success)
       end
 
+      it "immediately powers off a server successfully" do
+        ps = FactoryGirl.create(:physical_server)
+
+        api_basic_authorize action_identifier(:physical_servers, :power_off_now, :resource_actions, :post)
+        run_post(physical_servers_url(ps.id), gen_request(:power_off_now))
+
+        expect(response).to have_http_status(:success)
+      end
+
       it "restarts a server successfully" do
         ps = FactoryGirl.create(:physical_server)
 
         api_basic_authorize action_identifier(:physical_servers, :restart, :resource_actions, :post)
         run_post(physical_servers_url(ps.id), gen_request(:restart))
+
+        expect(response).to have_http_status(:success)
+      end
+
+      it "immediately restarts a server successfully" do
+        ps = FactoryGirl.create(:physical_server)
+
+        api_basic_authorize action_identifier(:physical_servers, :restart_now, :resource_actions, :post)
+        run_post(physical_servers_url(ps.id), gen_request(:restart_now))
+
+        expect(response).to have_http_status(:success)
+      end
+
+      it "restarts a server to the system setup successfully" do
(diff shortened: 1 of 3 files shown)


spec/requests/physical_servers_spec.rb
```diff
@@ -40,6 +49,33 @@
...
         expect(response).to have_http_status(:success)
       end
+
+      it "immediately restarts a server successfully" do
+        ps = FactoryGirl.create(:physical_server)
+
+        api_basic_authorize action_identifier(:physical_servers, :restart_now, :resource_actions, :post)
+        run_post(physical_servers_url(ps.id), gen_request(:restart_now))
+
+        expect(response).to have_http_status(:success)
```
abellotti: while I see these expectations here match the rest of the file, I think they should all check for the action result signature, looking at others, maybe they should all be something like:

```
          expected = { 
            "message"   => "Requested server state restart_now for Server instance ...",
            "success"   => true,
            "task_href" => a_string_matching(tasks_url),
            "task_id"   => anything
          }   
```

skovic: @abellotti I believe that the intent of these tests were to test that authorized actions succeed and unauthorized ones fail, so we were only really concerned about each call having a status of _success_ or _forbidden_ and not about the response details. I took a look at the responses of the _power_off_ action, as an example, and it appears that the various response fields differ between calls so simply passing a hash of expected values to _expect_ probably won't work.
```
{"success"=>true, "message"=>"Requested server state power_off for Server instance: 65000000000075 name:''", "task_id"=>"65r175", "task_href"=>"http://www.example.com/api/tasks/65r175", "href"=>"http://www.example.com/api/physical_servers/65r75"}
{"success"=>true, "message"=>"Requested server state power_off for Server instance: 65000000000091 name:''", "task_id"=>"65r183", "task_href"=>"http://www.example.com/api/tasks/65r183", "href"=>"http://www.example.com/api/physical_servers/65r91"}
{"success"=>true, "message"=>"Requested server state power_off for Server instance: 65000000000113 name:''", "task_id"=>"65r194", "task_href"=>"http://www.example.com/api/tasks/65r194", "href"=>"http://www.example.com/api/physical_servers/65r113"}
```

skovic:
@abellotti Can we get this PR merged now that https://github.com/ManageIQ/manageiq/pull/15683 has been merged? Thanks

spec/requests/physical_servers_spec.rb
abellotti: Other than forbidden, because these are action results, failures still come back as 200, but the "success" => false in the resultant hash. This is done for allowing clients deal with bulk requests. Thus the recommendation above.

skovic: @abellotti I updated the test cases to also check the response body for success to address situations such as the one you described.

juliancheal:
@abellotti @imtayadeway is this breaking for the same reason as https://github.com/ManageIQ/manageiq-api/pull/7#issuecomment-326411979

rodneyhbrown7:
@juliancheal , @blomquisg

abellotti:
Thanks @skovic for the API Enhancement !!  👍

PR #13, opened 2017-08-09
Check if the User.current_user is set before calling userid

### LINKED ISSUE #12: When no current_user defined exception raised
When logging in for the first time, there is no User.current_user set and the log_request line blows up with the exception
#<NoMethodError: undefined method `userid' for nil:NilClass>

Which leads to the render being called multiple times

 https://github.com/ManageIQ/manageiq-api/blob/33eb2d8780e9e9f3464adec6c032c54e21fb3560/app/controllers/api/base_controller.rb#L35

F, [2017-08-09T14:23:54.647550 #4476] FATAL -- : AbstractController::DoubleRenderError (Render and/or redirect were called multiple times in this action. Please note that you may only call render OR redirect, and at most once per action. Also note that neither redirect nor render terminate execution of the action, so if you want to exit an action after redirecting, you need to do something like "redirect_to(...) and return".):

mkanoor:
Fixes #12

### DIFF
--- a/app/controllers/api/base_controller/authentication.rb
+++ b/app/controllers/api/base_controller/authentication.rb
@@ -23,6 +23,7 @@ module Api
             User.current_user = auth_user_obj
           else
             request_http_basic_authentication
+            return
           end
         end
         log_api_auth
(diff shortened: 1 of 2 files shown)


mkanoor:
@abellotti as discussed I have added a return when requesting credentials

abellotti:
Thanks @mkanoor for the fix.  👍

PR #14, opened 2017-08-10
add flavors create, delete to api

alexander-demichev:
Add flavors actions(create,delete) to api

### DIFF
--- a/app/controllers/api/subcollections/flavors.rb
+++ b/app/controllers/api/subcollections/flavors.rb
@@ -1,0 +1,31 @@
+module Api
+  module Subcollections
+    module Flavors
+      def flavors_query_resource(object)
+        object.flavors
+      end
+
+      def flavors_create_resource(parent, _type, _id, data)
+        task_id = Flavor.create_flavor_queue(User.current_user.id, parent, data)
+        action_result(true, 'Creating Flavor', :task_id => task_id)
+      rescue => err
+        action_result(false, err.to_s)
+      end
+
+      def delete_resource_flavors(_parent, type, id, _data)
+        flavor = resource_search(id, type, collection_class(type))
+        task_id = flavor.delete_flavor_queue(User.current_user.id)
+        action_result(true, "Deleting #{flavor_ident(flavor)}", :task_id => task_id)
+      rescue => err
+        action_result(false, err.to_s)
+      end
+      alias flavors_delete_resource delete_resource_flavors
+
+      private
+
+      def flavor_ident(flavor)
+        "Flavor id:#{flavor.id} name: '#{flavor.name}'"
+      end
+    end
+  end
+end
--- a/config/api.yml
+++ b/config/api.yml
@@ -864,7 +864,8 @@
     :identifier: flavor
     :options:
     - :collection
-    :verbs: *gp
+    - :subcollection
+    :verbs: *gpd
     :klass: Flavor
     :collection_actions:
       :get:
@@ -876,7 +877,28 @@
     :resource_actions:
       :get:
       - :name: read
+        :identifier: flavor_show    
+    :subcollection_actions:
+      :get:
+      - :name: read
+        :identifier: flavor_show_list
+      :post:
+      - :name: delete
+        :identifier: flavor_delete
+      - :name: query
+        :identifier: flavor_show_list
+      - :name: create
+        :identifier: flavor_create
+    :subresource_actions:
+      :get:
+      - :name: read
         :identifier: flavor_show
+      :post:
+      - :name: delete
+        :identifier: flavor_delete
+      :delete:
+      - :name: delete
+        :identifier: flavor_delete
   :floating_ips:
     :description: Floating IPs
     :identifier: floating_ip
@@ -1655,6 +1677,7 @@
     - :load_balancers
     - :security_groups
     - :vms
+    - :flavors
     :collection_actions:
       :get:
       - :name: read
--- a/spec/requests/flavors_spec.rb
+++ b/spec/requests/flavors_spec.rb
@@ -1,0 +1,171 @@
+RSpec.describe "Flavors API" do
+  describe "as a subcollection of providers" do
+    describe "GET /api/providers/:c_id/flavors" do
+      it "can list the flavors of a provider" do
+        api_basic_authorize(action_identifier(:flavors, :read, :subcollection_actions, :get))
+        ems = FactoryGirl.create(:ems_cloud)
+        flavor = FactoryGirl.create(:flavor, :ext_management_system => ems)
+
+        get(api_provider_flavors_url(nil, ems))
+
+        expected = {
+          "count"     => 1,
+          "name"      => "flavors",
+          "resources" => [
+            {"href" => api_provider_flavor_url(nil, ems, flavor)}
+          ]
+        }
+        expect(response.parsed_body).to include(expected)
+        expect(response).to have_http_status(:ok)
+      end
+
+      it "will not list flavors unless authorized" do
+        api_basic_authorize
+        ems = FactoryGirl.create(:ems_cloud)
+        FactoryGirl.create(:flavor, :ext_management_system => ems)
+
+        get(api_provider_flavors_url(nil, ems))
+
+        expect(response).to have_http_status(:forbidden)
+      end
+    end
+
+    describe "GET /api/providers/:c_id/flavors/:id" do
+      it "can show a provider's flavor" do
+        api_basic_authorize(action_identifier(:flavors, :read, :subresource_actions, :get))
+        ems = FactoryGirl.create(:ems_cloud)
+        flavor = FactoryGirl.create(:flavor, :ext_management_system => ems)
+
+        get(api_provider_flavor_url(nil, ems, flavor))
+
+        expected = {
+          "href" => api_provider_flavor_url(nil, ems, flavor),
+          "id"   => flavor.id.to_s
+        }
+        expect(response.parsed_body).to include(expected)
+        expect(response).to have_http_status(:ok)
+      end
+
+      it "will not show a flavor unless authorized" do
+        api_basic_authorize
+        ems = FactoryGirl.create(:ems_cloud)
+        flavor = FactoryGirl.create(:flavor, :ext_management_system => ems)
+
+        get(api_provider_flavor_url(nil, ems, flavor))
+
+        expect(response).to have_http_status(:forbidden)
+      end
+    end
+
+    describe "POST /api/providers/:c_id/flavors" do
+      it "can queue the creation of a flavors" do
+        api_basic_authorize(action_identifier(:flavors, :create, :subcollection_actions))
+        ems = FactoryGirl.create(:ems_cloud)
+
+        post(api_provider_flavors_url(nil, ems), :params => { :name => "test-flavor" })
+
+        expected = {
+          "results" => [
+            a_hash_including(
+              "success"   => true,
+              "message"   => "Creating Flavor",
+              "task_id"   => anything,
+              "task_href" => a_string_matching(api_tasks_url)
+            )
+          ]
+        }
+        expect(response.parsed_body).to include(expected)
+        expect(response).to have_http_status(:ok)
+      end
+
+      it "will not create a flavor unless authorized" do
+        api_basic_authorize
+        ems = FactoryGirl.create(:ems_cloud)
+
+        post(api_provider_flavors_url(nil, ems), :params => { :name => "test-flavor" })
+
+        expect(response).to have_http_status(:forbidden)
+      end
+    end
+
+    describe "POST /api/providers/:c_id/flavors/:s_id with delete action" do
+      it "can queue a flavor for deletion" do
+        api_basic_authorize(action_identifier(:flavors, :delete, :subresource_actions))
+
+        ems = FactoryGirl.create(:ems_cloud)
+        flavor = FactoryGirl.create(:flavor, :ext_management_system => ems)
+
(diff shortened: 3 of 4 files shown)


app/controllers/api/flavors_controller.rb
```diff
@@ -1,4 +1,30 @@
...
+    end
+
+    private
+
+    def flavor_ident(flavor)
+      "Flavor id:#{flavor.id} name: '#{flavor.name}'"
+    end
+
+    def validate_flavor_attrs(data)
+      data.dup
```
jntullo: not sure this method is necessary if no validation is being done. If you aren't manipulating the data, you shouldn't need to dup it either

app/controllers/api/flavors_controller.rb
```diff
@@ -1,4 +1,30 @@
...
+      attrs = validate_flavor_attrs(data)
+      task_id = Flavor.create_flavor_queue(User.current_user.id, EmsCloud.find_by_id(attrs['ems']['id']), attrs.deep_symbolize_keys)
+      action_result(true, 'Creating Flavor', :task_id => task_id)
+    rescue => err
+      action_result(false, err.to_s)
+    end
+
+    def delete_resource(type, id, _data = {})
+      flavor = resource_search(id, type, collection_class(:flavors))
+      raise "Delete not supported for #{flavor_ident(flavor)}" unless flavor.respond_to?(:delete_flavor_queue)
```
jntullo: is this supposed to be `delete_in_provider_queue`? it's checking a different method than is used below

alexander-demichev:
Thank you for review. I added some specs.

app/controllers/api/flavors_controller.rb
jntullo: @alexander-demichev I may be incorrect, but is this method `delete_flavor_queue` the correct method? I don't see any flavor descendants responding to it

alexander-demichev:
Flavors actions are implemented here https://github.com/ManageIQ/manageiq/pull/15552

app/controllers/api/flavors_controller.rb
alexander-demichev: @jntullo I made comment with link to flavors actions PR.

alexander-demichev:
@jntullo Hi, can you review it again?

imtayadeway:

```
vms_url => api_vms_url
```

Many thanks!
Tim

alexander-demichev:
@imtayadeway Hi, one of the tests is failing, can you take a look at it? Thanks. :-)

spec/requests/flavors_spec.rb
```diff
@@ -0,0 +1,76 @@
...
+
+      expect(response).to have_http_status(:no_content)
+    end
+
+    it 'will not delete a flavor without an appropriate role' do
+      flavor = FactoryGirl.create(:flavor)
+
+      api_basic_authorize
+
+      run_delete api_flavors_url(nil, flavor)
```
jntullo: I believe you want `api_flavor_url(nil, flavor) here

app/controllers/api/flavors_controller.rb
imtayadeway: Why don't all flavors respond to this? How does the user know which ones can be deleted? Is it possible that flavors could utilize the supports_feature? feature?

spec/requests/flavors_spec.rb
alexander-demichev: @jntullo Silly mistake :) Thank you, now it`s ok!

app/controllers/api/flavors_controller.rb
alexander-demichev: All flavors support delete. I should delete this line of code.

aufi:
Looks good to me, looking forward to see this merged :-)

abellotti:
Maybe add 2 tests for delete single flavor (POST /api/flavors/:id - action edit) and do a bulk delete (POST /api/flavors - action edit).

Otherwise looks good,

@imtayadeway and @jntullo your 👀  when you get a chance.  Thanks!!

spec/requests/flavors_spec.rb
```diff
@@ -0,0 +1,76 @@
...
+
+      api_basic_authorize collection_action_identifier(:flavors, :read, :get)
+
+      run_get(api_flavors_url)
+
+      expected = {
+        'count'     => 1,
+        'subcount'  => 1,
+        'name'      => 'flavors',
+        'resources' => [hash_including('href' => a_string_matching(api_flavors_url(nil, flavor.compressed_id)))]
```
jntullo: I wouldn't use `compressed_id` here, and instead just pass in `flavor`

alexander-demichev:
@jntullo @imtayadeway Hi, can you take a look, I edited tests and during CI it's throwing 'NoMethodError', but it works if I run test on my machine.

jntullo:
hi @alexander-demichev, if you rebase and update to use `get` instead of `run_get`, etc, you should be good to go 👍

alexander-demichev:
Added some tests :) @imtayadeway @jntullo

app/controllers/api/flavors_controller.rb
```diff
@@ -1,4 +1,24 @@
 module Api
   class FlavorsController < BaseController
+    def create_resource(_type, _id, data)
+      task_id = Flavor.create_flavor_queue(User.current_user.id, EmsCloud.find(data['ems']['id']), data)
```
jntullo: Rather than `EmsCloud.find` here, I think you may want to use `resource_search(:id, :providers, collection_class(:providers)`. Also, what do you think about supporting both `id` and `href`? Something along the lines of 
`{ "provider": { "href" : "/api/providers/:id" }` or ` { "provider": { "id": ":id" } }`. This [example](https://github.com/ManageIQ/manageiq-api/blob/master/app/controllers/api/configuration_script_sources_controller.rb#L25-L27) may help

alexander-demichev:
@jntullo Hi, I refactored code can you check it again?

app/controllers/api/flavors_controller.rb
```diff
@@ -1,4 +1,28 @@
 module Api
   class FlavorsController < BaseController
+    def create_resource(_type, _id, data)
+      manager_id = parse_id(data['ems'], :providers)
+      raise 'Must specify a valid manager_resource href or id' unless manager_id
```
jntullo: this is looking good! 😄  one minor thing - the error message does not match the data it is expecting. Perhaps change it to `data['manager_resource']`?

spec/requests/flavors_spec.rb
```diff
@@ -0,0 +1,163 @@
...
+  end
+
+  describe 'POST /api/flavors' do
+    let(:create_params) do
+      {
+        :action      => 'create',
+        :description => "Description",
+        :name        => "A Flavor",
+        :related     => {},
+        :ems         => ems,
```
jntullo: a little confused how this is working. should this be `ems => { 'href' => api_provider_url(nil, ems) }`? One last minor request as well, could you add a test for the validation of the ems provider? Thanks! 👍

spec/requests/flavors_spec.rb
```diff
@@ -0,0 +1,163 @@
...
+
+      api_basic_authorize collection_action_identifier(:flavors, :read, :get)
+
+      get(api_flavors_url)
+
+      expected = {
+        'count'     => 1,
+        'subcount'  => 1,
+        'name'      => 'flavors',
+        'resources' => [hash_including('href' => a_string_matching(api_flavors_url(nil, flavor)))]
```
imtayadeway: This `a_string_matching` isn't needed - it should match the whole of the string

spec/requests/flavors_spec.rb
```diff
@@ -0,0 +1,163 @@
+RSpec.describe 'Flavors API' do
+  let(:ems) { FactoryGirl.create(:ems_cloud) }
+  let(:flavor) { FactoryGirl.create(:flavor_openstack, :ext_management_system => ems) }
+  let(:flavor_2) { FactoryGirl.create(:flavor_openstack, :ext_management_system => ems) }
+
+  describe 'GET/api/flavors' do
+    it 'lists all the flavor bases with an appropriate role' do
```
imtayadeway: It's important to create a unified language both when writing tests and creating an API. Are they flavors or flavor bases?

spec/requests/flavors_spec.rb
```diff
@@ -0,0 +1,163 @@
...
+        :related     => {},
+        :ems         => ems,
+        :type        => 'ManageIQ::Providers::Openstack::CloudManager::Flavor',
+      }
+    end
+
+    it 'will delete multiple flavors' do
+      api_basic_authorize collection_action_identifier(:flavors, :delete, :post)
+
+      expected = {
```
imtayadeway: Nit, but try to keep the steps for each phase (setup, execution, verification) grouped together, keeping each group separated by a space.

app/controllers/api/flavors_controller.rb
```diff
@@ -1,4 +1,28 @@
 module Api
   class FlavorsController < BaseController
+    def create_resource(_type, _id, data)
+      manager_id = parse_id(data['ems'], :providers)
```
imtayadeway: Since you require an `ems` attribute, I'm wondering if it would be more natural to create flavors as a subcollection of a provider, i.e.:

```
POST /api/providers/:pid/flavors
```

That should simplify things - you don't need special handling in the event that you don't have an `ems` attribute, and it helps keep things consistent (we call ems's "providers" in the API). Each time we add a new required attribute with a new name, it needs some documentation on how to use it. Whereas the workflow for creating subcollection resources is already well documented. 

/cc @abellotti any thoughts?

abellotti: If flavors are only created within the scope of a provider, then yes, that makes sense @imtayadeway and more elegant.

alexander-demichev: @imtayadeway Hi, thank you for advice and review! Where can I find documentation for creating subcollection resources?

imtayadeway: @alexander-demichev there a lots of examples - here's one I picked at random: https://github.com/ManageIQ/manageiq-api/blob/master/app/controllers/api/subcollections/snapshots.rb#L8-L23

This way you can create a snapshot of a vm/instance. It happens to do this asynchronously, but you probably don't need to do that. The most important part is the naming - once you've added flavors as a subcollection of providers in the api.yml - there's some metaprograming that routes the message to a `flavors_create_resource` method that you'll need to create with the same signature as above. That should be enough to get started - ping me on gitter if you need any other help. Thanks!

app/controllers/api/subcollections/flavors.rb
```diff
@@ -0,0 +1,30 @@
+module Api
+  module Subcollections
+    module Flavors
+      def flavors_query_resource(object)
+        object.flavors
+      end
+
+      def flavors_create_resource(parent, _type, _id, data)
+        task_id = Flavor.create_flavor_queue(User.current_user.id, parent, data)
```
alexander-demichev: @imtayadeway @jntullo Should I check 'data' for all needed attributes here? Openstack has it's own check and there will be a check on frontend level.

jntullo: @alexander-demichev no, best to keep that logic out of the controller, and it seems it's already being well handled by Openstack 👍

app/controllers/api/subcollections/flavors.rb
```diff
@@ -0,0 +1,30 @@
...
+        action_result(true, 'Creating Flavor', :task_id => task_id)
+      rescue => err
+        action_result(false, err.to_s)
+      end
+
+      def delete_resource_snapshots(_parent, type, id, _data)
```
jntullo: I believe this should be `flavors_delete_resource`

alexander-demichev:
@jntullo  @imtayadeway Can I ask you for another round of review? 😄

config/api.yml
```diff
@@ -861,20 +861,30 @@
     :description: Flavors
     :identifier: flavor
     :options:
-    - :collection
-    :verbs: *gp
+    - :subcollection
```
jntullo: I didn't realize the collection already exists, I am assuming we'll still want to keep the collection actions? And just add in the subcollection actions?

imtayadeway: 👍 Yeah I think if it already exists we can preserve the ability to retrieve flavors using only their id

alexander-demichev:
@imtayadeway @jntullo Is it ok now?

config/api.yml
```diff
@@ -862,19 +862,30 @@
     :identifier: flavor
     :options:
     - :collection
-    :verbs: *gp
+    - :subcollection
+    :verbs: *gpd
     :klass: Flavor
-    :collection_actions:
+    :subcollection_actions:
```
jntullo: if you just add the collection actions back in, should be good to go 👍

spec/requests/collections_spec.rb
```diff
@@ -114,11 +114,6 @@ def test_collection_bulk_query(collection, collection_url, klass, id = nil)
       test_collection_query(:features, api_features_url, MiqProductFeature)
     end
 
-    it "query Flavors" do
```
jntullo: and we'll want these once you add the collections actions back into the api.yml

abellotti:
@alexander-demichev can you add the test for bulk deletes ?  i.e. POST /api/providers/:id/flavors action "delete" with 2 flavors ?

Thanks.

alexander-demichev:
@abellotti Added tests for bulk delete.

abellotti:
Thanks @alexander-demichev for the added test and the API enhancement.  👍

PR #16, opened 2017-08-10
require credentials when creating a provider

jntullo:
`manageiq-api` version of https://github.com/ManageIQ/manageiq/pull/15528

Currently, credentials are not required when creating a new provider via the api, which differs from the Classic UI which requires that credentials be added. This PR makes credentials required *if* there are no connection attributes specified (container providers only supply connection attributes) 

BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1467692

@miq-bot assign @abellotti

### DIFF
--- a/app/controllers/api/providers_controller.rb
+++ b/app/controllers/api/providers_controller.rb
@@ -20,6 +20,7 @@ module Api
 
     def create_resource(type, _id, data = {})
       assert_id_not_specified(data, type)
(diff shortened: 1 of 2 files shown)


abellotti:
LGTM!!! Thanks @jntullo  👍

simaishi:
@jntullo @abellotti Can/should this be backported to Fine branch?  The associated BZ has the backport flag set.

simaishi:
@jntullo @abellotti ^ ping

PR #19, opened 2017-08-14
Middleware API Endpoints

cfcosta:
This is a port of https://github.com/ManageIQ/manageiq/pull/15644.

This PR is the first step into generating API endpoints for middleware-related entities. Currently it covers all entities that appear on the UI menu, albeit only read-only for now.

What is covered:
- Datasources
- Deployments
- Domains
- Messagings
- Servers

What is not covered:

- Providers: this is already another endpoint, no need to duplicate it.

Next steps:

- Check with @mtho11 what resources need destructive/mutable endpoints, so that the API mimics correctly what the UI needs to do.
- Implement provider-specific data, so that we have endpoints that return all the middleware servers for one provider, for instance.
- Implement power operations, this is going to be done in the same way as the power operations on Vms.


cfcosta:
Closed in favour of https://github.com/ManageIQ/manageiq/pull/15644

cfcosta:
@imtayadeway hey, build is green now. There's a detail that doesn't make sense to me: for some reason, id compression and decompression works inconsistently locally and on Travis. Are there guidelines to this? And if so, how should it be done?

Also, there's another problem, but I'm guessing this is either rspec-rails or rails fault: when using an URL helper like `middleware_servers_url` it is not showing the full url, instead it is not adding the `http://example.com` part. Is that something you guys experienced elsewhere?

imtayadeway:
> Also, there's another problem, but I'm guessing this is either rspec-rails or rails fault: when using an URL helper like middleware_servers_url it is not showing the full url, instead it is not adding the http://example.com part. Is that something you guys experienced elsewhere?

@cfcosta perhaps confusingly, `middleware_servers_url` isn't coming from Rails, but [this](https://github.com/ManageIQ/manageiq-api/blob/master/spec/support/api_helper.rb#L79-L84). As it stands, we don't use resourceful routes so don't get those path helpers for free. I did however begin some work to [change that](https://github.com/ManageIQ/manageiq-api/pull/23), so you might see this change (and fewer surprises) in the future.

> id compression and decompression works inconsistently locally and on Travis.

Can you expand on what's inconsistent? If I had to guess it might be that your local test environment might be set to a region of 0, while travis randomizes it. In 0, an id of 1 becomes "1", but in 1, an id of 1 becomes "1r1" - could that account for the behaviour you were seeing?

spec/requests/middleware_datasources_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+
+  # For some reason middleware_datasources_url is not returning the full
+  # url, just the path portion. This is a hack, but will do the trick.
+  def datasource_url
+    id = ApplicationRecord.compress_id(datasource.id)
+    "http://www.example.com#{middleware_datasources_url(id)}"
+  end
+
+  def match_id
+    eq ApplicationRecord.compress_id(datasource.id).to_s
```
imtayadeway: You can do `datasource.compressed_id` here - which might obviate the need for this matcher.

spec/requests/middleware_datasources_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+      api_basic_authorize collection_action_identifier(:middleware_datasources, :read, :get)
+
+      run_get middleware_datasources_url
+
+      expect(response).to have_http_status(:ok)
+      expect(response.parsed_body).to eq(
+        'name'      => 'middleware_datasources',
+        'count'     => 1,
+        'resources' => [{
+          'href' => datasource_url
```
imtayadeway: You could use `a_string_matching(middleware_datasources_url(datasource.compressed_id)` if you didn't want to write your own helper methods (this is pretty much the strategy we use elsewhere), but this is fine as is.

cfcosta: Ok, I'll leave it like this. After your change to use the actual rails routes for that I can remove them.

spec/requests/middleware_datasources_spec.rb
cfcosta: Oh, nice!

cfcosta:
Can you expand on what's inconsistent? If I had to guess it might be that your local test environment might be set to a region of 0, while travis randomizes it. In 0, an id of 1 becomes "1", but in 1, an id of 1 becomes "1r1" - could that account for the behaviour you were seeing?

@imtayadeway yeah, that's actually the case. Maybe it should be more consistent, the test suite should be configuration-independent, imho.

Anyway, fixed according to comments, thanks for reviewing!

abonas:
what's the status of this PR?

imtayadeway:

```
vms_url => api_vms_url
```

Many thanks!
Tim

cfcosta:
@imtayadeway hey Tim, just updated to use the new url helpers, great work! Can you review the last commit and see if everything is ok?

cfcosta:
@abellotti this is the PR, thanks!

config/api.yml
```diff
@@ -1134,6 +1134,56 @@
...
+  :middleware_servers:
+    :description: Middleware Servers
+    :klass: MiddlewareServer
+    :options:
+      - :collection
+    :verbs: *g
+    :collection_actions:
+      :get:
+        - :name: read
+          :identifier: middleware_server_show_list
```
abellotti: the new collections above need resource specific actions, for GET's they have different role identifiers (the _show ones instead of the _show_list from the miq_product_features.yml file)>

spec/requests/middleware_datasources_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+      expect(response).to have_http_status(:ok)
+      expect(response.parsed_body).to eq(
+        'name'      => 'middleware_datasources',
+        'count'     => 0,
+        'resources' => [],
+        'subcount'  => 0
+      )
+    end
+
+    it 'returns a a listing of datasources' do
```
abellotti: one too many a

spec/requests/middleware_deployments_spec.rb
```diff
@@ -0,0 +1,65 @@
...
+          'href' => api_middleware_deployment_url(nil, deployment.compressed_id)
+        }],
+        'subcount'  => 1
+      )
+    end
+  end
+
+  describe '/:id' do
+    it 'returns the attributes of one deployment' do
+      api_basic_authorize
```
abellotti: individual resource gets, /api/_collection_/id will need the proper role identified

abellotti: and the equivalent forbidden tests for each

config/api.yml
abellotti: we also need to support the query action on the new collections.

spec/requests/middleware_servers_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+      expect(response.parsed_body['id']).to eq server.compressed_id
+      expect(response.parsed_body).to include(
+        'href'       => api_middleware_server_url(nil, server.compressed_id),
+        'name'       => server.name,
+        'feed'       => server.feed,
+        'properties' => server.properties,
+      )
+    end
+  end
+end
```
abellotti: we need the primary collections_spec.rb updated for the new collections for completeness, as well as testing the query posts.

cfcosta:
Hello @abellotti, just pushed your proposed fixes and changes, can you take a look at it again?

/cc @abonas

spec/requests/middleware_datasources_spec.rb
```diff
@@ -0,0 +1,65 @@
...
+      api_basic_authorize collection_action_identifier(:middleware_datasources, :read, :get)
+
+      get api_middleware_datasources_url
+
+      expect(response).to have_http_status(:ok)
+      expect(response.parsed_body).to eq(
+        'name'      => 'middleware_datasources',
+        'count'     => 1,
+        'resources' => [{
+          'href' => api_middleware_datasource_url(nil, datasource.compressed_id)
```
imtayadeway: We recently removed support for compressed ids, so URLs can be simplified:

```ruby
api_middleware_datasource_url(nil, datasource)
```

spec/requests/middleware_datasources_spec.rb
```diff
@@ -0,0 +1,65 @@
...
+      expect(response).to have_http_status(:forbidden)
+    end
+
+    it 'returns the attributes of one datasource' do
+      api_basic_authorize action_identifier(:middleware_datasources, :read, :resource_actions, :get)
+
+      get api_middleware_datasource_url(nil, datasource.compressed_id)
+
+      expect(response).to have_http_status(:ok)
+      expect(response.parsed_body['id']).to eq datasource.compressed_id
```
imtayadeway: Ids in the response body are now simply the string form of the object's id:

```ruby
expect(response.parsed_body["id"]).to eq(datasource.id.to_s
```

Also, since you're changing this anyway, you could (optionally) combine the expectations:

```ruby
expect(response.parsed_body).to include(
  "id" => datasource.id.to_s,
  "href" => api_middleware_datasource_url(nil, datasource)
)
```

cfcosta:
@imtayadeway thanks for the review! Done the suggested changes, I agree they look a lot better (and not having to care about id compression is a big win).

abellotti:
Thanks @cfcosta for making the changes and this API Enhancement.  LGTM !! 👍

abonas:
@jstickler I think this needs documentation (new REST api for the middleware provider)

abonas:
@miq-bot add_label providers/middleware

PR #25, opened 2017-08-17
Move init_env code into after_initialize hook

imtayadeway:
This helps to isolate things a bit better  - I'd like to be able to (at least) in theory require lib/api.rb without having to load all of Rails with it. The `after_initialize` hook seems like a better place to put this code, rather than invoking it at the time when lib/api.rb is required

/cc @jrafanie 
@miq-bot assign @abellotti

### DIFF
--- a/lib/manageiq/api/engine.rb
+++ b/lib/manageiq/api/engine.rb
@@ -7,6 +7,21 @@ module ManageIQ
       config.autoload_paths << root.join('lib', 'services').expand_path
       config.autoload_paths << root.join('lib').expand_path
 
+      config.after_initialize do
+        $api_log.info("Initializing Environment for #{::Api::ApiConfig.base[:name]}")
+        $api_log.info("")
+        $api_log.info("Static Configuration")
+        ::Api::ApiConfig.base.each { |key, val| log_kv(key, val) }
+
+        $api_log.info("")
+        $api_log.info("Dynamic Configuration")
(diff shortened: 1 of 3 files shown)


lib/manageiq/api/engine.rb
```diff
@@ -7,6 +7,21 @@ class Engine < ::Rails::Engine
...
+        $api_log.info("Static Configuration")
+        ::Api::ApiConfig.base.each { |key, val| log_kv(key, val) }
+
+        $api_log.info("")
+        $api_log.info("Dynamic Configuration")
+        ::Api::Environment.user_token_service.api_config.each { |key, val| log_kv(key, val) }
+      end
+
+      def self.log_kv(key, val)
+        $api_log.info("  #{key.to_s.ljust([24, key.to_s.length].max, ' ')}: #{val}")
```
Fryguy: We have a thing in VMDBLogger that logs hashes, which might just simplify this...

```
$api_log.log_hashes(::Api::ApiConfig.base)
```

Fryguy:
To avoid The GlobalVars rubocop, just update your codeclimate.local file to ignore $api_log

imtayadeway:
@Fryguy thanks for suggestions and extra ✂️ !

Fryguy:
LGTM :+1:... I'll leave it to @abellotti to review and merge.

abellotti:
Thanks @imtayadeway for updating the PR.  🎹

PR #32, opened 2017-08-22
Send back full dialog on create and edit

jntullo:
When the dialog is created or edited, the entire contents of the dialog need to be returned.

cc: @romanblanco 

@miq-bot assign @abellotti

### DIFF
--- a/app/controllers/api/service_dialogs_controller.rb
+++ b/app/controllers/api/service_dialogs_controller.rb
@@ -18,7 +18,8 @@ module Api
     end
 
     def create_resource(_type, _id, data)
-      DialogImportService.new.import(data)
+      dialog = DialogImportService.new.import(data)
+      fetch_service_dialogs_content(dialog).first
     rescue => e
       raise BadRequestError, "Failed to create a new dialog - #{e}"
     end
@@ -31,7 +32,7 @@ module Api
       rescue => err
         raise BadRequestError, "Failed to update service dialog - #{err}"
       end
-      service_dialog
+      fetch_service_dialogs_content(service_dialog).first
     end
 
     def copy_resource(type, id, data)
--- a/spec/requests/service_dialogs_spec.rb
+++ b/spec/requests/service_dialogs_spec.rb
@@ -137,9 +137,12 @@ describe "Service Dialogs API" do
         }
 
         expected = {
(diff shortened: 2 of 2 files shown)


app/controllers/api/service_dialogs_controller.rb
```diff
@@ -18,7 +18,8 @@ def fetch_service_dialogs_content(resource)
     end
 
     def create_resource(_type, _id, data)
-      DialogImportService.new.import(data)
+      dialog = DialogImportService.new.import(data)
+      fetch_service_dialogs_content(dialog)
```
abellotti: don't we want to return both the resource + resource.content (i.e. as if we're specifying additional_attributes of content ?)

jntullo: @abellotti it does return both, there are tests for that as well

abellotti: I'm still a bit confused here, the fetch_<type>_<attr> are getters for virtual attributes. Why isn't the hook on line 3 getting this for us.  set_additional_attributes asks for content.

jntullo: @abellotti `fetch_` are getters for virtual attributes - and it works as expected on `:index` and `:show`, as specified on line 3. This, however, is for `create` and `update`. I played around with this quite a bit today, and unfortunately applying the logic of line 3 to `create` and `update` with additional attributes will require a bit of refactoring in renderer and virtual attribute selection. i'd be happy to go forward with that work, but want to get your thoughts on whether that would be necessary.

app/controllers/api/service_dialogs_controller.rb
```diff
@@ -31,7 +32,7 @@ def edit_resource(type, id, data)
       rescue => err
         raise BadRequestError, "Failed to update service dialog - #{err}"
       end
-      service_dialog
+      fetch_service_dialogs_content(service_dialog).first
```
abellotti: why is there a .first here and not above on line 22 ?

jntullo: @abellotti it returns an array, and it needs to return the first element which is both the service dialog and the content

abellotti:
Thanks @jntullo for the API enhancement.  LGTM!!

PR #33, opened 2017-08-23
metric rollups subcollection

jntullo:
Adds Metric Rollups as a subcollection to services and vms

### Required parameters
The following parameters are required to query metric rollups as a subcollection
* capture_interval (either 'hourly' or 'daily')
* start_date

### Optional parameters:
* end_date
  * if no end_date is specified, it will use the current date 
* limit
  * If no limit is specified, the default value in the api settings will be used.

### Paging 
* Paging will be automatically applied to these metric rollup requests 
* the limit may be overriden, but is set by default in the api settings (use parameter `limit` to override)

#### Minimum attributes:
`GET /api/services/:id/metric_rollups?capture_interval=hourly&start_date=2017-08-01`
#### End date:
`GET /api/services/:id/metric_rollups?capture_interval=daily&start_date=2017-08-01&end_date=2017-08-22`
#### Limit:
`GET /api/services/:id/metric_rollups?capture_interval=daily&start_date=2017-08-01&end_date=2017-08-22&limit=2000`

#### Dates are in the format:
`YYYY-MM-DD` or `YYYY-MM-DDTHH:MM:SSZ`

@miq-bot add_label wip, enhancement

### DIFF
--- a/lib/services/api/metric_rollups_service.rb
+++ b/lib/services/api/metric_rollups_service.rb
@@ -2,27 +2,33 @@ module Api
   class MetricRollupsService
     REQUIRED_PARAMS = %w(resource_type capture_interval start_date).freeze
 
-    def self.query_metric_rollups(params)
-      validate_params(params)
+    attr_reader :params
 
-      params[:offset] ||= 0
-      params[:limit] ||= Settings.api.metrics_default_limit
+    def initialize(params)
+      @params = params
+      validate_required_params
+      validate_capture_interval
+    end
 
+    def query_metric_rollups
       start_date = params[:start_date].to_date
       end_date = params[:end_date].try(:to_date)
 
       MetricRollup.rollups_in_range(params[:resource_type], params[:resource_ids], params[:capture_interval], start_date, end_date)
     end
 
-    def self.validate_params(params)
+    private
+
+    def validate_required_params
       REQUIRED_PARAMS.each do |key|
         raise BadRequestError, "Must specify #{REQUIRED_PARAMS.join(', ')}" unless params[key.to_sym]
       end
+    end
 
+    def validate_capture_interval
       unless MetricRollup::CAPTURE_INTERVAL_NAMES.include?(params[:capture_interval])
         raise BadRequestError, "Capture interval must be one of #{MetricRollup::CAPTURE_INTERVAL_NAMES.join(', ')}"
       end
     end
-    private_class_method :validate_params
   end
 end
--- a/lib/services/api/metric_rollups_service.rb
+++ b/lib/services/api/metric_rollups_service.rb
@@ -1,0 +1,28 @@
+module Api
+  class MetricRollupsService
+    REQUIRED_PARAMS = %w(resource_type capture_interval start_date).freeze
+
+    def self.query_metric_rollups(params)
+      validate_params(params)
+
+      params[:offset] ||= 0
+      params[:limit] ||= Settings.api.metrics_default_limit
+
+      start_date = params[:start_date].to_date
+      end_date = params[:end_date].try(:to_date)
+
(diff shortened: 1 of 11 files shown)


jntullo:
@miq-bot remove_label wip

lib/services/api/metric_rollups_service.rb
```diff
@@ -0,0 +1,28 @@
+module Api
+  class MetricRollupsService
+    REQUIRED_PARAMS = %w(resource_type capture_interval start_date).freeze
+
+    def self.query_metric_rollups(params)
+      validate_params(params)
+
+      params[:offset] ||= 0
```
imtayadeway: Not certain, but I don't see this or the following param being used anywhere

lib/services/api/metric_rollups_service.rb
```diff
@@ -0,0 +1,28 @@
...
+      params[:limit] ||= Settings.api.metrics_default_limit
+
+      start_date = params[:start_date].to_date
+      end_date = params[:end_date].try(:to_date)
+
+      MetricRollup.rollups_in_range(params[:resource_type], params[:resource_ids], params[:capture_interval], start_date, end_date)
+    end
+
+    def self.validate_params(params)
+      REQUIRED_PARAMS.each do |key|
```
imtayadeway: Since this service is larger than one method, WDYT about making it a stateful object? If you did that, you could for instance extract each validation into its own private method, with intention-revealing names, etc..

spec/lib/services/api/metric_rollups_service_spec.rb
```diff
@@ -0,0 +1,19 @@
+describe Api::MetricRollupsService do
+  before do
+    allow(Settings.api).to receive(:metrics_default_limit).and_return(1000)
```
imtayadeway: Is this stub important to the actual values returned, or more of a sane default? If the latter, I wonder if it  might be better placed as an actual configuration in the test environment, in `config/settings/test.yml`?

spec/lib/services/api/metric_rollups_service_spec.rb
```diff
@@ -0,0 +1,19 @@
...
+  end
+
+  it 'validates that all of the parameters are present' do
```
imtayadeway: Since all these tests are about validation, I'm wondering it that is suggesting the service's single responsibility should be validation, and that the rest could be pulled out either into the caller, or another service?

lib/services/api/metric_rollups_service.rb
jntullo: @imtayadeway it's being used by paging

spec/lib/services/api/metric_rollups_service_spec.rb
jntullo: @imtayadeway its single responsibility is not about validation. I just felt that it was better to test that the validation is done here rather than add individual tests for the various subcollections. The various subcollections do test the other functionality though. I can alter the tests.

lib/services/api/metric_rollups_service.rb
imtayadeway: Ah! That's quite surprising to me - that this service would mutate some input params for the caller - perhaps this could be handled closer to where paging happens, or at least outside of this service?

spec/lib/services/api/metric_rollups_service_spec.rb
imtayadeway: > I just felt that it was better to test that the validation is done here rather than add individual tests for the various subcollections. 

Definitely! I guess I was just thinking that the tests are telling me that validation is important enough to be its own service. If there's a responsibility besides validation that was important enough to extract, I think that behaviour deserves to also be tested at this level. It's OK to have some redundancy between tests of different sizes, since I'd expect a request spec to cover integration, and unit tests to cover all possible paths through the service

abellotti:
Thanks @jntullo for the updates. 👍

PR #39, opened 2017-08-31
Add a basic ping endpoint

bdunne:
Allows us to know whether the API server is ready to serve requests


config/routes.rb
```diff
@@ -5,6 +5,8 @@
     root :to => "api#index", :as => :entrypoint
     match "/", :to => "api#options", :via => :options
 
+    get "/ping" => "ping#index"
```
Fryguy: /ping or /api/ping?

Fryguy: or both?

bdunne: It's under the [api namespace](https://github.com/ManageIQ/manageiq-api/pull/39/files#diff-21497849d8f00507c9c8dcaf6288b136L4), so it's `/api/ping`

bdunne:
@abellotti 
- Correct, I didn't want authentication on this endpoint.
- I thought about the collection, but I didn't think that would conflict.  😉

abellotti:
ping @bdunne  😄   can you add a small test when you get a chance ?  Thanks.

spec/requests/ping_spec.rb
```diff
@@ -0,0 +1,8 @@
+describe "PingController" do
+  it "get" do
+    get("/api/ping")
```
abellotti: minor, could you update to use the rails helper for consistency with the other specs.

```
run_get api_ping_url
```

Thanks,

abellotti:
Thanks @bdunne for updating the test. will merge when 🍏

PR #23, opened 2017-08-16
Replace custom entrypoint_url with Rails helper

imtayadeway:
Could do a lot more with this, but starting small. This small, one line change enables us to:

* ✂️ some code
* leverage more of Rails
* avoid doing so much interpolation as in https://github.com/ManageIQ/manageiq-api/compare/master...imtayadeway:url-helpers?expand=1#diff-66949d857e0cae98626c54e33f6ebcd3L36

@miq-bot add-label refactoring, test, technical debt
@miq-bot assign @abellotti

PR #40, opened 2017-08-31
Use Rails path helpers for all routes

imtayadeway:
Follow up to #23 

This enables us to:

* delete the custom metaprogramming in api_helper.rb which offered a
  similar but limited implementation.
* take advantage of the default behavior when passed an object which
  is to form a URL from the object's id, meaning the id can be omitted
  for terseness.
* avoid having to do  fuzzy matching when comparing hrefs, because
  where our methods provided only the path, Rails' gives us the full
  URL
* avoid doing our own string concatenation for subcollection routes,
  e.g.:
  ```ruby
  # before
  "#{blueprints_url(blueprint.id)}/tags/#{tag.id}"
  # after
  api_blueprint_tag_url(nil, blueprint, tag)
* possibly leverage some of these helpers on the production side

This won't stay fresh for long, so ask me to rebase prior to merging.

@miq-bot add-label test, technical debt
@miq-bot assign @abellotti

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -13,27 +13,34 @@ Rails.application.routes.draw do
 
     Api::ApiConfig.collections.each do |collection_name, collection|
       # OPTIONS action for each collection
-      match collection_name.to_s, :controller => collection_name, :action => :options, :via => :options
+      match collection_name.to_s, :controller => collection_name, :action => :options, :via => :options, :as => nil
 
       scope collection_name, :controller => collection_name do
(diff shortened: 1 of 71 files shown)


abellotti:
@imtayadeway ❤️ this PR.  Can you resolve the conflicts/repush ?  Thanks.

abellotti:
Looks good to me, Thanks @imtayadeway for this refactoring !!  🎼

PR #41, opened 2017-08-31
Fix broken alert definition specs

imtayadeway:
`MiqAlert#expression` became `#hash_expression` and `#miq_expession`
in https://github.com/ManageIQ/manageiq-schema/pull/49 - so these
tests needed to be updated.

@miq-bot add-label test
@miq-bot assign @abellotti


Fryguy:
Could all of this be simplified with an `MiqAlert#expression=` redirection method that writes into the appropriate column?

abellotti:
ping @imtayadeway, we're still failing here.  Could you please take a look.  Thanks!!

PR #42, opened 2017-09-01
OPTIONS - Generic Object Definition

AparnaKarve:
We need support for `OPTIONS /api/generic_object_definitions` so that the Generic Object Definitions UI populates the dropdowns with the allowed `association types` and allowed `data types`

Requires https://github.com/ManageIQ/manageiq/pull/15922 to be merged first for the new Dictionary entries added for the allowed data types.

### DIFF
--- a/app/controllers/api/generic_object_definitions_controller.rb
+++ b/app/controllers/api/generic_object_definitions_controller.rb
@@ -25,14 +25,26 @@ module Api
       raise BadRequestError, "Failed to delete generic object definition - #{err}"
     end
 
+    def self.allowed_association_types
+      GenericObjectDefinition::ALLOWED_ASSOCIATION_TYPES.collect do |association_type|
+        [Dictionary.gettext(association_type, :type => :model, :notfound => :titleize, :plural => false), association_type]
+      end.sort
+    end
+
+    def self.allowed_types
+      GenericObjectDefinition::TYPE_MAP.keys.collect do |type|
+        [Dictionary.gettext(type.to_s, :type => :data_type, :notfound => :titleize, :plural => false), type.to_s]
+      end.sort
+    end
+
     def options
       render_options(:generic_object_definitions, build_generic_object_definition_options)
     end
 
     def build_generic_object_definition_options
       {
-        :allowed_association_types => allowed_association_types,
-        :allowed_types             => allowed_types
+        :allowed_association_types => GenericObjectDefinitionsController.allowed_association_types,
+        :allowed_types             => GenericObjectDefinitionsController.allowed_types
       }
     end
     private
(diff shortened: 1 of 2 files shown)


abellotti:
This LGTM!!  Thanks @AparnaKarve for the enhancement.

/cc @jntullo can I borrow your 👀  for a quick review ?

abellotti:
@AparnaKarve will need a rebase/repush when you get a chance.  Thanks.

imtayadeway:

```
vms_url => api_vms_url
```

Many thanks!
Tim

AparnaKarve:
@imtayadeway I did think about adding `allowed_association_types` and `allowed_types` in the model, but wanted to keep the model free of all the `Dictionary/i18n` related stuff.

In normal course (pre-API days), we would have implemented `allowed_association_types` and `allowed_types` in the view controller (`controllers/generic_object_definition_controller.rb`)
So to keep that parity, I decided to add the methods in `allowed_association_types` and `allowed_types` in the API controller (`api/generic_object_definitions_controller.rb `) instead.

That was mainly the rationale behind adding the methods in `api/generic_object_definitions_controller.rb `

imtayadeway:
> I did think about adding allowed_association_types and allowed_types in the model, but wanted to keep the model free of all the Dictionary/i18n related stuff.

@AparnaKarve is that because you didn't want it to be executed at load time? Aren't there ways to delay it until call time?

Either way, if you didn't want to update the model, you could add a service object instead?

AparnaKarve:
> is that because you didn't want it to be executed at load time? Aren't there ways to delay it until call time?

@imtayadeway If we add the methods to the model, we probably do not have to worry about delayed loading etc. It should just work.
Not adding it to the model was a conscious decision, but I think I can reconsider that.

@abellotti Can you please merge this and I can create a follow-up PR to move the methods `allowed_association_types` and `allowed_types` out of the API controller. Thanks.

> you could add a service object instead?

Not sure how applicable the service object is to this case.

abellotti:
I'm ok with keeping those here, but still not crazy with the dup. As an option (no pun intended), why not just define those as class instance methods for now.

```
module Api
   class GenericObjectDefinitionsController
      def self.allowed_types
      end
      ...
      end
      ...
```

and used by both api controller and specs.

AparnaKarve:
@abellotti When I make those methods as class instance methods, they work in the api controller when I access them like this -
```ruby
GenericObjectDefinitionsController::allowed_association_types
```

But `GenericObjectDefinitionsController::allowed_association_types` does not seem to work in the spec. I'm seeing this -

```
Unable to autoload constant GenericObjectDefinitionsController, expected ~/master/manageiq/plugins/manageiq-api/app/controllers/api/generic_object_definitions_controller.rb to define it
```

AparnaKarve:
@abellotti Thanks for the tip on class instance method and `Api::GenericObjectDefinitionsController.<method_name>`

abellotti:
Thanks @AparnaKarve for the API enhancement !!   will merge when 🍏

PR #46, opened 2017-09-06
Preserve contract for expressions of alert definitions

imtayadeway:
This requires setting the `@additional_attributes` to include
`"expression"` for GET and POST (update) requests. For POST (create)
requests, because the object is wrapped in a "results" object, we need
to do a bit of a hack to insert the expression into a serialized
version of the alert definition before it gets rendered.

@miq-bot add-label bug
@miq-bot assign @abellotti

### DIFF
--- a/app/controllers/api/alert_definitions_controller.rb
+++ b/app/controllers/api/alert_definitions_controller.rb
@@ -2,13 +2,15 @@ module Api
   class AlertDefinitionsController < BaseController
     REQUIRED_FIELDS = %w(description db expression options).freeze
 
+    before_action :set_additional_attributes
+
     def create_resource(type, id, data = {})
       assert_id_not_specified(data, type)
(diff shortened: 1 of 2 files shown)


app/controllers/api/alert_definitions_controller.rb
```diff
@@ -23,5 +25,11 @@ def edit_resource(type, id = nil, data = {})
         raise BadRequestError, "Failed to update alert definition - #{err}"
       end
     end
+
+    private
+
+    def set_additional_attributes
+      @additional_attributes = %w(expression notify_email)
```
abellotti: why was notify_email needed ?

Fryguy: Even though I'm curious about this as well, this gets us to green so I'm merging... Let's follow up.

PR #38, opened 2017-08-31
Add PUT to generic object definitions API

jntullo:
Add the ability to use PUT to edit a generic object definition (cc: @AparnaKarve)

### Example usage
```
PUT /api/generic_object_definitions/:id
{
   "name" : "updated name",
   "description": "updated description"
}
```

@miq-bot add_label enhancement
@miq-bot assign @abellotti

PR #48, opened 2017-09-07
Fix generic object definition spec

imtayadeway:
This went red with the collision of #38 and #40 - updated to use the
new helper.

@miq-bot add-label bug
@miq-bot assign @abellotti


abellotti:
Looks good. Thanks for catching this @imtayadeway !!

PR #49, opened 2017-09-07
Add an Uncompressor

imtayadeway:
### The problem
Since we're accepting incoming compressed ids in place of actual ids, they will occasionally be passed to things that aren't expecting them, and definitely shouldn't. We don't want to update every method that deals with an id somewhere to uncompress it. (see Dialogs/Dialog Fields)

### The solution
I've added an `Uncompressor` service which gets executed when the body is parsed. It will traverse the body, looking for things to uncompress, and will uncompress them. This seems like the right place to do it, since it only needs to happen once and nothing else will need to know about compressed ids 

/cc @jntullo @eclarizio 
@miq-bot add-label enhancement, technical debt
@miq-bot assign @abellotti

### DIFF
--- a/app/controllers/api/base_controller/parser/request_adapter.rb
+++ b/app/controllers/api/base_controller/parser/request_adapter.rb
@@ -94,7 +94,7 @@ module Api
         def json_body
           @json_body ||= begin
                            body = @request.body.read if @request.body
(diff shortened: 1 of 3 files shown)


abellotti:
Any chance this will be morphing data blocks, not necessarily MiQ object resource attributes ?
Also, what's the performance impact.

PR #37, opened 2017-08-31
Generic Object Definition Actions

jntullo:
Generic Object Definition Actions

Dependent upon https://github.com/ManageIQ/manageiq/pull/15893

@miq-bot add_label enhancement, wip
@miq-bot assign @abellotti 

### General Usage Information
* Can reference a generic object definition by name, id, or href (see "add_attributes" example for how to specify)

#### add_attributes 
```
POST /api/generic_object_definitions
{
  "action": "add_attributes",
  "resources": {
    "attributes": [
    {
      "name": "name",
      "resource": {
        "attributes": {
          "attr1": "string",
          "another": "boolean"
        }
      }
    },
    {
      "id": ":id",
      "resource": {
        "attributes": {
          "attr2": "string"
        }
      }
    },
    {
      "href": "/generic_object_definition/:id",
      "resource": {
        "attributes": {
          "attr3": "string"
        }
      }
    }
  ]}
}
```

```
POST /generic_object_definitions/:id
{
  "action"   : "add_attributes",
  "resource" : {
    "attributes" : {
      "added_attribute1" : "string",
      "added_attribute2" : "boolean"
    }
  }
}
```
#### remove_attributes 
* Accepts same payload as "add_attributes", with different action specified
```
POST /generic_object_definitions/:id
{
  "action"   : "remove_attributes",
  "resource" : {
    "attributes" : {
      "added_attribute1" : "string",
      "added_attribute2" : "boolean"
    }
  }
}
```
### add_methods 
```
POST /generic_object_definitions
{
  "action"    : "add_methods",
  "resources" : [
    { "id" : "id", "resource" : { "methods" : ["method1", "method2"] } }
  ]
}
```
```
POST /generic_object_definitions/:id
{
  "action"   : "add_methods",
  "resource" : {
    "methods" : ["foo", "bar"]
  }
}
```
### remove_methods
* accepts same payload as add_methods with different action specified 
```
POST /generic_object_definitions
{
  "action"    : "remove_methods",
  "resources" : [
    { "id" : "id", "resource" : { "methods" : ["method1", "method2"] } }
  ]
}
```
```
POST /generic_object_definitions/:id
{
  "action"   : "remove_methods",
  "resource" : {
    "methods" : ["foo", "bar"]
  }
}
```
### add_associations
```
POST /generic_object_definitions
{
  "action"    : "add_associations",
  "resources" : [
[...]

PR #15, opened 2017-08-10
Generic Object Definition CRUD

jntullo:
`manageiq-api` version of https://github.com/ManageIQ/manageiq/pull/15374

This PR adds:
```
GET /api/generic_object_definitions/
GET /api/generic_object_definitions/:id
GET /api/generic_object_definitions/:name

DELETE /api/generic_object_definitions/:id
DELETE /api/generic_object_definitions/:name
POST /api/generic_object_definitions/:id      action: "delete"
POST /api/generic_object_definitions/:name    action: "delete"
POST /api/generic_object_definitions          action: "delete" (works with name, id, href)

POST /api/generic_object_definitions

POST /api/generic_object_definitions/:id      action: "edit"
POST /api/generic_object_definitions/:name    action: "edit"
POST /api/generic_object_definitions          action: "edit" (works with name, id, href)
```

Example of create:
```
POST /api/generic_object_definitions
{
  "name"        : "LoadBalancer",
  "description" : "LoadBalancer description",
  "properties"  : {
    "attributes"   : {
      "address"      : "string",
      "last_restart" : "datetime"
    },
    "associations" : {
      "vms"      : "Vm",
      "services" : "Service"
    },
    "methods"      : [
      "add_vm",
      "remove_vm"
    ]
  }
}
```
Editing a generic object definition. Properties will be updated as is - IE if you remove "methods" as in the below example, properties will no longer have methods and contain the attributes and associations passed in.
```
POST /api/generic_object_definitions/:id
{
 "action"       : "edit",
  "name"        : "LoadBalancer",
  "description" : "LoadBalancer description",
  "properties"  : {
    "attributes"   : {
      "address"      : "string",
      "last_restart" : "datetime"
    },
    "associations" : {
      "vms"      : "Vm",
      "services" : "Service"
    }
  }
}
```

@miq-bot add_label enhancement, wip
@miq-bot assign @abellotti

PR #50, opened 2017-09-11
Generic Objects API

jntullo:
Generic Objects API CRUD 
(will make separate PRs for subcollection actions and calling generic object methods)

### GET
* Returns property_attributes by default
* can specify associations to return 
```
GET /generic_objects/:id?associations=vms,services

response:
{
  "id": "<id>",
  "name": "generic_object_name",
  "generic_object_definition_id": "id",
  "property_attributes": { "widget": "test", "is_something": true },
  "vms": [ { "id": "id" ...  } ],
  "services": [ { "id": "id" ... }]  
}
```

#### Creating a new generic object
```
{  
   "generic_object_definition":{  
      "href":"/api/generic_object_definitions/:id"
   },
   "name":"go_name1",
   "uid":"optional_uid",
   "property_attributes":{  
      "widget":"widget value",
      "is_something":false
   },
   "associations":{  
      "vms":[  
         {  
            "href":"/api/vms/:id"
         },
         {  
            "href":"api/vms/:id"
         }
      ],
      "services":[  
         {  
            "href":"api/service/:id"
         }
      ]
   }
}
```

#### Editing a Generic Object
* Much like create - accepts `associations` and `property_attributes`
```
{  
   "action":"edit",
   "resources":[  
      {  
         "href":"/api/generic_objects/:id",
         "name":"updated name",
         "property_attributes":{  
            "widget":"updated widget"
         }
      }
   ]
}
```
#### Deleting a Generic Object
```
{  
   "action":"delete",
   "resources":[  
      {  
         "href":"/api/generic_objects/:id"
      }
   ]
}
```
#### Editing a generic object
```
{  
   "action":"edit",
   "resource":{  
      "name":"updated object",
      "property_attributes":{  
         "widget":"updated widget val",
         "is_something":false
      },
      "associations":{  
         "vms":[  
            {  
               "href":"/api/vms/:id"
            }
         ],
         "services":[  

         ]
      }
   }
}
```
#### Deleting a generic object
```
{
   "action": "delete"
}
```

### DELETE api/generic_objects/:id

@miq-bot assign @abellotti

### DIFF
--- a/app/controllers/api/base_controller/normalizer.rb
+++ b/app/controllers/api/base_controller/normalizer.rb
@@ -112,7 +112,7 @@ module Api
       end
 
       def normalize_array(obj)
-        type = @req.subcollection || @req.collection
+        type = @req.subject
         obj.collect { |item| normalize_attr(get_reftype(type, type, item), item) }
       end
 
--- a/app/controllers/api/base_controller/normalizer.rb
+++ b/app/controllers/api/base_controller/normalizer.rb
@@ -112,7 +112,8 @@ module Api
       end
 
       def normalize_array(obj)
-        obj.collect { |item| normalize_attr(@req.subcollection || @req.collection, item) }
+        type = @req.subcollection || @req.collection
+        obj.collect { |item| normalize_attr(get_reftype(type, type, item), item) }
       end
 
       def new_href(type, current_id, current_href)
--- a/app/controllers/api/generic_objects_controller.rb
+++ b/app/controllers/api/generic_objects_controller.rb
@@ -2,6 +2,8 @@ module Api
   class GenericObjectsController < BaseController
     EXCEPTION_ATTRS = %w(generic_object_definition associations).freeze
 
+    before_action :set_additional_attributes, :only => [:index, :show]
+
     def create_resource(_type, _id, data)
       object_def = retrieve_generic_object_definition(data)
       generic_object = object_def.create_object(data.except(*EXCEPTION_ATTRS))
@@ -11,6 +13,10 @@ module Api
     end
 
     private
+
+    def set_additional_attributes
+      @additional_attributes = %w(property_attributes)
+    end
 
     def retrieve_generic_object_definition(data)
       definition_id = parse_id(data['generic_object_definition'], :generic_object_definitions)
--- a/app/controllers/api/generic_objects_controller.rb
+++ b/app/controllers/api/generic_objects_controller.rb
@@ -8,7 +8,7 @@ module Api
     def create_resource(_type, _id, data)
       object_def = retrieve_generic_object_definition(data)
       object_def.create_object(data.except(*ADDITIONAL_ATTRS)).tap do |generic_object|
-        add_associations(generic_object, data)
+        add_associations(generic_object, data, object_def) if data.key?('associations')
         generic_object.save!
       end
     rescue => err
@@ -18,7 +18,7 @@ module Api
     def edit_resource(type, id, data)
       resource_search(id, type, collection_class(type)).tap do |generic_object|
         generic_object.update_attributes!(data.except(*ADDITIONAL_ATTRS))
-        add_associations(generic_object, data)
+        add_associations(generic_object, data, generic_object.generic_object_definition) if data.key?('associations')
         generic_object.save!
       end
     rescue => err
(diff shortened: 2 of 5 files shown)


app/controllers/api/generic_objects_controller.rb
```diff
@@ -0,0 +1,53 @@
...
+    rescue => err
+      raise BadRequestError, "Failed to update generic object - #{err}"
+    end
+
+    private
+
+    def set_additional_attributes
+      @additional_attributes = %w(property_attributes)
+      return unless params[:property_associations]
+      params[:property_associations].split(',').each do |prop|
```
jntullo: @abellotti wanted to get your thoughts on this approach to returning associations since it's slightly different than we talked about. With `GET /generic_objects/:id?property_associations=vms,services`, and by adding them to the additional attributes, they can then get returned as if they were a normal relationship, ie:
```
{
  "id": "<generic_object_id>",
   ...
  "vms" [ {"id": "vm_id" }, ... ]
  "services": [ ... ]
}
```
thoughts?

abellotti: I like it, I wonder if the parameter shouldn't just be associations=vms,services,... to match the add_associations/remove_associations actions.

jntullo: @abellotti yeah, I like that. will update! 👍

abellotti:
Thanks @jntullo for the PR. this LGTM 👍 

/cc @imtayadeway can I borrow your 👀 for a quick look here.  Thanks!!

app/controllers/api/base_controller/normalizer.rb
```diff
@@ -112,7 +112,8 @@ def normalize_select_attributes(obj, opts)
       end
 
       def normalize_array(obj)
-        obj.collect { |item| normalize_attr(@req.subcollection || @req.collection, item) }
+        type = @req.subcollection || @req.collection
```
imtayadeway: If you wish, you can use `type = @req.subject` here ☺️

abellotti: 👍 🎵

abellotti:
@imtayadeway waiting for your approval when you get a chance.  Thanks !!

chriskacerguis:
Echoing a previous question but how do you (1) get GO’s for a service and (2) how does one associate a GO with a service?

/cc @chalettu

jntullo:
@chriskacerguis @chalettu services don't have GO's so much as a GO can have a service. A service can be added to a Generic Object via association so long as that association is defined in the generic object definition of the GO (see https://github.com/ManageIQ/manageiq-api/pull/15 or https://github.com/ManageIQ/manageiq-api/pull/37). Then if that relationship exists, the services can be added as seen in the examples I have documented above (via both create and edit).

@lfu and @bzwei would be best to describe it though, and to inform if there is a straightforward way in which to grab generic objects of a service.

chriskacerguis:
@jntullo thanks for the clarification.  So, given a service, how can I get the associated GO's?

lfu:
>  A service can be added to a Generic Object via association so long as that association is defined in the generic object definition of the GO (see #15 or #37). 

@jntullo This is correct. A service can be connected to a generic object as an association. 

Generic objects can also be added to a service as resources. @chriskacerguis  To get the associated GO's of a given service, just call:
```
service.generic_objects
```
(see https://github.com/ManageIQ/manageiq/pull/15490)

jntullo:
@lfu thank you! I did not see that relationship. 

@chriskacerguis I will create a PR for a subcollection on services. ie `/services/:id/generic_objects`

chriskacerguis:
@jntullo much appreciated!!

@lfu thank you for pointing that out.

You both are awesome.

jntullo:
@chriskacerguis @chalettu  just realized, it's not even necessary for the subcollection. you'll be able to specify `attributes=generic_objects` since that relationship already exists.

chriskacerguis:
nice.  Thanks @jntullo

chriskacerguis:
@chalettu does all that address your concerns?

chalettu:
I believe this does address my concerns .  Thank you so much @jntullo for helping to clarify

abellotti:
@jntullo agreeing with you above, simply accessing via attributes=generic_objects.  Otherwise, we'd need to add the subcollection to all  API CI's which map to an Miq reportable model type.

abellotti:
LGTM !!!  Thanks @jntullo for this API enhancement 👍

PR #51, opened 2017-09-11
Use wrapped Rails request helpers instead of custom ones

imtayadeway:
By wrapping `Session#process` to do a little extra work for us (namely defaulting to `:as => :json` for PUT/PATCH/POST and merging in the request headers that are created by the `basic_authorize` method), the `run_get`, `run_post`, etc. helpers become transparent wrappers for the request helpers provided by Rails, meaning that we can simply use those ones instead. I think this will provide a better experience for new contributors who will be more familiar with these helpers (if they are a Rails developer).

@miq-bot add-label refactoring, test
@miq-bot assign @abellotti


imtayadeway:
Ah yes, the new syntax. I'll see if I can do something about that.

@miq-bot add-label wip

imtayadeway:
@miq-bot rm-label wip

PR #52, opened 2017-09-13
add image scanning endpoint

ilackarms:
this PR adds an API endpoint for initiating Container Image scans (SSA)

relevant RFE: https://bugzilla.redhat.com/show_bug.cgi?id=1486362

### DIFF
--- a/app/controllers/api/subcollections/container_images.rb
+++ b/app/controllers/api/subcollections/container_images.rb
@@ -1,0 +1,27 @@
+# frozen_string_literal: true
+
+module Api
+  module Subcollections
+    module ContainerImages
+      def container_images_scan_resource(provider_id, type, image_id, _payload)
+        api_action(type, image_id, :skip_href => true) do |klass|
+          image = resource_search(image_id, type, klass)
+          begin
+            desc = "#{container_image_ident(image)} scanning"
+            task_id = image.scan
+            url = api_provider_container_image_url(nil, provider_id, image_id)
+            action_result(true, desc).merge(:task_id => task_id, :href => url)
+          rescue => err
+            action_result(false, err.to_s)
+          end
+        end
+      end
+
+      private
+
+      def container_image_ident(image)
+        "ContainerImage id:#{image.id} name:'#{image.name}'"
+      end
+    end
+  end
+end
--- a/config/api.yml
+++ b/config/api.yml
@@ -19,6 +19,8 @@
     :gp: &gp
     - :get
     - :post
+    :p: &p
+    - :post
     :gpd: &gpd
     - :get
     - :post
@@ -686,6 +688,19 @@
       :get:
       - :name: read
         :identifier: container_deployment_show
+  :container_images:
+    :description: ContainerImages
+    :options:
+    - :subcollection
+    :verbs: *gp
+    :klass: ContainerImage
+    :subresource_actions:
+      :get:
+      - :name: show
+        :identifier: container_image_show
+      :post:
+      - :name: scan
+        :identifier: container_image_scan
   :container_nodes:
     :description: Container Nodes
     :identifier: container_node
@@ -1714,6 +1729,7 @@
     - :vms
     - :flavors
     - :cloud_templates
+    - :container_images
     :collection_actions:
       :get:
       - :name: read
--- a/spec/requests/container_images_spec.rb
+++ b/spec/requests/container_images_spec.rb
@@ -1,0 +1,48 @@
+# frozen_string_literal: true
+
+#
+# REST API Request Tests - /api/container_images
+#
+describe "ContainerImages API" do
+  let(:provider) { FactoryGirl.create(:ems_kubernetes) }
+  let(:container_image) { FactoryGirl.create(:container_image, :ext_management_system => provider) }
+  let(:conatiner_id) { container_image.id }
+  let(:invalid_container_image) { container_image.id + 100 }
+  let(:invalid_image_url) { api_provider_container_image_url(nil, provider, invalid_container_image) }
+
+  context "ContainerImages show action" do
+    it "read a Container Image" do
+      api_basic_authorize(action_identifier(:container_images, :show, :subresource_actions, :get))
+      get api_provider_container_image_url(nil, provider, container_image)
+
+      expect(response).to have_http_status(:ok)
+      expect_single_resource_query("id" => container_image.id.to_s)
+    end
+  end
+
+  context "ContainerImages scan action" do
+    it "responds with 404 Not Found for an invalid container image" do
+      api_basic_authorize(action_identifier(:container_images, :scan, :subresource_actions, :post))
+
+      post(invalid_image_url, :params => { :action => "scan" })
+
+      expect(response).to have_http_status(:not_found)
+    end
+
+    it "scans a Container Image without appropriate role" do
+      api_basic_authorize
+
+      post(invalid_image_url, :params => { :action => "scan" })
+
+      expect(response).to have_http_status(:forbidden)
+    end
+
+    it "scan a Container Image" do
+      api_basic_authorize(action_identifier(:container_images, :scan, :subresource_actions, :post))
+      url = api_provider_container_image_url(nil, provider, container_image)
+      post url, :params => { :action => "scan" }
+
+      expect_single_action_result(:success => true, :message => "ContainerImage id:#{container_image.id} name:'#{container_image.name}' scanning", :href => url, :task_id => true)
+    end
+  end
+end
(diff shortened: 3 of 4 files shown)


config/api.yml
```diff
@@ -770,6 +772,15 @@
       :post:
       - :name: assign
       - :name: unassign
+  :container_images:
+    :description: ContainerImages
+    :options:
+    - :subcollection
+    :verbs: *p
```
jntullo: it seems that not only would you want to scan them, but you would also want to be able to get them (ie `GET /api/providers/container_images`) after the scan.

config/api.yml
```diff
@@ -1476,6 +1488,10 @@
         :identifier: ems_infra_protect
       - :name: unassign
         :identifier: ems_infra_protect
+    :container_images_subcollection_actions:
```
jntullo: you shouldn't have to define these again here, unless they were specific to the collection. so defining them above would be just fine

config/api.yml
```diff
@@ -770,6 +772,15 @@
       :post:
       - :name: assign
       - :name: unassign
+  :container_images:
```
imtayadeway: @ilackarms it looks like you just need to move this section to https://github.com/ManageIQ/manageiq-api/pull/52/files#diff-15502b585d5033aa8043cc26237a06baL686 to get this to 💚

ilackarms: can you explain what you mean? shouldn't `  :container_images:` be a `collection`?

ilackarms: currently the api endpoint is `/api/providers/<provider_id/container_images/<image_id>`

config/api.yml
ilackarms: what should the `GET` return?

config/api.yml
ilackarms: nevermind, i see what you mean :+1:

config/api.yml
imtayadeway: I'm surprised that this works, as `p` isn't an alias for anything

simon3z:
@ilackarms is there a BZ to link here?

app/controllers/api/subcollections/container_images.rb
```diff
@@ -0,0 +1,43 @@
+# frozen_string_literal: true
+
+module Api
+  module Subcollections
+    module ContainerImages
+      def container_images_scan_resource(provider_name, resource_name = nil, image_id = nil, _payload = nil)
+        raise BadRequestError, "Must specify an id for scanning a #{provider_name} resource" unless resource_name
+        image = ContainerImage.find(image_id)
```
jntullo: This should probably use a `resource_search` rather than a find here

app/controllers/api/subcollections/container_images.rb
```diff
@@ -0,0 +1,43 @@
...
+        image_name = image.name
+        task = nil
+        begin
```
jntullo: You probably don't need the begin statement, and can just rescue the errors similarly to how `scan_container_image` is structured below, with the action results for both success and failure.

app/controllers/api/subcollections/container_images.rb
```diff
@@ -0,0 +1,43 @@
...
+          task = image.scan
+        rescue => bang
+          raise ApiError, "Error during 'Analysis' of #{image_name}: #{bang.message}"
+        end
+        { :msg => "Analysis of #{image_name} successfully initiated", :task_id => task}
+      end
+
+      private
+
+      def container_image_ident(vm)
```
jntullo: Is this method being used? I'm not sure I see it being used. Same for the other private methods

ilackarms:
@simon3z yes; bz: https://bugzilla.redhat.com/show_bug.cgi?id=1486362

simon3z:
> @simon3z yes; bz: https://bugzilla.redhat.com/show_bug.cgi?id=1486362

@ilackarms add it to the PR description

app/controllers/api/subcollections/container_images.rb
ilackarms: if i remove the `begin` block and rescue anything, we wind up catching 404s which get converted to 200s. so i think it's correct to leave the begin block in

config/api.yml
```diff
@@ -1608,6 +1623,7 @@
     - :custom_attributes
     - :load_balancers
     - :vms
+    - :container_images
```
Fryguy: This should be alphbetical.

Fryguy:
Just to clarify, this PR is doing 2 separate things, right?  1. Adding the show part for retrieving container_images, and 2. introducing the scan.  I wonder if that should be 2 separate PRs just to make it clearer.

Fryguy:
> > @simon3z yes; bz: https://bugzilla.redhat.com/show_bug.cgi?id=1486362
> 
> @ilackarms add it to the PR description

Add it to the commit.  Then the bot will auto-do things to the BZ on merge.

ilackarms:
@Fryguy I agree that logically show and scan are logically separate. However the only feature I'm trying to implement is `scan`; it's necessary to add `show` in order for `scan` to work. if i do split this into 2 PRs, the `scan` PR will depend on `show`

config/api.yml
```diff
@@ -24,6 +24,8 @@
     :gp: &gp
     - :get
     - :post
+    :p: &p
```
imtayadeway: I don't see this being used anywhere. Was it necessary?

spec/requests/container_images_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+      api_basic_authorize(action_identifier(:container_images, :show, :subresource_actions, :get))
+      get api_provider_container_image_url(nil, provider, container_image)
+
+      expect(response).to have_http_status(:ok)
+      expect_single_resource_query("id" => container_image.id.to_s)
+    end
+  end
+
+  context "ContainerImages scan action" do
+    it "scans an invalid container_image" do
```
imtayadeway: Reading this description, it looks like you're saying it does the opposite of what's intended. Perhaps:

```ruby
it "responds 404 Not Found for an invalid container image"
```

app/controllers/api/subcollections/container_images.rb
```diff
@@ -0,0 +1,28 @@
+# frozen_string_literal: true
+
+module Api
+  module Subcollections
+    module ContainerImages
+      def container_images_scan_resource(provider_id, type = nil, image_id = nil, _payload = nil)
+        raise BadRequestError, "Must specify an id for scanning a #{provider_id} resource" unless type
```
imtayadeway: Can you add a test for this? Honestly, I'd be surprised if you could be routed here without a provider id

app/controllers/api/subcollections/container_images.rb
```diff
@@ -0,0 +1,28 @@
+# frozen_string_literal: true
+
+module Api
+  module Subcollections
+    module ContainerImages
+      def container_images_scan_resource(provider_id, type = nil, image_id = nil, _payload = nil)
```
imtayadeway: Is there a reason all these defaults are here? Is somewhere that's calling this method with just the first arg?

ilackarms: good point, fixing in revision

app/controllers/api/subcollections/container_images.rb
ilackarms: you're right; I picked up this line when looking at another endpoint as a template (the VM scanning one). since the path includes a provider ID, this should never be reachable code

spec/requests/container_images_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+  context "ContainerImages scan action" do
+    it "responds with 404 Not Found for an invalid container image" do
+      api_basic_authorize(action_identifier(:container_images, :scan, :subresource_actions, :post))
+
+      post(invalid_image_url, :params => { :action => "scan" })
+
+      expect(response).to have_http_status(:not_found)
+    end
+
+    it "scans an invalid Container Image without appropriate role" do
```
abellotti: does not scan an invalid ...

ilackarms:
@imtayadeway please re-review. thank you

app/controllers/api/subcollections/container_images.rb
```diff
@@ -0,0 +1,27 @@
...
+  module Subcollections
+    module ContainerImages
+      def container_images_scan_resource(provider_id, type, image_id, _payload)
+        api_action(type, image_id, :skip_href => true) do |klass|
+          image = resource_search(image_id, type, klass)
+          begin
+            desc = "#{container_image_ident(image)} scanning"
+            task_id = image.scan
+            url = api_provider_container_image_url(nil, provider_id, image_id)
+            action_result(true, desc).merge(:task_id => task_id, :href => url)
```
abellotti: no need for the :href here and related url in 12,  this can be done with  :parent_id => provider_id instead in action_result.

abellotti:
@ilackarms can you take care of the requested changes and rebase ?  Thanks.

abellotti:
closing this PR as it is re-submitted as #245

PR #55, opened 2017-09-18
Express ids as uncompressed strings

imtayadeway:
Support for compressed ids was added and enabled by default (rendering
all ids in their compressed form) with the thought of killing two birds with one stone:

1. JavaScript doesn't support very large numbers
2. we wanted to add support for compressed ids

It turns out that (2) wasn't solving any problems in UI land
anymore. With the added complexity and emerging issues¹ that came as a
result of this change, it seemed better to remove the compressed id
support and simply render all ids as strings instead, which is
sufficient to solve the javascript large numbers issue.

Support for incoming compressed ids in URLs has been left, since this
has already been incorporated into the Fine release.

1. See #49, ManageIQ/manageiq#15658

@miq-bot assign @abellotti 
/cc @chrisarcand @Fryguy

### DIFF
--- a/spec/requests/alert_definitions_spec.rb
+++ b/spec/requests/alert_definitions_spec.rb
@@ -24,10 +24,10 @@ describe "Alerts Definitions API" do
       "subcount"  => 2,
       "resources" => a_collection_containing_exactly(
         {
-          "href" => api_alert_definition_url(nil, alert_definitions[0].compressed_id)
+          "href" => api_alert_definition_url(nil, alert_definitions[0])
         },
         {
-          "href" => api_alert_definition_url(nil, alert_definitions[1].compressed_id)
+          "href" => api_alert_definition_url(nil, alert_definitions[1])
         }
       )
     )
@@ -49,8 +49,8 @@ describe "Alerts Definitions API" do
     get(api_alert_definition_url(nil, alert_definition))
     expect(response).to have_http_status(:ok)
     expect(response.parsed_body).to include(
-      "href"        => api_alert_definition_url(nil, alert_definition.compressed_id),
-      "id"          => alert_definition.compressed_id,
+      "href"        => api_alert_definition_url(nil, alert_definition),
+      "id"          => alert_definition.id.to_s,
       "description" => alert_definition.description,
       "guid"        => alert_definition.guid,
       "expression"  => {"exp" => {"=" => {"field" => "Vm-name", "value" => "foo"}}, "context_type" => nil}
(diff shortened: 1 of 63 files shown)


imtayadeway:
@chrisarcand curses. We should be using the path helpers in a lot of those cases, so I get to make it even better 😀

imtayadeway:
Leaving the remaining cop complaints as that's the builder DSL....we should configure rubocop to ignore these.....or I can just use parens. No strong feelings.

abellotti:
@imtayadeway I'm ok with the Builder DSL rubocop complaints for now, (enough changes as is in this PR).

abellotti:
tested both inbound GET /api/vms/10000000127 and /api/vms/1r127, both still good 👍   
all outbound responses have the expand 10000... in the id's, href's.   Now time to browse at 62 files 😏

imtayadeway:
@abellotti I would suggest making a pot of ☕️ 😉

spec/requests/alert_definitions_spec.rb
```diff
@@ -224,8 +224,8 @@
     expect(response).to have_http_status(:ok)
     expect_result_resources_to_include_hrefs(
       "resources",
-      [api_alert_definition_profile_alert_definitions_url(nil, alert_definition_profile.compressed_id, alert_definitions.first.compressed_id),
-       api_alert_definition_profile_alert_definitions_url(nil, alert_definition_profile.compressed_id, alert_definitions.first.compressed_id)]
+      [api_alert_definition_profile_alert_definitions_url(nil, alert_definition_profile.id.to_s, alert_definitions.first),
+       api_alert_definition_profile_alert_definitions_url(nil, alert_definition_profile.id.to_s, alert_definitions.first)]
```
abellotti: I see some usage of the helpers, with the object itself, and in the second argument here, the id specifically.  can this be called as:   api_alert_definition_profile_alert_definitions_url(nil, alert_definition_profile, alert_definitions.first) ?

spec/requests/alerts_spec.rb
```diff
@@ -80,7 +80,7 @@
         "subcount"  => 1,
         "resources" => [
           {
-            "href" => api_alert_alert_action_url(nil, alert.compressed_id, alert_action.compressed_id)
+            "href" => api_alert_alert_action_url(nil, alert.id.to_s, alert_action)
```
abellotti: same question as above for all collection/subcollection _url calls.

imtayadeway: Yeah, these should be. Will update!

abellotti:
@imtayadeway other than above calls to collection/subcollection url helpers, this looks great !! 🎼 

Once updated, can this be merged as is or do we need to synchronize with the UI team ?

imtayadeway:
@abellotti great, thanks! I can push up those changes in a new commit if that's easier to review and then squash after 👍 ?

This should be able to be merged as is :D

imtayadeway:
@abellotti updated in https://github.com/ManageIQ/manageiq-api/pull/55/commits/f2cc6a3dbaf3b185322b2ef33cdc669d9e1836b7 - please let me squash that after you're done :D

abellotti:
Thanks @imtayadeway looks good. (thanks for not squashing).   Can can squash/rebase (Generic objects PR merged)/repush ?   Thanks.

imtayadeway:
@abellotti rebased and made https://github.com/ManageIQ/manageiq-api/pull/55/commits/f5849e3b6d11c3cf4e92bac904356e55ff41b3f9 so you could more easily see what I changed (/cc @jntullo), just give me the word and I will squash those last two

abellotti:
Thanks @imtayadeway for this PR, will merge when 🍏

PR #56, opened 2017-09-18
Simplify href matching helper

imtayadeway:
Because we are now using the Rails path helpers everywhere, we no
longer need to match so flexibly in this helper for checking hrefs.

Doing so revealed that a couple of tests were giving false positives
and needed to be updated to faithfully test the actual behavior.

This also allowed for some deletions (it no longer depends on the
`resources_include_suffix?` helper) and simplified to make one
expectation where three were before. This should yield better test
feedback in the event of a regression - it'll provide a diff instead
of the old "expected true, got false".

@miq-bot add-label test, technical debt
@miq-bot assign @abellotti


abellotti:
LGTM!! :+1:

PR #57, opened 2017-09-18
Generic Objects Subcollection

jntullo:
This PR adds a Generic Objects Subcollection on the Generic Object Definition

#### GET /api/generic_object_definitions/:id/generic_objects
will return the collection of generic objects of that definition type:
```
{
  "name"      :"generic_objects",
  "count"     :2,
  "subcount"  :2,
  "resources" :[
   { "href" : "api/generic_object_definitions/:id/generic_objects/:id" },
   { "href" : "api/generic_object_definitions/:id/generic_objects/:id" },
  ]
}
```
#### GET /api/generic_object_definitions/:id/generic_objects/:id
will return property attributes by default:
```
{
  "href"                : "/api/generic_object_definitions/:id/generic_objects/:id",
  "id"                  : "1000000001",
  "name"                : "bar",
  ... 
  "property_attributes" : { "is_something" : true }
}
```
can specify `associations` on the request, ie: 
`GET /api/generic_object_definitions/:id/generic_objects/:id?associations=vms,services` 
returns:
```
{
  "href"                : "api/generic_object_definitions/:id/generic_objects/:id",
  "id"                  : "10000001",
  "name"                : "bar",
  "property_attributes" : { "is_something" : true },
  "vms"                 : [{ "id": "<vm id>" ...}],
  "services"            : [{"id": "<service id>"...}]
}

```
#### GET /api/services/:id/generic_objects
* Returns the generic objects associated with a service 
#### GET /api/services/:id?expand=resources,generic_objects&attributes=generic_objects.generic_object_definition
* will return the service as well as the generic objects with associated generic object definition:
```
{'name':'services',
 'count':1,
 'subcount':1,
 'resources':
   [{'href':'/api/services/:id',
     'id':':id',
     'name':'svc',
     ...
     'generic_objects':
       [{
         'id':':id',
         'name':'generic_object_0000000000001',
         ...
         'generic_object_definition' : {
            'id' : ':id',
            'name' : 'generic_object_definition_name'
            ...
          }
        }
       ]
    }
   ]
}
```

#### POST /api/generic_object_definitions/:id/generic_objects
it can create a new generic object for the specified generic object definition:
```
{
  "name"                : "go_name1",
  "uid"                 : "optional_uid",
  "property_attributes" : {
    "widget" : "widget value",
  },
  "associations"        : {
    "vms" : [
      { "href" : "/api/vms/:id" },
    ]
  }
}

```
#### Background
~There is shared code between the generic objects controller and the generic objects subcollection. The code is tightly coupled with the controllers, and therefore not suitable for a service, which was the design choice behind the mixin.~

~When writing the mixin, the initial intent was to not have it as a part of BaseController, it was instead to have `Api::Mixins::GenericObjects`, however, this caused sporadic test failures when included as `include Mixins::GenericObjects`.  I think this may require greater refactoring to achieve it as `Api::Mixins`, which might be better suited in another PR.~

### DIFF
--- a/app/controllers/api/generic_object_definitions_controller.rb
+++ b/app/controllers/api/generic_object_definitions_controller.rb
@@ -1,5 +1,11 @@
 module Api
   class GenericObjectDefinitionsController < BaseController
+    include Api::Mixins::GenericObjects
+    include Subcollections::GenericObjects
+
+    before_action :set_additional_attributes, :if => :generic_objects_request?
+    before_action :set_associations, :only => [:index, :show], :if => :generic_objects_request?
+
     def create_resource(_type, _id, data)
       klass = collection_class(:generic_object_definitions)
       klass.create!(data.deep_symbolize_keys)
@@ -113,6 +119,10 @@ module Api
     end
 
     private
+
+    def generic_objects_request?
+      @req.subject == 'generic_objects'
+    end
 
     def fetch_generic_object_definition(type, id, data)
       id ||= data['name']
--- a/app/controllers/api/subcollections/generic_objects.rb
+++ b/app/controllers/api/subcollections/generic_objects.rb
@@ -2,7 +2,15 @@ module Api
   module Subcollections
     module GenericObjects
       def generic_objects_query_resource(object_definition)
-        object_definition.generic_objects
+        generic_objects = object_definition.generic_objects
+        go_attrs = attribute_selection_for('generic_objects')
+
+        return generic_objects if go_attrs.blank?
+
+        generic_objects.collect do |go|
+          attributes_hash = create_resource_attributes_hash(go_attrs, go)
+          go.as_json.merge(attributes_hash)
+        end
       end
 
       def generic_objects_create_resource(object, _type, _id, data)
--- a/app/controllers/api/subcollections/generic_objects.rb
+++ b/app/controllers/api/subcollections/generic_objects.rb
@@ -1,0 +1,13 @@
+module Api
+  module Subcollections
+    module GenericObjects
+      def generic_objects_query_resource(object_definition)
+        object_definition.generic_objects
+      end
+
+      def generic_objects_create_resource(object, _type, _id, data)
+        create_generic_object(object, data)
+      end
+    end
+  end
+end
--- a/spec/requests/generic_object_definitions_spec.rb
+++ b/spec/requests/generic_object_definitions_spec.rb
@@ -76,6 +76,81 @@ RSpec.describe 'GenericObjectDefinitions API' do
     end
   end
 
+  describe 'GET /api/generic_object_definitions/:id/generic_objects' do
+    it 'can return the generic objects of a definition with appropriate role' do
+      object = FactoryGirl.create(:generic_object, :name => 'bar', :generic_object_definition => object_def)
+      api_basic_authorize subcollection_action_identifier(:generic_object_definitions, :generic_objects, :read, :get)
+
+      get(api_generic_object_definition_generic_objects_url(nil, object_def))
+
+      expected = {
+        'name'      => 'generic_objects',
+        'count'     => 1,
+        'subcount'  => 1,
+        'resources' => [
+          { 'href' => api_generic_object_definition_generic_object_url(nil, object_def, object) }
+        ]
+      }
+      expect(response).to have_http_status(:ok)
+      expect(response.parsed_body).to include(expected)
+    end
+
+    it 'does not return the generic objects without an appropriate role' do
+      api_basic_authorize
+
+      get(api_generic_object_definition_generic_objects_url(nil, object_def))
+
+      expect(response).to have_http_status(:forbidden)
+    end
+  end
+
+  describe 'GET /api/generic_object_definitions/:id/generic_objects/:id' do
+    let(:object) { FactoryGirl.create(:generic_object, :name => 'bar', :generic_object_definition => object_def) }
+    let(:vm) { FactoryGirl.create(:vm_amazon) }
+
+    before do
+      object_def.add_property_attribute('is_something', 'boolean')
+      object_def.add_property_association('services', 'Service')
+      object_def.add_property_association('vms', 'Vm')
+
+      object.is_something = true
+      object.save!
+
+      object.add_to_property_association('vms', [vm])
+    end
+
+    it 'returns the generic object with property attributes by default' do
+      api_basic_authorize subcollection_action_identifier(:generic_object_definitions, :generic_objects, :read, :get)
+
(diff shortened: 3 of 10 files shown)


spec/requests/generic_object_definitions_spec.rb
```diff
@@ -688,4 +763,39 @@
...
+
+      generic_object = {
+        'name'                => 'go_name1',
+        'uid'                 => 'optional_uid',
+        'property_attributes' => {
+          'widget' => 'widget value',
+        },
+        'associations'        => {
+          'vms' => [
+            { 'href' => api_vm_url(nil, vm.id) },
```
abellotti: with these helpers, you can simply pass in the object, i.e. api_vm_url(nil, vm)

spec/requests/generic_object_definitions_spec.rb
```diff
@@ -76,6 +76,81 @@
...
+      expect(response).to have_http_status(:ok)
+    end
+
+    it 'will return the specified associations' do
+      api_basic_authorize subcollection_action_identifier(:generic_object_definitions, :generic_objects, :read, :get)
+
+      get(api_generic_object_definition_generic_object_url(nil, object_def.id, object.id), :params => {:associations => 'vms'})
+
+      expected = {
+        'href'                => api_generic_object_definition_generic_object_url(nil, object_def.id, object.id),
```
abellotti: same comment throughout, with these helpers just the object so ...(nil, object_def, object)

app/controllers/api/subcollections/generic_objects.rb
```diff
@@ -0,0 +1,13 @@
+module Api
+  module Subcollections
+    module GenericObjects
+      def generic_objects_query_resource(object_definition)
+        object_definition.generic_objects
```
abellotti: what if the generic object definition has no generic object, is this nil here ? what is returned by API ?

app/controllers/api/base_controller/generic_objects_mixin.rb
```diff
@@ -0,0 +1,40 @@
+module Api
```
abellotti: I wonder if this should go in a new folder: app/controllers/api/mixins/ 
and dropping the BaseController in the module hierarchy, it would just be a submodule of the
class including the mixin.

app/controllers/api/generic_object_definitions_controller.rb
```diff
@@ -114,6 +120,10 @@ def build_generic_object_definition_options
 
     private
 
+    def generic_objects_request
```
abellotti: not quite sure what this does. why is it needed ?

app/controllers/api/base_controller/generic_objects_mixin.rb
jntullo: @abellotti I explained a bit more in the PR description, but I have been looking into this (which is why it's still WIP) ...however, without the BaseController module hierarchy and with it in `app/controllers/api/mixins`, it has been creating sporadic test failures that I have been trying to figure out.

jntullo: It seems to have to do with the way things are loaded, but haven't figured out the solution just yet.

app/controllers/api/subcollections/generic_objects.rb
jntullo: @abellotti if it has no generic objects, it just returns an empty array, so it just returns `"resources": []`

app/controllers/api/generic_object_definitions_controller.rb
jntullo: I know, it's a bit funky, but basically this controller needs to have the `before_action` hooks to allow for the subcollection to set the `associations` for requests such as `GET /api/generic_object_definitions/:id/generic_objects?associations=vms`as well as to set the hook to set `@additional_attributes`...we don't want to set the hooks unless the subject is `generic_objects`, so this method is the `if` condition on the hook. updated the name to `is_generic_objects_request`

jntullo:
@abellotti exhibit A of sporadic test failures. Thought everything was fine with the mixin, finally, but Travis is now having failures I don't get locally 😭

jntullo:
@miq-bot assign @abellotti

app/controllers/api/generic_object_definitions_controller.rb
```diff
@@ -1,5 +1,11 @@
 module Api
   class GenericObjectDefinitionsController < BaseController
+    include Api::Mixins::GenericObjects
```
imtayadeway: In short, I think you needed the `Api` qualification because the classic UI has a top-level `Mixins` namespace. If something in there gets loaded first, it will break the autoloading of this module. I'm proposing that the classic UI not do that and put their mixins in their own namespace.

jntullo: thanks @imtayadeway !!! ✨

imtayadeway: Or you could use `include_concern Mixins::GenericObjects` which I guess is the standard way that we deal with this problem (and may convey to the next person that there is some kind of contention over the `Mixins` namespace)

jntullo: @imtayadeway unfortunately I tried this, and the API is not happy :(

bascar:
@abellotti can we get this merged please? it is blocking other Generic Object items

jntullo:
just squashed some commits.

abellotti:
LGTM!!  

@imtayadeway ?

abellotti:
ping @imtayadeway can we borrow your 👀 please ?

spec/requests/generic_object_definitions_spec.rb
```diff
@@ -76,6 +76,81 @@
...
+      expect(response).to have_http_status(:forbidden)
+    end
+  end
+
+  describe 'GET /api/generic_object_definitions/:id/generic_objects/:id' do
+    let(:object) { FactoryGirl.create(:generic_object, :name => 'bar', :generic_object_definition => object_def) }
+    let(:vm) { FactoryGirl.create(:vm_amazon) }
+
+    before do
+      object_def.add_property_attribute('is_something', 'boolean')
```
imtayadeway: The extensive setup here seems to suggest that the GO API is missing something....perhaps we could at a later point create a factory, or a builder, or at least enhance the factory girl definition to help simplify this process?

abellotti:
Thank you @imtayadeway for your review 🙇

spec/requests/generic_object_definitions_spec.rb
Fryguy: Yeah, some factories for "simple" objects with dummy attributes would be useful

PR #60, opened 2017-09-20
Support more cases for parse_href

NickLaMuro:
When the path for a href is passed in without a leading `'/'`, the value can turn into something like `'/api/api/collection/:id'`, which is not desired and will break following methods in the API in a non-obvious way.

By normalizing all paths to make sure they start with a slash, we can then assume that is what we are working with when we append '/api' to the path if it is missing, allowing a more flexible interface for `#parse_href`

In the final commit, I also to a stab at reducing some object allocations a bit, but it is far from necessary.  Can be removed if desired.

Spec File Location
------------------

In a side conversation, we discussed that the `spec/requests/` dir is probably not the best place for this test to live, as no real "requests" are being made to confirm functionality, and a different dir should probably be used.  Other ideas were tossed around like this parser should probably be a service object, like `HrefParser`, and then it can be tested as a typical "plain old ruby object" instead of how it is now.

For now, I have followed the conventions of putting the spec file in the `spec/requests/` dir, but this ideally would change in the future (out of scope).

Links
-----
* Gitter conversation that raised this issue:  [link](https://gitter.im/ManageIQ/api?at=59c180007b7d98d30d15dcf7)

### DIFF
--- a/app/controllers/api/base_controller/parser.rb
+++ b/app/controllers/api/base_controller/parser.rb
@@ -62,7 +62,8 @@ module Api
       def parse_href(href)
         if href
           path = href.match(/^http/) ? URI.parse(href).path.sub(/\/*$/, '') : href
-          path = "/api/#{path}" unless path.match("/api")
+          path.prepend("/")     unless path.start_with?("/")
+          path = "/api#{path}" unless path.match("/api")
           path = path.sub(/\/*$/, '')
           return href_collection_id(path)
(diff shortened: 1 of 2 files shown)


NickLaMuro:
cc @jntullo @imtayadeway

abellotti:
Thanks @NickLaMuro for the API improvement and tests !!  👍

PR #62, opened 2017-09-20
Update API to support LXCA config patterns

skovic:
This PR adds API support for Lenovo XClarity Administrator (LXCA) configuration patterns which can be thought of as templates describing the configuration of physical systems. These "templates" are used by LXCA to configure hardware such as servers, switches, and storage.

### DIFF
--- a/spec/requests/configuration_templates_spec.rb
+++ b/spec/requests/configuration_templates_spec.rb
@@ -1,0 +1,30 @@
+RSpec.describe "configuration templates API" do
+  describe "display a configuration template's details" do
+    context "with the user authorized" do
+      it "responds with configuration template details" do
+        config_template = FactoryGirl.create(:configuration_template,
+                                             :ems_id  => 1,
+                                             :ems_ref => "65",
+                                             :name    => "server-template")
+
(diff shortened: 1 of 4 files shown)


skovic:
@imtayadeway This PR has the test cases using the factory. Thanks

spec/requests/configuration_templates_spec.rb
```diff
@@ -0,0 +1,30 @@
+RSpec.describe "configuration templates API" do
+  describe "display a configuration template's details" do
+    context "with the user authorized" do
+      it "responds with configuration template details" do
+        config_template = FactoryGirl.create(:configuration_template,
+                                             :ems_id  => 1,
```
imtayadeway: Even though this may work, setting ids yourself can cause surprises down the road. It may be best to create a real ems and let Rails handle which ids need to be set:

```ruby
FactoryGirl.create(:configuration_template, :ems => FactoryGirl.create(:ems_lenovo), :etc => "etc")
```

imtayadeway:
@skovic the above approach looks good....my thought earlier was that you may not need a factory, but it's probably best to create one (even if it is an empty definition) to avoid brittleness 👍

PR #63, opened 2017-09-20
Update references to OrchstrationTemplate subclasses due to rename

bzwei:
As described in https://github.com/ManageIQ/manageiq/issues/15985 all subclasses of `OrchestrationTemplate` were moved to provider repos. This PR updates the only one explicit reference to `OrchestrationTemplate` subclasses in manageiq-api. 

Please advice how to keep the backward compatibility for OrchestrationTemplate api. The only thing I can think of is the create method where the type is required. We can have a hard-coded hash to translate old type to new type.

@miq-bot assign @abellotti

### DIFF
--- a/app/controllers/api/orchestration_templates_controller.rb
+++ b/app/controllers/api/orchestration_templates_controller.rb
@@ -4,7 +4,7 @@ module Api
       klass    = collection_class(type)
       resource = resource_search(id, type, klass)
       result = super
-      resource.raw_destroy if resource.kind_of?(OrchestrationTemplateVnfd)
+      resource.raw_destroy if resource.kind_of?(ManageIQ::Providers::Openstack::CloudManager::VnfdTemplate)
       result
     end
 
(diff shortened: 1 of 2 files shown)


abellotti:
Thanks @bzwei for this PR.

yes, definitely a concern, specially for folks that have templates stashed somewhere that they will feed into our API.

type translation inbound will be needed as you mentioned, at least for next release where we can deprecate its use and log/doc accordingly.

Not sure where would be best to do this translation (API or Model).  /cc @imtayadeway for your thoughts.

bzwei:
ping @abellotti @imtayadeway 
Can we merge this PR and have another PR to address the backward compatibility concern?
If we merge the clean-up PR https://github.com/ManageIQ/manageiq/pull/15964, it will break the build of api without this fix.

abellotti:
@bzwei I'm ok with this for now, I'll go ahead and merge.

We do need the mapping done on creates so we don't break current API customers with templates they have stored/use.    Do you mind creating another PR to do the mapping on creates ?   This will need to be logged as such/deprecated in the next release.

bzwei:
@abellotti I will try. Thanks.

PR #65, opened 2017-09-20
Add a rudimentary event streams API

imtayadeway:
Adds an Event Streams API with the rudimentary features as stipulated in https://www.pivotaltracker.com/story/show/149754397. If still required, I'll follow this up with event streams subcollection endpoints in another PR. Marking this as WIP as I have a few questions I need addressed before officially submitting this.

@miq-bot add-label wip, enhancement
@miq-bot assign @abellotti 

/cc @AllenBW

PR #68, opened 2017-09-22
s/metrics_default_limit/large_collection_default_limit/

imtayadeway:
Since we want to use the same value/concept for the upcoming event
streams API, we need to give this configurable a more abstract name.

Disclaimer: naming is hard. If you have a good suggestion, please come forward. Also considered: `firehose_default_limit` and `big_$@!_collection_default_limit` 😁 

/cc @jntullo 
@miq-bot assign @abellotti

### DIFF
--- a/app/controllers/api/metric_rollups_controller.rb
+++ b/app/controllers/api/metric_rollups_controller.rb
@@ -2,7 +2,7 @@ module Api
   class MetricRollupsController < BaseController
     def index
       params[:offset] ||= 0
-      params[:limit] ||= Settings.api.metrics_default_limit
+      params[:limit] ||= Settings.api.large_collection_default_limit
 
       rollups_service = MetricRollupsService.new(params)
       resources = rollups_service.query_metric_rollups
--- a/app/controllers/api/subcollections/metric_rollups.rb
+++ b/app/controllers/api/subcollections/metric_rollups.rb
@@ -7,7 +7,7 @@ module Api
 
       def metric_rollups_query_resource(object)
         params[:offset] ||= 0
-        params[:limit] ||= Settings.api.metrics_default_limit
+        params[:limit] ||= Settings.api.large_collection_default_limit
         params[:resource_type] = RESOURCE_TYPES[@req.collection] || object.class.to_s
         params[:resource_ids] ||= [object.id]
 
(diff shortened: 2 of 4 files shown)


abellotti:
I'm thinking (uh oh) more like: `paging_default_limit` 
since we're enforcing paging on metrics, event streams.

imtayadeway:
@abellotti I wanted to convey something of the where/why we are applying pagination by default in the name (hence qualifying with "large collections"), since we don't apply it everywhere?

abellotti:
I understand that but things like vms can also be large and we don't enforce paging so it wasn't clear as to its usage other than us documenting it.  We could go back to having one for each `metrics_default_limit`, `events_default_limit`, .... then from looking at the settings one can tell which collection is enforcing it.

yes, naming is hard 😉

imtayadeway:
@abellotti on my development machine I have ~5 miilion event stream rows compared to 714 Vms....I'd expect anyone else's to be (similarly) several orders of magnitude bigger? I think *not* qualifying it in some way could be more confusing?

abellotti:
another option is to have a collection option in the api.yml
 \:metrics
   \:options
      - enforce_paging

for the collections that need it, then leverage the `paging_default_limit` setting.

chrisarcand:
Suggestion: `Settings.api.max_results_per_page` or similar makes more sense to me for this particular option. I didn't have any clue what `large_collection_default_limit` could mean opening this page.

> another option is to have a collection option in the api.yml

Yes, but be able to override that as you see fit (give it a hash)

```
:metrics
:options
:paging
  :enabled: true
  :max_results_per_page: 1000
```

`max_results` in this override is optional. It would default to `api.max_results_per_page`.

Just some thoughts.

imtayadeway:
@abellotti that's a good suggestion, but that would really be a new feature and well out of the scope of this PR. At present we don't have the ability to switch pagination on/off arbitrarily on collections. If we call it the `pagination_default_limit` that implies that we have a default page size. At present we don't in general - it's only a feature of metric rollups. That is something we should consider but I don't want it to be a blocker for the event streams API.

Perhaps not trying to generalize this (yet) would make the most sense, and just living with two separate configurables for now (while it's not officially released) would be better?

imtayadeway:
@chrisarcand I love that approach but I think I am going to close this for now in favor of https://github.com/ManageIQ/manageiq-api/pull/65/files#diff-646ad6b10917e4385cbcc8c8524e24a2R5. I'll try to incorporate in a future PR when I'm able to properly generalize max page size for all/any collections 👍

PR #69, opened 2017-09-22
Config specs are not request specs

imtayadeway:
Moving this file into a path that matches the subject.

@miq-bot add-label test
@miq-bot assign @abellotti


abellotti:
👍  each test in their own home.

PR #70, opened 2017-09-22
Spec helpers housekeeping

imtayadeway:
Some refactorings that remove some obsolete code and low-value abstractions resulting in quite a few ✂️ and IMHO improved readability.

@miq-bot add-label test, refactoring
@miq-bot assign @abellotti

### DIFF
--- a/spec/requests/providers_spec.rb
+++ b/spec/requests/providers_spec.rb
@@ -162,12 +162,12 @@ describe "Providers API" do
 
     context 'with restricted user' do
       let(:user) do
-        FactoryGirl.create(:user, :miq_groups => [group], :password => api_config(:password), :userid => api_config(:user), :name => api_config(:user_name))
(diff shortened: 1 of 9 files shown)


spec/requests/providers_spec.rb
```diff
@@ -161,13 +161,9 @@ def have_endpoint_attributes(expected_hash)
...
     context 'with restricted user' do
-      let(:user) do
-        FactoryGirl.create(:user, :miq_groups => [group], :password => api_config(:password), :userid => api_config(:user), :name => api_config(:user_name))
-      end
-
-      def define_user
+      before do
+        @user.update(:miq_groups => [group])
         @role = role
-        user
```
chrisarcand: Don't change it in this PR but...

While this is better, it still makes two writes to make a user. In the future, instead of an ivar in setup ("`init_api_spec_env`") we should have a top level memoized helper that gets called before an example is run. Then this would become the following:

```ruby

context 'with restricted user' do
  let(:user) do
    create(:restricted_user)
  end
end
```

The restricted user is then just created, not created and then updated.

PR #71, opened 2017-09-22
Extract HrefParser service

imtayadeway:
Builds on https://github.com/ManageIQ/manageiq-api/pull/60 by extracting that logic into a reasonably-isolated service (it only knows about `ApplicationRecord.compressed_id` which will go away eventually, and `ApiConfig`). The benefits of this are:

1. Allows us to test the service in isolation/outside of the integration test framework. Here's a before/after with times:
Before:
```
Finished in 0.64315 seconds (files took 4.63 seconds to load)
```
After:
```
Finished in 0.01073 seconds (files took 4.61 seconds to load)
```
2. Allows us to refactor for readability while burying the details in something outside of the controller.
3. Fixes some code climate issues

As you can see, I took the opportunity to refactor quite a lot, trading some LOC for readability. I did this in very small deliberate steps....happy to squash if preferred but I left as is for the reviewer as the before/after look quite different.

@miq-bot assign @abellotti


abellotti:
LGTM!!  squash would be good, but maybe after all reviewed.

/cc @NickLaMuro can we borrow your 👀  ?  Thanks.

NickLaMuro:
Yeah, I will take a look in a bit, but from the PR description alone, I probably already am for doing this.

lib/services/api/href_parser.rb
```diff
@@ -0,0 +1,72 @@
...
+
+    private
+
+    attr_reader :href
+
+    def subcollection?
+      !!subcollection
+    end
+
+    def path
```
NickLaMuro: This could use some kind of caching, and how you go about it depends on how you decide to go forward with my previous comment I guess.

But for example, if you had a `href` that was a `subcollection`, and `#parse` was called this method would get executed 3 times and return the same value each time, but require doing the transformations and regexp evaluations each time:

* First time would be in line `13` when `subcollection?` is called (from `#subcollection => #path_parts => #path`)
* Second would be `subcollection.to_sym` in line 14
* Third would be from `subcollection_id` in line 14 as well

Oh!  And I lied, it would actually call this method 6 times because for each of the above, the call to `version?` also hits `path` (again, through `#path_parts` in that method).  Either caching the value of `#path_parts` via an instance variable, or storing the value of `path_parts` as something like `parts` (if you go with the "class method approach" from above) would avoid this duplicated work.

lib/services/api/href_parser.rb
```diff
@@ -0,0 +1,72 @@
+module Api
+  class HrefParser
+    def self.parse(href)
```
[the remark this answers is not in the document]
imtayadeway: Mainly because I wanted a way to express a very narrow public API for this service, which is a bit more gracefully expressed as an instance. I can also anticipate some further work being done here where instead of an `HrefParser` it might simply embody an `Href` (or `MiqHref`) so we could ask it some questions that are currently expressed private methods. We duplicate some of this knowledge in various places (see, for example, https://github.com/ManageIQ/manageiq-api/blob/master/app/controllers/api/base_controller/parser/request_adapter.rb).

lib/services/api/href_parser.rb
imtayadeway: I am happy to memoize if keeping the instance implementation is agreeable?

lib/services/api/href_parser.rb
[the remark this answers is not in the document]
NickLaMuro: Cool, makes sense.  Works for me, and a single object allocation on each request probably isn't going to cause problems here.

My comment was more of a "is there a reason for this that I am missing", and less of a "you should change this" one (though that probably wasn't clear).

lib/services/api/href_parser.rb
NickLaMuro: Yup, using an instance is fine (elaborated in my response above).  👍

imtayadeway:
@NickLaMuro thanks for feedback! Updated in https://github.com/ManageIQ/manageiq-api/pull/71/commits/d4872fd4f215147ec3b482f4463f3d31f5498d2d

NickLaMuro:
@imtayadeway I was going to suggest you use `Regexp.match?` instead of `=~` since it is a bit of a performance boost, and Rails seems to backport that functionality (see: https://github.com/rails/rails/commit/575dbee), but looks like that is only in Rails 5.1, so nevermind... 😞

chrisarcand:
> but looks like that is only in Rails 5.1

Whoa, TIL. Coming soon!

lib/services/api/href_parser.rb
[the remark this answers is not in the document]
chrisarcand: @NickLaMuro FWIW this is a really common pattern that I preach. ([example](https://github.com/ManageIQ/manageiq/pull/16023#pullrequestreview-64897892)) Looks great to me.

lib/services/api/href_parser.rb
```diff
@@ -0,0 +1,71 @@
...
+    def subcollection
+      path_parts[version? ? 5 : 4]
+    end
+
+    def subcollection_id
+      path_parts[version? ? 6 : 5]
+    end
+
+    def version?
+      @version ||= !!(ApiConfig.version.regex =~ path_parts[2])
```
chrisarcand: I know there's another PR that might change this, so really just waiting on that.

imtayadeway: @chrisarcand rebased/updated 😁

chrisarcand:
Actually, wanna add some yardoc to this to clarify what you get back from `parse` (similar to the doc you removed)? You'd make @durandom proud.

chrisarcand:
After discussion with @imtayadeway, he can add docs in a followup PR, as the parse method itself will probably disappear in favor of more specific methods to extract info from said href. Out of scope of this specific change, so :shipit:

PR #72, opened 2017-09-23
Remove sqlfilter warning

imtayadeway:
Support for the sqlfilter parameter was removed in 2014 in version
1.1 (see
http://manageiq.org/docs/reference/latest/api/overview/versioning). Now
on version 3.0.0-pre, it seems appropriate to silently ignore this
instead of returning a 400.

@miq-bot assign @abellotti


abellotti:
LGTM!!  👍

PR #44, opened 2017-09-05
Allow current group update 

jntullo:
This PR allows editing a users current group via:

```
{
   "action": "edit", 
  "current_group": { "href": "api/groups/:compressed_id" }
}
```

and also allows for editing of all `miq_groups` via edit. By passing an array of groups, the new `miq_groups` will be set.

```
{
   "action": "edit",
   "miq_groups": [ { "href": "/api/groups/:compressed_id" }. { "id": ":compressed_id" } ]
}
```

@miq-bot add_label enhancement
cc: @AllenBW

PR #73, opened 2017-09-25
Allow update of miq_groups by authenticated User

jntullo:
Although we allowed the authenticated user to update their current group in a [previous PR](https://github.com/ManageIQ/manageiq-api/pull/44), the `miq_groups` are still not editable by the current user, which is necessary for updating the current group.

#### Example of editing miq_groups and current_group 
```
{
  'action'    : 'edit',
  'resources' : [{
                    'href'          : '/api/groups/:id'
                    'miq_groups'    : [
                      { 'id' : '<id>' }, 
                      { 'href' : '/api/groups/:id' },
                    ],
                    'current_group' : {
                      'href' : '/api/groups/:id'
                    }
                  }]
}
```

cc: @AllenBW 
@miq-bot assign @abellotti

### DIFF
--- a/app/controllers/api/users_controller.rb
+++ b/app/controllers/api/users_controller.rb
@@ -2,7 +2,7 @@ module Api
   class UsersController < BaseController
     INVALID_USER_ATTRS = %w(id href current_group_id settings).freeze # Cannot update other people's settings
     INVALID_SELF_USER_ATTRS = %w(id href current_group_id).freeze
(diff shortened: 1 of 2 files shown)


abellotti:
I'm a bit confused here, users should not be able to edit their list of groups if they are not entitled as such (e.g. super admins), but always be able to edit their current group regardless.

jntullo:
closing, change is not needed.

PR #75, opened 2017-09-25
Fix api.yml trailing whitespace

imtayadeway:
Because I *will* commit this in a totally unrelated revision by
mistake if I don't do this here.

:heart: :rainbow: :lollipop:

@miq-bot assign @abellotti

### DIFF
--- a/config/api.yml
+++ b/config/api.yml
@@ -1099,7 +1099,7 @@
     :options:
     - :collection
     :subcollections:
-    - :custom_attributes 
+    - :custom_attributes
     - :load_balancers
     - :snapshots
     :collection_actions:


abellotti:
I donno, so much to review in this one PR 😄 

will merge when 🍏

imtayadeway:
@abellotti makes quite a change, no? 😉

chrisarcand:
@imtayadeway Before we merge this, I'd ask you to please first consider what exactly whitespace is.

In computer programming, white space is any character or series of characters that represent horizontal or vertical space in typography. When rendered, a whitespace character does not correspond to a visible mark, but typically does occupy an area on a page. For example, the common whitespace symbol U+0020   SPACE (HTML &#32;), also ASCII 32, represents a blank space punctuation character in text, used as a word divider in Western scripts.

With many keyboard layouts, a horizontal whitespace character may be entered through the use of a spacebar. Horizontal white space may also be entered on many keyboards through the use of the Tab ↹ key, although the length of the space may vary. Vertical white space is a bit more varied as to how it is encoded, but the most obvious in typing is the ↵ Enter result which creates a 'newline' code sequence in applications programs. Older keyboards might instead say Return, abbreviating the typewriter keyboard meaning 'Carriage-Return' which generated an electromechanical return to the left stop (CR code in ASCII-hex &0D;) and a line feed or move to the next line (LF code in ASCII-hex &0A;); in some applications these were independently used to draw text cell based displays on monitors or for printing on tractor-guided printers—which might also contain reverse motions/positioning code sequences allowing yesterdays text base fancier displays. Many early computer games used such codes to draw a screen (e.g. Kingdom of Kroz).

The term "white space" is based on the resulting appearance on ordinary paper. However they are coded inside an application, white space can be processed the same as any other character code and programs can do the proper action as defined for the context in which they occur.

The most common whitespace characters may be typed via the space bar or the tab key. Depending on context, a line-break generated by the return or enter key may be considered white space as well.

In programming language syntax, spaces are frequently used to explicitly separate tokens. Runs of whitespace characters (beyond the first) occurring within source code written in computer programming languages (outside of strings and other quoted regions) are ignored by most languages; such languages are called free-form. In a few languages, including Haskell, occam, ABC, and Python, white space and indentation are used for syntactical purposes. In the satirical language called Whitespace, whitespace characters are the only valid characters for programming, while any other characters are ignored.

Still, for most programming languages, excessive use of white space, especially trailing white space at the end of lines, is considered a nuisance.[by whom?] However correct use of white space can make the code easier to read and help group related logic.

The C language defines white space characters to be "space, horizontal tab, new-line, vertical tab, and form-feed". The HTTP network protocol requires different types of white space to be used in different parts of the protocol, such as: only the space character in the status line, CRLF at the end of a line, and "linear white space" in header values. If you still happen to be reading this I'm super impressed, :trollface: :shipit:

abellotti:
Thank you @chrisarcand for the feedback, I think I can now in good conscience merge this 😏

PR #76, opened 2017-09-26
Alert Definitions: add hash_expression

zakiva:
Depends on: https://github.com/ManageIQ/manageiq/pull/16397

Changes included: 

* Currently, it is possible to create an alert definition via API only with an expression as MiqExpression. Following the changes in https://github.com/ManageIQ/manageiq/pull/15315, this PR adds the ability to create an alert with hash expression. 

* Backward Compatibility: we would like to keep the option to create an alert with expression as MiqExpression without explicitly passing a `miq_expression` attribute. So the user will have now 3 options when sending a create request and will need to pass exactly one of the following:
1. `expression` which will be saved as MiqExpression (same as before).
2. `miq_expression`.
3. `hash_expression`.  

* Allowing editing an alert definition (with similar logic). 

* Deep symbolize for data. 

(This PR replaces https://github.com/ManageIQ/manageiq/pull/14979 with different implementation)

### DIFF
--- a/app/controllers/api/alert_definitions_controller.rb
+++ b/app/controllers/api/alert_definitions_controller.rb
@@ -1,16 +1,17 @@
 module Api
   class AlertDefinitionsController < BaseController
-    REQUIRED_FIELDS = %w(description db expression options).freeze
-
     before_action :set_additional_attributes
 
     def create_resource(type, id, data = {})
       assert_id_not_specified(data, type)
-      assert_all_required_fields_exists(data, type, REQUIRED_FIELDS)
       begin
-        data["expression"] = MiqExpression.new(data["expression"])
-        data["enabled"] = true if data["enabled"].nil?
-        super(type, id, data).serializable_hash.merge("expression" => data["expression"])
+        update_miq_expression(data) if data["expression"]
+        alert = if data["hash_expression"]
+                  super(type, id, data.deep_symbolize_keys).serializable_hash
+                else
+                  super(type, id, data).serializable_hash
+                end
+        alert.merge("expression" => alert["miq_expression"] || alert["hash_expression"])
       rescue => err
         raise BadRequestError, "Failed to create a new alert definition - #{err}"
       end
@@ -19,8 +20,12 @@ module Api
     def edit_resource(type, id = nil, data = {})
       raise BadRequestError, "Must specify an id for editing a #{type} resource" unless id
       begin
-        data["expression"] = MiqExpression.new(data["expression"]) if data["expression"]
-        super(type, id, data)
+        update_miq_expression(data) if data["expression"]
+        if data["hash_expression"]
+          super(type, id, data.deep_symbolize_keys)
+        else
+          super(type, id, data)
+        end
       rescue => err
         raise BadRequestError, "Failed to update alert definition - #{err}"
       end
@@ -30,6 +35,11 @@ module Api
 
     def set_additional_attributes
       @additional_attributes = %w(expression)
+    end
+
+    def update_miq_expression(data)
+      data["miq_expression"] = data["expression"]
+      data.delete("expression")
     end
   end
 end
--- a/spec/requests/alert_definitions_spec.rb
+++ b/spec/requests/alert_definitions_spec.rb
@@ -71,7 +71,7 @@ describe "Alerts Definitions API" do
     sample_alert_definition = {
       "description" => "Test Alert Definition",
       "db"          => "ContainerNode",
-      "expression"  => { "eval_method" => "dwh_generic", "mode" => "internal", "options" => {} },
+      "expression"  => {"exp" => {"=" => {"field" => "Vm-name", "value" => "foo"}}, "context_type" => nil},
       "options"     => { "notifications" => {"delay_next_evaluation" => 600, "evm_event" => {} } },
       "enabled"     => true
     }
@@ -89,6 +89,66 @@ describe "Alerts Definitions API" do
         "exp" => sample_alert_definition["expression"]
       )
     )
+  end
+
+  it "creates an alert definition with miq_expression" do
+    sample_alert_definition = {
+      "description"    => "Test Alert Definition",
+      "db"             => "ContainerNode",
+      "miq_expression" => {"exp" => {"=" => {"field" => "Vm-name", "value" => "foo"}}, "context_type" => nil},
+      "options"        => { "notifications" => {"delay_next_evaluation" => 600, "evm_event" => {} } },
+      "enabled"        => true
+    }
+    api_basic_authorize collection_action_identifier(:alert_definitions, :create)
+    post(api_alert_definitions_url, :params => sample_alert_definition)
+    expect(response).to have_http_status(:ok)
+    alert_definition = MiqAlert.find(response.parsed_body["results"].first["id"])
+    expect(alert_definition).to be_truthy
+    expect(alert_definition.expression.class).to eq(MiqExpression)
+    expect(alert_definition.expression.exp).to eq(sample_alert_definition["miq_expression"])
+    expect(response.parsed_body["results"].first).to include(
+      "description" => sample_alert_definition["description"],
+      "db"          => sample_alert_definition["db"],
+      "expression"  => a_hash_including(
+        "exp" => sample_alert_definition["miq_expression"]
+      )
+    )
+  end
+
+  it "creates an alert definition with hash_expression" do
+    sample_alert_definition = {
+      "description"     => "Test Alert Definition",
+      "db"              => "ContainerNode",
+      "hash_expression" => { "eval_method" => "dwh_generic", "mode" => "internal", "options" => {} },
+      "options"         => { "notifications" => {"delay_next_evaluation" => 0, "evm_event" => {} } },
+      "enabled"         => true
+    }
+    api_basic_authorize collection_action_identifier(:alert_definitions, :create)
+    post(api_alert_definitions_url, :params => sample_alert_definition)
+    expect(response).to have_http_status(:ok)
+    alert_definition = MiqAlert.find(response.parsed_body["results"].first["id"])
+    expect(alert_definition).to be_truthy
+    expect(alert_definition.expression.class).to eq(Hash)
+    expect(alert_definition.expression).to eq(sample_alert_definition["hash_expression"].deep_symbolize_keys)
+    expect(response.parsed_body["results"].first).to include(
+      "description" => sample_alert_definition["description"],
+      "db"          => sample_alert_definition["db"],
+      "expression"  => sample_alert_definition["hash_expression"]
+    )
+  end
+
+  it "fails to create an alert definition with more than one expression" do
+    sample_alert_definition = {
+      "description"     => "Test Alert Definition",
+      "db"              => "ContainerNode",
+      "hash_expression" => { "eval_method" => "nothing", "mode" => "internal", "options" => {} },
+      "miq_expression"  => { "exp" => {"=" => {"field" => "Vm-name", "value" => "foo"}}, "context_type" => nil },
+      "options"         => { "notifications" => {"delay_next_evaluation" => 600, "evm_event" => {} } },
+      "enabled"         => true
+    }
+    api_basic_authorize collection_action_identifier(:alert_definitions, :create)
+    post(api_alert_definitions_url, :params => sample_alert_definition)
+    expect(response).to have_http_status(:bad_request)
   end
 
   it "deletes an alert definition via POST" do
@@ -122,7 +182,7 @@ describe "Alerts Definitions API" do
     api_basic_authorize(action_identifier(:alert_definitions, :edit, :resource_actions, :post))
     alert_definition = FactoryGirl.create(
       :miq_alert,
-      :expression => { :eval_method => "mw_heap_used", :mode => "internal", :options => {} },
+      :expression => { "exp" => {"=" => {"field" => "Vm-name", "value" => "foo"}}},
       :options    => { :notifications => {:delay_next_evaluation => 0, :evm_event => {} } }
     )
 
@@ -140,11 +200,7 @@ describe "Alerts Definitions API" do
     )
 
     expected = {
-      "expression" => {
-        "eval_method" => "mw_heap_used",
-        "mode"        => "internal",
-        "options"     => {}
-      },
+      "expression" => alert_definition.expression,
       "options"    => {
         "notifications" => {
           "delay_next_evaluation" => 60,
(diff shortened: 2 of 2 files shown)


zakiva:

zakiva:
@imtayadeway @moolitayer Please review
cc @simon3z @ilackarms

app/controllers/api/alert_definitions_controller.rb
```diff
@@ -1,25 +1,39 @@
...
     end
 
     def edit_resource(type, id = nil, data = {})
       raise BadRequestError, "Must specify an id for editing a #{type} resource" unless id
+      exp_values = expression_values(data)
+      if exp_values.size > 1
+        raise BadRequestError, "Must not specify more than one expression for editing a #{type} resource"
+      elsif exp_values.size == 1
+        if data["hash_expression"]
+          data["miq_expression"] = nil
```
moolitayer: So as we discussed with @zakiva this handles an edge case where a user wants to replace a hash expression with a miq expression or vice versa. Honestly I don't know if we should support that

moolitayer:
@imtayadeway @abellotti We need an API review here.

ilackarms:
@zakiva can you provide an example of a `curl` / HTTP request to add an alert definition?

zakiva:
> @zakiva can you provide an example of a curl / HTTP request to add an alert definition?

`curl -X POST -H "Content-Type: application/json" -d "@alert.json" http://localhost:3000/api/alert_definitions --user admin:smartvm`

example of `alert.json` file: 

```
{
   "action" : "create",
   "resource" : {
   	"description" : "Alert Description",
   	"db": "ContainerNode",
   	"miq_expression": {
   		"mode": "internal",
   		"options": {
   			},
   		"eval_method": "hostd_log_threshold"
   	},
   	"options": {
   		"notifications": {
   			"email": {
   				"from": "",
                                "to": []
            }
          }
        }
   }
}
```

zakiva:
ping @imtayadeway Can you please review?

zakiva:
ping @abellotti @imtayadeway

app/controllers/api/alert_definitions_controller.rb
```diff
@@ -1,26 +1,39 @@
...
   class AlertDefinitionsController < BaseController
-    REQUIRED_FIELDS = %w(description db expression options).freeze
+    REQUIRED_FIELDS = %w(description db options).freeze
 
     before_action :set_additional_attributes
 
     def create_resource(type, id, data = {})
       assert_id_not_specified(data, type)
       assert_all_required_fields_exists(data, type, REQUIRED_FIELDS)
+      if expression_values(data).size != 1
```
imtayadeway: This looks like business logic to me. What's more - it's duplicated on line 26. This should be added to the model as a validation, which will greatly simplify this code. The code here should be almost entirely concerned with deserializing the JSON request body and passing data to the model.

app/controllers/api/alert_definitions_controller.rb
```diff
@@ -1,26 +1,39 @@
 module Api
   class AlertDefinitionsController < BaseController
-    REQUIRED_FIELDS = %w(description db expression options).freeze
+    REQUIRED_FIELDS = %w(description db options).freeze
```
imtayadeway: I know that you're only changing this part - but I'm also confused as to why these "required fields" aren't expressed in the model too.

zakiva: Agreed, I can remove them in this PR and then add the validations in the model.

app/controllers/api/alert_definitions_controller.rb
zakiva: removed that, will add a validation in the model later on.

zakiva:
@imtayadeway PTAL

app/controllers/api/alert_definitions_controller.rb
```diff
@@ -1,26 +1,31 @@
 module Api
   class AlertDefinitionsController < BaseController
-    REQUIRED_FIELDS = %w(description db expression options).freeze
```
imtayadeway: You can nix this empty line too 😁

app/controllers/api/alert_definitions_controller.rb
```diff
@@ -1,26 +1,31 @@
...
 
     before_action :set_additional_attributes
 
     def create_resource(type, id, data = {})
       assert_id_not_specified(data, type)
-      assert_all_required_fields_exists(data, type, REQUIRED_FIELDS)
       begin
-        data["expression"] = MiqExpression.new(data["expression"])
+        create_miq_expression(data)
         data["enabled"] = true if data["enabled"].nil?
```
imtayadeway: Again, I know you're not touching this here (and feel free to leave for a follow up if appropriate) but this is more business logic - "an alert definition is enabled by default" - that belongs in the model

app/controllers/api/alert_definitions_controller.rb
```diff
@@ -31,5 +36,14 @@ def edit_resource(type, id = nil, data = {})
     def set_additional_attributes
       @additional_attributes = %w(expression)
     end
+
+    def expression_values(data)
+      data.slice("expression", "miq_expression", "hash_expression").values.compact
+    end
+
+    def create_miq_expression(data)
+      data["expression"] = MiqExpression.new(data["expression"]) if data["expression"]
```
imtayadeway: I'm not sure I understand why this is necessary. Type coercion should be handled by the setter in the model - if that's not currently the case then that needs to be added so we don't have to munge the data here (and anywhere else for that matter).

app/controllers/api/alert_definitions_controller.rb
```diff
@@ -1,26 +1,31 @@
...
-        super(type, id, data).serializable_hash.merge("expression" => data["expression"])
+        super(type, id, data.deep_symbolize_keys).serializable_hash.merge("expression" => expression_values(data).first)
       rescue => err
         raise BadRequestError, "Failed to create a new alert definition - #{err}"
       end
     end
 
     def edit_resource(type, id = nil, data = {})
       raise BadRequestError, "Must specify an id for editing a #{type} resource" unless id
+      if expression_values(data).size == 1
```
imtayadeway: Is this still needed? Again, this is another business rule that needs to be encapsulated in the model.

app/controllers/api/alert_definitions_controller.rb
```diff
@@ -1,26 +1,31 @@
...
 
     def create_resource(type, id, data = {})
       assert_id_not_specified(data, type)
-      assert_all_required_fields_exists(data, type, REQUIRED_FIELDS)
       begin
-        data["expression"] = MiqExpression.new(data["expression"])
+        create_miq_expression(data)
         data["enabled"] = true if data["enabled"].nil?
-        super(type, id, data).serializable_hash.merge("expression" => data["expression"])
+        super(type, id, data.deep_symbolize_keys).serializable_hash.merge("expression" => expression_values(data).first)
```
imtayadeway: Why is the `deep_symbolize_keys` needed? If it's for a particular serialized column that is keyed with symbols, I think it would be better to make that explicit rather than converting everything wholesale

app/controllers/api/alert_definitions_controller.rb
zakiva: done :-)

app/controllers/api/alert_definitions_controller.rb
zakiva: Right, removing that too. This should be handled explicitly by the user.
cc @moolitayer @elad661

app/controllers/api/alert_definitions_controller.rb
zakiva: We want the created alerts via API to be identical to alerts created in other places. 
It is not a particular (single) column, so I think it will be nicer not to list them.

app/controllers/api/alert_definitions_controller.rb
zakiva: OK, moved it to the model: https://github.com/ManageIQ/manageiq/pull/16397

app/controllers/api/alert_definitions_controller.rb
zakiva: I will remove that too.
The intention here was to handle the case of replacing a miq expression with hash expression (and vice versa) when editing an alert, but we can leave that to the user to explicitly nullify these fields. 
cc @moolitayer @elad661

zakiva:
@imtayadeway Thanks for the review! I made the requested changes.

elad661:
@imtayadeway now that the change in the model ( https://github.com/ManageIQ/manageiq/pull/16397 ) is merged, can we get this PR merged too, or are there any other required changes?

bazulay:
@imtayadeway @abellotti  PTAL

elad661:
@imtayadeway Can you please take a look? This PR has been waiting for a re-review for almost a month now...

bazulay:
@gtanzillo  ping

app/controllers/api/alert_definitions_controller.rb
```diff
@@ -31,5 +28,10 @@ def edit_resource(type, id = nil, data = {})
     def set_additional_attributes
       @additional_attributes = %w(expression)
     end
+
+    def update_miq_expression(data)
+      data["miq_expression"] = data["expression"]
```
imtayadeway: I'm still not sure why this is needed. Doesn't the model support this via both `#expression=` and `#miq_expression=` ?

app/controllers/api/alert_definitions_controller.rb
```diff
@@ -1,16 +1,13 @@
...
 
     def create_resource(type, id, data = {})
       assert_id_not_specified(data, type)
-      assert_all_required_fields_exists(data, type, REQUIRED_FIELDS)
       begin
-        data["expression"] = MiqExpression.new(data["expression"])
-        data["enabled"] = true if data["enabled"].nil?
-        super(type, id, data).serializable_hash.merge("expression" => data["expression"])
+        update_miq_expression(data) if data["expression"]
+        alert = super(type, id, data.deep_symbolize_keys).serializable_hash
```
imtayadeway: I'm concerned that this `deep_symbolize_keys` will corrupt an incoming `MiqExpression`:

```ruby
{"=" => {"field" => "foo", "value" => "bar"}}.deep_symbolize_keys # => {:"=" => {:field => "foo", :value => "bar"}}
```

spec/requests/alert_definitions_spec.rb
```diff
@@ -81,7 +81,7 @@
     alert_definition = MiqAlert.find(response.parsed_body["results"].first["id"])
     expect(alert_definition).to be_truthy
     expect(alert_definition.expression.class).to eq(MiqExpression)
-    expect(alert_definition.expression.exp).to eq(sample_alert_definition["expression"])
+    expect(alert_definition.expression.exp).to eq(sample_alert_definition["expression"].deep_symbolize_keys)
```
imtayadeway: This looks wrong, because miq expressions are not keyed with symbols (see above)

spec/requests/alert_definitions_spec.rb
```diff
@@ -91,6 +91,66 @@
...
+      "options"        => { "notifications" => {"delay_next_evaluation" => 600, "evm_event" => {} } },
+      "enabled"        => true
+    }
+    api_basic_authorize collection_action_identifier(:alert_definitions, :create)
+    post(api_alert_definitions_url, :params => sample_alert_definition)
+    expect(response).to have_http_status(:ok)
+    alert_definition = MiqAlert.find(response.parsed_body["results"].first["id"])
+    expect(alert_definition).to be_truthy
+    expect(alert_definition.expression.class).to eq(MiqExpression)
+    expect(alert_definition.expression.exp).to eq(sample_alert_definition["miq_expression"].deep_symbolize_keys)
```
imtayadeway: As above

spec/requests/alert_definitions_spec.rb
```diff
@@ -91,6 +91,66 @@
     )
   end
 
+  it "creates an alert definition with miq_expression" do
+    sample_alert_definition = {
+      "description"    => "Test Alert Definition",
+      "db"             => "ContainerNode",
+      "miq_expression" => { "eval_method" => "dwh_generic", "mode" => "internal", "options" => {} },
```
imtayadeway: Is this supposed to be a hash expression standing in place of an miq expression? Even if this "works", I wouldn't want us to officially support this surprising behavior

spec/requests/alert_definitions_spec.rb
```diff
@@ -91,6 +91,66 @@
...
+      "expression"  => sample_alert_definition["hash_expression"]
+    )
+  end
+
+  it "fails to create an alert definition with more than one expression" do
+    sample_alert_definition = {
+      "description"     => "Test Alert Definition",
+      "db"              => "ContainerNode",
+      "hash_expression" => { "eval_method" => "nothing", "mode" => "internal", "options" => {} },
+      "miq_expression"  => { "eval_method" => "dwh_generic", "mode" => "internal", "options" => {} },
```
imtayadeway: As above

spec/requests/alert_definitions_spec.rb
```diff
@@ -156,6 +216,77 @@
...
+      :options         => { :notifications => {:delay_next_evaluation => 0, :evm_event => {} } }
+    )
+
+    exp = { :eval_method => "mw_heap_used", :mode => "internal", :options => {} }
+
+    post(
+      api_alert_definition_url(nil, alert_definition),
+      :params => {
+        :action          => "edit",
+        :miq_expression  => exp,
```
imtayadeway: As above

spec/requests/alert_definitions_spec.rb
```diff
@@ -156,6 +216,77 @@
...
+      :hash_expression => { :eval_method => "nothing", :mode => "internal", :options => {} },
+      :options         => { :notifications => {:delay_next_evaluation => 0, :evm_event => {} } }
+    )
+
+    post(
+      api_alert_definition_url(nil, alert_definition),
+      :params => {
+        :action          => "edit",
+        :hash_expression => { :eval_method => "event_threshold", :mode => "internal", :options => {} },
+        :miq_expression  => { :eval_method => "mw_heap_used", :mode => "internal", :options => {} }
```
imtayadeway: As above

app/controllers/api/alert_definitions_controller.rb
elad661: This is needed because the setter in the model for `expression=` looks like this: https://github.com/ManageIQ/manageiq/blob/master/app/models/miq_alert.rb#L59-L65
```ruby
    if exp.kind_of?(MiqExpression)
      self.miq_expression = exp
    elsif exp.kind_of?(Hash)
      self.hash_expression = exp
```

And the data we get from the api user is always a "hash" (but not a Hash Expression), because it's parsed json.

The model will only try to create a new MiqExpression from hash data if you explicitly use the `miq_expression` setter, so we need to check if the user provided us an "expression" parameter, and if so *explicitly* say we want a miq_expression and not an hash expression.

Since hash expressions are just ruby hashes, there's no cleaner way to do this - I don't think we'd want the model to attempt parsing hashes to see if they fit the "miq expression" or "hash expression" forms and magically decide which type to use.

elad661:
Thanks for the review. I fixed the tests and changed the controller to only do `.deep_symbolize_keys` on hash expressions (since those do need to be symbolized).

app/controllers/api/alert_definitions_controller.rb
imtayadeway: This is OK for now....can we look at deprecating the old API (i.e. `MiqAlert#expression`) in a follow up PR? Then I think a lot of the complexity of this controller could go away.

spec/requests/alert_definitions_spec.rb
imtayadeway: It looks like this is still a "hash_expression" in place of an "miq_expression"?

elad661: oh, you're right, I didn't actually fix this one when I said it was. I fixed it now.

app/controllers/api/alert_definitions_controller.rb
elad661: I'm fine with deprecating the `MiqAlert#expression`, and I can do a follow-up PR.

@moolitayer is there any reason to keep the old `MiqAlert#expression` API around?

moolitayer: In the model we would need to keep the `expression` due to the alerts screen.
In the API I agree we should deprecate and then drop.
(deprecate in Gaprindashvili and Drop in H?)

spec/requests/alert_definitions_spec.rb
```diff
@@ -71,7 +71,7 @@
     sample_alert_definition = {
       "description" => "Test Alert Definition",
       "db"          => "ContainerNode",
-      "expression"  => { "eval_method" => "dwh_generic", "mode" => "internal", "options" => {} },
+      "expression"  => {"exp" => {"=" => {"field" => "Vm-name", "value" => "foo"}}, "context_type" => nil},
```
abellotti: Just curious about these changes, description mentioned being backward compatible, why are these needed ?

elad661: Because, as @imtayadeway noted in this comment https://github.com/ManageIQ/manageiq-api/pull/76#discussion_r154723513, the expression there was not actually a MiqExpression, so that means the test was wrong. It wasn't supposed to "work" like that.

abellotti: 👍 thanks for clarifying.

abellotti:
Thanks @zakiva for the API Enhancement and everyone for the reviews.  This LGTM!! 👍

PR #78, opened 2017-09-26
Added support for collection specific hide_resources option.

abellotti:
This is needed for the new automate_workspaces collection where
we don't want to return current set of workspaces as they
are only accessible by resource guid from automate.

We're leveraging the current mechanism we have in place for
the hide=resources parameter and supported at the collection
level via the api.yml.

We've contemplated the 404 on the collection but that breaks
the clients and inhibits it from getting collection level
actions and such.


mkanoor:
👍

PR #79, opened 2017-09-26
Fixed a key name in the output field

mkanoor:
We had to rename a key in the hash in the automate workspace. 
https://github.com/ManageIQ/manageiq/pull/15977

Fixed spec


abellotti:
LGTM!! Thanks @mkanoor for updating this.

PR #81, opened 2017-09-26
Add security group subcollection to providers, cloud tenants, and vms

tzumainn:


tzumainn:
Required by https://github.com/ManageIQ/manageiq-ui-classic/pull/1997

app/controllers/api/subcollections/security_groups.rb
```diff
@@ -0,0 +1,9 @@
+module Api
+  module Subcollections
+    module SecurityGroups
+      def security_groups_query_resource(object)
+        object.respond_to?(:security_groups) ? object.security_groups : []
```
imtayadeway: Can you add a test that executes the case when the object does not respond to `:security_groups`?

tzumainn: Done - thanks!

abellotti:
@tzumainn when you get a chance, can you add tests for accessing security_groups subcollection for vms and cloud_tenants, similar to what you did for providers ?  Thanks.

tzumainn:
@abellotti Thanks for taking a look! I've added the tests; they depend on https://github.com/ManageIQ/manageiq/pull/16127 being merged in manageiq, though.

abellotti:
Thanks @tzumainn for the update, I'll look at both.

abellotti:
@tzumainn can you update the Style/Align rubocop issue and repush ?  Thanks.

tzumainn:
@abellotti Done! (I think - I'm a bit uncertain about the frozen string comment, but I guess I'll see).

app/controllers/api/subcollections/security_groups.rb
```diff
@@ -2,7 +2,7 @@ module Api
   module Subcollections
     module SecurityGroups
       def security_groups_query_resource(object)
-        object.respond_to?(:security_groups) ? object.security_groups : []
+        object.respond_to?(:security_groups) ? object.security_groups : [].freeze
```
abellotti: this .freeze is not needed. can you remove it/repush ?  Thanks.  the frozen literal rubocop is unrelated.

tzumainn:
@abellotti Ah, that's good to know. I've removed it!

himdel:
Looks like create & update (& delete) will also be needed for https://github.com/ManageIQ/manageiq-ui-classic/pull/1997 .

Issue #74, opened 2017-09-25
GET api?attributes=authorization should return more than an array of strings for `groups`


AllenBW:
While mildly helpful, the array of strings, better would be an array of full group objects that below to the user when this call is made:`GET http://localhost:3000/api?attributes=authorization`

```
  "identity": {
    "userid": "admin",
    "name": "Administrator",
    "user_href": "http://localhost:3000/api/users/10000000000001",
    "group": "EvmGroup-super_administrator",
    "group_href": "http://localhost:3000/api/groups/10000000000002",
    "role": "EvmRole-super_administrator",
    "role_href": "http://localhost:3000/api/roles/10000000000001",
    "tenant": "Red Hat",
    "groups": [
      "Tenant My Company access",
      "EvmGroup-super_administrator"
    ]
  }
```

@imtayadeway 👾 💅 ‼️

abellotti:
keeping current groups array returned as is (ye-old compatibility) but maybe some additional details separately.  

@AllenBW what info out of groups do you need ?  with all groups details, might be tons of data, we may want to only return that if attributes=authorization, "identity" is always returned but we can extend with an miq_groups element when attributes=authorization

AllenBW:
see i did that! added the `miq_groups` attribute... it didn't do much... all i need/want is an id or href for the groups, of course can be gotten by doing a call on `api/groups` and matching in the client... but that kinda seems crazy to make another call :-/

abellotti:
If this is something you need back with each call with attributes=authorization,

Maybe something like this ?

```
  "identity": {
    "userid": "admin",
    "name": "Administrator",
    "user_href": "http://localhost:3000/api/users/10000000000001",
    "group": "EvmGroup-super_administrator",
    "group_href": "http://localhost:3000/api/groups/10000000000002",
    "role": "EvmRole-super_administrator",
    "role_href": "http://localhost:3000/api/roles/10000000000001",
    "tenant": "Red Hat",
    "groups": [
      "Tenant My Company access",
      "EvmGroup-super_administrator"
    ],
    "miq_groups" : [
        {
           "description" : "Tenant My Company access",
           "href" : "http://localhost:3000/api/groups/1000000000021",
           "id" : "1000000000021"
        },
        {
            "description" : "EvmGroup-super_administrator",
            "href" : "http://localhost:3000/api/groups/1000000000002",
             "id": "1000000000002",
        }
    ]
  }
```

AllenBW:
[...]

PR #84, opened 2017-09-27
Return miq_groups on api entrypoint 

jntullo:
Resolves https://github.com/ManageIQ/manageiq-api/issues/74 

This enhancement returns `miq_groups` on the entrypoint
```
GET /api
{
   "name": "API",
   "description": "REST API",
   "settings": {...},
   "identity": {
      ...
      "groups" [...],
      "miq_groups": [
         { "href": "/api/groups/:id", "id": ":id" ... } ...
      ]
   }
   ...
}
```

@miq-bot assign @abellotti 
cc: @AllenBW

### DIFF
--- a/app/controllers/api/base_controller/normalizer.rb
+++ b/app/controllers/api/base_controller/normalizer.rb
@@ -111,8 +111,8 @@ module Api
         end
       end
 
-      def normalize_array(obj)
-        type = @req.subject
+      def normalize_array(obj, type = nil)
+        type ||= @req.subject
         obj.collect { |item| normalize_attr(get_reftype(type, type, item), item) }
       end
 
--- a/spec/requests/authentication_spec.rb
+++ b/spec/requests/authentication_spec.rb
@@ -104,7 +104,11 @@ describe "Authentication API" do
         "group_href" => "/api/groups/#{group2.id}",
         "role"       => @role.name,
         "role_href"  => "/api/roles/#{group2.miq_user_role.id}",
-        "tenant"     => @group.tenant.name
+        "tenant"     => @group.tenant.name,
+        "groups"     => @user.miq_groups.pluck(:description),
+        "miq_groups" => a_collection_including(
+          hash_including("href" => api_group_url(nil, @user.miq_groups.first))
+        )
       )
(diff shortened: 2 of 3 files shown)


spec/requests/authentication_spec.rb
```diff
@@ -115,7 +115,12 @@
       get api_entrypoint_url, :params => { :attributes => "authorization" }
 
       expect(response).to have_http_status(:ok)
-      expected = {"authorization" => hash_including("product_features")}
+      expected = {
+        "authorization" => hash_including("product_features"),
+        "identity"      => hash_including("miq_groups" => a_collection_including(
+          hash_including("href" => api_users_url(nil, @user.miq_groups.first))
+        ))
+      }
```
abellotti: I think the miq_groups in identity is returned with all calls  (i.e. without attributes=authorization), in which case if you can update the description and move the test here to the regular entrypoint (not authorization).

spec/requests/authentication_spec.rb
```diff
@@ -104,7 +104,10 @@
         "group_href" => "/api/groups/#{group2.id}",
         "role"       => @role.name,
         "role_href"  => "/api/roles/#{group2.miq_user_role.id}",
-        "tenant"     => @group.tenant.name
+        "tenant"     => @group.tenant.name,
+        "miq_groups" => a_collection_including(
```
abellotti: Thanks @jntullo for the update.  Maybe we should also add the groups =>  [ ] here before miq_groups for completion.  It may never have been added to the spec.

app/controllers/api/base_controller/normalizer.rb
```diff
@@ -111,8 +111,8 @@ def normalize_select_attributes(obj, opts)
         end
       end
 
-      def normalize_array(obj)
-        type = @req.subject
+      def normalize_array(obj, type = nil)
+        type ||= @req.subject
```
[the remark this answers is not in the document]
jntullo: @abellotti I used this for rendering the `miq_groups` properly, and since the `/api` endpoint doesn't have the `groups` subject, I pass in the :groups parameter

abellotti:
LGTM!! Thanks @jntullo for the enhancement.

AllenBW:
O frabjous day! Callooh! Callay!

PR #85, opened 2017-09-27
add ability to set custom attributes on services via api

ailisp:
Issue: missing ability to set custom attributes on services via api, which is now fixed

BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1481547

\cc @abellotti @gtanzillo 
@miq-bot add-label wip, bug


abellotti:
api.yml declaration (identifiers for the subcollection actions, see vms/instances/providers for example), also, the specs would be needed, also look at instances for reference.  Thanks!!

ailisp:
@abellotti Hi, should we do similar spec instances_spec.rb and services_spec.rb for custom attributes. Or, since they're almost the same, test for custom_attributes mixin.

ailisp:
@miq-bot remove-label wip

abellotti:
Thanks @ailisp for the API Enhancement !! 👍 

The Hakiri warning above is unrelated and was taken care in a separate PR so I'll go ahead and merge.

PR #86, opened 2017-09-28
Return correct automate_domains and policy_actions href_slugs

jntullo:
Currently, the automate domain href_slugs return `automate/:id` because the Api config finds the first matching class. Similarly, policy_actions returns `actions/:id`. For now, it seems the best way to resolve this issue is to override the default href_slug virtual attribute itself.

BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1485310

@miq-bot add_label bug
@miq-bot assign @abellotti

### DIFF
--- a/app/controllers/api/automate_domains_controller.rb
+++ b/app/controllers/api/automate_domains_controller.rb
@@ -27,6 +27,10 @@ module Api
       end
     end
 
+    def fetch_automate_domains_href_slug(resource)
+      "automate_domains/#{resource.id}"
+    end
+
     private
 
     def automate_domain_ident(domain)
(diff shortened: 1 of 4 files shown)


app/controllers/api/automate_domains_controller.rb
```diff
@@ -27,6 +27,10 @@ def refresh_from_source_resource(type, id = nil, data = nil)
       end
     end
 
+    def fetch_automate_domains_href_slug(resource)
```
imtayadeway: Does this mean this method gets invoked via some meta programming? Seems like this could be better implemented using polymorphism. But definitely not as part of this PR ;)

abellotti:
LGTM!! 👍

PR #88, opened 2017-09-29
DRY up version regex

imtayadeway:
We currently have two things claiming to be the version regex -
`Api::VERSION_REGEX` and `ApiConfig.version.regex`. It seems sensible
to remove one of them. Somewhat alarmingly they are not exactly the
same, but the tests do pass. If it turns out we do need two different
ones, we should probably:

* put them in the same place
* use better naming to convey intent
* add tests that cover the unique behavior

@miq-bot assign @abellotti

### DIFF
--- a/app/controllers/api/base_controller/parser.rb
+++ b/app/controllers/api/base_controller/parser.rb
@@ -64,7 +64,7 @@ module Api
 
       def href_collection_id(path)
         path_array = path.split('/')
-        cidx = path_array[2] && path_array[2].match(ApiConfig.version.regex) ? 3 : 2
+        cidx = path_array[2] && path_array[2].match(Api::VERSION_REGEX) ? 3 : 2
 
         collection, c_id    = path_array[cidx..cidx + 1]
         subcollection, s_id = path_array[cidx + 2..cidx + 3]
--- a/app/controllers/api/base_controller/parser/request_adapter.rb
+++ b/app/controllers/api/base_controller/parser/request_adapter.rb
@@ -123,7 +123,7 @@ module Api
         end
 
         def version_override?
-          @params[:version] && @params[:version].match(ApiConfig.version[:regex]) # v#.# version signature
+          @params[:version] && @params[:version].match(Api::VERSION_REGEX) # v#.# version signature
         end
 
         def fullpath
--- a/config/api.yml
+++ b/config/api.yml
@@ -5,7 +5,6 @@
   :description: REST API
   :version: '3.0.0-pre'
 :version:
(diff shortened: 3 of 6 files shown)


abellotti:
I _think_ the one in the api was more accurate wrt semver.  Can't we ditch both and use something like https://github.com/canton7/sem_version ?  This would allow us to do version comparison and such if we need to differentiate in code based on /api/v#

imtayadeway:
@abellotti can you provide an example showing that the one I selected is not accurate enough? If so, I'd prefer to add a test for that as part of this PR.

IMO we could use a gem to do this, but I think this is a) beyond the scope of this PR, and b) possibly overkill right now, since our present needs do not extend beyond detecting whether we have a version string or not

Fryguy:
For those who don't want to search:

| source | regex |
| --- | --- |
| version | `/v[\d]+(\.[\da-zA-Z]+)*(\-[\da-zA-Z]+)?/` |
| config | `/^v[0-9]+(?>\.[0-9a-zA-Z]+)*(-[0-9A-Za-z-]+(\.[0-9A-Za-z-]+)*)?/`

Fryguy:
I agree with @imtayadeway that we should just stick to one regex unless there are specific reasons for having a difference.  I don't really understand the point of a "configurable" regex.

One major difference is the leading `^`, which implies it's anchored to the start of a String and this might be important as the changes here are using them for route validation.  That is, someone could tack in extra data to the front of the regex string, and don't know what havoc that would entail (incidentally, we *should* be using `\A` and `\Z` over `^` and `?` because of multiline concerns, so that should be changed regardless).

The other major difference is the trailing `-` in the config version (e.g. `[0-9A-Za-z-]`).  That dash is no in the version one.  To me this shows that the config one supports string based suffixes with dashes inside of them (e.g. `3.0.0-pre-1`), but the version one does not, however both support non-dash strings in the suffix (e.g. `3.0.0-pre1`)

I think there are also differences in "how much" of the version can be asked for.  For example `v1`, `v1.1`, or `v1.1.1`.  I *think* they both support those combinations, but it's hard to tell...would need tests to verify.

Also, very minor, but `[\d]+` should be represented instead as `\d+`, since there's no "choice" there.

imtayadeway:
@Fryguy I think I took care of all your points above (very helpful BTW). I added tests for all those use cases *and* I discovered why there were two regexes - you can't have anchors in routing constraints. So we are now back to having two, but they are both constants and in the same place, and one is defined in terms of the other.

Fryguy:
> you can't have anchors in routing constraints

O_O WAT

imtayadeway:
@Fryguy relevant section: http://guides.rubyonrails.org/routing.html#segment-constraints

PR #89, opened 2017-09-29
Use HTTP status to indicated that cids have been deprecated

imtayadeway:
Using the HTTP status is the Right Way :tm: to show that a resource
has been moved (which, to the outside world, it has - it has a
different id). With this change we can remove some of the handling
done in the request parser that checks for compressed ids. We still
need some of the custom handling, however, since we still support
compressed ids in hrefs in the request body, which cannot use the same
strategy.

Thanks @chrisarcand for reminding me that HTTP exists 😁 

@miq-bot assign @abellotti

### DIFF
--- a/app/controllers/api/base_controller.rb
+++ b/app/controllers/api/base_controller.rb
@@ -25,6 +25,7 @@ module Api
                   :validate_api_request
     before_action :validate_api_action, :except => [:options]
     before_action :validate_response_format, :except => [:destroy]
+    before_action :redirect_on_compressed_path
     after_action :log_api_response
 
     respond_to :json
@@ -45,6 +46,13 @@ module Api
     rescue_from(ArgumentError)                  { |e| api_error(:bad_request, e) }
 
     private
+
+    def redirect_on_compressed_path
+      return unless [params[:c_id], params[:s_id]].any? { |id| Api.compressed_id?(id) }
+      url = request.original_url.sub(params[:c_id], Api.uncompress_id(params[:c_id]).to_s)
+      url.sub!(params[:s_id], Api.uncompress_id(params[:s_id]).to_s) if params[:s_id]
+      redirect_to(url, :status => :moved_permanently)
+    end
 
     def set_gettext_locale
       FastGettext.set_locale(LocaleResolver.resolve(User.current_user, headers))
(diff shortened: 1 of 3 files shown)


spec/requests/queries_spec.rb
```diff
@@ -62,12 +62,13 @@ def create_vms(count)
       expect_single_resource_query("id" => vm1.id.to_s, "href" => api_vm_url(nil, vm1), "guid" => vm1.guid)
     end
 
-    it 'supports compressed ids' do
+    specify 'support for compressed ids has been deprecated' do
```
chrisarcand: I'm *super* impressed you changed it from `it` to `specify`/`example` 😍

app/controllers/api/base_controller.rb
```diff
@@ -46,6 +47,13 @@ class BaseController < ActionController::API
 
     private
 
+    def redirect_on_compressed_path
+      return unless [params[:c_id], params[:s_id]].any? { |id| Api.compressed_id?(id) }
+      url = request.original_url.sub(params[:c_id], Api.uncompress_id(params[:c_id]).to_s)
+      url.sub!(params[:s_id], Api.uncompress_id(params[:s_id]).to_s) if params[:s_id]
```
chrisarcand: ? What's `sub`/`sub!`?

imtayadeway: Same as `gsub`/`gsub` but without the g - will substitute the first found match, and quit: https://ruby-doc.org/core-2.4.2/String.html#method-i-sub

chrisarcand: ...I think my tags are broken again. 🙄 Thanks

spec/requests/queries_spec.rb
Fryguy: Ugh *ANOTHER* rspec syntax...please no :(

abellotti:
Is the Hakiri security warning a false positive here or a real issue ?

imtayadeway:
@abellotti AFAICT it's a false positive. It seems to be tripped on all redirects - so needs to be marked as false positive. Here's the description: 

> Sometimes redirect_to can be used with a user-supplied value that may allow the attacker to change the :host option and load a malicious script from a third party website.

spec/requests/queries_spec.rb
chrisarcand: TBH I actually didn't know that there was an alias as `specify` either lol. I usually use `example`...

Fryguy:
We should get those hakiri warnings fixed, but definitely in a separate PR

PR #90, opened 2017-09-29
Validate limit if offset is specified

jntullo:
Currently if offset is specified and limit is set to 0 (ie `limit=0&offset=5`), the following error is returned due to division by 0:
```
{
    "error": {
        "message": "Infinity",
        "klass": "FloatDomainError"
    }
}
```

This updates the `LinksBuilder` to validate that the `limit` is greater than 0 if offset is specified.

BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1489885

@miq-bot add_label bug
@miq-bot assign @abellotti

### DIFF
--- a/lib/api/links_builder.rb
+++ b/lib/api/links_builder.rb
@@ -5,6 +5,7 @@ module Api
       @limit = params["limit"].to_i if params["limit"]
       @href = href
       @counts = counts if counts
+      validate_limit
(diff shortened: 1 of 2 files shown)


abellotti:
LGTM!! Thanks @jntullo for the fix.  will merge when 🍏

PR #97, opened 2017-10-03
DRY up RequestParser and HrefParser logic

imtayadeway:
A lot of the same work was duplicated across `RequestParser` and `HrefParser`. By turning the `HrefParser` service into more of an object (and expanding its API by exposing some of its private methods) it was possible to delegate many things in the `RequestParser` to it, resulting in many ✂️ 

Builds on the work done in #71 

/cc @chrisarcand 

@miq-bot assign @abellotti

### DIFF
--- a/app/controllers/api/authentications_controller.rb
+++ b/app/controllers/api/authentications_controller.rb
@@ -55,9 +55,9 @@ module Api
     def validate_auth_attrs(data)
       raise 'must supply a manager resource' unless data['manager_resource']
       attrs = data.dup.except('manager_resource')
-      manager_collection, manager_id = HrefParser.parse(data['manager_resource']['href'])
-      raise 'invalid manger_resource href specified' unless manager_collection && manager_id
-      manager_resource = resource_search(manager_id, manager_collection, collection_class(manager_collection))
+      href = HrefParser.new(data['manager_resource']['href'])
+      raise 'invalid manger_resource href specified' unless href.subject && href.subject_id
+      manager_resource = resource_search(href.subject_id, href.subject, collection_class(href.subject))
       [manager_resource, attrs]
     end
   end
--- a/app/controllers/api/authentications_controller.rb
+++ b/app/controllers/api/authentications_controller.rb
@@ -55,7 +55,7 @@ module Api
     def validate_auth_attrs(data)
       raise 'must supply a manager resource' unless data['manager_resource']
       attrs = data.dup.except('manager_resource')
-      href = HrefParser.new(data['manager_resource']['href'])
+      href = Href.new(data['manager_resource']['href'])
       raise 'invalid manger_resource href specified' unless href.subject && href.subject_id
       manager_resource = resource_search(href.subject_id, href.subject, collection_class(href.subject))
       [manager_resource, attrs]
--- a/app/controllers/api/base_controller/parser/request_adapter.rb
+++ b/app/controllers/api/base_controller/parser/request_adapter.rb
@@ -10,7 +10,7 @@ module Api
         def to_hash
           [:method, :action, :fullpath, :url, :base,
            :path, :prefix, :version, :api_prefix,
-           :collection, :c_suffix, :c_id, :subcollection, :s_id]
+           :collection, :c_suffix, :collection_id, :subcollection, :subcollection_id]
             .each_with_object({}) { |attr, hash| hash[attr] = send(attr) }
         end
 
@@ -57,7 +57,7 @@ module Api
           @params[:c_suffix]
         end
 
-        def c_id
+        def collection_id
           href.collection_id
         end
 
@@ -65,7 +65,7 @@ module Api
           href.subcollection
         end
 
-        def s_id
+        def subcollection_id
           href.subcollection_id
         end
 
--- a/app/controllers/api/base_controller/parser/request_adapter.rb
+++ b/app/controllers/api/base_controller/parser/request_adapter.rb
@@ -41,13 +41,6 @@ module Api
           url.partition(fullpath)[0] # http://target
         end
 
-        #
-        # c_path_parts returns: [collection, c_id, subcollection, s_id, ...]
-        #
-        def c_path_parts
-          @c_path_parts ||= version_override? ? path.split('/')[3..-1] : path.split('/')[2..-1]
-        end
-
         def subject
           href.subject
         end
--- a/app/controllers/api/base_controller/parser/request_adapter.rb
+++ b/app/controllers/api/base_controller/parser/request_adapter.rb
@@ -49,15 +49,15 @@ module Api
         end
 
         def subject
-          subcollection || collection
+          href.subject
         end
 
         def subject_id
-          subcollection? ? s_id : c_id
+          href.subject_id
         end
 
         def collection
-          @collection ||= c_path_parts[0]
+          href.collection
         end
 
         def c_suffix
@@ -65,19 +65,19 @@ module Api
         end
 
         def c_id
-          @params[:c_id] || c_path_parts[1]
+          href.collection_id
         end
 
         def subcollection
-          @subcollection ||= c_path_parts[2]
+          href.subcollection
         end
 
         def s_id
-          @params[:s_id] || c_path_parts[3]
+          href.subcollection_id
         end
 
         def subcollection?
-          !!subcollection
+          href.subcollection?
         end
 
         def expand?(what)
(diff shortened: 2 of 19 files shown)


lib/api/href.rb
```diff
@@ -1,34 +1,57 @@
...
 
-    def parse
-      return [nil, nil] unless href
-      if subcollection?
-        [subcollection.to_sym, ApplicationRecord.uncompress_id(subcollection_id)]
-      else
-        [collection.to_sym, ApplicationRecord.uncompress_id(collection_id)]
-      end
+    def resource
+      (subcollection? ? subcollection : collection)&.to_sym
```
chrisarcand: Do you actually need the call to `to_proc` (`&`) here to make this work? Doesn't (subcollection? ? subcollection : collection).to_sym` do the same thing?

lib/api/href.rb
```diff
@@ -1,34 +1,57 @@
...
       !!subcollection
     end
 
-    def path
-      @path ||= remove_trailing_slashes(fully_qualified? ? URI.parse(href).path : ensure_prefix(href))
+    def version
+      @version ||= if version?
+                     path_parts[2][1..-1]
+                   else
+                     ApiConfig.base.version
```
chrisarcand: Is this wise? If `Href` can't deduce the version from the string, I'd figure it would give me `nil` or something, not just YOLO and feed me the default version from the application configuration. If the callers are few, maybe let the callers default there instead of in here?

lib/api/href.rb
```diff
@@ -1,34 +1,57 @@
...
+                   end
+    end
+
+    def version?
+      @has_version ||= !!(Api::VERSION_REGEX =~ path_parts[2])
```
chrisarcand: Sorta see how you can work around my previous comment with this (`version if version?`) but...previous comment still stands.

lib/api/href.rb
```diff
@@ -44,24 +67,8 @@ def ensure_prefix(str)
...
-      path_parts[version? ? 6 : 5]
-    end
-
-    def version?
-      @version ||= !!(Api::VERSION_REGEX =~ path_parts[2])
+    def ensure_uncompressed(id)
+      Api.compressed_id?(id) ? Api.uncompress_id(id) : id
     end
 
     def path_parts
```
chrisarcand: (I know it's not part of this PR but now you're changing callers of this anyway)

I actually had no idea what `path_parts` would mean by it's title until I saw the code here. I feel like URL *segments* is more of a common name, which seems to be confirmed by hits on Google between those two. Maybe change this to `path_segments`?

lib/api/href.rb
chrisarcand: Heh, and I just noticed that Rubocop is exploding with a syntax error after I stared at this for a bit wondering if the omission of `:` here is intended. Looks like it's not ;) Indeed I think you can just do away with the ampersand.

imtayadeway: I needed this in the case that the expression in parentheses evaluates to `nil` : https://github.com/ManageIQ/manageiq-api/pull/97/commits/e88cef9e103d4ed17b5289160dbfb3a0159290fe

*Having said that*, I'm kind of questioning why this needs to be a symbol....I've mostly tried to preserve the API from what this started as. We will either be interpolating it or using it to lookup some configuration. In the latter case, if we don't do this already, it should be flexible enough for us to provide either a symbol/string. I'll look into removing the `to_sym`

lib/api/href.rb
imtayadeway: You are totally right. I got a bit carried away with the delegation. I think there is one caller, and I'm going to update it to do exactly as you suggested 😁

lib/api/href.rb
imtayadeway: 👍 I love `path_segments`

imtayadeway:
Here's what my yardoc looks like when rendered:

class_api_href_ _documentation_for_manageiq-api_-_2017-10-03_12 09 02

lib/api/href.rb
```diff
@@ -0,0 +1,107 @@
+module Api
+  class Href
+    # @param href [String] the href string
+    #
+    #   Supports the following forms:
+    #   - http://www.example.com/api
```
imtayadeway: I can't find any RDoc documentation on how to prevent this from being rendered as a link 😕

app/controllers/api/automate_workspaces_controller.rb
```diff
@@ -1,9 +1,9 @@
 module Api
   class AutomateWorkspacesController < BaseController
     def show
-      obj = AutomateWorkspace.find_by(:guid => @req.c_id)
+      obj = AutomateWorkspace.find_by(:guid => @req.collection_id)
```
Fryguy: THANK YOU... c_id and those shortcuts have been killing me forever :)

app/controllers/api/base_controller/parser/request_adapter.rb
```diff
@@ -2,6 +2,8 @@ module Api
   class BaseController
     module Parser
       class RequestAdapter
+        delegate :resource, :resource_id, :collection, :collection_id, :subcollection, :subcollection_id, :subcollection?, :path, :version?, :to => :href
+
```
Fryguy: 😍

app/controllers/api/vms_controller.rb
```diff
@@ -10,7 +10,7 @@ class VmsController < BaseController
     include Subcollections::MetricRollups
 
     VALID_EDIT_ATTRS = %w(description child_resources parent_resource).freeze
-    RELATIONSHIP_COLLECTIONS = [:vms, :templates].freeze
+    RELATIONSHIP_COLLECTIONS = %w(vms templates).freeze
```
Fryguy: Does the change from symbol to string matter?

imtayadeway: It's only referenced [here](https://github.com/ManageIQ/manageiq-api/pull/97/files/5cfadbf45766890b6a86b78f35188bd23186cea5#diff-032d88eafbded3d355900924d1239226R293)...I changed it because every other consumer of `Href.resource` doesn't care because it's either intepolated or converted by the caller to a symbol anyway, so it helped simplify.

app/controllers/api/authentications_controller.rb
```diff
@@ -55,9 +55,9 @@ def build_additional_fields
     def validate_auth_attrs(data)
       raise 'must supply a manager resource' unless data['manager_resource']
       attrs = data.dup.except('manager_resource')
-      manager_collection, manager_id = HrefParser.parse(data['manager_resource']['href'])
-      raise 'invalid manger_resource href specified' unless manager_collection && manager_id
-      manager_resource = resource_search(manager_id, manager_collection, collection_class(manager_collection))
+      href = Href.new(data['manager_resource']['href'])
+      raise 'invalid manger_resource href specified' unless href.resource && href.resource_id
```
abellotti: typo, it's wrong in the original line 59, but let's fix this since we're updating the line anyways.

app/controllers/api/automate_workspaces_controller.rb
abellotti: Who are these people that like to have shortcuts 😉

lib/api/href.rb
```diff
@@ -0,0 +1,107 @@
...
+    #   i.e. /api/collection/:cid. The preceding /, /api may also be
+    #   omitted, and all examples may be suffixed with a /.
+    def initialize(href)
+      @href = href
+    end
+
+    # @return [String, nil] the name of the subcollection from
+    #   the href if there is one, otherwise returns the name of the
+    #   collection. If neither are present, returns nil
+    def resource
```
abellotti: Why this change ?  This is confusing for me, I prefer the current subject. For the API a resource is the managed CI, i.e. "provider/1" or "provider/1/vms/20",  not the targeted collection "provider" or subcollection "vms".

app/controllers/api/authentications_controller.rb
imtayadeway: @abellotti there's a test for this typo! Which I don't want to incorporate in this PR....I'm very happy to do a follow up

lib/api/href.rb
imtayadeway: Any ideas @durandom , @chrisarcand ?

chrisarcand:
@imtayadeway Sorry this keeps getting merged over. 

 Check out http://docs.seattlerb.org/rdoc/RDoc/Markup.html#class-RDoc::Markup-label-Escaping+Text+Markup

Escape those links and we'll get this merged.

imtayadeway:
Here's the updated docs: 

class_api_href_ _documentation_for_manageiq-api_-_2017-10-05_09 55 01

chrisarcand:
LGTM, @abellotti's feedback has been addressed 👍

abellotti:
Thanks @imtayadeway for updating the PR. 👍

PR #102, opened 2017-10-04
Fix failing tests from changed ovirt provider

NickLaMuro:
The manageiq-providers-ovirt had a new change recently that caused issues with the tests:

https://github.com/ManageIQ/manageiq-providers-ovirt/commit/87703c6

This causes the provider tests to fail because the ovirt provider does not properly handle a missing/not present SDK when asking for api versions by default.  This adds a stub to support that.

_**Note:**  We returning the v3 version from the stub since the :sample_rhevm attributes are only setup to work with that version of the API, and it fails to work with the v4 adapter._

Links:
-----
* https://github.com/ManageIQ/manageiq-providers-ovirt/pull/90

### DIFF
--- a/spec/requests/providers_spec.rb
+++ b/spec/requests/providers_spec.rb
@@ -415,6 +415,13 @@ describe "Providers API" do
   end
 
   describe "Providers create" do
+    before(:each) do
+      require "ovirtsdk4" # incase it hasn't been autoloaded yet
+
+      allow(OvirtSDK4::Probe).to receive(:probe)
+        .and_return([OvirtSDK4::ProbeResult.new(:version => '3')])
+    end
+
     it "rejects creation without appropriate role" do
       api_basic_authorize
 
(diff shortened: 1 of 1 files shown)


abellotti:
Thanks @NickLaMuro for fixing these. :heart_eyes:

NickLaMuro:
@abellotti Forgot to mention this in the description, but this felt like a bit of a poor work around in a few ways:

* Being forced to know so much about the provider code/sdk is troubling, and it seems like the API test suite shouldn't have be that aware of the other providers to work.
* I had to specifically omit the v4 API as an option (compared to the [original stub code](https://github.com/ManageIQ/manageiq-providers-ovirt/blob/87703c6/spec/models/manageiq/providers/redhat/infra_manager_spec.rb#L210-L212) I ripped this from), because without doing so also causes the test to fail with a different error.

Can't say whether either of these is something that is even something we can address or if I am just being naive with respect to my knowledge of the API and ovirt provider and wanting a clear separation of concerns.

Regardless, just wanted you to know about those concerns so you can decide whether to address them if you so choose.  Thanks.

abellotti:
main thing is that you got the tests back to :green_apple: so we're not blocked in the other PRs.  Your first bullet makes sense, whether the API is testing too much or it's just how the provider gets initialized for us to test our in/outbound signatures, we can surely revisit. Thanks again!!

himdel:
The same problems happens in ui-classic .. ems_infra#create fails on `Unitialized constant ManageIQ::Providers::Redhat::InfraManager::OvirtServices::Strategies::V`, even when creating a provider of a different type.

PR #103, opened 2017-10-04
Brakeman ignore redirect warning

imtayadeway:
This should be marked as a false positive. It's raised to alert you
of a possible exploit on the host of the redirect, but this should not
be possible.

@miq-bot assign @abellotti

### DIFF
--- a/config/brakeman.ignore
+++ b/config/brakeman.ignore
@@ -1,0 +1,26 @@
+{
+  "ignored_warnings": [
+    {
+      "warning_type": "Redirect",
+      "warning_code": 18,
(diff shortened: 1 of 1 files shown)


config/brakeman.ignore
```diff
@@ -0,0 +1,26 @@
...
+      "code": "redirect_to(request.original_url.sub(params[:c_id], Api.uncompress_id(params[:c_id]).to_s), :status => :moved_permanently)",
+      "render_path": null,
+      "location": {
+        "type": "method",
+        "class": "Api::BaseController",
+        "method": "redirect_on_compressed_path"
+      },
+      "user_input": "params[:c_id]",
+      "confidence": "Weak",
+      "note": ""
```
Fryguy: Can you add a note describing why you excluded this?

abellotti:
LGTM!! Thanks @imtayadeway

PR #105, opened 2017-10-04
Don't compress ids when generating href slugs

imtayadeway:
Part of continuing work to deprecate compressed ids

@miq-bot assign @abellotti


abellotti:
was about to do the same.  👍

PR #107, opened 2017-10-05
Do not allow removal of all miq groups 

jntullo:
Validate that the request is not trying to set MiqGroups to an empty array, thus removing all of the User's groups.

BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1497295

@miq-bot add_label bug
@miq-bot assign @abellotti


abellotti:
LGTM!! Thanks @jntullo for the API fix.

PR #109, opened 2017-10-05
Search for resources based on href_slug  for automate workspace

mkanoor:
Automate workspaces can only be accessed using GUID's and not ID's. The href_slug for Automate Workspaces includes the GUID. The search function has been modified to account for that.

### DIFF
--- a/lib/api/utils.rb
+++ b/lib/api/utils.rb
@@ -16,7 +16,9 @@ module Api
       raise _("User must be defined") unless user
 
       klass = collection_config.klass(collection)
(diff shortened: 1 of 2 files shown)


abellotti:
LGTM!! Thanks @mkanoor for the enhancement.  will merge when all is 🍏

imtayadeway:
> Automate workspaces can only be accessed using GUID's and not ID's

If this is the case, have we considered making the GUID the primary key for automate workspaces, so that we don't have to duplicate the knowledge of how to look up automate workspaces everywhere we need one?

abellotti:
Could be revisited in the future, I'll leave that decision to @mkanoor. Thanks.

PR #112, opened 2017-10-05
Remove required filters from event streams

imtayadeway:
Opening for discussion. My thoughts were that since we have a default limit/page size in place, that should be sufficient for preventing the user from asking for anything crazy. IMO having required filters seems a little draconian in forcing the user to use this API a certain way. For instance, one of the requirements was to enforce a minimum timestamp filter to be present. If we do this, that prevents the user from writing a more elegant query in the case that they want event streams for a certain day. It would be more elegant to use the `=` filter, but they have to use two - specifying both a `<` and `>` value.

/cc @AllenBW @gtanzillo 

@miq-bot assign @abellotti

### DIFF
--- a/app/controllers/api/event_streams_controller.rb
+++ b/app/controllers/api/event_streams_controller.rb
@@ -1,19 +1,8 @@
 module Api
   class EventStreamsController < BaseController
-    before_action :validate_filters, :ensure_pagination, :only => :index
+    before_action :ensure_pagination, :only => :index
 
     private
-
-    def validate_filters
-      messages = []
-      messages << "must specify target_type" unless filter_contains?("target_type")
-      messages << "must specify a minimum value for timestamp" unless filter_contains?("timestamp>")
-      raise BadRequestError, messages.join(", ").capitalize if messages.any?
-    end
-
-    def filter_contains?(filter)
-      Array(params["filter"]).any? { |f| f.start_with?(filter) }
-    end
 
     def ensure_pagination
       params["limit"] ||= Settings.api.event_streams_default_limit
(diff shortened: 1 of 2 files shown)


AllenBW:
Agree, getting a single day is a little more painful than one might think it should be, off the top i cant recall any, but are there other endpoints that support `=` date filtering?

imtayadeway:
@AllenBW any collection that has a times or dates (and I think all of them have created/updated on timestamps) will support it

AllenBW:
@imtayadeway so theeeeen i doubly agree with your initial sentiment, in order to maintain continuity throughout the api, an essential behavior! long live the **Equal Sign**™

abellotti:
what about MetricRollupsService ? it does similar validation, should that go too ? /cc @jntullo

jntullo:
@abellotti we *can* change the MetricRollupsService, and probably should if this is the direction we choose to go. However, since the SUI is currently using it as is, I am not sure now would be the right time to implement that change since that would require changes on their part (thoughts @AllenBW ?). Since those are specific parameters, removing validation would likely be confusing for the user. If the validation were to go, we should then remove the parameter usage completely and make it work entirely off of the filters.

AllenBW:
Side note, preference is for everything to be a parameter...

ALSO all the work being done that uses both of these services [is here, still in wip](https://github.com/ManageIQ/manageiq-ui-service/pull/910) so changing it isn't that big of a deal at the moment... 

Is it possible to only enable timestamp filter to support `=`?

abellotti:
This LGTM!! Thanks @imtayadeway.

PR #113, opened 2017-10-05
Paginate all the things

imtayadeway:
Unifies the separate collection default settings by providing a single
configurable value for the max results per page, that's now applied to
all collections.

Thanks to @chrisarcand for naming 💙 

@miq-bot assign @abellotti

### DIFF
--- a/app/controllers/api/base_controller.rb
+++ b/app/controllers/api/base_controller.rb
@@ -26,6 +26,7 @@ module Api
     before_action :validate_api_action, :except => [:options]
     before_action :validate_response_format, :except => [:destroy]
     before_action :redirect_on_compressed_path
+    before_action :ensure_pagination, :only => :index
     after_action :log_api_response
 
     respond_to :json
@@ -117,6 +118,11 @@ module Api
 
       render :json => ErrorSerializer.new(type, error).serialize, :status => Rack::Utils.status_code(type)
       log_api_response
+    end
+
+    def ensure_pagination
+      params["limit"] ||= Settings.api.max_results_per_page
+      params["offset"] ||= 0
     end
   end
 end
--- a/app/controllers/api/base_controller.rb
+++ b/app/controllers/api/base_controller.rb
@@ -121,6 +121,9 @@ module Api
     end
 
     def ensure_pagination
+      if params["limit"].to_i > Settings.api.max_results_per_page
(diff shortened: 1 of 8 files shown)


spec/requests/querying_spec.rb
```diff
@@ -116,14 +116,6 @@ def create_vms_by_name(names)
...
     end
 
-    it "only returns paging links if both offset and limit are specified" do
-      create_vms_by_name %w(aa bb)
-
-      get api_vms_url, :params => { :offset => 0, :expand => :resources }
-
-      expect(response.parsed_body.keys).to eq(%w(name count subcount resources actions))
-    end
-
```
abellotti: maybe replace with a simple query returns self/first/last.  (next/previous won't be there for a small data set).

abellotti:
I like 😍 , starting to feel like LDAP 😉 

Definitely a breaking change though.  For customers currently expecting the whole data set for whatever they are querying, they will need to change their code. This most likely affects CLI clients and not web.

so 2 things:
- maybe a quick write-up here on current vs. new behavior and recommendation on changes (i.e. follow next links if there).
- to help minimize changes to customers currently expecting whole data set, maybe code/ocument how they would ask for that.   Think about /api/v2.4.0/ (actually anything < v3.0.0*) or maybe some value for limit (-1 ?).

app/controllers/api/base_controller.rb
```diff
@@ -118,5 +119,10 @@ def api_error(type, error)
       render :json => ErrorSerializer.new(type, error).serialize, :status => Rack::Utils.status_code(type)
       log_api_response
     end
+
+    def ensure_pagination
+      params["limit"] ||= Settings.api.max_results_per_page
```
Fryguy: Should a user be allowed to pass a value *higher* that the max_results_per_page?  That is, if they pass `:limit => 1000000`, it could bring down the server, which seems like a possible DoS.

config/settings.yml
```diff
@@ -1,5 +1,4 @@
 :api:
   :token_ttl: 10.minutes
   :authentication_timeout: 30.seconds
-  :metrics_default_limit: 1000
-  :event_streams_default_limit: 1000
```
Fryguy: If you are removing settings, you should probably create a data migration against the settings_changes in case someone has modified these values previously.

imtayadeway: @Fryguy AFAIK neither of these settings have made it into an official release yet - does that change anything?

abellotti:
Thanks @imtayadeway for this enhancement.  👍

PR #116, opened 2017-10-10
Allow generic object definition picture to be removed via edit

jntullo:
#### Usage
```
{
  "action"    : "edit",
  "resources" : [
  { "id" : ":id", "resource" : { "name" : "updated 1", "picture" : {} } }
  ]
}
```

@miq-bot assign @abellotti 
cc: @AparnaKarve

### DIFF
--- a/app/controllers/api/generic_object_definitions_controller.rb
+++ b/app/controllers/api/generic_object_definitions_controller.rb
@@ -123,6 +123,7 @@ module Api
     private
 
     def add_picture_resource(data)
+      return nil if data.empty?
       id = parse_id(data, :pictures)
       return resource_search(id, :pictures, collection_class(:pictures)) if id
       Picture.create_from_base64(data)
--- a/spec/requests/generic_object_definitions_spec.rb
+++ b/spec/requests/generic_object_definitions_spec.rb
@@ -275,6 +275,28 @@ RSpec.describe 'GenericObjectDefinitions API' do
       expect(response.parsed_body).to include(expected)
     end
 
+    it 'can remove a picture via edit' do
+      api_basic_authorize collection_action_identifier(:generic_object_definitions, :edit)
+      object_def.update_attributes!(:picture => picture)
+
+      request = {
+        'action'    => 'edit',
+        'resources' => [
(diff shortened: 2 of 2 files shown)


abellotti:
I think this is fine, essentially un-assigning the picture from the generic object definition.

If someone wants to also delete the picture from the vmdb, would that be the separate call to the picture resource ?  I'm curious as to what happens if someone deletes a picture then we fetch a generic object definition asking for the image, is that nil or bogus ?

AparnaKarve:
> If someone wants to also delete the picture from the vmdb, would that be the separate call to the picture resource

Good question.
I think it needs to be a separate call, since it is a separate concern.
But then OTOH, the Generic Object Definition `Create` API was responsible for creating the picture, so why not let the API un-assign the picture record and also delete it? 
( I don't know :) )

@abellotti I created this PR - https://github.com/ManageIQ/manageiq/pull/16163 that will enable us to get the `picture_url_path` for an existing Generic Object Definition record.

What I noticed was, it is important to set `:allow_nil => true`
Otherwise we end up getting this -
```
{
  "error": {
    "message": "GenericObjectDefinition#picture_url_path delegated to picture.url_path, but picture is nil: #<GenericObjectDefinition id: 10000000000080, name: \"a11\", description: \"a\", properties: {:attributes=>{}, :associations=>{}, :methods=>[]}, created_at: \"2017-10-09 22:31:05\", updated_at: \"2017-10-09 22:31:05\">",
    "klass": "Module::DelegationError"
  }
}
```

jntullo:
@abellotti @AparnaKarve meant to post this earlier, sorry

but it actually removes and deletes the resource all in one call, since on the model it is set as `:dependent => :destroy` here https://github.com/ManageIQ/manageiq/blob/master/app/models/generic_object_definition.rb#L20

AparnaKarve:
@jntullo That's great. Problem solved :)
Thanks for that link.

abellotti:
so the remaining question is the travis failure.

jntullo:
@abellotti working on it 👍

jntullo:
Fix: https://github.com/ManageIQ/manageiq/pull/16164

abellotti:
LGTM!! Thanks @jntullo for the API enhancement.

PR #119, opened 2017-10-12
ActionsBuilder

jntullo:
Breaking out a lot of the action logic that is held in the `renderer` into an `ActionsBuilder`. We have a lot of tests in random places for actions, so this will allow us to centralize those tests as well as reduce what the renderer does. 

@miq-bot add_label wip, refactoring

### DIFF
--- a/lib/api/actions_builder.rb
+++ b/lib/api/actions_builder.rb
@@ -12,30 +12,16 @@ module Api
       @resource = resource
     end
 
-    def collection_actions
+    def actions
       return unless cspec
-      targets = subcollection.nil? ? cspec[:collection_actions] : subcollection_actions
+      targets = collection? ? collection_targets : subcollection_targets
       return [] unless targets
-      targets.each.collect do |method, action_definitions|
+      targets.each.collect do |method, actions|
         next unless render_actions_for_method?(method)
-        action_defs = typed_subcollection_actions(method) || action_definitions
-        action_defs.each.collect do |action|
-          next unless !action[:disabled] && user_role_allows?(action[:identifier])
-          { 'name' => action[:name], 'method' => method, 'href' => href }
-        end
-      end.flatten.compact
-    end
-
-    def resource_actions
-      targets = subcollection.nil? ? cspec[:resource_actions] : subresource_actions
-      return [] unless targets
-      targets.each.collect do |method, action_definitions|
-        next unless render_actions_for_method?(method)
-        action_defs = action_definitions || typed_subcollection_actions(method)
-        action_defs.each.collect do |action|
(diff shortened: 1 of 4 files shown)


jntullo:
@miq-bot assign @imtayadeway

lib/api/actions_builder.rb
```diff
@@ -0,0 +1,93 @@
+module Api
+  class ActionsBuilder
+    attr_reader :config, :href, :type, :cspec, :collection, :subcollection, :resource
+
+    def initialize(request, href, type, resource = nil)
+      @config = CollectionConfig.new
```
imtayadeway: Similarly, you could pass this into the constructor (but maybe leaving it as the default), meaning you could test this with a fake config that conforms to the same structure. I think this would be OK since there should be good coverage that this works in general in the request specs, but would enable you to create edge cases that are not easy to find in the wild, and also decouples the tests from the specifics of the api.yml. Small changes there could cause these tests to fail, which would be unfortunate.

spec/lib/api/actions_builder_spec.rb
```diff
@@ -0,0 +1,118 @@
...
+  let(:role) { FactoryGirl.create(:miq_user_role) }
+  let(:group) { FactoryGirl.create(:miq_group, :miq_user_role => role) }
+  let(:user) { FactoryGirl.create(:user, :miq_groups => [group]) }
+  let(:request) { instance_double(Api::BaseController::Parser::RequestAdapter) }
+
+  before do
+    User.current_user = user
+  end
+
+  describe '#collection_actions' do
```
imtayadeway: It looks like the only public method is `#actions` - perhaps these example groups need to be grouped under a single `#actions` group?

spec/requests/custom_actions_spec.rb
```diff
@@ -128,7 +128,7 @@ def expect_result_to_have_custom_actions_hash
 
       expect_result_to_have_keys(%w(id href actions))
       action_specs = response.parsed_body["actions"]
-      expect(action_specs.size).to eq(1)
+      expect(action_specs.size).to eq(3)
```
imtayadeway: It looks like this behavior changed during a refactoring. Nit, but it might help to break up this commit into the refactoring part and the part that introduced the change in behavior.

spec/requests/querying_spec.rb
```diff
@@ -822,108 +822,6 @@ def create_vms_by_name(names)
   end
 
   describe "Querying resources" do
-    it "does not return actions if not entitled" do
```
imtayadeway: I think if we have good unit tests for the extracted object, we probably don't need exhaustive edge cases tested at this level 👍 I would however leave one or two general cases to show that everything works when integrated. WDYT?

PR #121, opened 2017-10-12
Loosen JBuilder dependency

chrisarcand:
No reason this can't allow minor updates (and currently blocks me on dependency resolutions for Rails 5.1)


abellotti:
👍  Thanks @chrisarcand

PR #122, opened 2017-10-13
Fix Generic Object Creation

jntullo:
Fix so that a generic object definition is not created if it is a bad request, ie if there are invalid associations.

@miq-bot add_label bug
@miq-bot assign @abellotti 

bz: https://bugzilla.redhat.com/show_bug.cgi?id=1500865


abellotti:
LGTM!! Thanks @jntullo for fixing this.

PR #123, opened 2017-10-13
Do not allow user to have multiple carts

jntullo:
Fixes a bug where a user is able to create multiple carts, when they are only supposed to have one cart at a time.

BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1493788

@miq-bot add_label bug
@miq-bot assign @abellotti

### DIFF
--- a/app/controllers/api/service_orders_controller.rb
+++ b/app/controllers/api/service_orders_controller.rb
@@ -7,6 +7,7 @@ module Api
       raise BadRequestError, "Can't create an ordered service order" if data["state"] == ServiceOrder::STATE_ORDERED
       service_requests = data.delete("service_requests")
       data["state"] ||= ServiceOrder::STATE_CART
+      raise BadRequestError, "Cart for user #{User.current_user.name} already exists" if cart?
       if service_requests.blank?
         super
       else
@@ -55,6 +56,12 @@ module Api
     end
 
     private
+
+    def cart?
+      ServiceOrder.cart_for(User.current_user)
+    rescue ActiveRecord::RecordNotFound
+      false
+    end
 
     def add_request_to_cart(workflow)
       workflow.add_request_to_cart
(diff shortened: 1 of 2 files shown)


abellotti:
I don't mind having this here, but should this be in the model ?  /cc @gmcculloug

gmcculloug:
The back-end currently does not enforce this because it is possible for a user to have multiple carts.  When ordering through the ServiceUI this restriction should be true, but when orders are created directly from the API, the OpsUI or automate we also create a new cart at that time and order it immediately.

If we want to move this check to the back-end then I guess we would need to start passing a flag to create in the places where we do need to allow multiple carts active carts, otherwise the validation could happen then.

@abellotti thoughts?

abellotti:
Interesting, the issue at hand is that the API is not restricted to SUI usage only, so we really shouldn't put that restriction here either.

Fryguy:
IMO, the shopping cart creation should not be in the API.  When would a user ever create a cart?  It should be auto-created by the backend on the first addition to a cart.

I don't completely understand the SUI restriction.  Does that mean that if someone has a cart ordered and in flight, and then tries to add another service, they can't because they only have 1 cart?  If so, that doesn't make sense...that's like you are waiting for your Amazon shipment to arrive, and you can't order anything else from Amazon until you receive it.

gmcculloug:
@Fryguy That is an great point, and what you describe about cart creation is what we do today.  The SUI order comes through this `add_to_cart` in the `ServiceOrder` model: 
 https://github.com/ManageIQ/manageiq/blob/master/app/models/service_order.rb#L72

The cart logic is that you should only have one cart in the "cart" state, pre-user/per-tenant.  You can have other carts in the "ordered" state and the "wish" state for future wish-list support.

jntullo:
closing in favor of https://github.com/ManageIQ/manageiq/pull/16351

PR #124, opened 2017-10-13
encrypt/decrypt for Automate Workspace objects

mkanoor:
This PR exposes the 
* encrypt
* decrypt

Functions on the Automate Workspace objects.

The input parameters are
* object        name
* attribute    name
* value          clear text string

Depends on PR https://github.com/ManageIQ/manageiq/pull/16202

* Input Body
{
  "action" : "decrypt",
  "resource" : { "object" : "root", "attribute" : "ipam_password" }
}

Response:
       { "object" : "root", "attribute" : "ipam_password", "value" : "decrypted_value" }

### DIFF
--- a/app/controllers/api/automate_workspaces_controller.rb
+++ b/app/controllers/api/automate_workspaces_controller.rb
@@ -17,19 +17,14 @@ module Api
 
       obj = resource_search(id, type, collection_class(type))
       obj.encrypt(data['object'], data['attribute'], data['value'])
-      obj.reload
-      obj
     end
 
     private
 
     def decrypt(obj, data)
-      begin
-        value = obj.decrypt(data['object'], data['attribute'])
-      rescue
-        value = ""
-      end
-      {'object' => data['object'], 'attribute' => data['attribute'], 'value' => value}
+      {'object'    => data['object'],
+       'attribute' => data['attribute'],
+       'value'     => obj.decrypt(data['object'], data['attribute'])}
     end
   end
 end
--- a/app/controllers/api/automate_workspaces_controller.rb
+++ b/app/controllers/api/automate_workspaces_controller.rb
@@ -9,7 +9,7 @@ module Api
       raise BadRequestError, "Must specify an id for starting a #{type} resource" unless id
 
       obj = resource_search(id, type, collection_class(type))
-      data['resources'] ? decrypt_all(obj, data) : decrypt_one(obj, data)
+      decrypt(obj, data)
     end
 
     def encrypt_resource(type, id = nil, data = nil)
@@ -23,11 +23,7 @@ module Api
 
     private
 
-    def decrypt_all(obj, data)
-      { "results" => data["resources"].collect { |res| decrypt_one(obj, res) } }
-    end
-
-    def decrypt_one(obj, data)
+    def decrypt(obj, data)
       begin
         value = obj.decrypt(data['object'], data['attribute'])
       rescue
--- a/app/controllers/api/automate_workspaces_controller.rb
+++ b/app/controllers/api/automate_workspaces_controller.rb
@@ -1,8 +1,26 @@
 module Api
   class AutomateWorkspacesController < BaseController
     def edit_resource(type, id, data = {})
+      raise BadRequestError, "Must specify an id for starting a #{type} resource" unless id
+
       obj = resource_search(id, type, collection_class(type))
       obj.merge_output!(data)
+    end
+
+    def decrypt_resource(type, id = nil, data = nil)
+      raise BadRequestError, "Must specify an id for starting a #{type} resource" unless id
+
+      obj = resource_search(id, type, collection_class(type))
+      { 'result' => obj.decrypt(data['object'], data['attribute']) }
+    end
+
+    def encrypt_resource(type, id = nil, data = nil)
+      raise BadRequestError, "Must specify an id for starting a #{type} resource" unless id
+
+      obj = resource_search(id, type, collection_class(type))
+      obj.encrypt(data['object'], data['attribute'], data['value'])
+      obj.reload
+      obj
     end
   end
 end
--- a/app/controllers/api/automate_workspaces_controller.rb
+++ b/app/controllers/api/automate_workspaces_controller.rb
@@ -1,8 +1,6 @@
 module Api
   class AutomateWorkspacesController < BaseController
     def edit_resource(type, id, data = {})
-      raise BadRequestError, "Must specify an id for starting a #{type} resource" unless id
-
       obj = resource_search(id, type, collection_class(type))
       obj.merge_output!(data)
     end
--- a/app/controllers/api/automate_workspaces_controller.rb
+++ b/app/controllers/api/automate_workspaces_controller.rb
@@ -11,7 +11,7 @@ module Api
       raise BadRequestError, "Must specify an id for starting a #{type} resource" unless id
 
       obj = resource_search(id, type, collection_class(type))
-      { 'result' => obj.decrypt(data['object'], data['attribute']) }
+      data['resources'] ? decrypt_all(obj, data) : decrypt_one(obj, data)
     end
 
     def encrypt_resource(type, id = nil, data = nil)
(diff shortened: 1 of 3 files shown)


spec/requests/automate_workspaces_spec.rb
```diff
@@ -66,5 +99,35 @@
...
       expect(response).to have_http_status(:bad_request)
     end
+
+    it 'decrypt a password' do
+      api_basic_authorize action_identifier(:automate_workspaces, :decrypt)
+
+      post(api_automate_workspace_url(nil, aw.guid), :params => decrypt_params)
+
+      expect(response).to have_http_status(:ok)
+      expect(JSON.parse(response.body)['value']).to eq(password)
```
imtayadeway: Prefer `expect(response.parsed_body)` here and else where

spec/requests/automate_workspaces_spec.rb
```diff
@@ -66,5 +99,35 @@
...
+    end
+
+    it 'decrypt collection of passwords' do
+      api_basic_authorize action_identifier(:automate_workspaces, :decrypt)
+
+      post(api_automate_workspace_url(nil, aw.guid), :params => decrypt_params_a)
+
+      expect(response).to have_http_status(:ok)
+      expect(JSON.parse(response.body)['results'][0]['value']).to eq(password)
```
imtayadeway: Can the order of these be guaranteed?

Also, prefer one expectation where possible:

```ruby
expected = {
  "results" => a_collection_containing_exactly( # assuming order can't be guaranteed
    a_hash_including("value" => password"),
    a_hash_including("value" => p45)
    # etc...
  )
}

File: app/controllers/api/automate_workspaces_controller.rb [hunk 3]
```diff
@@ -1,8 +1,41 @@
...
+
       obj = resource_search(id, type, collection_class(type))
       obj.merge_output!(data)
     end
+
+    def decrypt_resource(type, id = nil, data = nil)
+      raise BadRequestError, "Must specify an id for starting a #{type} resource" unless id
+
+      obj = resource_search(id, type, collection_class(type))
+      data['resources'] ? decrypt_all(obj, data) : decrypt_one(obj, data)
```
imtayadeway: Is this necessary? IIRC the update code should call this for each set of data in the "resources" node, so this should never have to handle the `decrypt_all` case

app/controllers/api/automate_workspaces_controller.rb
```diff
@@ -1,8 +1,41 @@
 module Api
   class AutomateWorkspacesController < BaseController
     def edit_resource(type, id, data = {})
+      raise BadRequestError, "Must specify an id for starting a #{type} resource" unless id
```
imtayadeway: Has this been copy/pasted from elsewhere? Do you mean s/starting/editing/, etc..? Can you add a test for these?

app/controllers/api/automate_workspaces_controller.rb
mkanoor: @imtayadeway 
These resources are not like vmdb objects, these are embedded objects in the AutomateWorkspace input column. Which stores a Hash with the objects and attributes.

app/controllers/api/automate_workspaces_controller.rb
mkanoor: @imtayadeway That line is not needed, I had accidentally added it. The resource_search raises the error.

mkanoor:
@imtayadeway 
Please review

app/controllers/api/automate_workspaces_controller.rb
imtayadeway: I get that part - what I don't understand is `data` should represent the thing *under* "resources", so I don't understand why you're expecting that here.

abellotti: I don't think our API manager supports "resources" other than collection or subcollection targets (i.e. bulk requests for custom_attributes, tags, etc), above is needed to support multiple decrypts on a single automate workspace resource.  Maybe the name "resources" is misleading  and we need something else (i.e. attribute_references, etc).

mkanoor: @abellotti @imtayadeway 
Do we need to support bulk decrypts? I don't think it makes much sense since we are doing this on a single method by method basis. Each Automate method might be doing something specific e.g
* Getting IP Address from IPAM system
* Getting some objects from LDAP system

They would typically be defined in 2 separate methods.

So can we not do bulk decrypts? and keep it simple by doing single decrypts (one attribute at a time)

abellotti: Ok by me.   We can always revisit later if it becomes an issue and we see many API round trips for this. Thanks.

app/controllers/api/automate_workspaces_controller.rb
```diff
@@ -4,5 +4,32 @@ def edit_resource(type, id, data = {})
...
+      obj
+    end
+
+    private
+
+    def decrypt(obj, data)
+      begin
+        value = obj.decrypt(data['object'], data['attribute'])
+      rescue
+        value = ""
```
imtayadeway: Could you add a test for this case where we rescue the error? What kind of error are you anticipating? This seems like the sort of thing that `obj` here should handle internally

app/controllers/api/automate_workspaces_controller.rb
```diff
@@ -4,5 +4,32 @@ def edit_resource(type, id, data = {})
...
+      obj = resource_search(id, type, collection_class(type))
+      decrypt(obj, data)
+    end
+
+    def encrypt_resource(type, id = nil, data = nil)
+      raise BadRequestError, "Must specify an id for starting a #{type} resource" unless id
+
+      obj = resource_search(id, type, collection_class(type))
+      obj.encrypt(data['object'], data['attribute'], data['value'])
+      obj.reload
```
imtayadeway: Why does `obj` need to be reloaded? If `encrypt` is doing something to the object's data that isn't synced with its attributes, it seems like it should be `encrypt`'s responsiblity to maintain that consistency

app/controllers/api/automate_workspaces_controller.rb
```diff
@@ -4,5 +4,32 @@ def edit_resource(type, id, data = {})
       obj = resource_search(id, type, collection_class(type))
       obj.merge_output!(data)
     end
+
+    def decrypt_resource(type, id = nil, data = nil)
+      raise BadRequestError, "Must specify an id for starting a #{type} resource" unless id
```
imtayadeway: s/starting/decrypting/

app/controllers/api/automate_workspaces_controller.rb
```diff
@@ -4,5 +4,32 @@ def edit_resource(type, id, data = {})
...
+
+      obj = resource_search(id, type, collection_class(type))
+      decrypt(obj, data)
+    end
+
+    def encrypt_resource(type, id = nil, data = nil)
+      raise BadRequestError, "Must specify an id for starting a #{type} resource" unless id
```
imtayadeway: s/starting/encrypting/

app/controllers/api/automate_workspaces_controller.rb
imtayadeway: Also, is it possible for this error to raise? This needs a test proving why it is needed (as does the one below)

mkanoor:
@abellotti @imtayadeway 
Please review this is needed for G release.
I have changed to decrypt a single obj/attribute

abellotti:
Thanks @mkanoor for the update, new signature looks good.

app/controllers/api/automate_workspaces_controller.rb
mkanoor: @imtayadeway Catching and returning an empty string was suggested by @Fryguy to thwart Oracle attack, trying to decipher the encryption key, which has a very remote possibility here.
The app/models code raises the exception when the object, attribute is missing or when the attribute is not a password type.

imtayadeway: Maybe you could extract a `#decrypt!` method?

```ruby
def decrypt!(object, attribute)
  # whatever old #decrypt does
end

def decrypt(object, attribute)
  decrypt!(object, attribute)
rescue TheSpecificError
  ""
end
```

I don't think handling this in the controller is a good idea - it's a piece of knowledge about how to recover from such an event that needs to be encapsulated in the model.

mkanoor:
@abellotti @imtayadeway
I am waiting for this PR https://github.com/ManageIQ/manageiq/pull/16335 to get merged then this PR should go green

mkanoor:
@imtayadeway 
Please review model changes done and the build is green

imtayadeway:
@mkanoor looks good but there are still some outstanding requests https://github.com/ManageIQ/manageiq-api/pull/124#discussion_r147478490, https://github.com/ManageIQ/manageiq-api/pull/124#discussion_r147478898 and https://github.com/ManageIQ/manageiq-api/pull/124#discussion_r147478550

abellotti:
Thanks @mkanoor for the update and @imtayadeway for the review.

will merge when 🍏

PR #130, opened 2017-10-18
Add images api

andyvesel:
Add Images to API

### DIFF
--- a/app/controllers/api/cloud_templates_controller.rb
+++ b/app/controllers/api/cloud_templates_controller.rb
@@ -1,0 +1,4 @@
+module Api
+  class CloudTemplatesController < BaseController
+  end
+end
--- a/config/api.yml
+++ b/config/api.yml
@@ -486,6 +486,33 @@
       :get:
       - :name: read
         :identifier: cloud_tenant_show
+  :cloud_templates:
+    :description: Cloud Templates
+    :identifier: image
+    :options:
+      - :collection
+      - :subcollection
+    :verbs: *gp
+    :klass: ManageIQ::Providers::CloudManager::Template
+    :collection_actions:
+      :get:
+      - :name: read
+        :identifier: image_show_list
+      :post:
+      - :name: query
+        :identifier: image_show_list
+    :resource_actions:
+      :get:
+      - :name: read
+        :identifier: image_show
+    :subcollection_actions:
+      :get:
+      - :name: read
+        :identifier: image_show_list
+    :subresource_actions:
+      :get:
+      - :name: read
+        :identifier: image_show
   :cloud_tenants:
     :description: Cloud Tenants
     :identifier: cloud_tenant
@@ -1696,6 +1723,7 @@
     - :security_groups
     - :vms
     - :flavors
+    - :cloud_templates
     :collection_actions:
       :get:
       - :name: read
--- a/app/controllers/api/providers_controller.rb
+++ b/app/controllers/api/providers_controller.rb
@@ -20,6 +20,7 @@ module Api
     include Subcollections::SecurityGroups
     include Subcollections::Vms
     include Subcollections::Flavors
+    include Subcollections::CloudTemplates
 
     def create_resource(type, _id, data = {})
       assert_id_not_specified(data, type)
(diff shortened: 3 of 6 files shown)


app/controllers/api/cloud_templates_controller.rb
```diff
@@ -0,0 +1,24 @@
+module Api
+  class CloudTemplates < BaseController
```
jntullo: should be `CloudTemplatesController`

app/controllers/api/cloud_templates_controller.rb
```diff
@@ -0,0 +1,24 @@
...
+    def cloud_templates_create_resource
```
jntullo: you'll want `create_resource(type, id, data)` here

app/controllers/api/cloud_templates_controller.rb
```diff
@@ -0,0 +1,24 @@
...
+      task_id = Template.create_image(User.current_user.id, parent, data)
+      action_result(true, 'Creating Flavor', :task_id => task_id)
+    rescue => err
+      action_result(false, err.to_s)
+    end
+
+    def cloud_templates_delete_resource(type, id, _data = {})
```
jntullo: and similarly to create, should be `delete_resource(type, id, data)` here

config/api.yml
```diff
@@ -2315,6 +2316,53 @@
...
+      - :subcollection
+    :verbs: *gpd
+    :klass: ManageIQ::Providers::CloudManager::Template
+    :collection_actions:
+      :get:
+      - :name: read
+        :identifier: miq_template_show_list
+      :post:
+      - :name: query
+        :identifier: miq_template_show_list
```
jntullo: Is there a reason why you are using `miq_template_show_list` instead of the `image_show_list`? The product features seem to be inconsistent

config/api.yml
```diff
@@ -2315,6 +2316,53 @@
...
+      - :name: query
+        :identifier: miq_template_show
+      - :name: create
+        :identifier: image_create
+    :subcollection_actions:
+      :get:
+      - :name: read
+        :identifier: miq_template_show_list
+      :post:
+      - :name: delete
```
jntullo: Can you add tests for these subcollection actions?

config/api.yml
```diff
@@ -2315,6 +2316,53 @@
...
+    :resource_actions:
+      :get:
+      - :name: read
+        :identifier: miq_template_show
+      :post:
+      - :name: delete
+        :identifier: image_delete
+      - :name: query
+        :identifier: miq_template_show
+      - :name: create
```
jntullo: create would not be a resource action but a collection action. Can you also add tests for this?

andyvesel: I'll just remove `create` and `delete` actions for now, and add it, when it's ready. Does it make sense?

config/api.yml
andyvesel: @jntullo yes, replaced it

config/api.yml
andyvesel: @jntullo added tests for provider's image, and removed `delete` action for now

config/api.yml
```diff
@@ -2315,6 +2316,35 @@
...
+      - :name: read
+        :identifier: image_show_list
+      :post:
+      - :name: query
+        :identifier: image_show_list
+    :resource_actions:
+      :get:
+      - :name: read
+        :identifier: image_show
+      - :name: query
```
jntullo: query should only be a collection action

andyvesel: @jntullo Ah, sorry missed it.
Fixed.

andyvesel:
@imtayadeway fixed! Hope to get Travis passed

abellotti:
@andyvesel can you add the POST /api/cloud_templates bulk "query" test, that can be added to the collections_spec.rb like the others, Thanks.

andyvesel:
@abellotti tests added

abellotti:
Thanks @andyvesel for the update.  

@imtayadeway anything else here ?

abellotti:
LGTM !! Thanks @andyvesel for the API Enhancement.

PR #131, opened 2017-10-18
[WIP] Cleanup Generic Object Specs

jntullo:
Dependent on https://github.com/ManageIQ/manageiq/pull/16227 

Using a new trait takes a lot of the setup that is currently present in our generic object specs

@miq-bot add_label cleanup, test

### DIFF
--- a/spec/requests/generic_object_definitions_spec.rb
+++ b/spec/requests/generic_object_definitions_spec.rb
@@ -1,8 +1,6 @@
 # rubocop:disable Style/WordArray
 RSpec.describe 'GenericObjectDefinitions API' do
-  let(:object_def) { FactoryGirl.create(:generic_object_definition, :name => 'foo') }
-  let(:object_def2) { FactoryGirl.create(:generic_object_definition, :name => 'foo 2') }
-  let(:object_def3) { FactoryGirl.create(:generic_object_definition, :name => 'foo 3') }
+  let(:object_def) { FactoryGirl.create(:generic_object_definition, :with_methods_attributes_associations, :name => 'foo') }
   let(:picture) { FactoryGirl.create(:picture, :extension => 'png') }
   let(:content) do
     "iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAYAAAAfFcSJAAAABGdBTUEAALGP"\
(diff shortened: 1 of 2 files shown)


spec/requests/generic_object_definitions_spec.rb
```diff
@@ -397,15 +392,14 @@
...
-      object_def.add_property_attribute('attr1', 'string')
       object_def2.add_property_attribute('attr2', 'string')
       object_def3.add_property_attribute('attr3', 'string')
       api_basic_authorize collection_action_identifier(:generic_object_definitions, :add_associations)
 
       request = {
         'action'    => 'remove_attributes',
         'resources' => [
-          { 'name' => object_def.name, 'resource' => { 'attributes' => { 'attr1' => 'string' } } },
+          { 'name' => object_def.name, 'resource' => { 'attributes' => { 'widget' => 'string' } } },
```
imtayadeway: I think this test might now be less clear because it's no longer obvious what the connection is between this attribute and `object_def` - one of the main players in this test. It's now implicit that it has this specific attribute. I think traits are fine if they help you create objects that are suited for certain use cases, but where the specifics don't matter too much ("an object that has associations, attributes, etc..."). But when those specifics are the thing under test, I think those details are often best left in the test body, even if that means duplicating things from the trait definition.

spec/requests/generic_object_definitions_spec.rb
```diff
@@ -414,7 +408,7 @@
 
       expected = {
         'results' => a_collection_including(
-          a_hash_including('id' => object_def.id.to_s, 'properties' => a_hash_including('attributes' => {})),
+          a_hash_including('id' => object_def.id.to_s, 'properties' => a_hash_including('attributes' => {'powered_on' => 'boolean'})),
```
imtayadeway: Similarly, this test had to be updated with the specifics of an object that was meant to be standing in for something ("an object that has attributes"), but any change to that trait would cause this to fail, even though could still serve perfectly well as "an object that has attributes". I think here and in any similar test it would be better to be decoupled from the trait.

spec/requests/generic_object_definitions_spec.rb
jntullo: @imtayadeway I agree, was actually working on these changes.

chrisarcand:
> I think traits are fine if they help you create objects that are suited for certain use cases, but where the specifics don't matter too much ("an object that has associations, attributes, etc..."). But when those specifics are the thing under test, I think those details are often best left in the test body, even if that means duplicating things from the trait definition.

I agree with this. I'd normally not care that much and be ok with a spec depending on data in a factory (`object.powered_on = true`, the object must have `powered_on` but the spec has no say in that. If the factory is changed, the specs will fail), but I'm concerned since the factory doesn't actually live in this project. Seems like a coupling that we can avoid 😕 Thoughts on that?

PR #132, opened 2017-10-18
Remove some indirection in rendering

imtayadeway:
These methods add little while also harboring comment-lies ("Render nothing for normal update.") Removing them is, IMO, preferable.

@miq-bot assign @abellotti

### DIFF
--- a/app/controllers/api/auth_controller.rb
+++ b/app/controllers/api/auth_controller.rb
@@ -17,8 +17,7 @@ module Api
 
     def destroy
       api_token_mgr.invalidate_token(request.headers[HttpHeaders::AUTH_TOKEN])
-
-      render_normal_destroy
+      head :no_content
     end
 
     private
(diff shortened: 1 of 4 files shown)


abellotti:
This is fine, if you can just take care of Rubocop, I see other places calling head(:ok), let's just be consistent. Thanks.

imtayadeway:
@abellotti I'd rather we didn't `head(:ok)` as it's DSL - we should probably mark that as such for the bot so that it doesn't continue to complain, even if we have added parentheses elsewhere.

abellotti:
Can you or @chrisarcand update the bot ?

chrisarcand:
> we should probably mark that as such for the bot so that it doesn't continue to complain

Yeah for sure. @imtayadeway if you send the PR I'll merge and deploy.

imtayadeway:
@chrisarcand https://github.com/ManageIQ/guides/pull/263

@abellotti are you OK to merge?

abellotti:
Yep, with the rubocop changes merged.  Thanks @imtayadeway

Issue #134, opened 2017-10-19
Rein in message passing to `Object` in collection config


chrisarcand:
https://github.com/ManageIQ/manageiq-api/blob/c5cc04a2ae95b8ac89436ceb25b774f41eff1a58/lib/api/collection_config.rb#L7-L13

As brought up on Gitter by @himdel, we should rein in how we handle the issue fixed by this rescue. We're currently still sending the message (calling the method) unrestricted on the config object, so you can send things like `display` (which writes to stdout), etc. That's less than ideal.

At the very least, if this is indeed the best solution, I wonder what error is being rescued here.

himdel:
Traceback of the case where this makes the API output a lot of stuff on any /api access :+1: 

```
 "/home/milan/.rbenv/versions/2.4.1/lib/ruby/2.4.0/ostruct.rb:281:in `inspect'",
 "/home/milan/.rbenv/versions/2.4.1/lib/ruby/gems/2.4.0/gems/config-1.3.0/lib/config/options.rb:135:in `write'",
 "/home/milan/.rbenv/versions/2.4.1/lib/ruby/gems/2.4.0/gems/config-1.3.0/lib/config/options.rb:135:in `display'",
 "/home/milan/.rbenv/versions/2.4.1/lib/ruby/gems/2.4.0/gems/config-1.3.0/lib/config/options.rb:135:in `[]'",
 "/home/milan/src/manage-iq/manageiq-api/lib/api/collection_config.rb:10:in `[]'",
 "/home/milan/src/manage-iq/manageiq-api/lib/api/collection_config.rb:98:in `resource_identifier'",
 "/home/milan/src/manage-iq/manageiq-api/app/controllers/api/base_controller/normalizer.rb:14:in `normalize_hash'",
 "/home/milan/src/manage-iq/manageiq-api/app/controllers/api/base_controller/normalizer.rb:35:in `normalize_attr'",
 "/home/milan/src/manage-iq/manageiq-api/app/controllers/api/base_controller/normalizer.rb:22:in `block in normalize_hash'",
 "/home/milan/src/manage-iq/manageiq-api/app/controllers/api/base_controller/normalizer.rb:21:in `each'",
 "/home/milan/src/manage-iq/manageiq-api/app/controllers/api/base_controller/normalizer.rb:21:in `normalize_hash'",
 "/home/milan/src/manage-iq/manageiq-api/app/controllers/api/base_controller/normalizer.rb:35:in `normalize_attr'",
[...]

PR #135, opened 2017-10-19
Blacklist Config Values

jntullo:
Proposed solution to resolve https://github.com/ManageIQ/manageiq-api/issues/134 

Rather than rescuing an error, blacklist the `:system` and `:display` values. Not allowing the `:display` value to be called will resolve issues with the config being displayed in the console output, and removing the call to `:system` removes the need for the rescue block.

cc: @himdel @chrisarcand 

@miq-bot add_label bug 
@miq-bot assign @abellotti

### DIFF
--- a/app/controllers/api/base_controller/normalizer.rb
+++ b/app/controllers/api/base_controller/normalizer.rb
@@ -30,7 +30,7 @@ module Api
         return if value.nil?
         if value.kind_of?(Array) || value.kind_of?(ActiveRecord::Relation)
           normalize_array(value)
-        elsif value.respond_to?(:attributes) || value.respond_to?(:keys)
+        elsif !attr.nil? && (value.respond_to?(:attributes) || value.respond_to?(:keys))
           normalize_hash(attr, value)
         elsif attr == "id" || attr.to_s.ends_with?("_id")
           value.to_s
--- a/lib/api/collection_config.rb
+++ b/lib/api/collection_config.rb
@@ -1,16 +1,11 @@
 module Api
   class CollectionConfig
-    BLACKLISTED = [:system, :display].freeze
-
     def initialize
       @cfg = ApiConfig.collections
     end
 
     def [](collection_name)
-      # Config::Options implements a system method causing self[:system] to error.
-      # i.e. subcollection name system in an arbitraty path /api/automate/manageiq/system
-      collection_name = collection_name.to_sym
-      return if BLACKLISTED.include?(collection_name)
+      return unless include?(collection_name)
       @cfg[collection_name.to_sym]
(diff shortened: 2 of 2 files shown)


himdel:
@jntullo I don't think this is a good solution.

As mentioned in the issue, this is not about 2 methods. This is about *every* method defined on Object, OpenStruct or Config::Options. Right now, we're hitting problems with 2 of them but that's not really future-proof.

(I'm almost sure if we ever add a gem that extends Object to add a method named after anything already in the config, it will break the API equally well.)

Can we use a whitelist instead?

chrisarcand:
> I don't think this is a good solution.

I think this is a great solution - when replacing the `config` gem quickly or easily is certainly *not* an option.

himdel:
@chrisarcand I don't know how much it's being used and what all is used from it, but.. would it really be that hard to replace?

But.. assuming the gem can't be replaced, I'm good with this. I still think we'll hit these problems again, but .. eh, maybe not soon.

Fryguy:
I agree with @himdel as this is not API's concern to blacklist the config gem.  If anything, it belongs in Vmdb::Settings.  See my comment in the issue : https://github.com/ManageIQ/manageiq-api/issues/134#issuecomment-337954371

imtayadeway:
> As mentioned in the issue, this is not about 2 methods. This is about every method defined on Object, OpenStruct or Config::Options. Right now, we're hitting problems with 2 of them but that's not really future-proof.

I think a better solution to this is: don't do that. Having a blacklist elsewhere to ensure we don't add conflicting keys to the config I'm all for. The problem in this case was that we were looking for settings that weren't there, and that's the part I'm suggesting we tackle.

Fryguy:
> The problem in this case was that we were looking for settings that weren't there, and that's the part I'm suggesting we tackle.

I don't disagree but this is not API's concern...it's a *general* concern, and can happen to anyone that tries to access the config.  Even so (and I commented in the other thread with this), if all that's trying to be done is verify the presence of a key, then you can also use `Settings.to_h.include?(key)`, which avoids the "doing a key lookup by actually calling a method", which in general you shouldn't do with Settings anyway since any random key is available on an OpenStruct.

```ruby
Settings.sdkfjhsdkfjh
# => nil
Settings.display
# OMGSOMUCHOUTPUT=>nil
Settings.to_h.include?(:sdkfjhsdkfjh)
# => false
Settings.to_h.include?(:display)
# => false
```

abellotti:
Thanks @jntullo for fixing this.  👍

cben:
Neat!  Could you rename the PR and maybe edit description to reflect the final solution?

PR #137, opened 2017-10-21
Add security group subcollection actions

tzumainn:
https://bugzilla.redhat.com/show_bug.cgi?id=1487103

### DIFF
--- a/app/controllers/api/subcollections/security_groups.rb
+++ b/app/controllers/api/subcollections/security_groups.rb
@@ -4,6 +4,30 @@ module Api
       def security_groups_query_resource(object)
         object.respond_to?(:security_groups) ? object.security_groups : []
       end
+
+      def security_groups_add_resource(parent, _type, _id, data)
+        security_group = data["security_group"]
+
+        begin
+          message = "Adding security group #{security_group} to #{parent.name}"
+          task_id = queue_object_action(parent, message, {:method_name => "add_security_group", :args => [security_group]})
+          action_result(true, message, :task_id => task_id)
+        rescue => e
+          action_result(false, e.to_s)
+        end
+      end
+
+      def security_groups_remove_resource(parent, _type, _id, data)
+        security_group = data["security_group"]
+
+        begin
+          message = "Removing security group #{security_group} from #{parent.name}"
+          task_id = queue_object_action(parent, message, {:method_name => "remove_security_group", :args => [security_group]})
+          action_result(true, message, :task_id => task_id)
+        rescue => e
+          action_result(false, e.to_s)
+        end
+      end
     end
   end
 end
--- a/spec/requests/instances_spec.rb
+++ b/spec/requests/instances_spec.rb
@@ -616,4 +616,95 @@ RSpec.describe "Instances API" do
       expect(instance.reload.custom_attributes.pluck(:value).sort).to eq(["new value1", "new value2"])
     end
   end
+
+  context 'security groups subcollection' do
+    before do
+      @network_port = FactoryGirl.create(:network_port, :device => instance)
+      @security_group = FactoryGirl.create(:security_group, :cloud_tenant => @cloud_tenant)
+      @security_group_new = FactoryGirl.create(:security_group, :cloud_tenant => @cloud_tenant)
+      @network_port_security_group = FactoryGirl.create(:network_port_security_group,
+                                                        :network_port   => @network_port,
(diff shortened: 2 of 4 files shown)


app/controllers/api/subcollections/security_groups.rb
```diff
@@ -4,6 +4,30 @@ module SecurityGroups
       def security_groups_query_resource(object)
         object.respond_to?(:security_groups) ? object.security_groups : []
       end
+
+      def security_groups_add_resource(parent, _type, _id, data)
+        security_group = data["security_group"]
+
+        begin
+          message = "Adding security group #{security_group} to #{parent.name}"
+          task_id = queue_object_action(parent, message, :method_name => "add_security_group", :args => [security_group])
```
jntullo: I see the `query_resource` checks to see if it responds to `:security_groups` - I am assuming this means not all Instances have security groups, so should it be checking to see if the parent also responds to `add_security_group` before the action is queued? Same for the `remove_resource`

spec/requests/instances_spec.rb
```diff
@@ -616,4 +616,95 @@ def update_raw_power_state(state, *instances)
...
+    it 'queries all security groups from an instance' do
+      api_basic_authorize subcollection_action_identifier(:instances, :security_groups, :read, :get)
+
+      get(api_instance_security_groups_url(nil, instance))
+
+      expected = {
+        'resources' => [
+          { 'href' => api_instance_security_group_url(nil, instance, @security_group) }
+        ]
+
```
jntullo: nit, but is this just an empty line?

tzumainn:
@jntullo thanks for the review! I've made the requested changes, let me know if you think there should be more!

app/controllers/api/subcollections/security_groups.rb
```diff
@@ -4,6 +4,32 @@ module SecurityGroups
       def security_groups_query_resource(object)
         object.respond_to?(:security_groups) ? object.security_groups : []
       end
+
+      def security_groups_add_resource(parent, _type, _id, data)
+        security_group = data["security_group"]
+        raise "Cannot add #{security_group} to #{parent.name}" unless parent.respond_to? :add_security_group
```
jntullo: Any reason why this raise is outside of the rescue block? Returning action results for everything plays nicer with bulk actions.

jntullo:
@tzumainn can you add tests for the new check you did for whether it responds to adding or removing security groups?

app/controllers/api/subcollections/security_groups.rb
abellotti: was gonna ask, we'd need specs for adding/removing in bulk.

tzumainn:
@jntullo thanks for the additional review! I've moved the raises statements and added the tests. Let me know what you think!

@abellotti this operation can't be done in bulk - let me know if I'm misunderstanding your comment?

abellotti:
why can't it be done in bulk ?  instances can have multiple security groups unless I am missing something. Standard signature would be as follows:

```
POST /api/instances/:id/security_groups

{
  "action" : "add",
  "resources" : [
      { "security_group" : "security_group_A" },
      { "security_group" : "security_group_B" }
  ]
}
```

same with the remove signature,

```
POST /api/instances/:id/security_groups

{
  "action" : "remove",
  "resources" : [
      { "security_group" : "security_group_D" },
      { "security_group" : "security_group_E" }
  ]
}
```

tzumainn:
@abellotti Ah, thanks for the explanation; I've added bulk tests. Thanks!

tzumainn:
the remaining style issues look like false positives due to nested hashes

tzumainn:
@abellotti let me know if you'd like further changes - thanks!

abellotti:
the one change that I'd like is to use "name" instead of "security_group" in the code/specs.

```
{
   "action" : "add",
   "resources" : [
      { "name" : "security_group_name1" },
      { "name" : "security_group_name2" }
   ]
}
```

would allow us to support href's, id's, later.

also, if you could take care of rubocop warnings (I saw extra spaces before =>, and params not aligned with first method argument for example).

tzumainn:
@abellotti fixed, thanks for the review! I think the last rubocop issues are incorrect - things look correctly aligned to me - but let me know if I'm wrong!

abellotti:
Thanks @tzumainn for the API Enhancement 👍

PR #139, opened 2017-10-23
Remove blueprint

bzwei:
Blueprint feature has been removed. Dependencies on API need to be removed before I can remove blueprint from the model.


bzwei:
@miq-bot add_label technical_debt
@miq-bot assign @abellotti 
cc @gmcculloug

bzwei:
@miq-bot add_label technical debt

abellotti:
_Note_  manageiq.org documentation should be updated to remove blueprint from the 3.0.0 api docs.

abellotti:
Thanks @bzwei for this PR.  Looks like it's not used at all from SUI.  We're good.

I'll just mention it in the manageiq.org API doc for 3.0.0.

PR #140, opened 2017-10-23
Custom Button CRUD API

AparnaKarve:
CRUD API for Custom Buttons
```
Create -

POST /api/custom_buttons
{
  "name"        : "CustomButton",
  "description" : "CustomButtondescription",
  "options": {
    "button_icon": "ff ff-view-expanded",
    "button_color": "#4727ff",
    "display": true,
  },
  "applies_to_class": "GenericObjectDefinition",
  "applies_to_id": "10000000000050",
}


Retrieve -

GET /api/custom_buttons/
GET /api/custom_buttons/:id


Update -

POST /api/custom_buttons/:id      action: "edit"
PUT /api/custom_buttons/:id
{
  "name" : "CustomButtons Updated",
}


Delete -

DELETE /api/custom_buttons/:id
POST /api/custom_buttons/:id      action: "delete"

Bulk delete -
POST /api/custom_buttons
{
  "action" : "delete",
  "resources": [{"id": "101"}, {"id": "102"}]
}
```

### DIFF
--- a/app/controllers/api/custom_buttons_controller.rb
+++ b/app/controllers/api/custom_buttons_controller.rb
@@ -1,0 +1,54 @@
+module Api
+  class CustomButtonsController < BaseController
+    def create_resource(_type, _id, data)
+      klass = collection_class(:custom_buttons)
+      data['resource_action'] = add_resource_action_resource(data['resource_action']) if data.key?('resource_action')
+      data['userid'] = User.current_user.userid
+      klass.create!(data.deep_symbolize_keys)
+    rescue => err
+      raise BadRequestError, "Failed to create new custom button - #{err}"
+    end
+
+    def edit_resource(type, id, data)
+      custom_button = fetch_custom_button(type, id)
+      updated_data = data['resource'] || data
+      updated_data['resource_action'] = add_resource_action_resource(updated_data['resource_action']) if updated_data.key?('resource_action')
+      custom_button.update_attributes!(updated_data.deep_symbolize_keys) if data.present?
+      custom_button
+    rescue => err
+      raise BadRequestError, "Failed to update custom button - #{err}"
+    end
+
+    def options
+      render_options(:custom_buttons, build_custom_button_options)
+    end
+
+    def build_custom_button_options
+    {
+      :custom_button_types => CustomButton::TYPES
+    }
+    end
+
+    private
+
+    def add_resource_action_resource(data)
+      return nil if data.empty?
+      id = parse_id(data, :resource_actions)
+      return resource_search(id, :resource_actions, collection_class(:resource_actions)) if id
+      ResourceAction.create(data)
+    end
+
+    def fetch_custom_button(type, id)
+      resource_search(id, type, collection_class(type))
+    end
+
+    def resource_search(id, type, klass)
+      if ApplicationRecord.compressed_id?(id)
+        super
+      else
+        resource_action = klass.find_by!(:name => id)
+        filter_resource(resource_action, type, klass)
+      end
+    end
+  end
+end
--- a/app/controllers/api/custom_buttons_controller.rb
+++ b/app/controllers/api/custom_buttons_controller.rb
@@ -1,10 +1,10 @@
 module Api
   class CustomButtonsController < BaseController
     def create_resource(_type, _id, data)
-      klass = collection_class(:custom_buttons)
-      data['resource_action'] = add_resource_action_resource(data['resource_action']) if data.key?('resource_action')
-      data['userid'] = User.current_user.userid
-      klass.create!(data.deep_symbolize_keys)
+      custom_button = CustomButton.new(data.except("resource_action"))
+      custom_button.userid = User.current_user.userid
+      custom_button.resource_action = find_or_create_resource_action(data["resource_action"]) if data.key?("resource_action")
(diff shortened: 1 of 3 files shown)


app/controllers/api/custom_buttons_controller.rb
```diff
@@ -0,0 +1,54 @@
...
+      id = parse_id(data, :resource_actions)
+      return resource_search(id, :resource_actions, collection_class(:resource_actions)) if id
+      ResourceAction.create(data)
+    end
+
+    def fetch_custom_button(type, id)
+      resource_search(id, type, collection_class(type))
+    end
+
+    def resource_search(id, type, klass)
```
imtayadeway: Do your tests work if you omit this method? I don't see any for looking up by name - is that something you wanted? I'd prefer is we weren't trying to support that unless expressly needed :)

app/controllers/api/custom_buttons_controller.rb
```diff
@@ -0,0 +1,54 @@
...
+
+    def build_custom_button_options
+      {
+        :custom_button_types => CustomButton::TYPES
+      }
+    end
+
+    private
+
+    def add_resource_action_resource(data)
```
imtayadeway: WDYT about calling this `find_or_create_resource_action` ?

app/controllers/api/custom_buttons_controller.rb
```diff
@@ -0,0 +1,54 @@
+module Api
+  class CustomButtonsController < BaseController
+    def create_resource(_type, _id, data)
+      klass = collection_class(:custom_buttons)
+      data['resource_action'] = add_resource_action_resource(data['resource_action']) if data.key?('resource_action')
```
imtayadeway: Rather than mutate the data that's passed into this method (which can lead to surprises for the caller). what do you think about creating a new object:

```ruby
custom_button = CustomButton.new(data.except("resource_action"))
custom_button.userid = User.current_user.userid
custom_button.resource_action = add_resource_action_resource(data["resource_action"]) if data.key?("resource_action")
custom_button.save!
```

app/controllers/api/custom_buttons_controller.rb
AparnaKarve: The tests do work minus this. I will remove it. Thanks :)

app/controllers/api/custom_buttons_controller.rb
imtayadeway: Also, did you consider adding `accepts_nested_attributes_for :resource_action` to `CustomButton`? I think that could be a good idea, since we don't expose `ResourceAction` in the API - this code is jumping through a lot of hoops just to extract the id from the request body (it shouldn't ever be an href, which is half of what `parse_id` does). Accepting nested attributes would make this controller incredibly simple 😁

spec/requests/custom_buttons_spec.rb
```diff
@@ -0,0 +1,206 @@
...
+
+  describe 'POST /api/custom_buttons' do
+    it 'can create a new custom button' do
+      api_basic_authorize collection_action_identifier(:custom_buttons, :create)
+
+      cb_rec = {
+        'name'             => 'Generic Object Custom Button',
+        'description'      => 'Generic Object Custom Button description',
+        'applies_to_class' => 'GenericObjectDefinition',
+        'options'          => {
```
imtayadeway: Can you add a resource action to this request just to prove that that works?

spec/requests/custom_buttons_spec.rb
```diff
@@ -0,0 +1,206 @@
...
+  end
+
+  describe 'POST /api/custom_buttons/:id' do
+    it 'can update a custom buttons by id' do
+      api_basic_authorize action_identifier(:custom_buttons, :edit)
+
+      request = {
+        'action'      => 'edit',
+        'name'        => 'Generic Object Custom Button Updated',
+        'description' => 'Generic Object Custom Button description Updated',
```
imtayadeway: Similarly if you can edit the resource action details here too that would be great 😁

app/controllers/api/custom_buttons_controller.rb
AparnaKarve: Ok, addressed both in 1d85410
The change above, required an alteration in the spec.

app/controllers/api/custom_buttons_controller.rb
AparnaKarve: Addressed in 1d85410.
This change required a spec alteration as well.

spec/requests/custom_buttons_spec.rb
AparnaKarve: in 5104806

spec/requests/custom_buttons_spec.rb
AparnaKarve: in 5104806

AparnaKarve:
@imtayadeway I created https://github.com/ManageIQ/manageiq/pull/16272 per your recommendation.

abellotti:
LGTM!!  The only thing is the codeclimate issue, looks like a similar pattern with pictures controller (3-liner which can be generalized), we can refactor in a separate PR.

Thanks @AparnaKarve for the API Enhancement  (and @imtayadeway for the diligent review 🎶 ) 👍

PR #141, opened 2017-10-24
Change from Ruby 2.3.1 to 2.3.5

chessbyte:


chrisarcand:
@chessbyte We don't actually want to do this. The appliance version is currently 2.3.1, bound by SCL.

PR #142, opened 2017-10-24
Move validate_provider_class into providers controller

imtayadeway:
This only relates to providers so that's where it should go.

@miq-bot assign @abellotti


abellotti:
LGTM!! Thanks @imtayadeway   🎶

PR #143, opened 2017-10-24
Extract a #create action

imtayadeway:
There are currently a few different ways to create a resource:

```http
POST /api/<collection>

{
  "action": "create", 
  "some": "attributes"
}
```

```http
POST /api/<collection>

{
  "action": "create", 
  "resource": {
    "some": "attributes"
  }
}
```

```http
POST /api/<collection>

{
  "action": "create", 
  "resources": [
    {"some": "attributes"},
    {"some": "other attributes"}
  ]
}
```

In all three, the "action" can also be "add", or omitted entirely.

Perhaps surprisingly, these all get routed through the `#update` action, which implicitly handles both the creation and updating of resources.

Most Rails developers would expect this to be handled through the more conventional `#create` action. I've added this functionality here, extracting the create-specific code into this new action.

There are a few benefits that I'd like to highlight:

1. it's more familiar to Rails developers. Having an `#update` action that is also responsible for creating resources is very confusing/surprising
1. the newly extracted `#create` action can be hooked into, overridden, etc.. without disturbing any of the deeply-nested `#update` logic
1. this teases apart some of that logic so that we're not mixing in knowledge of how to create resources in a low-level method called `#update_one_collection`, etc..
1. it should now be much easier to understand/read since most of the indirection involved in the `#update` action could be eliminated for `#create`

@miq-bot add-label refactoring, technical debt
@miq-bot assign @chrisarcand

### DIFF
--- a/app/controllers/api/base_controller.rb
+++ b/app/controllers/api/base_controller.rb
@@ -59,6 +59,10 @@ module Api
       render_collection(@req.subject, res, opts)
     end
 
+    def create
+      render_resource(@req.collection.to_sym, update_collection(@req.subject.to_sym, @req.subject_id))
+    end
+
     def show
       klass = collection_class(@req.subject)
       opts  = {:name => @req.subject, :is_subcollection => @req.subcollection?, :expand_actions => true}
--- a/app/controllers/api/base_controller.rb
+++ b/app/controllers/api/base_controller.rb
@@ -77,9 +77,8 @@ module Api
           create_resource(@req.subject.to_sym, nil, r)
         end
       end
-      created = {"results" => results}
 
-      render_resource(@req.collection.to_sym, created)
+      render_resource(@req.collection.to_sym, "results" => results)
     end
 
     def show
--- a/app/controllers/api/base_controller.rb
+++ b/app/controllers/api/base_controller.rb
@@ -61,10 +61,8 @@ module Api
 
     def create
       target = target_resource_method(@req.subject.to_sym, @req.action)
-      render_resource(
-        @req.collection.to_sym,
-        get_and_update_multiple_collections(@req.subcollection?, target, @req.subject.to_sym)
-      )
+      created = get_and_update_multiple_collections(@req.subcollection?, target, @req.subject.to_sym)
+      render_resource(@req.collection.to_sym, created)
     end
 
     def show
--- a/app/controllers/api/base_controller.rb
+++ b/app/controllers/api/base_controller.rb
@@ -60,18 +60,11 @@ module Api
     end
 
     def create
-      resources = []
-      if @req.json_body.key?("resources")
-        resources += @req.json_body["resources"]
-      else
-        resources << json_body_resource
-      end
-
-      if resources.all?(&:blank?)
+      if @req.resources.all?(&:blank?)
         raise BadRequestError, "No #{@req.subject} resources were specified for the create action"
       end
 
-      results = resources.collect do |r|
+      results = @req.resources.collect do |r|
         next if r.blank?
         if parse_id(r, @req.subject.to_sym)
           raise BadRequestError, "Resource id or href should not be specified for creating a new #{@req.subject}"
--- a/app/controllers/api/base_controller.rb
+++ b/app/controllers/api/base_controller.rb
@@ -77,7 +77,12 @@ module Api
         if parse_id(r, @req.subject.to_sym)
           raise BadRequestError, "Resource id or href should not be specified for creating a new #{@req.subject}"
         end
-        update_one_collection(@req.subcollection?, target, @req.subject.to_sym, nil, r)
+
+        if @req.subcollection?
+          send(target, parent_resource_obj, @req.subject.to_sym, nil, r)
+        else
+          send(target, @req.subject.to_sym, nil, r)
+        end
       end
       created = {"results" => results}
 
--- a/app/controllers/api/base_controller.rb
+++ b/app/controllers/api/base_controller.rb
@@ -60,7 +60,6 @@ module Api
     end
 
     def create
-      target = target_resource_method(@req.subject.to_sym, @req.action)
       resources = []
       if @req.json_body.key?("resources")
         resources += @req.json_body["resources"]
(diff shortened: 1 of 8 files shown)


app/controllers/api/base_controller.rb
```diff
@@ -59,6 +59,22 @@ def index
       render_collection(@req.subject, res, opts)
     end
 
+    def create
+      if @req.resources.any? { |r| parse_id(r, @req.subject.to_sym) }
+        raise BadRequestError, "Resource id or href should not be specified for creating a new #{@req.subject}"
+      end
```
chrisarcand: You should leave this for a followup PR. There's a whole can of worms to fix here to try and stop such errors (we should really be doing a bulk update in a transaction to rollback on *any* failures, not just specifying an ID when you shouldn't)

app/controllers/api/base_controller.rb
```diff
@@ -59,6 +59,22 @@ def index
...
+
+      results = @req.resources.collect do |r|
+        if @req.subcollection?
+          send("#{@req.subject}_create_resource", parent_resource_obj, @req.subject.to_sym, nil, r)
```
chrisarcand: Continue using `target_resource_method` instead of interpolating yourself here; if we want to remove that it's out of scope for what you're doing here.

app/controllers/api/base_controller.rb
```diff
@@ -59,6 +59,22 @@ def index
       render_collection(@req.subject, res, opts)
     end
 
+    def create
```
chrisarcand: This method should probably raise on trying to add an empty set of things, like the update version does, to keep parity (even if temporary).

app/controllers/api/base_controller/manager.rb
```diff
@@ -122,14 +111,7 @@ def update_multiple_collections(is_subcollection, target, type, resources)
...
         results = resources.each.collect do |r|
           next if r.blank?
-
-          rid = parse_id(r, type)
-          create_or_add_action = %w(create add).include?(action)
-          if rid && create_or_add_action
-            raise BadRequestError, "Resource id or href should not be specified for creating a new #{type}"
-          elsif !rid && !create_or_add_action
-            rid = parse_by_attr(r, type)
-          end
```
chrisarcand: This logic probably can't be removed, because in the routes you don't actually have all types of legacy requests routing to the new create method; that's because you need to handle passing an ID (on create... I know...), but doing so also ruins the point of the PR because you're trying to do something more conventional. 

This logic will have to remain until we're ready to actually state breaking changes and remove it.

app/controllers/api/base_controller/parser/request_adapter.rb
```diff
@@ -59,7 +59,7 @@ def hide?(thing)
         def json_body
           @json_body ||= begin
                            body = @request.body.read if @request.body
-                           body.blank? ? {} : JSON.parse(body)
+                           body.blank? ? {} : Hash(JSON.parse(body))
```
Fryguy: Now that we are on Ruby 2.3, we can use Array#to_h which will convert associative arrays into Hashes.  I'm assuming that's why you have this Hash() bit here?

config/routes.rb
```diff
@@ -68,6 +69,7 @@
             when :delete
               delete "/:c_id/#{subcollection_name}/:s_id", :action => :destroy
             when :post
+              post "/:c_id/#{subcollection_name}", :action => :create, :constraints => Api::CreateConstraint.new
```
Fryguy: I wonder if we can use similar constraints to deal with the exploding number of routes on the UI side 🤔  cc @jrafanie

Fryguy:
❤️  this moving towards a more "standard" Rails experience 👍

app/controllers/api/base_controller/parser/request_adapter.rb
imtayadeway: I'm actually putting this back for now because it causes other existing tests to fail 😩 . Basically:

* When the request body is empty, `@request.body # => "null"`, which is non-blank, but `JSON.parse("null") # => nil`
* Sometimes we are sending a top-level array (:scream:), which can't be converted implicitly into a hash

I made the test I wrote pending instead, because I don't want to fix that as part of this refactoring: https://github.com/ManageIQ/manageiq-api/pull/143/files#diff-c31ee124141903c07b03421343749667R2

lib/api/create_constraint.rb
```diff
@@ -0,0 +1,9 @@
+module Api
+  class CreateConstraint
+    def matches?(request)
+      Hash(JSON.parse(request.body.read)).fetch("action", "create").in?(%w(create add))
```
chrisarcand: @imtayadeway Why do you need the `Hash()` call here? `parse` defaults to `Hash`, no?

spec/requests/creating_spec.rb
```diff
@@ -0,0 +1,56 @@
+RSpec.describe "Creating" do
+  it "responds with Bad Request if there are no resources at the top level" do
+    skip("Currently results in an internal server error")
```
chrisarcand: 😬 ??

spec/requests/creating_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+
+  it "responds with Bad Request if we specify an href in the body" do
+    api_basic_authorize(action_identifier(:roles, :create, :collection_actions))
+
+    post(api_roles_url, :params => {:href => api_vm_url(nil, 123), :name => "Alice's Group"})
+
+    expect(response.parsed_body).to include_error_with_message("Resource id or href should not be specified for creating a new roles")
+    expect(response).to have_http_status(:bad_request)
+  end
+end
```
chrisarcand: No tests for the successful case?

lib/api/create_constraint.rb
imtayadeway: It does, unless the body is empty:

```ruby
JSON.parse("null") # => nil
```

spec/requests/creating_spec.rb
imtayadeway: Yepppppp. See https://github.com/ManageIQ/manageiq-api/pull/143#discussion_r146954819 for more on this

config/routes.rb
jrafanie: Hopefully... I don't really know routes that well.

PR #144, opened 2017-10-25
Return only id attributes if specified

jntullo:
Fixes [BZ](https://bugzilla.redhat.com/show_bug.cgi?id=1437201) where if you specify `expand=resources&attributes=id`, it returns all of the attributes.

Started to go down the 🐰 hole of refactoring attribute selection (it could use it), but felt it was best to get this fixed separately.

@miq-bot assign @imtayadeway

### DIFF
--- a/app/controllers/api/base_controller/renderer.rb
+++ b/app/controllers/api/base_controller/renderer.rb
@@ -367,6 +367,7 @@ module Api
         physical_attrs, virtual_attrs = [], []
         attrs = attribute_selection
         return [physical_attrs, virtual_attrs] if resource.kind_of?(Hash) || attrs == 'all'
+        return [attrs, virtual_attrs] if (attrs - ID_ATTRS).empty?
 
         attrs.each do |attr|
           if attr_physical?(resource, attr) || attr == 'actions'
--- a/spec/requests/querying_spec.rb
+++ b/spec/requests/querying_spec.rb
@@ -85,6 +85,25 @@ describe "Querying" do
       expect(response).to have_http_status(:bad_request)
     end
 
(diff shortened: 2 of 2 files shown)


jntullo:
@miq-bot add_label bug

spec/requests/querying_spec.rb
```diff
@@ -85,6 +85,25 @@ def create_vms_by_name(names)
       expect(response).to have_http_status(:bad_request)
     end
 
+    it 'returns only id attributes if specified on a collection' do
+      create_vms_by_name %w(bb)
```
imtayadeway: Since the name isn't important here, WDYT about `vm = FactoryGirl.create(:vm)` here, then check for

```ruby
{"href" => api_vm_url(nil, vm), "id" => vm.id}
```
below?

app/controllers/api/base_controller/renderer.rb
```diff
@@ -367,6 +367,7 @@ def validate_attr_selection(resource)
         physical_attrs, virtual_attrs = [], []
         attrs = attribute_selection
         return [physical_attrs, virtual_attrs] if resource.kind_of?(Hash) || attrs == 'all'
+        return [attrs, virtual_attrs] if (attrs - ID_ATTRS).empty?
```
imtayadeway: Does removing the condition in the ternary on 382/3 (and defaulting to the RHS) have the same effect? I'm wondering if that would be preferable - if you could fix this by only removing code 😁 ✂️

jntullo: @imtayadeway I tried that 😢  it was the 🐰 hole I started to go down, actually. I can follow up with refactoring, though

PR #145, opened 2017-10-25
Return the correct href slugs for requests and service requests

jntullo:
Another [bug](https://bugzilla.redhat.com/show_bug.cgi?id=1503852) that could probably use some refactoring...

When the `href_slug` is generated, it is using [name_for_subclass](https://github.com/ManageIQ/manageiq-api/blob/70e15767af07a45a3d49af87a6b34ade789ea440/lib/api/collection_config.rb#L80) which is essentially finding the collection name from the resource class.

This is causing problems in numerous places where resource classes are applicable to multiple collections - as can be seen [below](https://github.com/ManageIQ/manageiq-api/compare/master...jntullo:bz/return_correct_request_slug?expand=1#diff-8174336b6ca531d40c4f4ca9b7f06979) where an automation request being returned from the `/requests` endpoint returns the `automation_request/:id` slug, when it should be `requests`

@miq-bot add_label bug
@miq-bot assign @imtayadeway

### DIFF
--- a/app/controllers/api/requests_controller.rb
+++ b/app/controllers/api/requests_controller.rb
@@ -66,6 +66,10 @@ module Api
       {:requester => User.current_user}
     end
 
+    def fetch_requests_href_slug(resource)
+      "requests/#{resource.id}"
+    end
+
     private
 
     def authorize_request(typed_request_klass)
--- a/spec/requests/requests_spec.rb
+++ b/spec/requests/requests_spec.rb
@@ -474,6 +474,17 @@ RSpec.describe "Requests API" do
     end
   end
 
+  context 'href_slug' do
+    it 'returns the correct slug for a request' do
+      api_basic_authorize action_identifier(:requests, :read, :resource_actions, :get)
+      request = FactoryGirl.create(:automation_request, :requester => @user)
+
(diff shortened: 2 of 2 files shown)


abellotti:
ping @imtayadeway

imtayadeway:
@abellotti I chatted to @jntullo offline about this but didn't summarize anywhere so I'll do this now. I've added a comment in the BZ here: https://bugzilla.redhat.com/show_bug.cgi?id=1503852#c3. To summarize: it describes two behaviors that it considers to be bugs - I think only one of them is. I think it's also a bug to do with `Api::Utils.build_href_slug` - we will still need to fix that for cases where we're generating href slugs outside of an HTTP request.

app/controllers/api/requests_controller.rb
```diff
@@ -66,6 +66,10 @@ def requests_search_conditions
       {:requester => User.current_user}
     end
 
+    def fetch_requests_href_slug(resource)
+      "requests/#{resource.id}"
```
imtayadeway: I still feel that it's not up to the controller to change the resource's href slug on the fly - is this open for discussion?

jntullo:
Closing in favor of https://github.com/ManageIQ/manageiq-api

PR #149, opened 2017-10-29
Allow assigning/un-assigning of alert definitions to alert profiles

elad661:
This commit allows assigning and un-assigning of alert definitions to alert definition profiles.

**Usage examples:**
to assign an alert definition to a profile:

POST /api/alert_definition_profiles/:id/alert_definitions
```json 
{"action": "assign",
 "resource": {"href": "/api/alert_definition/:alert_definition_id"}}
```

to un-assign an alert definition from a profile:

POST /api/alert_definition_profiles/:id/alert_definitions
```json 
{"action": "unassign",
 "resource": {"href": "/api/alert_definition/:alert_definition_id"}}
```

### DIFF
--- a/app/controllers/api/subcollections/alert_definitions.rb
+++ b/app/controllers/api/subcollections/alert_definitions.rb
@@ -4,6 +4,26 @@ module Api
       def alert_definitions_query_resource(object)
         object.respond_to?(:miq_alerts) ? object.miq_alerts : []
       end
+
+      def alert_definitions_assign_resource(object, type, id = nil, _data = nil)
+        alert = resource_search(id, type, collection_class(type))
+        object.add_member(alert)
+        result = action_result(true, "Assigning alert_definition #{id} to profile #{object.id}")
+        add_parent_href_to_result(result)
+        add_subcollection_resource_to_result(result, type, alert)
+      rescue => err
+        action_result(false, err.to_s)
+      end
+
+      def alert_definitions_unassign_resource(object, type, id = nil, _data = nil)
+        alert = resource_search(id, type, collection_class(type))
+        success = object.remove_member(alert).present?
+        result = action_result(success, "Unassigning alert_definition #{id} from profile #{object.id}")
+        add_parent_href_to_result(result)
+        add_subcollection_resource_to_result(result, type, alert)
+      rescue => err
+        action_result(false, err.to_s)
+      end
     end
   end
 end
--- a/config/api.yml
+++ b/config/api.yml
@@ -116,6 +116,12 @@
       :delete:
       - :name: delete
         :identifier: alert_definition_profile_delete
+    :alert_definitions_subcollection_actions:
+      :post:
+      - :name: assign
+        :identifier: alert_definition_profile_edit
+      - :name: unassign
+        :identifier: alert_definition_profile_edit
   :alert_definitions:
     :description: Alert Definitions
     :identifier: alert_definition
--- a/spec/requests/alert_definitions_spec.rb
+++ b/spec/requests/alert_definitions_spec.rb
@@ -306,4 +306,101 @@ describe "Alerts Definition Profiles API" do
     expect(alert_definition_profiles.first.reload.description).to eq("Updated Test Alert Profile 1")
     expect(alert_definition_profiles.second.reload.description).to eq("Updated Test Alert Profile 2")
   end
+
+  it "assigns alert definitions to a profile" do
+    api_basic_authorize collection_action_identifier(:alert_definition_profiles, :edit)
+    alert_definition_profile = FactoryGirl.create(:miq_alert_set)
+    alert_definitions = FactoryGirl.create_list(:miq_alert, 2)
+
+    post(
+      api_alert_definition_profile_alert_definitions_url(nil, alert_definition_profile),
+      :params => {
+        :action    => "assign",
+        :resources => [
+          {:id => alert_definitions[0].id},
+          {:href => api_alert_definition_url(nil, alert_definitions[1])}
+        ]
+      }
+    )
+
+    expect_multiple_action_result(2)
+    expected = {
+      'results' => [
+        a_hash_including("success"               => true,
+                         "message"               => "Assigning alert_definition #{alert_definitions[0].id} to profile #{alert_definition_profile.id}",
+                         "href"                  => api_alert_definition_profile_url(nil, alert_definition_profile),
+                         "alert_definition_href" => api_alert_definition_url(nil, alert_definitions[0])),
+        a_hash_including("success"               => true,
+                         "message"               => "Assigning alert_definition #{alert_definitions[1].id} to profile #{alert_definition_profile.id}",
+                         "href"                  => api_alert_definition_profile_url(nil, alert_definition_profile),
+                         "alert_definition_href" => api_alert_definition_url(nil, alert_definitions[1]))
+      ]
+    }
+    expect(response.parsed_body).to include(expected)
+  end
+
+  it "unassigns alert definitions from a profile" do
+    api_basic_authorize collection_action_identifier(:alert_definition_profiles, :edit)
+    alert_definition_profile = FactoryGirl.create(:miq_alert_set)
+    alert_definitions = FactoryGirl.create_list(:miq_alert, 2)
+    alert_definition_profile.add_member(alert_definitions[0])
+    alert_definition_profile.add_member(alert_definitions[1])
+
+    post(
+      api_alert_definition_profile_alert_definitions_url(nil, alert_definition_profile),
+      :params => {
+        :action    => "unassign",
+        :resources => [
+          {:id => alert_definitions[0].id},
+          {:href => api_alert_definition_url(nil, alert_definitions[1])}
+        ]
+      }
+    )
+
+    expect_multiple_action_result(2)
+    expected = {
+      'results' => [
+        a_hash_including("success"               => true,
+                         "message"               => "Unassigning alert_definition #{alert_definitions[0].id} from profile #{alert_definition_profile.id}",
+                         "href"                  => api_alert_definition_profile_url(nil, alert_definition_profile),
+                         "alert_definition_href" => api_alert_definition_url(nil, alert_definitions[0])),
+        a_hash_including("success"               => true,
+                         "message"               => "Unassigning alert_definition #{alert_definitions[1].id} from profile #{alert_definition_profile.id}",
+                         "href"                  => api_alert_definition_profile_url(nil, alert_definition_profile),
+                         "alert_definition_href" => api_alert_definition_url(nil, alert_definitions[1]))
+      ]
+    }
+    expect(response.parsed_body).to include(expected)
+  end
+
+  it "unassigns an alert definition from a profile to which it wasn't assigned" do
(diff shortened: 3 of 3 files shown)


elad661:

@moolitayer please review

elad661:
cc @abellotti

moolitayer:
@imtayadeway please review

config/api.yml
```diff
@@ -121,6 +121,10 @@
       :delete:
       - :name: delete
         :identifier: alert_definition_profile_delete
+    :alert_definitions_subcollection_actions:
+      :post:
+      - :name: assign
```
imtayadeway: Are there no RBAC identifiers for there actions? This seems surprising, since you need at least some privileges just to read an alert definition.

spec/requests/alert_definitions_spec.rb
```diff
@@ -306,4 +306,24 @@
...
+
+  it "assings an alert definition to a profile" do
+    api_basic_authorize
+    alert_definition_profile = FactoryGirl.create(:miq_alert_set)
+    alert_definition = FactoryGirl.create(:miq_alert)
+    post(api_alert_definition_profile_url(nil, alert_definition_profile) + "/alert_definitions", :params => gen_request(:assign, alert_definition))
+    expect(response).to have_http_status(:ok)
+    expect(response.parsed_body["results"].count).to eq(1)
+    expect(alert_definition_profile.reload.members.first).to eq(alert_definition)
+  end
```
imtayadeway: Please leave an empty line between tests.

spec/requests/alert_definitions_spec.rb
```diff
@@ -306,4 +306,24 @@
     expect(alert_definition_profiles.first.reload.description).to eq("Updated Test Alert Profile 1")
     expect(alert_definition_profiles.second.reload.description).to eq("Updated Test Alert Profile 2")
   end
+
+  it "assings an alert definition to a profile" do
+    api_basic_authorize
```
imtayadeway: It greatly helps with readibility if you can group these steps into each phase of the four-phase test, with an empty line separating each:

```ruby
api_basic_authorize
alert_definition_profile = FactoryGirl.create(:miq_alert_set)
alert_definition = FactoryGirl.create(:miq_alert)

post(api_alert_definition_profile_url(nil, alert_definition_profile) + "/alert_definitions", :params => gen_request(:assign, alert_definition))

expect(response.parsed_body["results"].count).to eq(1)
expect(alert_definition_profile.reload.members.first).to eq(alert_definition)
```

spec/requests/alert_definitions_spec.rb
```diff
@@ -306,4 +306,24 @@
...
+    alert_definition_profile = FactoryGirl.create(:miq_alert_set)
+    alert_definition = FactoryGirl.create(:miq_alert)
+    post(api_alert_definition_profile_url(nil, alert_definition_profile) + "/alert_definitions", :params => gen_request(:assign, alert_definition))
+    expect(response).to have_http_status(:ok)
+    expect(response.parsed_body["results"].count).to eq(1)
+    expect(alert_definition_profile.reload.members.first).to eq(alert_definition)
```
imtayadeway: Prefer to make this (and other) expectations on the response itself. Here, you're testing something about the implementation, which belongs in a lower level test. Request specs should be more concerned with how this behavior is revealed through the response - perhaps you can examine the contents of `"results"` and put some expectations there instead?

spec/requests/alert_definitions_spec.rb
```diff
@@ -306,4 +306,24 @@
     expect(alert_definition_profiles.first.reload.description).to eq("Updated Test Alert Profile 1")
     expect(alert_definition_profiles.second.reload.description).to eq("Updated Test Alert Profile 2")
   end
+
+  it "assings an alert definition to a profile" do
+    api_basic_authorize
+    alert_definition_profile = FactoryGirl.create(:miq_alert_set)
+    alert_definition = FactoryGirl.create(:miq_alert)
+    post(api_alert_definition_profile_url(nil, alert_definition_profile) + "/alert_definitions", :params => gen_request(:assign, alert_definition))
```
imtayadeway: Prefer `api_alert_definition_profile_alert_definitions_url(nil, alert_definition_profile)`

config/api.yml
elad661: I don't see any in `miq_product_features.yml`, so I don't think there are. Should I use the closest thing I can find, `alert_definition_edit`, or keep it without an identifier?

spec/requests/alert_definitions_spec.rb
elad661: done

spec/requests/alert_definitions_spec.rb
elad661: I changed the tests accordingly, and you were right - this helped me find a bug in the unassign method (which I now fixed).

config/api.yml
imtayadeway: It sounds like you may need to create new ones if they don't exist.

elad661: Oh, I was wrong, it does exist, there's `alert_profile_assign`. I'll modify the commit accordingly

elad661: `alert_profile_assign` appears to be unrelated. The UI uses `alert_definition_profile_edit` for assignment/unassignment, so I changed my code to use that instead.

elad661:
@imtayadeway can you please re-review? thanks

bazulay:
@imtayadeway @abellotti  PTAL

bazulay:
@gtanzillo  ping

spec/requests/alert_definitions_spec.rb
```diff
@@ -306,4 +306,46 @@
...
+  end
+
+  it "unassings an alert definition from a profile to which it wasn't assigned" do
+    api_basic_authorize collection_action_identifier(:alert_definition_profiles, :edit)
+    alert_definition_profile = FactoryGirl.create(:miq_alert_set)
+    alert_definition = FactoryGirl.create(:miq_alert)
+
+    post(api_alert_definition_profile_alert_definitions_url(nil, alert_definition_profile), :params => gen_request(:unassign, alert_definition))
+
+    expect(response).to have_http_status(:ok)
```
imtayadeway: This looks OK, but can you check whether this is consistent with unassigning behavior in other APIs (tags, for instance).

spec/requests/alert_definitions_spec.rb
```diff
@@ -306,4 +306,46 @@
...
+
+  it "assings an alert definition to a profile" do
+    api_basic_authorize collection_action_identifier(:alert_definition_profiles, :edit)
+    alert_definition_profile = FactoryGirl.create(:miq_alert_set)
+    alert_definition = FactoryGirl.create(:miq_alert)
+
+    post(api_alert_definition_profile_alert_definitions_url(nil, alert_definition_profile), :params => gen_request(:assign, alert_definition))
+
+    expect(response).to have_http_status(:ok)
+    expect(response.parsed_body["results"].count).to eq(1)
```
imtayadeway: You can express this and the following lines as one expectation:

```ruby
expect(response.parsed_body).to include("results" => [a_hash_including("href" => api_alert_definition_profile_alert_definition_url(nil, alert_definition_profile, alert_definition)]) # => guarantees that there is only one result
```

app/controllers/api/subcollections/alert_definitions.rb
```diff
@@ -4,6 +4,15 @@ module AlertDefinitions
       def alert_definitions_query_resource(object)
         object.respond_to?(:miq_alerts) ? object.miq_alerts : []
       end
+
+      def alert_definitions_assign_resource(object, _type, id = nil, _data = nil)
+        object.add_member(MiqAlert.find(id))
```
imtayadeway: `MiqAlert.find` will raise a `RecordNotFound` error in the case that this id does not match a record in the db. Can you ensure that there is test coverage for this and the unassign action and that you have the desired behavior?

You may also need to use `result_search` here (which delegates to RBAC), since a bare `find` will ignore all such authorization concerns.

spec/requests/alert_definitions_spec.rb
abellotti: assign, unassign actions are fine, consistent with tagging, policies, etc.  but the responses should use
the action result signature, i.e. ('success' : ..., 'message' : ..., 'href' : ..., 'alert_definition_href' : ....,)

abellotti: if these can have multiple alert_definition_profile -> alert_definition associations, bulk request tests are also needed.

app/controllers/api/subcollections/alert_definitions.rb
```diff
@@ -4,6 +4,15 @@ module AlertDefinitions
...
+      end
+
+      def alert_definitions_unassign_resource(object, _type, id = nil, _data = nil)
+        alert = MiqAlert.find(id)
```
abellotti: same as above, resource_search for RBAC instead of find. Thanks.

elad661:
Thanks for the review! I think I addressed all your feedback, but please let me know if I missed something.

spec/requests/alert_definitions_spec.rb
```diff
@@ -306,4 +306,83 @@
     expect(alert_definition_profiles.first.reload.description).to eq("Updated Test Alert Profile 1")
     expect(alert_definition_profiles.second.reload.description).to eq("Updated Test Alert Profile 2")
   end
+
+  it "assigns alert definitions to a profile" do
+    api_basic_authorize collection_action_identifier(:alert_definition_profiles, :edit)
+    alert_definition_profile = FactoryGirl.create(:miq_alert_set)
+    alert_definitions = FactoryGirl.create_list(:miq_alert, 2)
+
+    post(api_alert_definition_profile_alert_definitions_url(nil, alert_definition_profile), :params => gen_request(:assign, alert_definitions))
```
imtayadeway: If I understand this correctly, this request provides more information than is sufficient.

```ruby
post(
  api_alert_definition_profile_alert_definitions_url(nil, alert_definition_profile),
  :params => {
    :action => "assign",
    :resources => [
      {:id => alert_definitions[0].id},
      {:id => alert_definitions[1].id}
    ]
  }
)
```

...I think would suffice

spec/requests/alert_definitions_spec.rb
```diff
@@ -306,4 +306,83 @@
...
+  end
+
+  it "unassigns alert definitions from a profile" do
+    api_basic_authorize collection_action_identifier(:alert_definition_profiles, :edit)
+    alert_definition_profile = FactoryGirl.create(:miq_alert_set)
+    alert_definitions = FactoryGirl.create_list(:miq_alert, 2)
+    alert_definition_profile.add_member(alert_definitions[0])
+    alert_definition_profile.add_member(alert_definitions[1])
+
+    post(api_alert_definition_profile_alert_definitions_url(nil, alert_definition_profile), :params => gen_request(:unassign, alert_definitions))
```
imtayadeway: As above

elad661:
@imtayadeway Thanks for the review, much appreciated! I updated it to address your feedback, let me know if there's anything else needed.

elad661:
Thanks @imtayadeway!
rebased as requested.

abellotti:
Thanks @elad661 for the API Enhancement, LGTM !!  👍

PR #150, opened 2017-10-30
Return correct href for collections on Index

jntullo:
Fixes issues where the API prefix isn't passed back in the API index for collections

@miq-bot add_label bug
@miq-bot assign @abellotti

### DIFF
--- a/spec/requests/authentication_spec.rb
+++ b/spec/requests/authentication_spec.rb
@@ -55,6 +55,21 @@ describe "Authentication API" do
       }
       expect(response.parsed_body).to include(expected)
     end
+
+    it "returns a correctly formatted collection hrefs" do
+      api_basic_authorize
+
+      get api_entrypoint_url
+
+      collection_names = Api::ApiConfig.collections.to_h.select { |_, v| v.options.include?(:collection) }.keys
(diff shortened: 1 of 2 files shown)


spec/requests/authentication_spec.rb
```diff
@@ -55,6 +55,16 @@
       }
       expect(response.parsed_body).to include(expected)
     end
+
+    it "returns a correctly formatted collection hrefs" do
+      api_basic_authorize
+
+      get api_entrypoint_url
+
+      response.parsed_body["collections"].each do |collection|
```
imtayadeway: WDYT of doing something like this for the body here:

```ruby
      collection_names = Api::ApiConfig.collections.to_h.select { |_, v| v.options.include?(:collection) }.keys
      hrefs = collection_names.collect { |name| url_for(:controller => name, :action => "index") }
      expected = {
        "collections" => a_collection_containing_exactly(
          *hrefs.collect { |href| a_hash_including("href" => href) }
        )
      }
```

Aithough there's a bit more going on there, I like it because:

1. It verifies against an external source of truth (if for instance there were no collections in your response, you'd still pass)
2. It looks for an exact match on the href
3. It's a single expectation, which will give you one failure and a diff in the event of a regression

abellotti:
Thanks @jntullo for fixing this.

PR #153, opened 2017-10-30
Fix pending specs

imtayadeway:
Some internal server errors were discovered when testing in #143 for edge cases. This is the fix for both of those.

@miq-bot add-label bug
@miq-bot assign @abellotti


abellotti:
LGTM!! Thanks @imtayadeway for fixing the specs. 🎵

PR #159, opened 2017-10-31
Bumping up version to 3.1.0-pre for the H-release

abellotti:
continuing development on the master branch for the H-release.

### DIFF
--- a/config/api.yml
+++ b/config/api.yml
@@ -3,11 +3,11 @@
   :module: api
   :name: API
   :description: REST API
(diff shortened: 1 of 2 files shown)


chrisarcand:
Given that the API isn't truly versioned (due to the unrestricted git based dependencies we have going on), perhaps we should just assume 4.0.0-pre (in anticipation for the H release), and if we happen to not make any breaking changes at that time it can be amended.

PR #162, opened 2017-10-31
DRY up Api version

imtayadeway:
There should now be only one thing that needs to be updated when we bump the version 🎉 

TBH though I'm not totally convinced https://github.com/ManageIQ/manageiq-api/blob/master/lib/manageiq/api/version.rb#L3 is the right place for this. Are we saying that the manageiq-api gem's version is and will always be the same as the "API version"?

@miq-bot add-label refactoring, technical debt
@miq-bot assign @abellotti

### DIFF
--- a/lib/api.rb
+++ b/lib/api.rb
@@ -1,4 +1,5 @@
 module Api
+  SUPPORTED_VERSIONS = [ManageIQ::Api::VERSION].freeze
   VERSION_CONSTRAINT = /v\d+(\.[\da-zA-Z]+)*(\-[\da-zA-Z\-\.]+)?/
   VERSION_REGEX = /\A#{VERSION_CONSTRAINT}\z/
 
--- a/app/controllers/api/api_controller.rb
+++ b/app/controllers/api/api_controller.rb
@@ -23,10 +23,10 @@ module Api
     private
 
     def entrypoint_versions
-      ApiConfig.version.definitions.select(&:ident).collect do |version_specification|
+      Api::SUPPORTED_VERSIONS.collect do |version|
         {
-          :name => version_specification[:name],
-          :href => "#{@req.base}#{@req.prefix(false)}/#{version_specification[:ident]}"
+          :name => version,
+          :href => "#{@req.base}#{@req.prefix(false)}/v#{version}"
         }
       end
     end
--- a/app/controllers/api/api_controller.rb
+++ b/app/controllers/api/api_controller.rb
@@ -8,7 +8,7 @@ module Api
       res = {
         :name         => ApiConfig.base.name,
         :description  => ApiConfig.base.description,
-        :version      => ApiConfig.base.version,
+        :version      => ManageIQ::Api::VERSION,
         :versions     => entrypoint_versions,
         :settings     => user_settings,
         :identity     => auth_identity,
--- a/app/controllers/api/base_controller/parser.rb
+++ b/app/controllers/api/base_controller/parser.rb
@@ -11,7 +11,7 @@ module Api
         # API Version Validation
         if @req.version
           vname = @req.version
-          unless ApiConfig.version.definitions.collect { |vent| vent[:name] }.include?(vname)
+          unless Api::SUPPORTED_VERSIONS.include?(vname)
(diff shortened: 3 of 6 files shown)


lib/api.rb
```diff
@@ -1,4 +1,6 @@
 module Api
+  VERSION = ManageIQ::Api::VERSION
+  SUPPORTED_VERSIONS = [VERSION].freeze
```
abellotti: this is fine,  essentially yes, the Gem version is the current (default) version, the above SUPPORTED_VERSIONS could be extended to include versions in addition to VERSION that the code differentiates. i.e. /api/v4.0.0/collection  vs. /api/v3.0.0/collection.   Code defined vs. config defined is better anyways.

lib/api.rb
```diff
@@ -1,4 +1,5 @@
 module Api
+  VERSION = ManageIQ::Api::VERSION
```
abellotti: why is this needed ?

imtayadeway: Not sure exactly what you're asking, but it will fail without this line.

abellotti: Can't we reference ManageIQ::Api::VERSION where needed vs Api::VERSION ? I may be missing something.

imtayadeway: We *could*, but seems more intuitive to have it here as it's the only constant in that namespace. All others (including ones related to versions) are here.

I'm feeling more and more that perhaps these two things (`Api::VERSION` and `ManageIQ::Api::VERSION`) are *not the same thing* and one shouldn't be defined in terms of the other. The former represents the JSON API, the latter the programmatic API. Any thoughts @chrisarcand ?

abellotti: Not sure why you'd want to differentiate the two, pretty confusing.  Let's have the one, and if the above assignment makes it easier to reference, I'm 👍 for it.

chrisarcand: @imtayadeway I agree with @abellotti, there's no reason to differentiate, at least at this time. Might as well keep it simple, especially as the 'gem version' is especially worthless right now.

> The former represents the JSON API, the latter the programmatic API.

There is no contractual programmatic API from this project; it's solely built as a Rails engine of API endpoints, that's all that matters to 'our users', so I wouldn't worry about it. 

This particular line shouldn't need to exist anyway as really the API (and other 'gems' we've pulled) should be namespaced under `ManageIQ` (`ManageIQ::Api::Everything`), like the provider gems.

👉 As that's the desired end goal anyway, I'd probably just remove this assignment and refer to `ManageIQ::Api::Version`.

Quick aside: If you ever do go back and actually fix the namespacing here in the API, [it's standard Ruby convention to fully capitalize acronyms](https://github.com/bbatsov/ruby-style-guide#camelcase-classes), so the namespace for everything should be `ManageIQ::API`

imtayadeway: > There is no contractual programmatic API from this project; it's solely built as a Rails engine of API endpoints, that's all that matters to 'our users', so I wouldn't worry about it.

There is https://github.com/ManageIQ/manageiq-api/blob/master/lib/api/utils.rb which is used in the main repo and possibly the API client gem - whether that is contractual is perhaps up for debate?

chrisarcand: * For `build_href_slug`, 1) Building a href like that should not be the concern of the API, that's a concern for the client. If it's used in multiple places and the duplication per-client bothers people, then it should be a separate dependency (`manageiq-api-utils`, whatever). 2) Even if it is left in here, presumably the versioning aligns itself with the API that href refers to (the thing generating hrefs should align with the actual hrefs anyway)
* The `resource_search_by_href_slug` has code smell written all over it given this project is exposing a method that takes an href and runs RBAC, which it doesn't own in the slightest. It probably shouldn't be exposed externally like that anyway (though I'm sure it's used in core 😬 ).

TLDR both of those methods are href parsing stuff that should exist elsewhere or just be handled in the client anyway (and maybe even call the API for collections, whatever).

abellotti: resource_search_by_href_slug is used from manageiq-automation_engine.

abellotti:
Thanks @imtayadeway for this PR 👍  🎵

PR #168, opened 2017-11-01
Mask password attributes from AutomateWorkspace

mkanoor:
The Automate Workspace can optionally have objects and method
parameters that contain encrypted fields. When these attributes
parameters are sent back to the client we should mask it.
This PR masks the value to password::********

### DIFF
--- a/app/controllers/api/base_controller/normalizer.rb
+++ b/app/controllers/api/base_controller/normalizer.rb
@@ -34,6 +34,8 @@ module Api
           normalize_array(value)
         elsif value.respond_to?(:attributes) || value.respond_to?(:keys)
           normalize_hash(attr, value)
+        elsif @req.subject == "automate_workspaces" && value =~ /^password::/
+          "password::********"
         elsif attr == "id" || attr.to_s.ends_with?("_id")
           value.to_s
         elsif Api.time_attribute?(attr)
--- a/app/controllers/api/base_controller/normalizer.rb
+++ b/app/controllers/api/base_controller/normalizer.rb
@@ -34,7 +34,7 @@ module Api
           normalize_array(value)
         elsif value.respond_to?(:attributes) || value.respond_to?(:keys)
           normalize_hash(attr, value)
-        elsif @req.subject == "automate_workspaces" && value =~ /^password::/
+        elsif @req.subject == "automate_workspaces" && value =~ /\Apassword::/
           "password::********"
         elsif attr == "id" || attr.to_s.ends_with?("_id")
           value.to_s
--- a/app/controllers/api/base_controller/normalizer.rb
+++ b/app/controllers/api/base_controller/normalizer.rb
@@ -34,8 +34,6 @@ module Api
           normalize_array(value)
         elsif value.respond_to?(:attributes) || value.respond_to?(:keys)
           normalize_hash(attr, value)
-        elsif @req.subject == "automate_workspaces" && value =~ /\Apassword::/
-          "password::********"
         elsif attr == "id" || attr.to_s.ends_with?("_id")
           value.to_s
         elsif Api.time_attribute?(attr)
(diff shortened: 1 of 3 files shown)


spec/requests/automate_workspaces_spec.rb
```diff
@@ -40,6 +41,15 @@
       expect(response).to have_http_status(:ok)
     end
 
+    it 'should mask password attributes' do
+      api_basic_authorize action_identifier(:automate_workspaces, :read, :resource_actions, :get)
+      get(api_automate_workspace_url(nil, aw.guid))
+
+      expect(response).to have_http_status(:ok)
+      expect(response.parsed_body['input']['objects']['root']['var2']).to eq(masked_password)
+      expect(response.parsed_body['input']['method_parameters']['arg1']).to eq(masked_password)
```
abellotti: I'd prefer replacing 49-50 with expect(response.parsed_body).to include('input' => a_hash_including( ... )

app/controllers/api/base_controller/normalizer.rb
```diff
@@ -34,6 +34,8 @@ def normalize_attr(attr, value)
           normalize_array(value)
         elsif value.respond_to?(:attributes) || value.respond_to?(:keys)
           normalize_hash(attr, value)
+        elsif @req.subject == "automate_workspaces" && value =~ /^password::/
+          "password::********"
```
imtayadeway: This isn't the way we handle other passwords - we just omit them. Is there a special reason that they need to be handled differently?

mkanoor: @imtayadeway In other cases its our password say for the provider, storage. These passwords are created by customers in the Automate Model for their own use cases (connect to IPAM server, LDAP server etc) the decide the attribute name. Sending the password this way allows the client side to check if 
* the attribute exists
* the attribute has the correct data type

Once these pre conditions have been met it can make a REST API call to decrypt the password.

imtayadeway: @mkanoor can you add a test for the case in which the attribute does not exist? How is the data type revealed through the response? Can you also ensure there is test coverage for the happy/sad path there also?

mkanoor: @imtayadeway Thats already implemented here
https://github.com/ManageIQ/manageiq-api/blob/master/spec/requests/automate_workspaces_spec.rb#L103

If you send any bogus object/attribute pair we send back an empty string.

This PR is the about the "GET" request for Automate Workspace and filters/masks what is being sent out, its not about encrypting/decrypting passwords on demand which was done in PR.
https://github.com/ManageIQ/manageiq-api/pull/124

abellotti:
Ping @imtayadeway, has @mkanoor answered your question ?  if so can you 👍 when ready. Thanks.

app/controllers/api/base_controller/normalizer.rb
imtayadeway: I'm thinking that this may be a better candidate for inheritance - as it is, every single request is going to check if we're dealing with automate workspaces multiple times:

```ruby
# app/controllers/api/automate_workspaces_controller.rb

private

def normalize_attr(attr, value)
  return "password:***" if value /\Apassword/
  super
end
```

app/controllers/api/base_controller/normalizer.rb
```diff
@@ -34,6 +34,8 @@ def normalize_attr(attr, value)
           normalize_array(value)
         elsif value.respond_to?(:attributes) || value.respond_to?(:keys)
           normalize_hash(attr, value)
+        elsif @req.subject == "automate_workspaces" && value =~ /^password::/
```
imtayadeway: Avoid the use of the `^` anchor - prefer `\A` here

mkanoor:
@imtayadeway Please review

abellotti:
Thanks @mkanoor for the fix and @imtayadeway for the recommendation, love that normalize_attr in the controller that needed it 👍

simaishi:

```
$ git log -1
commit a1b4554c335fab557b9592ee39d4ccc8bbd0139f
Date:   Fri Nov 3 13:22:08 2017 -0400

    Merge pull request #168 from mkanoor/mask_automate_workspace_passwords
    
    Mask password attributes from AutomateWorkspace
    (cherry picked from commit 9a2767c050bf7bb943b0768ad650114c735ff445)
```

PR #170, opened 2017-11-01
Don't try to uncompress ids from href slugs

imtayadeway:
In f3b6b2ef1541a2d374d4fb9bd75be00f784b612a we stopped generating href
slugs with compressed ids. We should no longer need to uncompress the
id when parsing an href slug.


abellotti:
Thanks for cleaning 🏠 @imtayadeway 🎹

PR #171, opened 2017-11-02
Consolidate some API spec helpers

chrisarcand:
A few clean up items while I'm working on something else.

* `basic_authorize` is really just a helper method for `api_basic_authorize` and the thing it does itself is only used in a single place. The `api_basic_authorize` can just be more flexible about it, and we don't need two methods that are confusingly similar.
* `update_user_role` is a really simple thing by itself and parameterizes the role which you almost never need to do. You should basically never need to update a role mid-spec so it's a bit of an antipattern anyway.

### DIFF
--- a/spec/requests/authentication_spec.rb
+++ b/spec/requests/authentication_spec.rb
@@ -6,7 +6,7 @@ describe "Authentication API" do
 
   context "Basic Authentication" do
     it "test basic authentication with bad credentials" do
-      basic_authorize "baduser", "badpassword"
+      api_basic_authorize :user => 'baduser', :password => 'badpassword'
 
       get api_entrypoint_url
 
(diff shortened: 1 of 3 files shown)


abellotti:
Ping @imtayadeway

chrisarcand:
+1 on the naming, I thought exactly that myself but I decided changing all callers and stuff in this PR was a bad idea.

chrisarcand:
@simaishi The problem on Gaprindashvili is because this needs to be backported, my bad. Backport this one, see if it's green, and then I'll address Fine (it probably can't be cleanly backported to Fine so when I see that this actually is the issue then I'll make a PR)

simaishi:

```
$ git log -1
commit e39abd8e3bfaf0bb235deefdf6ad98ac6e4a763b
Author: Tim Wade <imtayadeway@users.noreply.github.com>
Date:   Fri Nov 3 07:55:51 2017 -0700

    Merge pull request #171 from chrisarcand/clean-up-api-helper
    
    Consolidate some API spec helpers
    (cherry picked from commit 7969144d9bec884c8ea250c602c35f68f706dda2)
```

PR #173, opened 2017-11-02
Fix coercing of string to 0 in parse_id

jntullo:
Currently, parse_id returns a 0 if a string is passed in, rather than a hash. This checks to ensure the resource passed in is a hash and returns nil otherwise, allowing normal error handling to proceed. 

I couldn't find a better place for the test, seems like it doesn't quite belong there and is more of a general case. Thoughts?

### DIFF
--- a/app/controllers/api/base_controller/parser.rb
+++ b/app/controllers/api/base_controller/parser.rb
@@ -39,7 +39,7 @@ module Api
       end
 
       def parse_id(resource, collection)
-        return nil if resource.blank?
+        return nil if !resource.kind_of?(Hash) || resource.blank?
 
         href_id = href_id(resource["href"], collection)
         case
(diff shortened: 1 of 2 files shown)


app/controllers/api/base_controller/parser.rb
```diff
@@ -39,7 +39,7 @@ def validate_api_action
       end
 
       def parse_id(resource, collection)
-        return nil if resource.blank?
+        return nil if !resource.kind_of?(Hash) || resource.blank?
```
chrisarcand: You don't need `resource.blank?` anymore, though it's fine that it short circuits the `case` below.

abellotti:
Thanks @jntullo for fixing this 👍

simaishi:

```
$ git log -1
commit fbe9f2164f84634695ae26aa4d8d1aebcd63a3e7
Date:   Thu Nov 2 10:45:11 2017 -0500

    Merge pull request #173 from jntullo/bug/fix_coerced_zero
    
    Fix coercing of string to 0 in parse_id
    (cherry picked from commit 42813a23b65b485609033cbaa8d65041549b2e6b)
```

PR #175, opened 2017-11-03
[WIP] Extract serializer

imtayadeway:
This is still very much a WIP/prototype, so please bear that in mind if you're planning to review this. I want to open this to facilitate a more high-level discussion over what we're currently doing, its shortcomings, and what might be the best way to solve them.

### What we're currently doing

We're currently grabbing the list of attributes from a model and creating hashes with names for keys and the result of `send`'ing that name to the model.

For each attribute we check to see if we should do something special with it, like coerce it into a different type, or omit/obscure it. There is a lot of guesswork going on here. For instance, we decide if something is a key if had has "id" in the name. We decide if something is a password if it's in the global list of known password attributes.

A user can also request virtual attributes to be returned in the response. These are not serialized by default.

### The shortcomings

Mainly it's the amount of guesswork that's going on, and the lack of customizability.

It might be nice if we were leveraging `Model.as_json` instead of `send`ing messages to the model, but I haven't attempted that here.

Currently, the only way to get non-attribute methods serialized is to add them as virtual attributes, which is currently taking virtual attributes well beyond their intended use.

### What to do

I've used `ActiveModel::Serializers` in the past and fount id acceptable, but don't think it will ever fit what we currently support. I know there are other/newer solutions to this problem that I can't speak about, but I would expect any such external library to be difficult to integrate without changing what we currently support. I've created a prototype using a similar pattern to AMS, bearing in mind:

1. We would need a serializer for anything that could be requested, and to be honest I don't even know how big that list is. You can request anything as long as it can be reached through the model corresponding to one of our primary endpoints. In my prototype I've elected to create serializers on the fly (unless one already exists) - the alternative to this is to create (potentially) hundreds of serializers.

2. It needs to be able to serialize additional attributes or select only specific attributes at request time

3. If it infers anything at runtime that's expensive (i.e. look up key names, etc..), that needs to be cached

4. If a serializer has any kind of default behavior (which it doesn't in AMS - you have to declare everything), that needs to be customizable

These are the problems that I've tried to solve here. There are still a few problems that are keeping this in the WIP phase for now, but I'll try to point those out in the code below.

/cc @abellotti @chrisarcand @jntullo @mkanoor


app/controllers/api/base_controller/normalizer.rb
```diff
@@ -19,9 +19,21 @@ def normalize_hash(type, obj, opts = {})
           end
         end
 
-        attrs.each do |k|
-          value = normalize_attr(k, obj.kind_of?(ActiveRecord::Base) ? obj.try(k) : obj[k])
-          result[k] = value unless value.nil?
+        case obj
```
imtayadeway: One of the biggest problems that remains to be solved is that certain models have already been serialized elsewhere before it even gets to this point, and arrive here as a hash. We need to address that so that we can eliminate most of the guesswork that's going on based only on attribute/key names

app/serializers/api/base_serializer.rb
```diff
@@ -0,0 +1,86 @@
...
+    end
+
+    def coerce(attr, value)
+      return if value.nil?
+      if key?(attr)
+        value.to_s
+      elsif date_or_time?(attr)
+        value.utc.iso8601
+      elsif value.kind_of?(Hash)
+        HashSerializer.serialize(value)
```
imtayadeway: It's regrettable that I need to do this here - i.e. resort to serializing something based on key names. The actual implementation of this is roughly what we currently do, but ideally I'd like to eliminate it and have each concrete serializer determine how exactly to deal with these.

PR #176, opened 2017-11-03
Add a set_current_group method for users

jntullo:
The current way of setting a user’s `current_group` via edit uses `resource_search`, which encounters some RBAC issues. For example, when a user changes from the super administrator group to the tenant group, they are no longer able to switch back even though they belong to the group because `resource_search` returns a forbidden error. By choosing a group based off of their current miq_groups, it resolves the issue and keeps the ability to change groups consistent with that in the classic-ui.

**How it works in the classic ui:**
group_switching_classic_ui

**The failure present in the SUI before the fix:**
group_switching_sui

**The successful API call with `set_current_group`:**
group_switching_api

In addition, with update_attributes! it raises a validation error saying that the superadministrator group is not in their groups, also appears to be due to RBAC. I looked to the [classic UI](https://github.com/ManageIQ/manageiq-ui-classic/blob/eac23c268907f22feec9f1109f80eb795ad43dd8/app/controllers/dashboard_controller.rb#L657) to see how their behavior works, and chose to use `update_attribute` based off of that. Because this is different behavior than a typical edit, I felt that it also deserved a specialized method, thus removing this functionality from edit and moving it to `set_current_group`. 

This is also a unique use case, and did not think it would apply as a collection action, but am open to discussion around that. 

https://bugzilla.redhat.com/show_bug.cgi?id=1467364

@miq-bot assign @abellotti 

cc: @AllenBW

### DIFF
--- a/app/controllers/api/users_controller.rb
+++ b/app/controllers/api/users_controller.rb
@@ -2,7 +2,7 @@ module Api
   class UsersController < BaseController
     INVALID_USER_ATTRS = %w(id href current_group_id settings).freeze # Cannot update other people's settings
     INVALID_SELF_USER_ATTRS = %w(id href current_group_id).freeze
-    EDITABLE_ATTRS = %w(password email settings group current_group).freeze
+    EDITABLE_ATTRS = %w(password email settings group).freeze
 
     include Subcollections::Tags
 
@@ -37,7 +37,6 @@ module Api
     def edit_resource(type, id, data)
       id == User.current_user.id ? validate_self_user_data(data) : validate_user_data(data)
       parse_set_group(data)
-      parse_set_current_group(data)
       parse_set_settings(data, resource_search(id, type, collection_class(type)))
       super
     end
@@ -46,6 +45,20 @@ module Api
       raise BadRequestError, "Must specify an id for deleting a user" unless id
       raise BadRequestError, "Cannot delete user of current request" if id.to_i == User.current_user.id
       super
+    end
+
+    def set_current_group_resource(_type, id, data)
+      User.current_user.tap do |user|
+        raise "Can only edit authenticated user's current group" unless user.id == id
+        group_id = parse_group(data["current_group"])
+        raise "Must specify a current_group" unless group_id
+        new_group = user.miq_groups.where(:id => group_id).first
+        raise "User must belong to group" unless new_group
+        # Cannot use update_attributes! due to the allowed ability to switch between groups that may have different RBAC visibility on a user's miq_groups
+        user.update_attribute(:current_group, new_group)
+      end
+    rescue => err
+      raise BadRequestError, "Cannot set current_group - #{err}"
     end
 
     private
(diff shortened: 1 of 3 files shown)


config/api.yml
```diff
@@ -2779,6 +2779,8 @@
         :identifier: rbac_user_edit
       - :name: delete
         :identifier: rbac_user_delete
+      - :name: set_current_group
+        :identifier: rbac_user_edit
```
abellotti: does this mean if a user does not have that role they cannot change their current group ?

jntullo: @abellotti yeah, I guess so. I am not sure if users should always be allowed to change their own group or if they need that particular role?

abellotti: May be easy to verify, in classic UI login as a user without that role but a member of a couple of groups, see if you can change the current group in the upper right hand pull-down. If you can, we'll need to somehow handle that.

jntullo: @abellotti verified that a user in the classic UI without this edit role is still able to change their group. updated accordingly 👍

abellotti:
With the role identifier removed, we probably need to make sure that the action is only run on the authenticated user, otherwise anyone can change anyone else's group.  Can you verify this and add a test pertaining to that ? Thanks.

app/controllers/api/users_controller.rb
```diff
@@ -48,6 +47,19 @@ def delete_resource(type, id = nil, data = nil)
       super
     end
 
+    def set_current_group_resource(_type, id, data)
+      User.current_user.tap do |user|
+        raise "Can only edit authenticated user's current group" unless user.id == id
+        group_id = parse_id(data["current_group"], :groups)
```
abellotti: can you use parse_group(data["current_group"]) here ? it support both id/href as well as the identifying attr,i.e. be able to set current group by description.

abellotti: or maybe the parse_fetch_group we used earlier.

jntullo: @abellotti can't use `parse_fetch_group` because it uses `resource_search`. Will update to use `parse_group` though

AllenBW:
I know everyone is da 🐝, wanted to do a quick status check on this 👶 

🙇‍♀️ ⭕️ :octocat:

abellotti:
I think I'm good with this, @jntullo can you take care of the rubocop warning ?  Thanks.

jntullo:
@abellotti the rubocop warning can't be fixed, it is part of the solution to the current issue

abellotti:
Hmm, can `user.current_group = new_group` be done instead ?  If not, maybe simply add a comment in the code that update_attributes! cannot be used just so that it's not inadvertently "fixed" in the future.  Thanks.

abellotti:
Thanks @jntullo for updating this. 👍

simaishi:

```
$ git log -1
commit 29014b7a216d843ee8a46106f0601d60e98aaa91
Date:   Mon Nov 13 16:47:43 2017 -0500

    Merge pull request #176 from jntullo/bug/change_current_group
    
    Add a set_current_group method for users
    (cherry picked from commit 8f30e9cd9894e302add1729d7aa7bb54764db33b)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1513191
```

PR #177, opened 2017-11-05
Allow assignments of alert profiles

elad661:
### Description
Alert profiles can be assigned to objects or to tags.

This functionality exists in the UI under Control->Explorer->Alert Profiles -> select an alert profile -> configuration -> edit assignments for this alert profile.

The API added in this pull request allows getting a list of all assignments for a profile, and to assign or unassign a profile.

This PR depends on https://github.com/ManageIQ/manageiq/pull/16401

### **Usage examples**
**assign profile to object**
POST /alert_definition_profiles/:id
{"action": "assign", "objects": ["http://localhost:3000/api/enterprises/1"]}

**unassign profile from object**
POST /alert_definition_profiles/:id
{"action": "unassign", "objects": ["http://localhost:3000/api/enterprises/1"]}

**assign profile to tags**
POST /alert_definition_profiles/:id
{"action": "assign", "tags": [{"href":"http://localhost:3000/api/tags/95","class": "host"}]}

**unassign profile from tags**
POST /alert_definition_profiles/:id
{"action": "unassign", "tags": [{"href":"http://localhost:3000/api/tags/95","class": "host"}]}

(see https://github.com/ManageIQ/manageiq/blob/master/app/models/mixins/assignment_mixin.rb#L62 for explanation of what "class" can be. Also, tag assignment only works for tags which are classification tags)

### DIFF
--- a/app/controllers/api/alert_definition_profiles_controller.rb
+++ b/app/controllers/api/alert_definition_profiles_controller.rb
@@ -4,6 +4,8 @@ module Api
 
     REQUIRED_FIELDS = %w(description mode).freeze
 
+    before_action :set_additional_attributes, :only => [:show]
+
     def create_resource(type, id, data = {})
       assert_all_required_fields_exists(data, type, REQUIRED_FIELDS)
       begin
@@ -20,6 +22,53 @@ module Api
       rescue => err
         raise BadRequestError, "Failed to update alert definition profile - #{err}"
       end
+    end
+
+    def get_resource_from_href(href)
+      hrefobj = Href.new(href)
+      resource_search(hrefobj.subject_id, hrefobj.subject, collection_class(hrefobj.subject))
+    end
+
+    def assign_resource(type, id = nil, data = {})
+      profile = resource_search(id, type, collection_class(type))
+      raise BadRequestError, "Must specify either objects or tags for the assignment target" unless data.include?("objects") || data.include?("tags")
+      if data.include?("objects")
+        objects = data["objects"].collect do |ref|
+          get_resource_from_href(ref)
+        end
+        profile.assign_to_objects(objects)
+      end
+
+      if data.include?("tags")
+        data["tags"].each do |tag|
+          profile.assign_to_tags([get_resource_from_href(tag["href"]).classification], tag["class"])
+        end
+      end
+      profile.get_assigned_tos
+    end
+
+    def unassign_resource(type, id = nil, data = {})
+      profile = resource_search(id, type, collection_class(type))
+      raise BadRequestError, "Must specify either objects or tags for the assignment target" unless data.include?("objects") || data.include?("tags")
+
+      if data.include?("objects")
+        objects_to_unassign = data["objects"].collect do |ref|
+          get_resource_from_href(ref)
+        end
+        profile.unassign_objects(objects_to_unassign)
+      end
+
+      if data.include?("tags")
+        data["tags"].each do |tag|
+          profile.unassign_tags(get_resource_from_href(tag["href"]).classification, tag["class"])
+        end
+      end
+
+      profile.get_assigned_tos
+    end
+
+    def set_additional_attributes
+      @additional_attributes = %w(get_assigned_tos)
     end
   end
 end
--- a/config/api.yml
+++ b/config/api.yml
@@ -113,6 +113,10 @@
         :identifier: alert_definition_profile_edit
       - :name: delete
         :identifier: alert_definition_profile_delete
+      - :name: assign
+        :identifier: alert_profile_assign
+      - :name: unassign
+        :identifier: alert_profile_assign
       :delete:
       - :name: delete
         :identifier: alert_definition_profile_delete
@@ -955,6 +959,15 @@
         :identifier: storage_tag
       - :name: unassign
         :identifier: storage_tag
+  :enterprises:
+    :description: Enterprise
+    :options:
+    - :collection
+    :verbs: *g
+    :klass: MiqEnterprise
+    :collection_actions:
+      :get:
+      - :name: read
   :event_streams:
     :description: Event Streams
     :options:
@@ -972,15 +985,6 @@
       :get:
       - :name: read
         :identifier: event_streams_show
-  :enterprises:
-    :description: Enterprise
-    :options:
-    - :collection
-    :verbs: *g
-    :klass: MiqEnterprise
-    :collection_actions:
-      :get:
-      - :name: read
   :events:
     :description: Events
     :identifier: event
--- a/config/api.yml
+++ b/config/api.yml
@@ -972,6 +972,15 @@
       :get:
       - :name: read
         :identifier: event_streams_show
+  :enterprises:
+    :description: Enterprise
+    :options:
+    - :collection
+    :verbs: *g
+    :klass: MiqEnterprise
+    :collection_actions:
+      :get:
+      - :name: read
   :events:
     :description: Events
     :identifier: event
--- a/spec/requests/alert_definitions_spec.rb
+++ b/spec/requests/alert_definitions_spec.rb
@@ -527,4 +527,77 @@ describe "Alerts Definition Profiles API" do
 
     expect(response).to have_http_status(:forbidden)
   end
+
+  it "gets assignments for an alert definition profile" do
+    api_basic_authorize action_identifier(:alert_definition_profiles, :read, :resource_actions, :get)
+    alert_definition_profile = FactoryGirl.create(:miq_alert_set)
+    enterprise = FactoryGirl.create(:miq_enterprise)
+    enterprise_href = api_enterprise_url(nil, enterprise)
+    alert_definition_profile.assign_to_objects(enterprise)
+
+    get(api_alert_definition_profile_url(nil, alert_definition_profile))
+
+    expect(response).to have_http_status(:ok)
+    expect(response.parsed_body["get_assigned_tos"]["objects"]).to include(a_hash_including("href"=>enterprise_href))
+  end
+
+  it "assigns an alert definition profile to object" do
+    api_basic_authorize :alert_profile_assign
+    alert_definition_profile = FactoryGirl.create(:miq_alert_set)
+    enterprise = FactoryGirl.create(:miq_enterprise)
+    enterprise_href = api_enterprise_url(nil, enterprise)
+
+    post(api_alert_definition_profile_url(nil, alert_definition_profile), :params => gen_request(:assign, "objects" => [enterprise_href]))
+
+    expect(response).to have_http_status(:ok)
+    expect(response.parsed_body["objects"]).to include(a_hash_including("href"=>enterprise_href))
+  end
+
+  it "unassigns an alert definition profile from object" do
+    api_basic_authorize :alert_profile_assign
+    alert_definition_profile = FactoryGirl.create(:miq_alert_set)
+    enterprise, enterprise2 = FactoryGirl.create_list(:miq_enterprise, 2)
+
+    enterprise_href = api_enterprise_url(nil, enterprise)
+    enterprise2_href = api_enterprise_url(nil, enterprise2)
+    expect(alert_definition_profile.get_assigned_tos).to eq(:objects => [], :tags => [], :labels => [])
+
+    alert_definition_profile.assign_to_objects([enterprise, enterprise2])
+
+    post(api_alert_definition_profile_url(nil, alert_definition_profile), :params => gen_request(:unassign, "objects" => [enterprise_href]))
+
+    expect(response).to have_http_status(:ok)
+    expect(response.parsed_body["objects"]).to include(a_hash_including("href"=>enterprise2_href))
+  end
+
+  it "assigns an alert definition profile to tag" do
+    api_basic_authorize :alert_profile_assign
+    alert_definition_profile = FactoryGirl.create(:miq_alert_set)
+    dept = FactoryGirl.create(:classification_department)
+    tag = FactoryGirl.create(:classification_tag, :name => 'test', :parent => dept).tag
+
+    tag_href = api_tag_url(nil, tag)
+
+    post(api_alert_definition_profile_url(nil, alert_definition_profile), :params => gen_request(:assign, "tags" => ["href" => tag_href, "class" => "vm"]))
+
+    expect(response).to have_http_status(:ok)
+    expect(response.parsed_body["tags"][0]).to include(a_hash_including("tag_id"=>String(tag.id)))
+  end
+
+  it "unassigns an alert definition profile from tag" do
+    api_basic_authorize :alert_profile_assign
+    alert_definition_profile = FactoryGirl.create(:miq_alert_set)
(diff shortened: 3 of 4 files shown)


elad661:

cc @moolitayer

app/controllers/api/alert_definition_profiles_controller.rb
```diff
@@ -21,5 +21,68 @@ def edit_resource(type, id = nil, data = {})
         raise BadRequestError, "Failed to update alert definition profile - #{err}"
       end
     end
+
+    def get_assignments_resource(_type, id, _data)
```
jntullo: Instead of using this, have you considered using [set_additional_attributes](https://github.com/ManageIQ/manageiq-api/blob/9a2767c050bf7bb943b0768ad650114c735ff445/app/controllers/api/alert_definitions_controller.rb#L5)? I just tested it out for `alert_definition_profiles` by setting `get_assigned_to` as an additional attribute, with a method `fetch_alert_definition_profiles_get_assigned_tos(resource)` and it worked well. You may want to consider restricting it to just `:show`

app/controllers/api/alert_definition_profiles_controller.rb
```diff
@@ -21,5 +21,68 @@ def edit_resource(type, id = nil, data = {})
...
+      profile.get_assigned_tos
+    end
+
+    def get_resource_from_href(href)
+      hrefobj = Href.new(href)
+      resource_search(hrefobj.subject_id, hrefobj.subject, collection_class(hrefobj.subject))
+    end
+
+    def assign_resource(_type, id = nil, data = {})
+      profile = MiqAlertSet.find(id)
```
jntullo: you will want to use `resource_search` here

app/controllers/api/alert_definition_profiles_controller.rb
```diff
@@ -21,5 +21,68 @@ def edit_resource(type, id = nil, data = {})
...
+      if data.include?("tags")
+        data["tags"].each do |tag|
+          profile.assign_to_tags([get_resource_from_href(tag["href"]).classification], tag["class"])
+        end
+      end
+      profile.get_assigned_tos
+    end
+
+    def unassign_resource(_type, id = nil, data = {})
+      profile = MiqAlertSet.find(id)
```
jntullo: same as above, should use `resource_search`

app/controllers/api/alert_definition_profiles_controller.rb
```diff
@@ -21,5 +21,68 @@ def edit_resource(type, id = nil, data = {})
...
+      raise BadRequestError, "Must specify either objects or tags for the assignment target" unless data.include?("objects") || data.include?("tags")
+      # the model does not have a method to remove a specific assignment
+      # so we save the current assignments, remove all assignments,
+      # and then re-add them without the one we wanted to remove
```
jntullo: In regards to this comment, have you considered adding a method on the model to perform this logic? Would be nice to keep this logic out of the controller - and would probably be easier to test as well.

config/api.yml
```diff
@@ -113,6 +113,9 @@
         :identifier: alert_definition_profile_edit
       - :name: delete
         :identifier: alert_definition_profile_delete
+      - :name: assign
```
jntullo: Do these actions not have identifiers?

app/controllers/api/alert_definition_profiles_controller.rb
elad661: I didn't know it's an option, I'll try using it and see if it works

elad661:
I've addressed all the feedback from the review, apart from moving the unassign logic to the model - so I still consider it WIP.

elad661:
Note, this PR now depends on https://github.com/ManageIQ/manageiq/pull/16401 for `unassign_objects` and `unassign_tags`

elad661:
Now that ManageIQ/manageiq#16401 is merged, this PR is ready for review/merge.

@jntullo, @moolitayer : please take a look.

moolitayer:
> Also, tag assignment only works for tags which are classification tags

Does that mean tags that have a related classification object?

elad661:
>Does that mean tags that have a related classification object?

Yeah, I think so. I don't entirely understand how this works though

moolitayer:
> get current assignments for a profile
> POST /alert_definition_profiles/:id
> {"action": "get_assignments"}

Can you include example output?
please remind me, is the "action": "get_assignments" needed? I can't find where that action is mentioned

elad661:
>  please remind me, is the "action": "get_assignments" needed? I can't find where that action is mentioned

Oops, I forgot to remove it from the PR description, it's from an older version of this PR.

The assignments are shown when you do a regular GET request for the specific alert profile, under the "get_assigned_tos" filed, for example:

```json
{
    "href": "http://localhost:3000/api/alert_definition_profiles/27",
    "id": "27",
    "name": "a3745ed9-f506-4298-8741-b2bbf275d1d8",
    "description": "test 3",
    "set_type": "MiqAlertSet",
    "created_on": "2017-11-02T14:28:27Z",
    "updated_on": "2017-11-02T14:28:27Z",
    "guid": "a3745ed9-f506-4298-8741-b2bbf275d1d8",
    "mode": "Vm",
    "get_assigned_tos": {
        "objects": [
            {
                "href": "http://localhost:3000/api/enterprises/1",
                "id": "1",
                "name": "Enterprise",
                "description": "Enterprise Root Object",
                "created_on": "2017-10-16T14:35:16Z",
                "updated_on": "2017-10-16T14:35:16Z"
            }
        ],
        "tags": [],
        "labels": []
    },
```

I don't like the name "get_assigned_tos" - that's the name of the method in the model that returns these assignments, and there doesn't seem to be a way to change it for the API

app/controllers/api/alert_definition_profiles_controller.rb
```diff
@@ -21,5 +23,52 @@ def edit_resource(type, id = nil, data = {})
...
+
+    def unassign_resource(type, id = nil, data = {})
+      profile = resource_search(id, type, collection_class(type))
+      raise BadRequestError, "Must specify either objects or tags for the assignment target" unless data.include?("objects") || data.include?("tags")
+
+      if data.include?("objects")
+        objects_to_unassign = data["objects"].collect do |ref|
+          get_resource_from_href(ref)
+        end
+        profile.unassign_objects(objects_to_unassign)
```
moolitayer: what happens if the object was not assigned to the tag?

elad661: Nothing would happen, you won't get an error or anything like it.

spec/requests/alert_definitions_spec.rb
```diff
@@ -306,4 +306,77 @@
...
+    api_basic_authorize action_identifier(:alert_definition_profiles, :read, :resource_actions, :get)
+    alert_definition_profile = FactoryGirl.create(:miq_alert_set)
+    enterprise = FactoryGirl.create(:miq_enterprise)
+    enterprise_href = api_enterprise_url(nil, enterprise)
+    alert_definition_profile.assign_to_objects(enterprise)
+
+    get(api_alert_definition_profile_url(nil, alert_definition_profile))
+
+    expect(response).to have_http_status(:ok)
+    expect(response.parsed_body["get_assigned_tos"]["objects"]).to include(a_hash_including("href"=>enterprise_href))
```
moolitayer: it would be nicer to have the api return 
s/get_assigned_tos/assigned_to/
(unless there is a convention for assignments i'm not aware of)

elad661: I agree it would be much nicer, but from what I understood when talking with @jntullo about this on gitter when I wrote this PR, it's not possible to for the API to rename anything defined as `additional_attributes` - the name has to be identical to the name of the method in the model.

config/api.yml
```diff
@@ -826,6 +830,15 @@
         :identifier: storage_tag
       - :name: unassign
         :identifier: storage_tag
+  :enterprises:
```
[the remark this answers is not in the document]
elad661: The enterprises controller? That's to expose MiqEnterprise in the API, so it'll have an href, so you could assign profiles to it.

spec/requests/alert_definitions_spec.rb
```diff
@@ -306,4 +306,77 @@
...
+  it "unassigns an alert definition profile from object" do
+    api_basic_authorize :alert_profile_assign
+    alert_definition_profile = FactoryGirl.create(:miq_alert_set)
+    enterprise = FactoryGirl.create(:miq_enterprise)
+    enterprise2 = FactoryGirl.create(:miq_enterprise)
+
+    enterprise_href = api_enterprise_url(nil, enterprise)
+    enterprise2_href = api_enterprise_url(nil, enterprise2)
+
+    alert_definition_profile.assign_to_objects([enterprise, enterprise2])
```
moolitayer: please add an expectation for this step (both profiles attached)

spec/requests/alert_definitions_spec.rb
```diff
@@ -306,4 +306,77 @@
...
+    expect(response).to have_http_status(:ok)
+    expect(response.parsed_body["get_assigned_tos"]["objects"]).to include(a_hash_including("href"=>enterprise_href))
+  end
+
+  it "assigns an alert definition profile to object" do
+    api_basic_authorize :alert_profile_assign
+    alert_definition_profile = FactoryGirl.create(:miq_alert_set)
+    enterprise = FactoryGirl.create(:miq_enterprise)
+    enterprise_href = api_enterprise_url(nil, enterprise)
+
```
moolitayer: please add an expectation for the empty assignments before the assigning request

spec/requests/alert_definitions_spec.rb
```diff
@@ -306,4 +306,77 @@
...
+    expect(response).to have_http_status(:ok)
+    expect(response.parsed_body["objects"]).to include(a_hash_including("href"=>enterprise_href))
+  end
+
+  it "unassigns an alert definition profile from object" do
+    api_basic_authorize :alert_profile_assign
+    alert_definition_profile = FactoryGirl.create(:miq_alert_set)
+    enterprise = FactoryGirl.create(:miq_enterprise)
+    enterprise2 = FactoryGirl.create(:miq_enterprise)
```
moolitayer: enterprise, enterprise2 = FactoryGirl.create(:miq_enterprise, 2)

spec/requests/alert_definitions_spec.rb
```diff
@@ -306,4 +306,77 @@
...
+    expect(response).to have_http_status(:ok)
+    expect(response.parsed_body["tags"][0]).to include(a_hash_including("tag_id"=>String(tag.id)))
+  end
+
+  it "unassigns an alert definition profile from tag" do
+    api_basic_authorize :alert_profile_assign
+    alert_definition_profile = FactoryGirl.create(:miq_alert_set)
+    dept = FactoryGirl.create(:classification_department)
+    tag = FactoryGirl.create(:classification_tag, :name => 'test1', :parent => dept).tag
+    tag2 = FactoryGirl.create(:classification_tag, :name => 'test2', :parent => dept).tag
```
moolitayer: you can use `FactoryGirl.create_list` here too, will be more complex so i'm good with this too
(not you can get different names from the `sequence` definition in the factory)

config/api.yml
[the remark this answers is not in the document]
elad661: The UI does something like `MiqEnterprise.first` or `MiqEnterprise.last` ( I don't remember if it's first or last) when assigning a profile. And yeah, API users don't really need to query it, but it has to be exposed to have a proper href.

spec/requests/alert_definitions_spec.rb
elad661: This is tested in the model, and the tests here are for the API. Do we really need an expectation in this step?

spec/requests/alert_definitions_spec.rb
elad661: Done.

spec/requests/alert_definitions_spec.rb
elad661: I prefer to keep it as it is right now unless it's absolutely required to use create_list.

moolitayer:
@jntullo @abellotti please review

imtayadeway:
@elad661 we should probably aim to keep assigning resources signature consistent with other APIs: e.g. http://manageiq.org/docs/reference/latest/api/reference/policies#assigning-policies-to-a-resource /cc @abellotti

abellotti:
Usage in the description does not seem consistent with other API signatures we have, I'm thinking more like tags, policies or policy_profiles. They would be exposed as subcollection of the intended class.

i.e.

```
/api/:collection/:id/alert_definition_profiles
```

That would allow for querying the alert_definition_profiles for the intended resources as well
as the assigning/unassigning of the same.

queries:

```
GET /api/enterprises/2/alert_definition_profiles

GET /api/enterprises?expand=resources,alert_definition_profiles
```

single assignment:

```
POST /api/enterprises/2/alert_definition_profiles

{
   "action" : "assign",
   "resource" : { "href" : "http://localhost:3000/api/alert_definition_profiles/11" }
}
```

or bulk:

```
POST /api/enterprises/2/alert_definition_profiles

{
   "action" : "assign",
   "resources" : [
      { "href" : "http://localhost:3000/api/alert_definition_profiles/11" },
      { "id" : "12" }
   ]
}
```

same with unassigns,

```
POST /api/enterprises/2/alert_definition_profiles

{
   "action" : "unassign",
   "resources" : [
      { "id" : "11" },
      { "href" : "http://localhost:3000/api/alert_definition_profiles/33" }
   ]
}
``` 

These would be exposed for each that we'd care about, /api/enterprises, /api/tags, etc.
_Note_:  Exposing enterprises as a primary collection would be a separate PR, not sure of relevance in pods though /cc @Fryguy 

Creates would continue as they are today via

```
POST /api/alert_definition_profiles
```

abellotti:
After meeting with @gtanzillo, @imtayadeway and @jntullo discussing this PR, we looked into the model implementation and went through several iterations on the API signature. The following is what was decided on:

1. Move the logic out of here that exposes enterprises into a separate PR.

```
GET /api/enterprises
GET /api/enterprises/:id
```

With the necessary specs in collection_specs.rb.

2. Update this PR to support the following signatures:

```
POST /api/altert_definition_profiles
{
  "action" : "assign",
  "resources" : [
    { "href" : "/api/hosts", "tag" : { "href" : "/api/tags/5" } },
    { "href" : "/api/enterprises/1" },
    { "href" : "/api/clusters/7" }
  ]
}
```

And similarly with the unassign:

```
POST /api/altert_definition_profiles
{
  "action" : "assign",
  "resources" : [
    { "href" : "/api/hosts", "tag" : { "name" : "dev" } },
    { "href" : "/api/clusters/3" }
  ]
}
```

Where the href can identify a single resource, i.e. /api/clusters/7, or a tagged collection where the href of the collection is specified, plus the tag reference, by either href, id or name.

The PR should support the collection types that is currently exposed by the UI (short of folders which is not currently exposed by the API).

abellotti:
Closing in preference to https://github.com/ManageIQ/manageiq-api/pull/348

Issue #146, opened 2017-10-25
n+1 query when requesting services


imtayadeway:
When I:

```http
GET http://localhost:3000/api/services?expand=resources&limit=20
```
I see that we have an n+1 problem:

#### The +1:
```
D, [2017-10-25T15:02:26.932359 #22628] DEBUG -- :   Service Load (0.3ms)  SELECT  "services".* FROM "services" LIMIT $1 OFFSET $2  [["LIMIT", 20], ["OFFSET", 0]]
```

#### For every n services:
```
D, [2017-10-25T15:02:26.996742 #22628] DEBUG -- :   Service Inst Including Associations (64.1ms - 20rows)
D, [2017-10-25T15:02:29.779592 #22628] DEBUG -- :   ServiceTemplate Load (0.2ms)  SELECT  "service_templates".* FROM "service_templates" WHERE "service_templates"."id" = $1 LIMIT $2  [["id", 10000000000005], ["LIMIT", 1]]
D, [2017-10-25T15:02:29.811939 #22628] DEBUG -- :   ServiceTemplate Inst Including Associations (32.1ms - 1rows)
D, [2017-10-25T15:02:29.814565 #22628] DEBUG -- :   ResourceAction Load (0.5ms)  SELECT  "resource_actions".* FROM "resource_actions" WHERE "resource_actions"."resource_id" = $1 AND "resource_actions"."resource_type" = $2 AND "resource_actions"."action" = $3 LIMIT $4  [["resource_id", 10000000000005], ["resource_type", "ServiceTemplate"], ["action", "Reconfigure"], ["LIMIT", 1]]
D, [2017-10-25T15:02:29.814762 #22628] DEBUG -- :   ResourceAction Inst Including Associations (0.0ms - 0rows)
D, [2017-10-25T15:02:29.817743 #22628] DEBUG -- :   CustomButton Load (0.2ms)  SELECT "custom_buttons".* FROM "custom_buttons" WHERE "custom_buttons"."applies_to_class" = $1 AND "custom_buttons"."applies_to_id" IS NULL  [["applies_to_class", "Service"]]
D, [2017-10-25T15:02:29.817925 #22628] DEBUG -- :   CustomButton Inst Including Associations (0.0ms - 0rows)
[...]

PR #186, opened 2017-11-08
Disable custom actions on /services

chrisarcand:
This is a temporary fix for https://bugzilla.redhat.com/show_bug.cgi?id=1499692

Including custom actions is incredibly slow due to the implementation of
custom buttons and how delegations happen, as well as a lack of
preloading on the API side. This causes massive N+1
queries and ultimately makes the query on /services 4-5x slower.

For now, disable custom actions on services only - but the entire
feature should be reviewed in the future.

This basically undoes the feature in https://github.com/ManageIQ/manageiq/pull/4394

I do not know of other things that might depend on this, and we'll need to review that quickly.

### DIFF
--- a/config/api.yml
+++ b/config/api.yml
@@ -2435,7 +2435,6 @@
     :identifier: service
     :options:
     - :collection
-    - :custom_actions
     :verbs: *gpppd
     :klass: Service
     :subcollections:


abellotti:
One option is only enabling custom_actions when querying a single resource and not a collection query. i.e. initial services list they won't get those actions, but if they drill down to the one, they would.  We'd need to double check usage with the SUI folks.  This could be applied to all collections that support custom_actions (tons being added shortly).

chrisarcand:
> One option is only enabling custom_actions when querying a single resource and not a collection query.

Honestly that's my preferred solution as well and regardless of preloading improvements we've made we should probably consider that.  cc/ @Fryguy 

In any case, we've started down the road of preloading stuff and have made great progress, so this nuclear option is likely taken off the table, so closing. See https://github.com/ManageIQ/manageiq-api/issues/146#issuecomment-342702844 ❤️

PR #193, opened 2017-11-09
Add delete for router

andyvesel:
Added `delete` action for Network Router, according to the [comment](https://bugzilla.redhat.com/show_bug.cgi?id=1494359#c3) of this [BZ](https://bugzilla.redhat.com/show_bug.cgi?id=1494359)

### DIFF
--- a/app/controllers/api/network_routers_controller.rb
+++ b/app/controllers/api/network_routers_controller.rb
@@ -1,4 +1,18 @@
 module Api
   class NetworkRoutersController < BaseController
+    def delete_resource(type, id, _data = {})
+      delete_action_handler do
+        network_router = resource_search(id, type, collection_class(:network_routers))
+        raise "Delete not supported for #{network_router_ident(network_router)}" unless network_router.respond_to?(:delete_network_router_queue)
+        task_id = network_router.delete_network_router_queue(User.current_user.id)
+        action_result(true, "Deleting #{network_router_ident(network_router)}", :task_id => task_id)
+      end
+    end
+
+    private
+
+    def network_router_ident(network_router)
+      "Network Router id:#{network_router.id} name: '#{network_router.name}'"
+    end
   end
 end
--- a/spec/requests/network_routers_spec.rb
+++ b/spec/requests/network_routers_spec.rb
@@ -3,6 +3,7 @@ RSpec.describe 'NetworkRouters API' do
     it 'lists all cloud subnets with an appropriate role' do
       network_router = FactoryGirl.create(:network_router)
       api_basic_authorize collection_action_identifier(:network_routers, :read, :get)
+
       get(api_network_routers_url)
 
       expected = {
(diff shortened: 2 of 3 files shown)


abellotti:
We need additional specs for POST based "delete" for a single resource as well as bulk deletes.  Thanks.

andyvesel:
Tests updated. @abellotti, could you review pls?

abellotti:
Hi @andyvesel, the PR looks good, thanks for adding the tests.

the one question I have is about the router.delete call we make. Can that call take a long time to complete ?  triggers communication with providers, etc.  If so we might want to update the return signature to be action based results.

andyvesel:
@abellotti Hello, thanks for comment, it makes sense. I've updated `delete` action for Router according to it

app/controllers/api/network_routers_controller.rb
```diff
@@ -1,4 +1,18 @@
 module Api
   class NetworkRoutersController < BaseController
+    def delete_resource_network_routers(type, id, _data = {})
+      network_router = resource_search(id, type, collection_class(type))
+      task_id = network_router.delete_network_router_queue(User.current_user.id)
+      action_result(true, "Deleting #{network_router_ident(network_router)}", :task_id => task_id)
+    rescue => err
+      action_result(false, err.to_s)
+    end
+    alias network_routers_delete_resource delete_resource_network_routers
```
abellotti: why the alias ?  also since this is a primary collection the method name itself should just be delete_resource. Thanks.

abellotti:
LGTM!!  Thanks for the updates @andyvesel 

@jntullo can you check ?  Thanks.

spec/requests/network_routers_spec.rb
```diff
@@ -48,4 +48,70 @@
       expect(response).to have_http_status(:forbidden)
     end
   end
+
+  describe "DELETE /api/network_routers" do
+    it "can delete a router" do
+      network_router = FactoryGirl.create(:network_router)
+
```
jntullo: nit, but can you remove this blank line here? This way all of the setup lines go together for the spec. Same for other ones

spec/requests/network_routers_spec.rb
```diff
@@ -48,4 +48,70 @@
...
+  describe "POST /api/network_routers with delete action" do
+    it "can delete a router" do
+      network_router = FactoryGirl.create(:network_router)
+
+      api_basic_authorize(action_identifier(:network_routers, :delete, :resource_actions))
+
+      post(api_network_router_url(nil, network_router), :params => gen_request(:delete))
+
+      expect(response).to have_http_status(:ok)
```
jntullo: Think you can add in a line to test the expected content that is passed back?

abellotti:
@andyvesel can you take care of @jntullo's comments ?  Thanks.

andyvesel:
@abellotti @jntullo PR updated

andyvesel:
Thanks for comments, folks!

PR #194, opened 2017-11-09
Custom actions on generic objects

jntullo:
Currently, the method overrides do not take into account the custom actions, only the property methods, therefore not listing custom actions for the Generic Objects. This update will return both property methods and custom actions. This update also calls custom actions the way they are expected to be called, while keeping the existing functionality for the calling of property methods.

~~Dependent upon https://github.com/ManageIQ/manageiq/pull/16433~~
As for the above PR, reloading the object before passing it to `super` resolves the issue of receiving `["invalid method name: [[\"button1\"]]"] ` when calling a custom button on a generic object. Seems to be a generic object specific issue, and confirmed it was not isolated just to the test environment as I was able to reproduce by making a call to a custom button. Should probably be investigated further, but for the time being reload works.

@miq-bot assign @abellotti 

### DIFF
--- a/spec/requests/custom_actions_spec.rb
+++ b/spec/requests/custom_actions_spec.rb
@@ -411,6 +411,35 @@ describe "Custom Actions API" do
     end
   end
 
+  describe "Generic Objects" do
+    before do
+      @object_definition = FactoryGirl.create(:generic_object_definition, :name => 'object def')
+      @resource = FactoryGirl.create(:generic_object, :generic_object_definition => @object_definition)
+      @button = define_custom_button1(@object_definition)
+    end
+
+    it "queries return custom actions and property methods defined" do
+      @object_definition.add_property_method("a_property_method")
+      api_basic_authorize(action_identifier(:generic_objects, :read, :resource_actions, :get))
+
+      get api_generic_object_url(nil, @resource)
+
+      expect(response.parsed_body).to include(
+        "id"      => @resource.id.to_s,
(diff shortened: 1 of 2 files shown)


spec/requests/custom_actions_spec.rb
```diff
@@ -220,4 +220,38 @@ def expect_result_to_have_custom_actions_hash
...
+    end
+
+    it "accept custom actions" do
+      api_basic_authorize
+
+      post api_generic_object_url(nil, @resource), :params => gen_request(:button1, "key1" => "value1")
+
+      expect_single_action_result(:success => true, :message => /.*/, :href => api_generic_object_url(nil, @resource))
+    end
+  end
```
abellotti: can you add a test that actions returns both custom actions from custom buttons as well as the actions from the generic object definition methods ?

abellotti:
Thanks @jntullo for the PR.  LGTM. 👍

abellotti:
LGTM!! Thanks @jntullo for making the updates. 👍

simaishi:

```
$ git log -1
commit 5a592d2f634d2389f528e2ff7a5dbb6457d9afe2
Date:   Mon Nov 20 18:49:16 2017 -0500

    Merge pull request #194 from jntullo/go_custom_actions
    
    Custom actions on generic objects
    (cherry picked from commit 4a24e71644b64917e739dfdff49686fcf8de4e4d)
```

PR #195, opened 2017-11-09
Gracefully fail on resource create

jntullo:
Calling create on an object with an incorrect `type` specified raises an internal server error. By leveraging `descendant_get` we can raise the appropriate bad request to notify the user that an incorrect type was specified. 

https://bugzilla.redhat.com/show_bug.cgi?id=1510215

@miq-bot add_label bug

### DIFF
--- a/app/controllers/api/base_controller/generic.rb
+++ b/app/controllers/api/base_controller/generic.rb
@@ -19,6 +19,7 @@ module Api
             data.delete(sc.to_s)
           end
         end
+        validate_type(klass, data['type']) if data['type']
         resource = klass.new(data)
         if resource.save
           add_subcollection_data_to_resource(resource, type, subcollection_data)
@@ -121,6 +122,12 @@ module Api
       end
 
       private
+
+      def validate_type(klass, type)
+        klass.descendant_get(type)
+      rescue ArgumentError => err
+        raise BadRequestError, "Invalid type #{type} specified - #{err}"
+      end
 
       def add_subcollection_data_to_resource(resource, type, subcollection_data)
         subcollection_data.each do |sc, sc_data|
--- a/spec/requests/orchestration_template_spec.rb
+++ b/spec/requests/orchestration_template_spec.rb
@@ -83,6 +83,21 @@ RSpec.describe 'Orchestration Template API' do
 
       expect_bad_request(/Resource id or href should not be specified/)
     end
+
+    it 'fails gracefully with invalid type specified' do
+      api_basic_authorize collection_action_identifier(:orchestration_templates, :create)
+
+      post(api_orchestration_templates_url, :params => { :type => 'OrchestrationTemplateCfn' })
+
+      expected = {
+        'error' => a_hash_including(
+          'kind'    => 'bad_request',
+          'message' => a_string_including('Invalid type OrchestrationTemplateCfn specified')
+        )
+      }
+      expect(response).to have_http_status(:bad_request)
+      expect(response.parsed_body).to include(expected)
+    end
   end
 
   context 'orchestration_template edit' do


jntullo:
@chrisarcand yeah, I agree completely. Thoughts about this change? We leverage this method for validation on other collections such as authentications, seems like this is an issue larger than just `OrchestrationTemplate` and we may want to ensure the type they're passing in is valid for the collection specified.

app/controllers/api/base_controller/generic.rb
```diff
@@ -19,6 +19,7 @@ def add_resource(type, _id, data)
             data.delete(sc.to_s)
           end
         end
+        klass = validate_type(klass, data['type']) if data['type']
```
chrisarcand: `descendant_get` returns the descendant, so I believe you're reassigning `klass` to the type class here, no? But if you are, how are tests passing? O_o  Confused.

app/controllers/api/base_controller/generic.rb
```diff
@@ -28,6 +29,12 @@ def add_resource(type, _id, data)
         end
       end
 
+      def validate_type(klass, type)
+        klass.descendant_get(type)
+      rescue => err
```
chrisarcand: Being as narrow as possible when rescuing exceptions is good, and since we know exactly the error we're expecting in this scenario, let's catch that specific one - `rescue ArgumentError => err`

app/controllers/api/base_controller/generic.rb
```diff
@@ -28,6 +29,12 @@ def add_resource(type, _id, data)
...
+        raise BadRequestError, "Invalid type #{type} specified - #{err}"
+      end
```
chrisarcand: This method is a helper and not intended for public consumption, so let's throw it under `private` below.

app/controllers/api/base_controller/generic.rb
```diff
@@ -19,6 +19,7 @@ def add_resource(type, _id, data)
             data.delete(sc.to_s)
           end
         end
+        validate_type(klass, data['type']) if data['type']
```
abellotti: are we saying data['type'] is always the class/subclass in all collections ?  or is this only for specific collections ?  If so, this might bite us later and we may want to implement in the relevant collection implementation with a super.

chrisarcand: To be honest I don't think I'd bother here. This hash is just passed raw in to the model in the next line and it's a fair to assume that passing `type` as an attribute to any ActiveRecord model is going to be problematic in some way and need adjustments if in the future there's some reason you want to be able to pass a `type` that isn't referring to a polymorphic type.

abellotti: 👍  in general too (future fine tuning as needed), I just didn't know offhand if we had such a beast today with all the collections we support. Maybe a quick perusal/sanity check.

chrisarcand: Yup, looked around already; my glance didn't yield anything that wasn't an overloaded STI class.

abellotti: Thanks !!  👍

chrisarcand:
@jntullo Thanks! Wanna add a gaprindashvili yes/no label? BZ is for master, so I wasn't sure.

jntullo:

simaishi:

```
$ git log -1
commit 2450b78c6289f37b352284de05c951affef19dca
Date:   Fri Nov 10 11:58:55 2017 -0600

    Merge pull request #195 from jntullo/bz/graceful_fail_on_delete
    
    Gracefully fail on resource create
    (cherry picked from commit 48128388cf0b47089c15c90e8caea65326b67b0e)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1513117
```

PR #196, opened 2017-11-09
Add support for /api/orchestration_stacks

jntullo:
- Add new collection
- GETs and queries
- supports custom_actions

As per earlier discussion, added in `orchestration_stacks` collection to work with custom actions. cc: @gmcculloug 

@miq-bot assign @abellotti


abellotti:
LGTM!!

The only thing needed is the followup PR to properly support subresource action identifiers and update the api.yml here and spec accordingly.  Thanks.

PR #201, opened 2017-11-12
Unmeta/expand routes

chrisarcand:
Take a deep breath. It's going to be ok, I promise.

### 'What...have...you done?'

This commit completely expands the metaprogramming based on the api.yml in the routes. It's a complete 1:1 expansion with absolutely no changes in route order or even differing ActionDispatch calls at all from the original.

### '...why?'

This is the first of many steps I'd like to take to move the API to more standard Rails patterns. It has quite a bit of metaprogramming based on the `api.yml` that is both incredibly difficult for people to follow and get up to speed with, and reinvents some things with its own abstractions that we simply don't need to maintain ourselves.

A return to normal Rails routing provides plenty of other direct benefits besides just better understanding on the part of developers. A couple random examples off the top of my head:

* **Route ordering**: There's currently no way, beyond perhaps confusingly 'ordering' the YAML, to order the router. Given the large number of routes and potential ubiquity of certain calls, there's a good chance some reordering could yield performance benefits. 

* **Non-'collection or subcollection' routes**: The simplest sort of routes that do *not* just return a resource as-is, which are easy to immediately understand in ActionDispatch, are convoluted by our own jargon of ''`arbitrary_resource_paths`' (what if you want an 'arbitrary path' on subcollections? `subcollection_arbitrary_resource_path`? Nope nope nope), a '`primary`' resource (which, as near as I can tell, is literally just a 'normal' route), etc. We're reinventing the wheel, and while these sorts of routes aren't super common at the moment, they will be as the API expands (*cough* advanced settings *cough*)

### 'But really...this actually doesn't look very good'

It's very important to note that this is the beginning and not the end result of using more standard Rails conventions - things can and should be refactored and combined in to `resource` style calls which will cut a lot of lines in this file - but to get started and avoid any future conflicts this is a good start. 

It should also be noted that this doesn't necessarily mean that no metaprogramming should be done whatsoever in the router - we may find a non-conventional pattern that makes sense to stop duplicating - but that logic should be self-contained within the router itself, and this is a good first step to just winding things back and finding those patterns, acknowledging that they are non-standard.

### 'Are you sure you did this right? I'm scared.'

Compare `rails routes` before and after this commit. They're identical, because of course I didn't actually write all of this by hand :) CC 'failing' because of intended duplication, I'm not removing it until this scripted change is merged to avoid any mistakes (see below).

### Merge plan

Because the api.yml currently controls all routing, you can actually cause a bad merge without an actual git conflict if you aren't careful. Two things need to happen:

* Before this is merged, ensure that no collections or routing options have been removed or added in the YAML since (currently) 4c17957d1a84ad2e1fac525e305fbc0963fde8ba
* After this is merged, bump every current PR that adds/removes collections or routing options to rerun their test run. (I will do this, and I mean there's only 25 open PRs anyway).

Thanks for reading my 📖  I hope I have made NickLamuro proud!

cc/ @abellotti @imtayadeway @jntullo @Fryguy

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -7,68 +7,1236 @@ Rails.application.routes.draw do
 
     get "/ping" => "ping#index"
 
-    Api::ApiConfig.collections.each do |collection_name, collection|
-      # OPTIONS action for each collection
-      match collection_name.to_s, :controller => collection_name, :action => :options, :via => :options, :as => nil
+    # /accounts
+    match 'accounts', :controller => 'accounts', :action => :options, :via => :options, :as => nil
 
-      scope collection_name, :controller => collection_name do
-        collection.verbs.each do |verb|
-          if collection.options.include?(:primary)
-            case verb
-            when :get
-              root :action => Api::VERBS_ACTIONS_MAP[verb], :as => collection_name
-            else
-              root :action => Api::VERBS_ACTIONS_MAP[verb], :via => verb
-            end
-          end
+    # /actions
+    match 'actions', :controller => 'actions', :action => :options, :via => :options, :as => nil
+    scope 'actions', :controller => 'actions' do
+      root :action => :index, :as => 'actions'
+      get '/:c_id', :action => :show, :as => 'action'
+      post '/', :action => :create, :constraints => Api::CreateConstraint.new
+      post '(/:c_id)', :action => :update
+      delete '/:c_id', :action => :destroy
+    end
 
-          next unless collection.options.include?(:collection)
+    # /alert_actions
+    match 'alert_actions', :controller => 'alert_actions', :action => :options, :via => :options, :as => nil
 
-          if collection.options.include?(:arbitrary_resource_path)
-            case verb
-            when :get
-              root :action => :index, :as => collection_name.to_s.pluralize
-              get "/*c_suffix", :action => :show, :as => collection_name.to_s.singularize
-            else
-              match "(/*c_suffix)", :action => Api::VERBS_ACTIONS_MAP[verb], :via => verb
-            end
-          else
-            case verb
-            when :get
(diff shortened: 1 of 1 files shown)


jntullo:
I think this is a great step in the right direction. Not only will this make it *much* easier to get up to speed with, it will also allow more flexibility than what we currently have when it comes to routes and actions, as mentioned. I will happily update blog post / contributing.md for this 😄

chrisarcand:
Note this isn't intended for Gaprindashvili, or at least the first release of it. For features dealing with collections or subcollections that need to be backported to Gaprindashvili (which seems pretty unlikely anyway) I'd suggest something like a `manual_routing: true` option in the yaml where you can then just paste in the routes written for master in to the Gaprindashvili routes.rb to make backporting anything exceptionally easy.

abellotti:
@chrisarcand Thanks for moving things forward. 

Can you put an example of how the post refactoring of routes.rb of a single collection and related subcollections would look like ?

Fryguy:
> CC 'failing' because of intended duplication

Can this file just be ignored in the codeclimate.yml?

chrisarcand:
> Can you put an example of how the post refactoring of routes.rb of a single collection and related subcollections would look like ?

Yeah, I'll do that, sure.

> Can this file just be ignored in the codeclimate.yml?

Good call.


This needs to be re-run again to be sure no routes are missing since I opened it; I'll circle back around soon.

chrisarcand:
I'm closing this for now as I don't have the bandwidth to finish it and follow through with the needed work to make it more manageable and Rails-like now. However, @jntullo let me know if you'd like to pick up the torch here, in which case I could probably dig up the script I used to generate these changes for me!

PR #202, opened 2017-11-13
Return 404 error if perform an action against a non existent Physical Server

douglasgabriel:
This PR is able to:
- Fix the error that returns a 200 code if try to perform an action against a non existent PH;
- Return 404 if the ID of the Physical Server doesn't exist.

### DIFF
--- a/app/controllers/api/physical_servers_controller.rb
+++ b/app/controllers/api/physical_servers_controller.rb
@@ -43,6 +43,8 @@ module Api
     def refresh_resource(type, id, _data = nil)
       raise BadRequestError, "Must specify an id for refreshing a #{type} resource" unless id
 
+      ensure_resource_exists(type, id) if single_resource?
+
       api_action(type, id) do |klass|
         physical_server = resource_search(id, type, klass)
         api_log_info("Refreshing #{physical_server_ident(physical_server)}")
@@ -55,6 +57,8 @@ module Api
     def change_resource_state(state, type, id)
       raise BadRequestError, "Must specify an id for changing a #{type} resource" unless id
 
+      ensure_resource_exists(type, id) if single_resource?
+
       api_action(type, id) do |klass|
         begin
           server = resource_search(id, type, klass)
@@ -66,6 +70,10 @@ module Api
           action_result(false, err.to_s)
         end
       end
+    end
+
+    def ensure_resource_exists(type, id)
+      raise NotFoundError unless collection_class(type).exists?(id)
     end
 
     def server_ident(server)
--- a/spec/requests/physical_servers_spec.rb
+++ b/spec/requests/physical_servers_spec.rb
@@ -237,6 +237,66 @@ RSpec.describe "physical_servers API" do
         expect(response.parsed_body["error"]).to include("kind" => "forbidden")
       end
     end
+
+    context "with a non existent physical server" do
+      actions = [
+        :power_on,
+        :power_off,
+        :power_off_now,
+        :restart,
+        :restart_now,
+        :restart_to_sys_setup,
+        :restart_mgmt_controller
+      ]
+
+      actions.each do |action|
+        it "fails to #{action} a server" do
+          api_basic_authorize(action_identifier(:physical_servers, action, :resource_actions, :post))
+
+          post(api_physical_server_url(nil, 999_999), :params => gen_request(action))
+
+          expect(response).to have_http_status(:not_found)
+        end
+      end
+    end
+
+    context "with an existent and a non existent physical server" do
+      actions = [
+        :power_on,
+        :power_off,
+        :power_off_now,
+        :restart,
+        :restart_now,
+        :restart_to_sys_setup,
+        :restart_mgmt_controller
+      ]
+
+      actions.each do |action|
+        it "returns status 200 and a failure message for the non existent physical server" do
+          ps = FactoryGirl.create(:physical_server)
+          api_basic_authorize(action_identifier(:physical_servers, action, :resource_actions, :post))
+
+          post(api_physical_servers_url, :params => gen_request(action, [{"href" => api_physical_server_url(nil, ps)}, {"href" => api_physical_server_url(nil, 999_999)}]))
+
+          expected = {
+            "results" => a_collection_containing_exactly(
+              a_hash_including(
+                "message" => a_string_matching(/#{ps.id}/),
+                "success" => true,
+                "href"    => api_physical_server_url(nil, ps)
+              ),
+              a_hash_including(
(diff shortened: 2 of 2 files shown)


douglasgabriel:

douglasgabriel:
Hi @imtayadeway @agrare, could you review this PR?

douglasgabriel:

abellotti:
while this might be ok for targeting a single resource, this will be incompatible with and break the behavior on bulk requests.  Note that for bulk requests, we don't want to interrupt the request mid-flight and return the 404, callers won't know what was, wasn't done.  The action result signature while is a 200 has the success attribute which can be queried.  This allows us to complete bulk requests even if a couple of resources in the middle of the request do not exist.

jntullo:
I agree with @abellotti. You may want to look into what has been done on [delete](https://github.com/ManageIQ/manageiq-api/blob/4c17957d1a84ad2e1fac525e305fbc0963fde8ba/app/controllers/api/base_controller/results.rb#L6), you may be able to do something similar to handle both scenarios if this is required.

douglasgabriel:
@abellotti and @jntullo I believe that with the changes on [here](https://github.com/ManageIQ/manageiq-api/pull/202/files#diff-ad8b01173fa2500da23b667a54ad56b0) this problem is solved, right?

jntullo:
@douglasgabriel that update has unintended consequences to other controllers as can be seen by the Travis failures. For now, I would keep the update contained to this particular method.

douglasgabriel:
Sorry about that @jntullo, I fixed this

jntullo:
@douglasgabriel can you please add tests? Thanks!

app/controllers/api/base_controller.rb
```diff
@@ -102,6 +102,10 @@ def options
       render_options(@req.collection)
     end
 
+    def single_resource?
+      !@req.json_body.key?('resources')
+    end
+
```
abellotti: I think this method should probably go in lib/api/request_adapter.rb,  WDYT @jntullo ?

jntullo: @abellotti yeah, i agree with that

douglasgabriel: @abellotti @jntullo Done :+1:

abellotti:
LGTM.  @jntullo ?

app/controllers/api/physical_servers_controller.rb
```diff
@@ -52,6 +52,8 @@ def change_resource_state(state, type, id)
           api_log_info(desc)
           task_id = queue_object_action(server, desc, :method_name => state, :role => :ems_operations)
           action_result(true, desc, :task_id => task_id)
+        rescue ActiveRecord::RecordNotFound => err
+          single_resource? ? raise(err) : action_result(false, err.to_s)
```
imtayadeway: I'm not sure I understand how this is working. If I understand this correctly, the call to `single_resource?` here should raise a `NoMethodError`

douglasgabriel: @imtayadeway it shouldn't raise any NoMethodError because this method is defined on its hierarchy on file [results.rb](https://github.com/ManageIQ/manageiq-api/blob/master/app/controllers/api/base_controller/results.rb), Besides this there is a spec for this method that confirms that it works well

abellotti:
ping @imtayadeway @jntullo any updates needed here ?

douglasgabriel:
Hm, good point and good suggestion, I agree with you

douglasgabriel:
It's done @imtayadeway :+1:

app/controllers/api/physical_servers_controller.rb
```diff
@@ -80,6 +78,18 @@ def refresh_physical_server(physical_server)
       action_result(false, err.to_s)
     end
 
+    def ensure_id_present(type, id)
+      raise BadRequestError, "Must specify an id for changing a #{type} resource" unless id
+    end
+
+    def ensure_resource_exists(type, id)
+      raise ActiveRecord::RecordNotFound if single_resource? && !collection_class(type).exists?(id)
```
abellotti: why is this method needed ?  resource_search would throw the NotFoundError plus do the necessary RBAC validation.

imtayadeway: @abellotti it does, but we don't want it caught by the `api_action` block, which will always respond with a 200. Performing another `resource_search` would be unnecessarily expensive if we only want to check for existence.

@douglasgabriel you may want to use our own `NotFoundError` here

app/controllers/api/physical_servers_controller.rb
```diff
@@ -41,7 +41,8 @@ def restart_mgmt_controller_resource(type, id, _data)
     end
 
     def refresh_resource(type, id, _data = nil)
-      raise BadRequestError, "Must specify an id for refreshing a #{type} resource" unless id
```
imtayadeway: I think it would be OK not to extract this into a private method, since that alters the message to talk about "changing" here, instead of "refreshing". Again, I think it's an acceptable structural similarity.

app/controllers/api/physical_servers_controller.rb
```diff
@@ -41,7 +41,8 @@ def restart_mgmt_controller_resource(type, id, _data)
...
+      ensure_id_present(type, id)
+      ensure_resource_exists(type, id)
```
imtayadeway: I think it may be more appropriate to say:

```ruby
ensure_resource_exists(type, id) if single_request?
```

Since we're trying to express that we don't care if a resource exists at this point for a bulk operation, because it's handled elsewhere.

app/controllers/api/physical_servers_controller.rb
```diff
@@ -68,10 +70,6 @@ def change_resource_state(state, type, id)
       end
     end
 
-    def server_ident(server)
```
imtayadeway: Was there a reason this was moved? I think it would be better to leave it here if it is not essential to fixing the bug

douglasgabriel: Well, ok. it was just for code organization reason

douglasgabriel:
Thank you for the support :+1:

simaishi:

```
$ git log -1
commit a5e20f83f4b4a760e4e1b839589a3496ed17e7e2
Date:   Mon Dec 18 08:42:29 2017 -0600

    Merge pull request #202 from douglasgabriel/XCS-503
    
    Return 404 error if perform an action against a non existent Physical Server
    (cherry picked from commit d17fd1859f46e7fb83f1ef604418efc669a3550d)
```

PR #204, opened 2017-11-13
Do not expand custom actions on collection searches

abellotti:
Do not expand custom actions if the user is querying a collection. Only show custom actions if an individual resource is fetched.

### DIFF
--- a/app/controllers/api/base_controller/renderer.rb
+++ b/app/controllers/api/base_controller/renderer.rb
@@ -45,7 +45,6 @@ module Api
 
           unless @req.hide?("resources") || collection_option?(:hide_resources)
             key_id = collection_config.resource_identifier(type)
-            opts[:collection_search] = true
             json.resources resources.collect do |resource|
               if opts[:expand_resources]
                 add_hash json, resource_to_jbuilder(type, reftype, resource, opts).attributes!
@@ -83,7 +82,7 @@ module Api
         expand_subcollections(json, type, resource) if resource.respond_to?(:attributes)
 
         expand_actions(resource, json, type, opts, physical_attrs) if opts[:expand_actions]
-        expand_resource_custom_actions(resource, json, type, physical_attrs) if opts[:expand_actions] && !opts[:collection_search]
+        expand_resource_custom_actions(resource, json, type, physical_attrs) if opts[:expand_custom_actions]
         json
       end
 
--- a/app/controllers/api/base_controller/renderer.rb
+++ b/app/controllers/api/base_controller/renderer.rb
@@ -45,6 +45,7 @@ module Api
 
           unless @req.hide?("resources") || collection_option?(:hide_resources)
             key_id = collection_config.resource_identifier(type)
+            opts[:collection_search] = true
             json.resources resources.collect do |resource|
               if opts[:expand_resources]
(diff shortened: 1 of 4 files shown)


abellotti:
Marking as WIP for now.

TODO:
- Verify UI usage/expectations
- Decide if the same needs to be done on bulk queries
- Add specs

app/controllers/api/base_controller/renderer.rb
```diff
@@ -45,6 +45,7 @@ def collection_to_jbuilder(type, reftype, resources, opts = {})
 
           unless @req.hide?("resources") || collection_option?(:hide_resources)
             key_id = collection_config.resource_identifier(type)
+            opts[:collection_search] = true
```
chrisarcand: * Hmm? It's not intuitive to me that this block of code directly relates to `collection_search` 😕  I probably just don't understand the implications of the logic on line 46
* Even if 'we don't want to expand custom actions on collection search' is your intention, don't you just want to pass an option for `expand_custom_actions` in this method to maintain clean semantics that are already present in `resource_to_jbuilder` instead of teaching `resource_to_jbuilder` that there's such a thing as a collection search that it seems to care about for some reason?

abellotti: both good 👍

abellotti:
@AllenBW I need to borrow your 👀  a bit.  Can you find out for us if currently in the services list in the SUI do you use/expose any of the custom actions of services as defined by the custom action buttons ? 🙇   I understand custom actions/buttons would be needed once you drill into a service but didn't know if you needed that information at the collection/list view.  Thanks.

AllenBW:
@abellotti on it! 👍  will report back whennnnn question has an answer

AllenBW:
@abellotti presently the call made on the service list view is 
`http://localhost:3001/api/services?expand=resources&limit=20&offset=0&attributes=picture%2Cpicture.image_href%2Cchargeback_report%2Cevm_owner.userid%2Cv_total_vms%2Cpower_state%2Call_service_children%2Ctags&filter[]=ancestry%3Dnull&filter[]=display%3Dtrue&sort_by=created_at&sort_options=&sort_order=desc`

so no custom_actions 👍

AllenBW:
@abellotti Nope!  does not.

abellotti:
Thanks @AllenBW looks like we're good here. 🙇 

Optimization/Preloading of additional attributes (as per attributes=...) would still be needed, but that's a separate PR.

Fryguy:
Does this remove all custom actions including the global ones?

abellotti:
yes sir, _all_ custom actions. these are returned to us via _custom_action_buttons_ from the CustomActionsMixin in the model.

abellotti:
@Fryguy marking as g/yes, as we _need_ this there before all the custom_actions enablement API PRs can be merged.  in G-release we've also upped the major semver for API to 3.0.0 so I'm ok with at least setting g/yes.  Thanks.

simaishi:

```
 $ git log -1
commit 127332b08bd10d91c27b31fe6d78345588baa684
Date:   Fri Nov 17 12:21:19 2017 -0500

    Merge pull request #204 from abellotti/do_not_expand_custom_actions_on_collections
    
    Do not expand custom actions on collection searches
    (cherry picked from commit a60b747e593ce75dea430aa13f85296bf44ad431)
```

PR #205, opened 2017-11-13
Return correct href slug on the VMs collection

jntullo:
For cloud vms, the href_slug returns `instances/:id` instead of `vms/:id`. As per BZ, the href slug is supposed to respect and match the href.

https://bugzilla.redhat.com/show_bug.cgi?id=1510238

@miq-bot assign @abellotti

### DIFF
--- a/app/controllers/api/vms_controller.rb
+++ b/app/controllers/api/vms_controller.rb
@@ -263,6 +263,10 @@ module Api
       action_result(false, "Failed to set miq_server - #{err}")
     end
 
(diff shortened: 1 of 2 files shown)


abellotti:
Same root cause problem as https://github.com/ManageIQ/manageiq-api/pull/145 ?

jntullo:
@abellotti yeah

jntullo:
Closing, resolved via https://github.com/ManageIQ/manageiq-api

PR #210, opened 2017-11-14
OPTIONS for Custom Button Sets

AparnaKarve:
Added `OPTIONS` for `Custom Button Sets`

To retrieve the CustomButtonSets for `GenericObjectDefinition` and `ServiceTemplate`

The `OPTIONS` call is required by this PR - https://github.com/ManageIQ/manageiq-ui-classic/pull/2733

### DIFF
--- a/app/controllers/api/custom_button_sets_controller.rb
+++ b/app/controllers/api/custom_button_sets_controller.rb
@@ -7,5 +7,18 @@ module Api
     def edit_resource(type, id, data)
       super(type, id, data.deep_symbolize_keys)
     end
+
+    def options
+      render_options(:custom_button_sets, build_custom_button_sets_options)
+    end
+
+    def build_custom_button_sets_options
+      {
+        :custom_button_sets => {
+          :generic_object_definitions => CustomButtonSet.find_all_by_class_name('GenericObjectDefinition'),
+          :service_templates          => CustomButtonSet.find_all_by_class_name('ServiceTemplate')
+        }
+      }
+    end
   end
 end
--- a/spec/requests/custom_button_sets_spec.rb
+++ b/spec/requests/custom_button_sets_spec.rb
@@ -199,4 +199,19 @@ RSpec.describe 'CustomButtonSets API' do
       expect(response.parsed_body).to include(expected)
     end
   end
+
+  describe 'OPTIONS /api/custom_button_sets' do
+    it 'returns custom_button_sets for Generic Object Definitions and Service Templates' do
+      options(api_custom_button_sets_url)
+
+      expected_data = {
+        'custom_button_sets' => {
+          'generic_object_definitions' => CustomButtonSet.find_all_by_class_name('GenericObjectDefinition'),
+          'service_templates'          => CustomButtonSet.find_all_by_class_name('ServiceTemplate')
+        }
+      }
+
+      expect_options_results(:custom_button_sets, expected_data)
+    end
+  end
 end


AparnaKarve:
@miq_bot add_label gaprindashvili/yes

chrisarcand:
⚠️  Note, as we've been having performance woes occur from custom button sets, this should be reviewed from a performance standpoint, too.

AparnaKarve:
@chrisarcand Any idea why are we having performance woes for custom button sets?

app/controllers/api/custom_button_sets_controller.rb
```diff
@@ -7,5 +7,16 @@ def create_resource(type, id, data)
...
+
+    def options
+      render_options(:custom_button_sets, build_custom_button_sets_options)
+    end
+
+    def build_custom_button_sets_options
+      {
+        :custom_button_sets_for_generic_object_definitions => CustomButtonSet.find_all_by_class_name('GenericObjectDefinition'),
+        :custom_button_sets_for_service_templates          => CustomButtonSet.find_all_by_class_name('ServiceTemplate')
+      }
```
abellotti: maybe:

```
{
   :custom_button_sets => {
      "generic_object_definitions" => CustomButtonSet.find_all....,
      "service_templates" => CustomButtonSet.find_all....,
   }
}
```

abellotti:
@AparnaKarve I don't have any in my appliance, but I did see 4 queries made, can you check on your env. that it's still just the 4 queries.  Thanks.

```
D, [2017-11-15T18:42:34.479341 #80262] DEBUG -- :   CustomButtonSet Load (0.5ms)  SELECT "miq_sets".* FROM "miq_sets" WHERE "miq_sets"."set_type" IN ('CustomButtonSet') AND (set_data like '%:applies_to_class: GenericObjectDefinition%') AND (set_data like '%:applies_to_id: 0%')
D, [2017-11-15T18:42:34.479527 #80262] DEBUG -- :   CustomButtonSet Inst Including Associations (0.0ms - 0rows)
D, [2017-11-15T18:42:34.480455 #80262] DEBUG -- :   CustomButtonSet Load (0.4ms)  SELECT "miq_sets".* FROM "miq_sets" WHERE "miq_sets"."set_type" IN ('CustomButtonSet') AND (set_data like '%:applies_to_class: GenericObject%') AND (set_data not like '%:applies_to_id%')
D, [2017-11-15T18:42:34.480640 #80262] DEBUG -- :   CustomButtonSet Inst Including Associations (0.0ms - 0rows)
D, [2017-11-15T18:42:34.481390 #80262] DEBUG -- :   CustomButtonSet Load (0.3ms)  SELECT "miq_sets".* FROM "miq_sets" WHERE "miq_sets"."set_type" IN ('CustomButtonSet') AND (set_data like '%:applies_to_class: ServiceTemplate%') AND (set_data like '%:applies_to_id: 0%')
D, [2017-11-15T18:42:34.481566 #80262] DEBUG -- :   CustomButtonSet Inst Including Associations (0.0ms - 0rows)
D, [2017-11-15T18:42:34.482343 #80262] DEBUG -- :   CustomButtonSet Load (0.4ms)  SELECT "miq_sets".* FROM "miq_sets" WHERE "miq_sets"."set_type" IN ('CustomButtonSet') AND (set_data like '%:applies_to_class: Service%') AND (set_data not like '%:applies_to_id%')
D, [2017-11-15T18:42:34.482535 #80262] DEBUG -- :   CustomButtonSet Inst Including Associations (0.0ms - 0rows)
I, [2017-11-15T18:42:34.483424 #80262]  INFO -- : Completed 200 OK in 9ms (Views: 0.3ms | ActiveRecord: 2.4ms)
```

AparnaKarve:
@abellotti I can confirm that only those 4 queries run for the `options` call.

app/controllers/api/custom_button_sets_controller.rb
AparnaKarve: Made the suggested change. Thanks.

AparnaKarve:
> I can confirm that only those 4 queries run for the options call.

@abellotti While I have confirmed that it's 2 queries being run for each call, I would also like to add that, in the UI, currently, I'm only concerned with the `GenericObjectDefinition` specific call.

The other call with `ServiceTemplate`, is not being used at the moment - but will be used when we angularize the Custom Button Groups for `ServiceTemplate`

abellotti:
Thanks @AparnaKarve for updating this.

While I'm ok with a couple here, will this call ever need to be expanded to return the CustomButtonSets of all the collections we're adding custom_actions to ?  I know of 25 being added shortly.  in which case it might be good to understand how expensive this call might end up being and if really bad, should we come up with an alternative if the information is not needed for all. Something to think about.

AparnaKarve:
> I know of 25 being added shortly. in which case it might be good to understand how expensive this call might end up being and if really bad, should we come up with an alternative if the information is not needed for all. Something to think about.

As far as I know, this call would be only needed in Angular forms. 
So if the other 25 end up being forms, then I agree - this will be an expensive operation.

The `OPTIONS` call does not usually take parameters, but if there was a way to specify that I need the call only for `GenericObjectDefinition` or `ServiceTemplate`, then we could limit the number of queries being run.

One other alternative I can think of is - drop the OPTIONS call completely, and let the JS do some parsing on the `response` object returned by `GET /api/custom_button_sets?expand=resources&attributes=set_data`

spec/requests/custom_button_sets_spec.rb
```diff
@@ -199,4 +199,19 @@
...
   end
+
+  describe 'OPTIONS /api/custom_button_sets' do
+    it 'returns custom_button_sets for Generic Object Definitions and Service Templates' do
+      options(api_custom_button_sets_url)
+
+      expected_data = {
+        'custom_button_sets' => {
+          'generic_object_definitions' => CustomButtonSet.find_all_by_class_name('GenericObjectDefinition'),
+          'service_templates'          => CustomButtonSet.find_all_by_class_name('ServiceTemplate')
```
Fryguy: This is not right to create a test that is the exact same code at the code itself.  You're not really testing anything.  cc @chrisarcand Thoughts?

AparnaKarve: Agree that the test here is not ideal - and mocking the expected result is also not an option.
So yes, ideas are welcome.

AparnaKarve: On second thoughts, I think we could write a proper test here.

Although I need to know if we are going ahead with this PR, before I rectify the spec.

app/controllers/api/custom_button_sets_controller.rb
```diff
@@ -7,5 +7,18 @@ def create_resource(type, id, data)
...
     end
+
+    def options
+      render_options(:custom_button_sets, build_custom_button_sets_options)
+    end
+
+    def build_custom_button_sets_options
+      {
+        :custom_button_sets => {
+          :generic_object_definitions => CustomButtonSet.find_all_by_class_name('GenericObjectDefinition'),
```
imtayadeway: If this returns `CustomButtonSet` objects - I'm wondering if this is appropriate for an OPTIONS request, which should only return metadata. Perhaps this could be solved with a filter on the GET #index call?

AparnaKarve: Yes, I suggested something on similar lines above -

```
GET /api/custom_button_sets?expand=resources&attributes=set_data
```

If it were possible to add a filter as shown below, that would address my requirement -
```
filter[]=set_data[:applies_to_class]=GenericObject
```

AparnaKarve:
@imtayadeway and I discussed an alternative for this, which was to add a virtual column to `CustomButtonSet` to retrieve the data that we need. 
But we are not pursuing this alternative.

As I had mentioned above in https://github.com/ManageIQ/manageiq-api/pull/210#issuecomment-345268206, I think I can get this done on the JS side itself with a little bit of creative parsing/filtering. So that's the approach I am going with.
Hence closing this PR...
Thanks everyone for your input!

spec/requests/custom_button_sets_spec.rb
chrisarcand: I know this PR is scrapped but just to answer the question for future knowledge...

Yes, you should be creating those custom button sets ahead of time with helpers/variables and referring to those directly in your test, ie:

```ruby
expected_data = {
  'custom_button_sets => {
    'generic_custom_object_definitions' => cb_set # You created cb_set earlier, now refer to it
# ......
```

PR #214, opened 2017-11-16
Adding support for optional token_ttl in the UserTokenService.

abellotti:
Adding support for optional token_ttl in the UserTokenService
- added optional token_ttl parameter to generate_token method
- added specs

### DIFF
--- a/lib/services/api/user_token_service.rb
+++ b/lib/services/api/user_token_service.rb
@@ -23,17 +23,19 @@ module Api
       @api_config ||= ::Settings[base_config[:module]].to_hash
     end
 
-    def generate_token(userid, requester_type)
+    def generate_token(userid, requester_type, token_ttl = nil)
       validate_userid(userid)
       validate_requester_type(requester_type)
-
-      $api_log.info("Generating Authentication Token for userid: #{userid} requester_type: #{requester_type}")
 
       # Additional Requester type token ttl's for authentication
       type_to_ttl_override = {'ui' => ::Settings.session.timeout}
 
-      token_mgr(requester_type).gen_token(:userid             => userid,
-                                          :token_ttl_override => type_to_ttl_override[requester_type])
+      # Let's honor the optional token_ttl
+      token_ttl ||= type_to_ttl_override[requester_type]
+
+      $api_log.info("Generating Authentication Token for userid: #{userid} requester_type: #{requester_type} token_ttl: #{token_ttl}")
+
+      token_mgr(requester_type).gen_token(:userid => userid, :token_ttl_override => token_ttl)
     end
 
     def validate_requester_type(requester_type)
--- a/lib/services/api/user_token_service.rb
+++ b/lib/services/api/user_token_service.rb
@@ -23,14 +23,13 @@ module Api
       @api_config ||= ::Settings[base_config[:module]].to_hash
     end
 
-    def generate_token(userid, requester_type, token_ttl = nil)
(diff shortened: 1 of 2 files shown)


abellotti:
Needed for Ansible playbooks.

/cc @mkanoor as discussed earlier, can you give it a try ?  Thanks.

lib/services/api/user_token_service.rb
```diff
@@ -23,17 +23,19 @@ def api_config
       @api_config ||= ::Settings[base_config[:module]].to_hash
     end
 
-    def generate_token(userid, requester_type)
+    def generate_token(userid, requester_type, token_ttl = nil)
```
chrisarcand: Use a keyword argument to make callers more explicit and allow for easier additions in the future?

spec/lib/services/api/user_token_service_spec.rb
```diff
@@ -0,0 +1,27 @@
+RSpec.describe Api::UserTokenService do
+  describe ".generate_token" do
+    before do
+      @user = FactoryGirl.create(:user_with_group)
+    end
+
+    it "uses default token_ttl" do
+      user_token_service = described_class.new
+      token = user_token_service.generate_token(@user.userid, 'api')
```
chrisarcand: How about removing these two lines and lines 18 and 19 below and instead doing setup only once and changing what you need:

```ruby
let(:user_token_service) { described_class.new }
let(:token) { user_token_service.generate_token(@user.userid, 'api', token_ttl: token_ttl) }

context "without token_ttl set" do
   let(:token_ttl) { nil }

   it "uses the default from settings" do
     # expect....
   end
end

# ...

File: lib/services/api/user_token_service.rb [hunk 3]
```diff
@@ -23,17 +23,19 @@ def api_config
...
       validate_requester_type(requester_type)
 
-      $api_log.info("Generating Authentication Token for userid: #{userid} requester_type: #{requester_type}")
-
       # Additional Requester type token ttl's for authentication
       type_to_ttl_override = {'ui' => ::Settings.session.timeout}
 
-      token_mgr(requester_type).gen_token(:userid             => userid,
-                                          :token_ttl_override => type_to_ttl_override[requester_type])
+      # Let's honor the optional token_ttl
```
chrisarcand: I don't think this comment is really necessarily, honestly.

jvlcek:
👍 LGTM

lib/services/api/user_token_service.rb
mkanoor: what would we do for backward compatibility if we use keyword argument, we would have to change all callers to use the keyword argument style

chrisarcand: huh? Can you tell me a hypothetical example of what you mean? I'm talking about using a keyword arg for `token_ttl` specifically, not the other two, if that isn't clear. See example in my comment below in the tests. The token_ttl is a defaulted, optional parameter (as opposed to the other two) which usually presents itself well as a keyword argument.

abellotti: Usage would be same as today:

```
UserTokenService.new.generate_token(userid, 'api')
```

With the keyword arg, if you need to pass an optional token, you'd pass by name:

```
UserTokenService.new.generate_token(userid, 'api', token_ttl:  ansible_playbook_ttl)
```

chrisarcand: Yup! ❤️

PR #220, opened 2017-11-21
Tasks: delete support

### LINKED ISSUE #203: Tasks - missing delete support
Right now, there is no support for `DELETE /api/tasks/123` (nor for `action: "delete"`).

We need one :). (for https://github.com/ManageIQ/manageiq/pull/16441 + https://github.com/ManageIQ/manageiq-ui-classic/pull/2708)

---

The idea is, when using tasks in the UI, via the API, there's no way to actually clear that task.

himdel:
This adds deletion for `MiqTask`, both via `DELETE` and via `POST {action:"delete"}`, *and* bulk deletion (`POST {action:"delete", resources: [...]}`).

Closes #203 

(Needed for https://github.com/ManageIQ/manageiq-ui-classic/pull/2708.)

---

Is it OK to reuse existing identifiers (`miq_task_delete`) or is that frowned upon? :)

### DIFF
--- a/spec/requests/tasks_spec.rb
+++ b/spec/requests/tasks_spec.rb
@@ -20,38 +20,42 @@ describe 'TasksController' do
   it 'deletes on POST' do
     api_basic_authorize resource_action_identifier(:tasks, :delete)
 
-    post(api_task_url(nil, task.id), :params => {
+    data = {
       :action => 'delete'
-    })
+    }
+    post(api_task_url(nil, task.id), :params => data)
 
     expect(response).to have_http_status(:ok) # 200
     expect_deleted(task)
 
-    expect(response.parsed_body).to include({
+    expected = {
       'success' => true,
       'message' => "tasks id: #{task.id} deleting"
-    })
+    }
+    expect(response.parsed_body).to include(expected)
   end
 
   it 'bulk deletes' do
     api_basic_authorize collection_action_identifier(:tasks, :delete)
 
-    post(api_tasks_url, :params => {
+    data = {
       :action    => 'delete',
       :resources => [
         {:href => api_task_url(nil, task.id)},
         {:href => api_task_url(nil, task2.id)}
       ]
-    })
+    }
+    post(api_tasks_url, :params => data)
 
     expect(response).to have_http_status(:ok) # 200
     expect_deleted(task, task2)
 
-    expect(response.parsed_body).to include({
+    expected = {
       'results' => a_collection_including(
         a_hash_including('success' => true, 'message' => "tasks id: #{task.id} deleting"),
         a_hash_including('success' => true, 'message' => "tasks id: #{task2.id} deleting")
       )
-    })
+    }
+    expect(response.parsed_body).to include(expected)
   end
 end
--- a/spec/requests/tasks_spec.rb
+++ b/spec/requests/tasks_spec.rb
@@ -11,7 +11,7 @@ describe 'TasksController' do
   it 'deletes on DELETE' do
     api_basic_authorize resource_action_identifier(:tasks, :delete, :delete)
 
-    delete(api_task_url(nil, task.id))
+    delete(api_task_url(nil, task))
 
     expect(response).to have_http_status(:no_content) # 204
     expect_deleted(task)
@@ -23,7 +23,7 @@ describe 'TasksController' do
     data = {
       :action => 'delete'
     }
-    post(api_task_url(nil, task.id), :params => data)
+    post(api_task_url(nil, task), :params => data)
 
     expect(response).to have_http_status(:ok) # 200
     expect_deleted(task)
@@ -41,8 +41,8 @@ describe 'TasksController' do
     data = {
       :action    => 'delete',
       :resources => [
-        {:href => api_task_url(nil, task.id)},
-        {:href => api_task_url(nil, task2.id)}
+        {:href => api_task_url(nil, task)},
+        {:href => api_task_url(nil, task2)}
       ]
     }
     post(api_tasks_url, :params => data)
--- a/spec/requests/tasks_spec.rb
+++ b/spec/requests/tasks_spec.rb
@@ -1,0 +1,57 @@
+describe 'TasksController' do
+  let(:task) { FactoryGirl.create(:miq_task, :state => MiqTask::STATE_FINISHED) }
+  let(:task2) { FactoryGirl.create(:miq_task, :state => MiqTask::STATE_FINISHED) }
+
+  def expect_deleted(*args)
+    args.each do |arg|
+      expect(MiqTask.find_by(:id => arg.id)).to be_nil
+    end
+  end
+
+  it 'deletes on DELETE' do
+    api_basic_authorize resource_action_identifier(:tasks, :delete, :delete)
+
+    delete(api_task_url(nil, task.id))
+
+    expect(response).to have_http_status(:no_content) # 204
+    expect_deleted(task)
+  end
+
+  it 'deletes on POST' do
+    api_basic_authorize resource_action_identifier(:tasks, :delete)
+
+    post(api_task_url(nil, task.id), :params => {
+      :action => 'delete'
+    })
+
+    expect(response).to have_http_status(:ok) # 200
+    expect_deleted(task)
+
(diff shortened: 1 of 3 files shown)


jntullo:
Hi @himdel can you add some specs for this? Is there any reason you didn't add delete on the collection (for bulk delete)? and you can reuse existing identifiers :)

himdel:
@jntullo Sure, will do :) (wasn't quite sure if we wanted to spec methods that aren't technically there, but definitely can't hurt :))

As for bulk delete.. I just didn't think of it :).. but there I have questions.. (I'm not seeing any docs on custom collection actions, sorry :).)

Collection actions are named `foo_collection`, similar to `foo_resource` or is there a different convention? I'm not seeing any such `delete_collection` in the generic code, should I add it there, or should this be strictly for task? Or .. am I misunderstanding and does it magically know to call the resource action on all the resources mentioned?

himdel:
... Never mind :).

It's really that easy, it does seem to call the resource action for each of the resources .. so while I still don't know how to custom collection action, I don't seem to need it right now :).

Addded, on to the specs now :)

spec/requests/tasks_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+  end
+
+  it 'deletes on DELETE' do
+    api_basic_authorize resource_action_identifier(:tasks, :delete, :delete)
+
+    url = api_task_url(nil, task.id)
+
+    delete(url)
+
+    expect(response).to have_http_status(204)
```
jntullo: For the http status, could you use `:forbidden` rather than `204`? will help to keep our specs consistent

spec/requests/tasks_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+
+  it 'deletes on POST' do
+    api_basic_authorize resource_action_identifier(:tasks, :delete)
+
+    url = api_task_url(nil, task.id)
+    data = {:action => 'delete'}
+
+    post(url, :params => data)
+
+    expect(response).to have_http_status(200)
```
jntullo: Similarly, you could use `:ok` instead of `200`

spec/requests/tasks_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+  it 'bulk deletes' do
+    api_basic_authorize collection_action_identifier(:tasks, :delete)
+
+    url = api_tasks_url
+    data = {
+      :action    => 'delete',
+      :resources => [
+        {:href => api_task_url(nil, task.id)},
+        {:href => api_task_url(nil, task2.id)},
+      ],
```
jntullo: nit, but no `,` here

spec/requests/tasks_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    post(url, :params => data)
+
+    expect(response).to have_http_status(200)
+    expect_deleted(task, task2)
```
jntullo: Generally, the API is concerned with the responses rather than validating the objects, something like [this](https://github.com/ManageIQ/manageiq-api/blob/1e91ed2b7ba2d20726ac8d76cdeb69e7bc8f7f3d/spec/requests/custom_button_sets_spec.rb#L136-L141)

spec/requests/tasks_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+    post(url, :params => data)
+
+    expect(response).to have_http_status(200)
+    expect_deleted(task)
+  end
+
+  it 'bulk deletes' do
+    api_basic_authorize collection_action_identifier(:tasks, :delete)
+
+    url = api_tasks_url
```
jntullo: are you only using this variable in one place? May not be necessary to assign it

spec/requests/tasks_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+    delete(url)
+
+    expect(response).to have_http_status(204)
+    expect_deleted(task)
+  end
+
+  it 'deletes on POST' do
+    api_basic_authorize resource_action_identifier(:tasks, :delete)
+
+    url = api_task_url(nil, task.id)
```
jntullo: You may not need to assign this to a variable if you're only calling it once, same as above and below

spec/requests/tasks_spec.rb
himdel: Probably not ... I guess `:empty`?

himdel: Turns out `:no_content`, according to http://billpatrianakos.me/blog/2013/10/13/list-of-rails-status-code-symbols/

That's why I prefer the numerical values, no confusion about what it could be :).

spec/requests/tasks_spec.rb
himdel: Sure I don't.

But is it not clearer to write..

```
url = function call with not-so clear params
data = create the structure

post(url, data)
```

than to have it all in one line, like

```ruby
  post(api_task_url(nil, task.id), {
      :action    => 'delete',
      :resources => [
        {:href => api_task_url(nil, task.id)},
        {:href => api_task_url(nil, task2.id)},
      ],
    })
```

But of course your repo, your rules, just tell me you're sure you want it that way :).

spec/requests/tasks_spec.rb
himdel: (fixed)

spec/requests/tasks_spec.rb
himdel: fixed :) Funnily enough, in ui, trailing commas are strongly encouradged :).

spec/requests/tasks_spec.rb
himdel: Well, I don't care about the response at all, only that it actually deleted the thing :).

Should I add something like that then?

But.. more and more this looks like I'm testing your generic functions for one specific case, which is identical to all the other specific cases.

Would you be interested in specs for the generic `resource_delete` instead?

himdel: Or at least a set of helpers so that we can write the tests once, and then only provide that things that should be different?

That task_spec should not really have to be more code than something like..

```
describe 'TasksController' do
  let(:task) { FactoryGirl.create(:miq_task, :state => MiqTask::STATE_FINISHED) }
  let(:task2) { FactoryGirl.create(:miq_task, :state => MiqTask::STATE_FINISHED) }

  test_deletion(:methods => [:delete, :post, :bulk], :resources => [:task, :task2])
end
```

spec/requests/tasks_spec.rb
imtayadeway: My two penneth: I think "clearer" is always going to be a bit subjective in these cases. I personally find your counter example clearer.

himdel:
@jntullo Can you check again please?

I've added the spec for status and message in response, and removed the extra vars.

himdel:

himdel:
.. and I can't fix both miq-bot and not using vars for data while still being readable .. undoing that change.

And not touching this again :D.

spec/requests/tasks_spec.rb
```diff
@@ -0,0 +1,61 @@
...
+  end
+
+  it 'bulk deletes' do
+    api_basic_authorize collection_action_identifier(:tasks, :delete)
+
+    data = {
+      :action    => 'delete',
+      :resources => [
+        {:href => api_task_url(nil, task.id)},
+        {:href => api_task_url(nil, task2.id)}
```
abellotti: Hi @himdel minor, but on all the api_task_url calls, prefer just the object being specified i.e. 

```
api_task_url(nil, task)
```

abellotti:
LGTM!!  minor nit on the rails url helper usage in specs.

himdel:
Thanks @abellotti, updated :)

abellotti:
Thanks @himdel for the update and the API enhancement !!  👍 will merge when 🍏

simaishi:

```
$ git log -1
commit 80834992e9cfb29b48b8cf37febf64d8ecdf2398
Date:   Tue Dec 5 13:17:36 2017 -0500

    Merge pull request #220 from himdel/tasks-delete
    
    Tasks: delete support
    (cherry picked from commit 85f315312fff5ba1fb41bbc71fb833766272d49e)
```

PR #225, opened 2017-11-22
Add subresource_action_identifier spec helper

jntullo:
The current `subcollection_action_identifier` spec helper does not account for subresource actions, which means that the subresource action cannot not be tested if they have a different identifier than the subcollection action. This adds in a new `subresource_action_identifier` which takes into account both subresource actions defined on the type and on the subtype. This will now allow the `orchestration_template` sub collection actions to be identified properly. This also adjusts a snapshots spec to ensure that the helper works for both cases of identifiers (on the type and subtype)

Follow up to https://github.com/ManageIQ/manageiq-api/pull/196

@miq-bot assign @abellotti

### DIFF
--- a/config/api.yml
+++ b/config/api.yml
@@ -1611,7 +1611,7 @@
     :subcollection_actions:
       :get:
       - :name: read
-        :identifier: orchestration_stack_show
+        :identifier: orchestration_stack_show_list
     :subresource_actions:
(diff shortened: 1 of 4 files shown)


abellotti:
Looks good.  Thanks @jntullo for fixing this 👍

Issue #125, opened 2017-10-15
double slash after host `http://127.0.0.1:3000//api` causes error


cben:
```
$ curl -u admin:smartvm 'http://127.0.0.1:3000//api'
{"error":{"kind":"bad_request","message":"Unsupported Collection api specified","klass":"Api::BadRequestError"}}
```
Not sure RFC-compliant to pass an extra slash, but easy to do it concatenating parts to base URL and user gives base url ending with `/`...  (seen in ansible based on python client)

Sounds like when it routes to controller it dons normalize to `/api`, but then when parsing which collection it refers to it does simple "after 2nd slash is collection name" on original url?

api.log:
```
[----] I, [2017-10-15T11:42:47.015562 #12446:2ab8b45f8b64]  INFO -- : MIQ(Api::ApiController.log_request_initiated)  
[----] I, [2017-10-15T11:42:47.015697 #12446:2ab8b45f8b64]  INFO -- : MIQ(Api::ApiController.log_request) API Request:    {:requested_at=>"2017-10-15 08:42:47 UTC", :method=>"GET", :url=>"http://127.0.0.1:3000//api"}
[----] I, [2017-10-15T11:42:47.117727 #12446:2ab8b45f8b64]  INFO -- : MIQ(Api::ApiController.log_request) Authentication: {:type=>"basic", :token=>nil, :x_miq_group=>nil, :user=>"admin"}
[----] I, [2017-10-15T11:42:47.119546 #12446:2ab8b45f8b64]  INFO -- : MIQ(Api::ApiController.log_request) Authorization:  {:user=>"admin", :group=>"EvmGroup-super_administrator", :role=>"EvmRole-super_administrator", :tenant=>"My Company"}
[----] I, [2017-10-15T11:42:47.120078 #12446:2ab8b45f8b64]  INFO -- : MIQ(Api::ApiController.log_request) Request:        {:method=>:get, :action=>"read", :fullpath=>"//api", :url=>"http://127.0.0.1:3000//api", :base=>"http://127.0.0.1:3000", :path=>"//api", :prefix=>"/", :version=>"3.0.0-pre", :api_prefix=>"http://127.0.0.1:3000/", :collection=>"api", :c_suffix=>nil, :collection_id=>nil, :subcollection=>nil, :subcollection_id=>nil}
[...]

PR #227, opened 2017-11-28
Squish consecutive slashes in the path portion of the URI

imtayadeway:
It looks like Rails/Rack already does some squishing:

```ruby
request.original_url # => "http://example.com//api"
request.url # => "http://example.com/api"
```

If there's a reason we're using the former over the latter, it's not
obvious and it's not covered by tests. Switching to the latter is
sufficient to resolve the issue below.

Closes https://github.com/ManageIQ/manageiq-api/issues/125

### DIFF
--- a/lib/api/request_adapter.rb
+++ b/lib/api/request_adapter.rb
@@ -74,7 +74,7 @@ module Api
     end
 
     def url
-      @request.original_url # http://target/api/...
+      @request.url # http://target/api/...
     end
 
     def prefix(version = true)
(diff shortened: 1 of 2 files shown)


imtayadeway:
I think in my description I meant s/squish/squeeze/, but, uh, hopefully it's obvious what was meant 😀

imtayadeway:
Ach, so I forgot that "squish" is a totally different thing from "squeeze", so closing in favor of https://github.com/ManageIQ/manageiq-api/pull/228

PR #228, opened 2017-11-28
Squeeze consecutive slashes in the path portion of the URI

imtayadeway:
It looks like Rails/Rack already does some squeezing:

```ruby
request.original_url # => "http://example.com//api"
request.url # => "http://example.com/api"
```

If there's a reason we're using the former over the latter, it's not
obvious and it's not covered by tests. Switching to the latter is
sufficient to resolve the issue below.

Closes https://github.com/ManageIQ/manageiq-api/issues/125

### DIFF
--- a/lib/api/request_adapter.rb
+++ b/lib/api/request_adapter.rb
@@ -74,7 +74,7 @@ module Api
     end
 
     def url
-      @request.original_url # http://target/api/...
+      @request.url # http://target/api/...
     end
 
     def prefix(version = true)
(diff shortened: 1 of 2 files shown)


abellotti:
@chrisarcand for api auth, we rely on basic auth (HttpAuthentication::basic.authenticate ...) as well as all the other token based variants including external auth via request headers so the url does not come in play.  Thanks.

imtayadeway:
@chrisarcand just rebased and incorporated suggestions. Thanks!

PR #232, opened 2017-11-30
Use type names directly from GenericObjectDefinition object

mzazrivec:
This change is to make sure API uses pretty (i.e. human readable) type names directly from the `GenericObjectDefinition` object, rather than from the `locale/en.yml` file. 

This PR depends on https://github.com/ManageIQ/manageiq/pull/16563

### DIFF
--- a/app/controllers/api/generic_object_definitions_controller.rb
+++ b/app/controllers/api/generic_object_definitions_controller.rb
@@ -104,9 +104,7 @@ module Api
     end
 
     def self.allowed_types
-      GenericObjectDefinition::TYPE_MAP.keys.collect do |type|
-        [Dictionary.gettext(type.to_s, :type => :data_type, :notfound => :titleize, :plural => false), type.to_s]
-      end.sort
+      GenericObjectDefinition::TYPE_NAMES.stringify_keys.invert.to_a.sort
     end
 
     def options


mzazrivec:
The core PR has been merged, this one is ready for review.

app/controllers/api/generic_object_definitions_controller.rb
```diff
@@ -104,8 +104,8 @@ def self.allowed_association_types
     end
 
     def self.allowed_types
-      GenericObjectDefinition::TYPE_MAP.keys.collect do |type|
-        [Dictionary.gettext(type.to_s, :type => :data_type, :notfound => :titleize, :plural => false), type.to_s]
+      GenericObjectDefinition::TYPE_NAMES.each.collect do |key, value|
```
imtayadeway: Does `GenericObjectDefinition::TYPE_NAMES.stringify_keys.invert.to_a.sort` suffice here?

Was this constant extracted for this use only? These transformations seem to suggest that its data structure is a poor fit 😉

mzazrivec: Yes, right now this is the only place where `enericObjectDefinition::TYPE_NAMES` is being used.

What you're suggesting definitely looks better.

abellotti:
LGTM!! Thanks @mzazrivec for the enhancement.

mzazrivec:
@imtayadeway Actually, I'd rather keep the model structures as they are. I think the hash used there is just fine. Yes, in the API there's this data transformation being done, which is not exactly pretty.

But if you look at the UI end consuming this data, the author of the code, for reasons I don't quite comprehend, decided to do the same transformation backwards: invert the fields + back to hash.

So I think the true fix is to avoid this data transformation back and forth, just send the hash over API, consume the same hash on the javascript end and that's the end of it. I'd like to do that in separate PRs though.

mzazrivec:
@imtayadeway https://github.com/ManageIQ/manageiq-api/pull/238

AparnaKarve:
> the author of the code, for reasons I don't quite comprehend, decided to do the same transformation backwards: invert the fields + back to hash.

Could it be that the author of the code initially intended to have an array across the board (API + UI), and later had to change it to a hash in the UI due to some reason?

mzazrivec:
@AparnaKarve Are you telling it as a matter of fact, or are you just guessing?

AparnaKarve:
@mzazrivec I was alluding to the fact that I could have answered whatever questions you had and you didn't have to speculate what the author of the code was trying to do.

In retrospect, and after a second look at the JS code, I think the hash conversion PR https://github.com/ManageIQ/manageiq-api/pull/238 wasn't really necessary - a few adjustments in the JS was what was required.

Hashes/Arrays - they both are fine, so I think we are good.

mzazrivec:
I wasn't really speculating. I simply stated that I don't understand the reasons for doing things this way.

You're welcome to explain your point on https://github.com/ManageIQ/manageiq-api/pull/238 in that PR.

PR #233, opened 2017-12-01
Specify a target and resource when refreshing a dialog field

eclarizio:
Related to https://github.com/ManageIQ/manageiq-ui-classic/pull/2885, this needs to be merged first so that the API then supports the parameters that are being passed in for refresh.

https://bugzilla.redhat.com/show_bug.cgi?id=1518390

### DIFF
--- a/app/controllers/api/service_dialogs_controller.rb
+++ b/app/controllers/api/service_dialogs_controller.rb
@@ -11,12 +11,12 @@ module Api
         service_dialog = resource_search(id, type, klass)
         api_log_info("Refreshing Dialog Fields for #{service_dialog_ident(service_dialog)}")
 
-        refresh_dialog_fields_service_dialog(service_dialog, data)
+        refresh_dialog_fields_service_dialog(data)
       end
     end
 
     def fetch_service_dialogs_content(resource)
-      target, resource_action = validate_dialog_content_params
+      target, resource_action = validate_dialog_content_params(params)
       resource.content(target, resource_action, true)
     end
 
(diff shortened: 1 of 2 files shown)


app/controllers/api/service_dialogs_controller.rb
```diff
@@ -62,26 +63,26 @@ def set_additional_attributes
...
 
-    def refresh_dialog_fields_service_dialog(service_dialog, data)
+    def refresh_dialog_fields_service_dialog(data)
       data ||= {}
       dialog_fields = Hash(data["dialog_fields"])
       refresh_fields = data["fields"]
       return action_result(false, "Must specify fields to refresh") if refresh_fields.blank?
 
-      define_service_dialog_fields(service_dialog, dialog_fields)
+      service_dialog = define_service_dialog(dialog_fields, data)
```
abellotti: with the service_dialog no longer passed in as parameter (which was RBAC checked earlier for the user via resource_search), is it possible we'll be updating a different service_dialog (data driven and not URL) ?

eclarizio: I thought about that too, but the way the classic-ui was doing the refresh before was pulling the dialog off of the resource_action, since it belongs to it. So I guess I could envision a scenario where someone could call, say, `/api/service_dialogs/123` with action set to refresh, and pass in a bogus `resource_action_id` that doesn't correspond to the dialog with the id of 123. We could compare ids, and return an error if they don't match, maybe?

abellotti: maybe pass the service_dialog to refresh_dialog_fields_service_dialog and have it check there against the one defined in data and raise an error accordingly, (with added test for that).  Thanks.

eclarizio:
@abellotti, addressed your concerns, added specs. It definitely would have been possible (although very unlikely) that the resource action id could have pointed to a different dialog, but when trying to update the fields, unless there was a field of the same name to attempt to be updated in the different dialog, the API would have returned an error as well.

This way, though, we should be safe if somehow that does happen.

abellotti:
Thanks @eclarizio just the picky rubocop ^

abellotti:
@eclarizio thanks for the update.

@eclarizio @gmcculloug Is this g/yes ?

eclarizio:

simaishi:

```
$ git log -1
commit bbd52eb2b6a17bafe8761974401460c9fae17e8f
Date:   Mon Dec 4 16:59:42 2017 -0500

    Merge pull request #233 from eclarizio/dialog_refresh_targets
    
    Specify a target and resource when refreshing a dialog field
    (cherry picked from commit d56b2a6214aa1d063ecc3ffb3ad03a8116fcc011)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1520678
```

PR #235, opened 2017-12-05
Remove remaining decorators from the API

jntullo:
With the addition of the `supported_consoles` virtual attribute on the VMs model, there is no longer a need to maintain the last remaining decorators in the API.

@miq-bot add_label cleanup

### DIFF
--- a/app/controllers/api/subcollections/vms.rb
+++ b/app/controllers/api/subcollections/vms.rb
@@ -5,52 +5,20 @@ module Api
         vms = object.try(:vms) || []
 
         vm_attrs = attribute_selection_for("vms")
-        vm_decorators = decorator_selection_for("vms")
-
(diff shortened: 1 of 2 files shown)


abellotti:
@jntullo I'm good with this 👍 , before this gets merged can we get a quick check that SUI/Classic-UI does not use this any longer ? Thanks.

We also allow decorators in the manageiq-api CLI, but that's ok to keep as we may use the CLI to access an earlier server.

AllenBW:
SUI sounding off here

we're **g2g** no decorators are present in master! (or anywhere else, but this is for master)

jntullo:
I don't see any references in the Classic UI either - they used the decorators directly, not via the API.

abellotti:
Thanks @AllenBW for checking for SUI.   did a quick grep in classic ui, no decorators there either.  Thanks.

PR #236, opened 2017-12-06
Don't include hrefs for non-existant collections

chrisarcand:
If a type wasn't a subcollection (but also wasn't a collection...) this method would blindly return a non-existant collection url. Objects which aren't valid collections or subcollections shouldn't be returning hrefs claiming they're available at such urls.

https://bugzilla.redhat.com/show_bug.cgi?id=1518283

### DIFF
--- a/app/controllers/api/base_controller/normalizer.rb
+++ b/app/controllers/api/base_controller/normalizer.rb
@@ -73,11 +73,14 @@ module Api
         svalue.match(pref) ? svalue : "#{pref}/#{svalue}#{suffix}"
       end
 
-      #
-      # Let's normalize an href based on type and id value
-      #
       def normalize_href(type, value)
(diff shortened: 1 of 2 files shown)


abellotti:
While I understand the change in normalize_href, I'm not sure offhand what caused the Travis failures.

chrisarcand:
Ditto. Looking further. Awkwardly, I found this sort of behavior *on master*



...but the result when requesting from the API is correct. Much confusion, digging further.

chrisarcand:
The problem is `@req.collection` doesn't match a type of `:collection/:id/:subcollection`.

I'm having a very hard time seeing what the expected format(s) of `type` are, or the scope of what this method was intended to 'normalize'; it's called for basically everything(?) Any insight you can provide here @abellotti ?

PR #237, opened 2017-12-06
Don't translate model names sent over API

mzazrivec:
This fix is to make sure the model names sent over API will remain in English and won't be translated.

The translation will happen on client side, in this particular case, with this PR: https://github.com/ManageIQ/manageiq-ui-classic/pull/2962

Also, I think this change should land in Gaprindashvili, unless we want to introduce this non-standard behavior in a released project / product version.

### DIFF
--- a/app/controllers/api/generic_object_definitions_controller.rb
+++ b/app/controllers/api/generic_object_definitions_controller.rb
@@ -99,7 +99,7 @@ module Api
 
     def self.allowed_association_types
       GenericObjectDefinition::ALLOWED_ASSOCIATION_TYPES.collect do |association_type|
-        [Dictionary.gettext(association_type, :type => :model, :notfound => :titleize, :plural => false), association_type]
+        [Dictionary.gettext(association_type, :type => :model, :notfound => :titleize, :plural => false, :translate => false), association_type]
       end.sort
     end
 


martinpovolny:
@miq-bot add_label internationalization

martinpovolny:
@abellotti : As far as I know the API responses are not supposed to be internationalized.

So this fixes the undesired situation.

Also I suggest pinging @mzazrivec or at least me in case of I18n related PRs.

app/controllers/api/generic_object_definitions_controller.rb
```diff
@@ -99,7 +99,7 @@ def remove_methods_resource(type, id, data)
 
     def self.allowed_association_types
       GenericObjectDefinition::ALLOWED_ASSOCIATION_TYPES.collect do |association_type|
-        [Dictionary.gettext(association_type, :type => :model, :notfound => :titleize, :plural => false), association_type]
+        [Dictionary.gettext(association_type, :type => :model, :notfound => :titleize, :plural => false, :translate => false), association_type]
```
abellotti: what about the other call to gettext in that file in the self.allowed_typed method ? maybe that too should get a :translate => false ?

abellotti:
Correct, the API responses are not internationalized. API is machine and not user centric, so messages are intended for clients to process.

app/controllers/api/generic_object_definitions_controller.rb
mzazrivec: I don't see a gettext call in `self.allowed_types` method.

Sure, `GenericObjectDefinition::TYPE_NAMES` does contain `N_()` in its definition, but `N_()` is quite harmless actually. It's just an identity function: no real translation is being made, it's just a marker for the gettext string collection process (to put the string inside `N_()` to a catalog). In fact, it's just this use case (delayed translation), where we're supposed to use `N_()` during declaration and then `_()` at the moment of translation.

abellotti: I saw it in the gaprindashvili branch, didn't know if that was an issue.

mzazrivec: Strictly speaking, it is an issue (for the reasons we already discussed) and can be resolved by cherry-picking the following 2 PRs into gaprindashvili: https://github.com/ManageIQ/manageiq/pull/16563 and https://github.com/ManageIQ/manageiq-api/pull/232

abellotti: got it, so keep this PR g/yes then ?

mzazrivec: Well, if we want to resolve this problem for gaprindashvili as well, then yes, we need to cherry-pick all the following PRs into gaprindashvili: https://github.com/ManageIQ/manageiq/pull/16563 https://github.com/ManageIQ/manageiq-api/pull/232 https://github.com/ManageIQ/manageiq-ui-classic/pull/2962 and the current one as well.

mzazrivec: @abellotti So should I label https://github.com/ManageIQ/manageiq/pull/16563 and https://github.com/ManageIQ/manageiq-api/pull/232 as g/yes ?

abellotti: I think so, so gapringashvili on it's consistent.   @martinpovolny thoughts ?

martinpovolny:
I just merged the UI PR. @abellotti : please, merge this one.

simaishi:

```
$ git log -1
commit a3b99fb5436bc3cc2d37c655bdb7df9bd48ee793
Date:   Wed Dec 6 13:31:00 2017 -0500

    Merge pull request #237 from mzazrivec/dont_translate_model_names
    
    Don't translate model names sent over API
    (cherry picked from commit 7d934d1ff4d28abae46dd51b2dd61b2dd3cb3d08)
```

PR #238, opened 2017-12-06
Generic Object Definition OPTIONS: send hashes rather than arrays

mzazrivec:
It's unnecessary to transform the hash obtained directly from model into an array, since on the client side, we'd have to do another transformation anyway.

Let's send the hash as it is.

Gaprindashvili/no

### DIFF
--- a/app/controllers/api/generic_object_definitions_controller.rb
+++ b/app/controllers/api/generic_object_definitions_controller.rb
@@ -98,13 +98,13 @@ module Api
     end
 
     def self.allowed_association_types
-      GenericObjectDefinition::ALLOWED_ASSOCIATION_TYPES.collect do |association_type|
-        [Dictionary.gettext(association_type, :type => :model, :notfound => :titleize, :plural => false, :translate => false), association_type]
-      end.sort
+      GenericObjectDefinition::ALLOWED_ASSOCIATION_TYPES.sort.each_with_object({}) do |association_type, result|
+        result[association_type] = Dictionary.gettext(association_type, :type => :model, :notfound => :titleize, :translate => false)
+      end
     end
 
     def self.allowed_types
-      GenericObjectDefinition::TYPE_NAMES.stringify_keys.invert.to_a.sort
+      GenericObjectDefinition::TYPE_NAMES
     end
 
     def options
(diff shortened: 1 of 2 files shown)


app/controllers/api/generic_object_definitions_controller.rb
```diff
@@ -98,13 +98,13 @@ def remove_methods_resource(type, id, data)
...
-        [Dictionary.gettext(association_type, :type => :model, :notfound => :titleize, :plural => false), association_type]
-      end.sort
+      GenericObjectDefinition::ALLOWED_ASSOCIATION_TYPES.sort.map do |association_type|
+        [association_type, Dictionary.gettext(association_type, :type => :model, :notfound => :titleize, :translate => false)]
+      end.to_h
     end
 
     def self.allowed_types
-      GenericObjectDefinition::TYPE_NAMES.stringify_keys.invert.to_a.sort
+      GenericObjectDefinition::TYPE_NAMES.stringify_keys
```
imtayadeway: On further consideration, I think the `stringify_keys` isn't needed, since Rails will know how to convert these to valid JSON

app/controllers/api/generic_object_definitions_controller.rb
```diff
@@ -98,13 +98,13 @@ def remove_methods_resource(type, id, data)
     end
 
     def self.allowed_association_types
-      GenericObjectDefinition::ALLOWED_ASSOCIATION_TYPES.collect do |association_type|
-        [Dictionary.gettext(association_type, :type => :model, :notfound => :titleize, :plural => false), association_type]
-      end.sort
+      GenericObjectDefinition::ALLOWED_ASSOCIATION_TYPES.sort.map do |association_type|
```
imtayadeway: Does this work?

```ruby
GenericObjectDefinition::ALLOWED_ASSOCIATION_TYPES.sort.each_with_object({}) do |association_type, result|
  result[association_type] = Dictionary.gettext(association_type, :type => :model, :notfound => :titleize, :translate => false)]
end
```

app/controllers/api/generic_object_definitions_controller.rb
mzazrivec: You're right, it's not needed (tested).

app/controllers/api/generic_object_definitions_controller.rb
mzazrivec: Yes, it works too. Changed as suggested.

PR #245, opened 2017-12-12
add image scanning endpoint

enoodle:
Adds https://bugzilla.redhat.com/show_bug.cgi?id=1486362

relevant RFE: Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1486362

This is a resubmission of: https://github.com/ManageIQ/manageiq-api/pull/52


enoodle:
@imtayadeway @jntullo Can you please take a look?

spec/requests/container_images_spec.rb
```diff
@@ -45,4 +45,41 @@
...
+
+      post(invalid_image_url, :params => { :action => "scan" })
+
+      expect(response).to have_http_status(:not_found)
+    end
+
+    it "doesn't scan a Container Image without appropriate role" do
+      api_basic_authorize
+
+      post(invalid_image_url, :params => { :action => "scan" })
```
imtayadeway: You might want a valid URL to ensure you're only changing one thing here, relative to the happy path?

app/controllers/api/subcollections/container_images.rb
```diff
@@ -0,0 +1,26 @@
...
+    module ContainerImages
+      def container_images_scan_resource(provider_id, type, image_id, _payload)
+        api_action(type, image_id, :skip_href => true) do |klass|
+          image = resource_search(image_id, type, klass)
+          begin
+            desc = "#{container_image_ident(image)} scanning"
+            task_id = image.scan
+            action_result(true, desc).merge(:task_id => task_id, :parent_id => provider_id)
+          rescue => err
+            action_result(false, err.to_s)
```
imtayadeway: Do you know what conditions might result in an error? If so, can you add a test for that case?

enoodle: I am not sure, this is basically creating a job and the rescue is here for unforeseen errors. I think we can use `allow(image).to receive(:scan) ....` magic to mimic this. wdy?

spec/requests/container_images_spec.rb
```diff
@@ -1,3 +1,16 @@
+class FailedScanContainerImage < ContainerImage
```
enoodle: @moolitayer cc

enoodle:
@imtayadeway updated, ptal

enoodle:
I am not sure what to do with codeclimate's error. It is suddenly complaining about `rescue` without a specific exception class, but this is exactly the goal here.

spec/requests/container_images_spec.rb
imtayadeway: IMO it's not a good idea to stub like this in an integration test. You're testing what happens if a container image raises `Exception` during a scan - which I would hope it would never do. It would be far better to create some state that results in an exception in the actual implementation of `ContainerImage`. If none can be created, it's possible you don't need the exception handling code above.

app/controllers/api/subcollections/container_images.rb
```diff
@@ -0,0 +1,26 @@
...
+  module Subcollections
+    module ContainerImages
+      def container_images_scan_resource(provider_id, type, image_id, _payload)
+        api_action(type, image_id, :skip_href => true) do |klass|
+          image = resource_search(image_id, type, klass)
+          begin
+            desc = "#{container_image_ident(image)} scanning"
+            task_id = image.scan
+            action_result(true, desc).merge(:task_id => task_id, :parent_id => provider_id)
+          rescue Exception => err
```
imtayadeway: Use `rescue => e` here, or `rescue SpecificError => e` if there's a specific error you're anticipating (see above for why you might not need to add exception handling here). `Exception` is not a specific error, and attempting to rescue it can cause some unexpected/unpleasant behavior.

spec/requests/container_images_spec.rb
enoodle: @imtayadeway the `scan` method raises and event to create a Job for the scan. It is not suppose to fail but the rescue is there for completeness. I don't the conditions that will make this fail but they are probably from outside of the `ContainerImage` class. I don't feel strongly either way, do you think the rescue above should be removed?

app/controllers/api/subcollections/container_images.rb
enoodle: Codeclimate didn't like `rescue => e` either, So if it stays here it will have to be against its suggestion.

spec/requests/container_images_spec.rb
imtayadeway: > rescue is there for completeness

It's not complete because you haven't demonstrated why it is needed. I don't feel strongly that you should remove it, but I do feel strongly that you should remove the stubbing here. You are stubbing the SUT here (since it is an integration test), which only serves to demonstrate that your code handles the stub correctly.

app/controllers/api/subcollections/container_images.rb
imtayadeway: I'm OK with that. Handling a specific error is the best practice that Codeclimate is trying to enforce. Rescuing `Exception` is far from a best practice and should be avoided at all cost.

> I don't the conditions that will make this fail but they are probably from outside of the ContainerImage class

If you don't know why you're adding this code, or what the conditions are that will make this fail, or what specific errors you might be anticipating, how can you convince me to merge it? We need answers to these questions (even if you can't encode it in an automated test) as a prerequisite to merging.

spec/requests/container_images_spec.rb
enoodle: This is a little similar to [1], This will create an automate event through a series for function calls that will create a scanning Job. I have noticed that no specific exception type is mentioned there and it is not mentioned in any test.

[1] https://github.com/ManageIQ/manageiq-api/blob/master/app/controllers/api/vms_controller.rb#L410

enoodle:
I have updated this with the case that the task failed to be created. I found that lacking the correct MiqEventDefinition will cause this and created a test for that, and updated the happy flow test to correctly verify the response.

app/controllers/api/subcollections/container_images.rb
enoodle: This is a little similar to [1], But this will go through automate and policy through a series for function calls that will eventually create a scanning Job through `MiqQueue.put` which is the same function that `queue_object_action` eventually calls. In the interest of continuity I suggest we catch errors that same way here, but we can ignore this possibility as well as the code on the `image.scan` function path will only raise an exception if `User.current_user` is not set, which is set for api calls.

[1] https://github.com/ManageIQ/manageiq-api/blob/master/app/controllers/api/vms_controller.rb#L410

app/controllers/api/subcollections/container_images.rb
```diff
@@ -0,0 +1,31 @@
...
+      def container_images_scan_resource(provider_id, type, image_id, _payload)
+        api_action(type, image_id, :skip_href => true) do |klass|
+          image = resource_search(image_id, type, klass)
+          begin
+            task_id = image.scan
+            if task_id.present?
+              desc = "#{container_image_ident(image)} scanning"
+              action_result(true, desc).merge(:task_id => task_id, :parent_id => provider_id)
+            else
+              desc = "#{container_image_ident(image)} failed to start scanning"
```
imtayadeway: Is there any way that we can bubble up to the user what the problem was?

spec/requests/container_images_spec.rb
```diff
@@ -45,4 +45,55 @@
       expect(response.parsed_body).to include(expected)
     end
   end
+
+  context 'POST /api/container_images/scan' do
+    let(:provider) { FactoryGirl.create(:ems_kubernetes) }
+    let(:container_image) { FactoryGirl.create(:container_image, :ext_management_system => provider) }
+    let(:conatiner_id) { container_image.id }
```
abellotti: remove, unused.

spec/requests/container_images_spec.rb
```diff
@@ -45,4 +45,55 @@
...
+    let(:invalid_container_image) { container_image.id + 100 }
+    let(:invalid_image_url) { api_provider_container_image_url(nil, provider, invalid_container_image) }
```
abellotti: since this is only used here, maybe just use a bogus number instead, like 999,999

config/api.yml
```diff
@@ -740,6 +740,7 @@
     :options:
     - :collection
     - :custom_actions
+    - :subcollection
```
abellotti: With this exposed as a subcollection, aren't these now fetchable ?  i.e. GET /api/container_nodes/:id/container_images ?  If so we'd need specs to test those happy (and unhappy / i.e. invalid role) paths.

enoodle: This means it will be available for any model that `has_many :container_images`  ? If so it will also be there for Container Groups/Projects. If so I will make tests for these cases as well, but seems like there should be a generic way to test it.

app/controllers/api/subcollections/container_images.rb
enoodle: This `image.scan` method only returns a task or nil. In the specific case of not defined MiqEventDefinition it will write to the evm log a warning but I don't see how to catch it here. I am hoping that if there will be a problem it will raise an exception.

config/api.yml
enoodle: I get 404 for all these cases so I guess it is not possible to access them this way.

enoodle:
@abellotti @imtayadeway PTAL

config/api.yml
abellotti: Could you add tests for GET /api/providers/:id/container_images and GET /api/providers/:id/container_images/:container_image_id ?

config/api.yml
```diff
@@ -750,6 +751,10 @@
       :get:
       - :name: read
         :identifier: container_image_show
+    :subresource_actions:
+      :post:
+      - :name: scan
+        :identifier: container_image_scan
```
abellotti: shouldn't we have the section:

```
:get:
- :name: read
  :identifier: container_image_show
```

in the subresource_actions section to allow one to do GET /api/providers/:id/container_images/:ci_id ?

enoodle:
@abellotti I agree with you, those tests seem to be missing. I think we should move this API to be under the `resource_actions` for container images instead of being a sub collection. We are not using the host provider anyway and it is not needed because the image can belong to only one provider anyway (currently I don't see us sharing entities between providers soon)

abellotti:
@enoodle can you take care of the rubocop issues ?  Thanks!!

@imtayadeway can you take a  👀  again ? were the changes made as you requested ok ?  Thanks.

app/controllers/api/subcollections/container_images.rb
```diff
@@ -0,0 +1,6 @@
+# frozen_string_literal: true
```
imtayadeway: Is this still needed?

config/api.yml
```diff
@@ -1789,6 +1792,7 @@
     - :vms
     - :flavors
     - :cloud_templates
+    - :container_images
```
imtayadeway: We need some tests exercising the behavior here - what can/can't you do with container images of providers?

enoodle: I am sorry, this should have been moved.

app/controllers/api/subcollections/container_images.rb
enoodle: This file is not relevant now that I changed this to be a ContainerImage resource_action and not a subcollection_action

abellotti:
LGTM!! Thanks 👍

simaishi:
@enoodle Can this be `gaprindashvili/yes` ?

enoodle:
@simaishi Yes, but this PR has to go with #307 and both should be back ported together to gaprindashvili.

simaishi:
@enoodle @abellotti Looks this is built on top of https://github.com/ManageIQ/manageiq-api/pull/185, which is `gaprindashvili/no`. Not sure if PR 185 should/can be changed to `gaprindashvili/yes`.. if not, please create a separate PR for gaprindashvili.

abellotti:
@simaishi container_images was not intended for gaprindashvili.  if this requirement has now changed, this may require additional changes in the model (Custom Button support and such) /cc  @gmcculloug

simaishi:
cc @Loicavenel

simaishi:
@Loicavenel @gtanzillo Please clarify if this needs to go to gaprindashvili or not.

gtanzillo:
@simaishi I will defer to @loicavenel to let us know whether he need it backported or not.

nimrodshn:
@Loicavenel should this be backported?

simaishi:
As per BZ, this should go to 'master' only.

Loicavenel:
@abellotti I think this one was never documented, do you have an exemple how to use it?
THANKS

PR #217, opened 2017-11-20
Use destroy_queue for provider delete

jntullo:
Changes to ExtManagementSystem now require the use of schedule_destroy_queue to queue the deletion of the provider. Without this change, provider deletion is silently failing due to a before_hook on destroy that checks if there are any active queues for the provider. The orchestrate_delete method that will now be called will resolve this.

https://bugzilla.redhat.com/show_bug.cgi?id=1501941

@miq-bot add_label bug, gaprindashvili/yes, blocker 

cc: @gtanzillo @zeari

PR #248, opened 2017-12-13
Do not return task ID on provider delete

jntullo:
Destroying a provider uses destroy_queue, however, that method does not create a task, but rather returns an [MiqQueue object](https://github.com/ManageIQ/manageiq/blob/master/app/models/ext_management_system.rb#L452). We are currently returning a task href with non-existent task id. This PR updates the returned task message and removes the task href reference.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1525498

More context into the way providers are deleted can be found [here](https://github.com/ManageIQ/manageiq/pull/14848) and [here](https://github.com/ManageIQ/manageiq-api/pull/217). I am not certain of the precedent of this as I do not see anything else like it in the API, or if we should be returning other information to the user. However, for the time being, I think the best course of action is to remove the improper task reference. 

@miq-bot add_label gaprindashvili/yes, bug, blocker 
@miq-bot assign @abellotti

### DIFF
--- a/app/controllers/api/providers_controller.rb
+++ b/app/controllers/api/providers_controller.rb
@@ -54,8 +54,8 @@ module Api
       delete_action_handler do
         raise BadRequestError, "Must specify an id for deleting a #{type} resource" unless id
         provider = resource_search(id, type, collection_class(type))
-        task = provider.destroy_queue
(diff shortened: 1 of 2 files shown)


jntullo:
Closing in favor of having the queue method return an MiqTask.

PR #252, opened 2017-12-18
Fix custom actions hrefs

imtayadeway:
Because custom actions are represented as a plain hash in the model,
we lose all context about the type of object we must add an href to in
the normalization process. This process tries, in lieu of this
context, to guess what the href path should look like, and is
incorrectly linking these objects to the parent object's
API (e.g. generic objects, services, etc..). Refactoring all of
normalization to get the proper context will be necessary in order to
fix this correctly. Instead, we intercept the result here, adding the
correct hrefs, which will not be overwritten later.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1518297

### DIFF
--- a/app/controllers/api/base_controller/renderer.rb
+++ b/app/controllers/api/base_controller/renderer.rb
@@ -266,9 +266,33 @@ module Api
         return unless attr_accessible?(resource, attr)
         virtattr_accessor = virtual_attribute_accessor(type, attr)
         value = virtattr_accessor ? send(virtattr_accessor, resource) : virtual_attribute_search(resource, attr)
+        value = add_custom_action_hrefs(value) if attr == "custom_actions"
         result = {attr => normalize_attr(attr, value)}
         # set nil vtype above to "#{type}/#{resource.id}/#{attr}" to support id normalization
         [value, result]
+      end
+
+      #
+      # HACK: Because custom actions are represented as a plain hash
+      # in the model, we lose all context about the type of object we
+      # must add an href to in the normalization process. Refactoring
+      # all of normalization to get the proper context will be
+      # necessary in order to fix this correctly. Instead, we
+      # intercept the result here, adding the correct hrefs, which
+      # will not be overwritten later.
+      #
+      def add_custom_action_hrefs(value)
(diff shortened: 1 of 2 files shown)


app/controllers/api/base_controller/renderer.rb
```diff
@@ -266,11 +266,37 @@ def fetch_direct_virtual_attribute(type, resource, attr)
...
         virtattr_accessor = virtual_attribute_accessor(type, attr)
         value = virtattr_accessor ? send(virtattr_accessor, resource) : virtual_attribute_search(resource, attr)
+        value = add_custom_action_hrefs(value) if attr == "custom_actions"
         result = {attr => normalize_attr(attr, value)}
         # set nil vtype above to "#{type}/#{resource.id}/#{attr}" to support id normalization
         [value, result]
       end
 
+      #
+      # WARNING: This is a hack.
```
Fryguy: Prefer `HACK: ` as a lead-in to the comment.  `HACK: `  (same with `TODO: ` and `FIXME: `) are recognized by a lot of IDEs and tools to allow you to keep a checklist of things to go back to.

jrafanie: I love that `# HACK:` is the official way to do it.

abellotti:
@imtayadeway looks good for a HACK 😉 

Can you fix the spec rubocop above since that's most likely staying around.  And since you're updating, can you update the comment to the "official" HACK: prefix ?  Thanks.

abellotti:
Thanks @imtayadeway for fixing this 🎵

simaishi:

```
$ git log -1
commit c07b0ee4c5cc6e1c7c01dbfe878e1e283e608f93
Date:   Mon Dec 18 17:34:18 2017 -0500

    Merge pull request #252 from imtayadeway/bug/custom-button-hrefs
    
    Fix custom actions hrefs
    (cherry picked from commit 8a3e3cd64225dc585e6d39dd02b7c093542895ec)
    
    Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1527564
```

PR #253, opened 2017-12-19
Updating the API so it now returns nil attributes.

abellotti:
- Return nil attributes
- Do not return encrypted attributes
- Added spec

### DIFF
--- a/app/controllers/api/base_controller/normalizer.rb
+++ b/app/controllers/api/base_controller/normalizer.rb
@@ -21,9 +21,11 @@ module Api
           end
         end
 
+        is_ar = obj.kind_of?(ActiveRecord::Base)
         attrs.each do |k|
-          value = normalize_attr(k, obj.kind_of?(ActiveRecord::Base) ? obj.try(k) : obj[k])
-          result[k] = value unless value.nil?
+          next if Api.encrypted_attribute?(k)
+          next if is_ar ? !obj.respond_to?(k) : !obj.key?(k)
+          result[k] = normalize_attr(k, is_ar ? obj.try(k) : obj[k])
         end
         result
       end
--- a/app/controllers/api/base_controller/renderer.rb
+++ b/app/controllers/api/base_controller/renderer.rb
@@ -32,7 +32,6 @@ module Api
       def collection_to_jbuilder(type, reftype, resources, opts = {})
         link_builder = Api::LinksBuilder.new(params, @req.url, opts[:counts])
         Jbuilder.new do |json|
-          json.ignore_nil!
           json.set! 'name', opts[:name] if opts[:name]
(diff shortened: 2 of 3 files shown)


abellotti:
Marking as WIP:
 - [ ] additional sanity tests with encrypted attributes 
 - [ ] performance comparison with current API (100 VMs)
 - [ ] payload size comparison with current API (100 VMs)

abellotti:
Performance delta seems negligible, the noticeable difference is the returned payload size. An increase of %24 which is understandable. Not a concern for local access but over the wire might be a factor. Though, clients should probably be asking for specific fields when doing bulk fetches anyway.

100 vms, expanded resources, first query (uncached):

Without nil | With nil |
-----------|---------|
432576 bytes | 535959 bytes
real	0m8.466s | real	0m8.664s
user	0m0.447s | user	0m0.448s
sys	0m0.178s | sys	0m0.180s

100 vms, expanded resources, secondary queries (cached):

Without nil | With nil |
-----------|---------|
432576 bytes | 535959 bytes
real	0m1.360s | real	0m1.397s
user	0m0.447s | user	0m0.444s
sys	0m0.129s | sys	0m0.129s

abellotti:
ok, this looks good to me.  @AllenBW if you don't mind could you give it a quick sanity check with the SUI 🙏  before I un-wip this ?

AllenBW:
@abellotti i wanta drop in on this one... just to be sure... un momento...

abellotti:
Ping @AllenBW

AllenBW:
OH GOSH SORRY THIS GOT FORGOTTEN !! 😱 

on it now

abellotti:
Thanks @AllenBW for checking it out 🙇

cben:
FYI, this exposed a bug in the python client — code assumed present fields are non-null.
This broke some QE workflows, and the ansible modules (at least manageiq_provider).
Client fix coming https://github.com/ManageIQ/manageiq-api-client-python/pull/39.

While conceptually adding nil fields feels backwards-compatible, this demonstrates it *could* break other code.
Is there a way to make this change opt-in for now?  Do we have some api versioning mechanism?  (may well be overkill for this change, just asking...)

(Also to be clear, I like this change and don't want it reverted!  Omitted nils have tripped me multiple times, in both manual and programatic API usage...)

abellotti:
As far as opting in not offhand. while we could have exposed the earlier signature via versioning i.e. /api/v2.4.0, but like you said exposing nil is backward compatible. But that would also mean you wouldn't get the v3.0.0 capabilities. In any case, the python client would have had to change to request the version or whatever opt-in parameter is implemented, so might as well fix to properly support the nils. my 2-cents.

PR #275, opened 2018-01-10
Add Settings API for servers, zones, and regions

chrisarcand:
This adds manipulation of all settings on servers, zones, and regions via the API in core's `ConfigurationManagementMixin` in the form of a new base controller method, `settings`, which routes to these three collections using the `settings` options keyword in the YAML.

* `GET /servers/:id/settings`
  Renders the server settings as provided by the settings API.

* `PATCH /zones/:id/settings`, body: `{"ems": { "ems_ansible_tower": "blacklisted_event_names": ["EventName1"] } }`
  Sets the ems.ems_ansible_tower.blacklisted_event_names to ["EventName1"] This action serves as both create and update.

* `DELETE /zones/:id/settings`, body: `["ems", "ems_ansible_tower"]`
  This deletes the settings change for this zone under this key path. All changes underneath this key path would be reset and the settings for this zone's ems.ems_ansible_tower subtree would revert to whatever is higher up the settings tree for this zone.

The interface here is very simplistic and just follows the API provided by the core mixin. It emulates the existing `/settings` endpoint.

Additionally,

* Only super admins can read, edit, and delete all settings
* All other users are read only and settings are scoped to only those set as categories in the settings collection.

https://bugzilla.redhat.com/show_bug.cgi?id=1486634

### DIFF
--- a/spec/requests/regions_spec.rb
+++ b/spec/requests/regions_spec.rb
@@ -40,4 +40,112 @@ describe "Regions API" do
       "id"   => region.id.to_s
     )
   end
+
+  describe "/api/regions/:id/settings" do
+    # let(:server) { FactoryGirl.create(:miq_server) }
+    # let(:region) { FactoryGirl.create(:miq_region, :region => FactoryGirl.create(:miq_region, :region => ApplicationRecord.my_region_number)) }
+    let(:region) { @region }
+
+    let(:original_timeout) { region.settings_for_resource[:api][:authentication_timeout] }
+    let(:super_admin) { FactoryGirl.create(:user, :role => 'super_administrator', :userid => 'alice', :password => 'alicepassword') }
+
+    # before do
+    #   allow(MiqServer).to receive(:in_region).with(ApplicationRecord.my_region_number).and_return(MiqServer.where(:id => server.id))
+    # end
+
+    it "shows the settings to an authenticated user with the proper role" do
+      api_basic_authorize(:ops_settings)
+
+      get(api_region_settings_url(nil, region))
+
+      expect(response).to have_http_status(:ok)
+    end
+
+    it "does not allow an authenticated user who doesn't have the proper role to view the settings" do
+      api_basic_authorize
+
+      get(api_region_settings_url(nil, region))
+
+      expect(response).to have_http_status(:forbidden)
+    end
+
+    it "does not allow an unauthenticated user to view the settings" do
+      get(api_region_settings_url(nil, region))
+
+      expect(response).to have_http_status(:unauthorized)
+    end
+
+    it "permits updates to settings for an authenticated super-admin user" do
+      api_basic_authorize(:user => super_admin.userid, :password => super_admin.password)
+
+      expect {
+        patch(api_region_settings_url(nil, region), :params => {:api => {:authentication_timeout => "1337.minutes"}})
+      }.to change { region.settings_for_resource[:api][:authentication_timeout] }.from(original_timeout).to("1337.minutes")
+
+      expect(response.parsed_body).to include("api" => a_hash_including("authentication_timeout" => "1337.minutes"))
+      expect(response).to have_http_status(:ok)
+    end
+
+    it "does not allow an authenticated non-super-admin user to update settings" do
+      api_basic_authorize
+
+      expect {
+        patch(api_region_settings_url(nil, region), :params => {:api => {:authentication_timeout => "10.minutes"}})
+      }.not_to change { region.settings_for_resource[:api][:authentication_timeout] }
+
+      expect(response).to have_http_status(:forbidden)
+    end
+
+    it "does not allow an unauthenticated user to update the settings" do
+      expect {
+        patch(api_region_settings_url(nil, region), :params => {:api => {:authentication_timeout => "10.minutes"}})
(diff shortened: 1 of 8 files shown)


chrisarcand:
cc/ @gtanzillo

chrisarcand:
Ready for review; the remaining rubocop stuff is exempt in RSpec DSL, IMO.

abellotti:
Thanks @chrisarcand for the quick turnaround with this PR 👍 

just a couple of notes:

- You probably meant DELETE instead of DESTROY in the description 😉 
- Non super-admin users should minimally have the ops_setting role to be able to read settings
- GETs of /api/settings should mimic GET /api/servers/:myserver_id/settings, so for super-admin, return all settings there and not just the white-listed ones.

Other things:

- would have been nice to also support
  - POST/PATCH/DELETE /api/settings/path/
  - POST/PATCH/DELETE /api/[zones|regions|servers]/:id/settings/path/

  but that can be a follow-up PR.

chrisarcand:
> You probably meant DELETE instead of DESTROY in the description

Ah ha, yup. 😄 

> Non super-admin users should minimally have the ops_setting role to be able to read settings

Ok! Will add

> GETs of /api/settings should mimic GET /api/servers/:myserver_id/settings, so for super-admin, return all settings there and not just the white-listed ones.

Ah yeah, I noticed that as well. However, the API that was built for that endpoint for some reason doesn't use the settings API and instead digs directly in to the database and parses records itself (and ignores the yaml layer), so I wasn't going to open that can of worms here. I guess I can correct it here though, that's cool 👍 

> would have been nice to also support
>    POST/PATCH/DELETE /api/settings/path/
>    POST/PATCH/DELETE /api/[zones|regions|servers]/:id/settings/path/

* This route is not RESTful, intentionally. The difference between POST and PATCH doesn't mean anything because from the view of the consumer, you're always *editing* a resource's settings (that the actual action you're taking is entering a record in the database for a particular server to override the environment settings is an implementation detail). In the case of adding a new settings that wasn't an override (like anything. { "my_new_setting" => true }), it's not worth being pedantic to the consumer by saying 'When it's an existing entry in the hash you should use PATCH but when it's a completely new one that isn't defaulted you should use POST'. So no POSTs, is my point here.

* I originally had /path stuff so you could do arbitrary subtrees like `/api/settings/ems/ems_ansible_tower/` but decided that it doesn't actually add that much value for the consumer and then handcuffs us as *no* new paths could be added to this endpoint since a user can arbitrarily add any settings. I guess I'm not completely against it but I just didn't see the value. 

Perhaps we should actually ask consumers of the API if they think that would be helpful, as someone like QE for example might just want all the settings back anyway. 


Addressing actionable feedback now! ❤️

chrisarcand:
@abellotti Ready for 👀 again

abellotti:
Thanks @chrisarcand for the updates. Lookin' good 👍 

As chatted in Gitter, we'd need a follow-up in Settings API land to not allow disabling web_services in servers role if this was the last server.

chrisarcand:
:shipit:

spec/requests/regions_spec.rb
```diff
@@ -40,4 +40,112 @@
...
+    # let(:server) { FactoryGirl.create(:miq_server) }
+    # let(:region) { FactoryGirl.create(:miq_region, :region => FactoryGirl.create(:miq_region, :region => ApplicationRecord.my_region_number)) }
+    let(:region) { @region }
+
+    let(:original_timeout) { region.settings_for_resource[:api][:authentication_timeout] }
+    let(:super_admin) { FactoryGirl.create(:user, :role => 'super_administrator', :userid => 'alice', :password => 'alicepassword') }
+
+    # before do
+    #   allow(MiqServer).to receive(:in_region).with(ApplicationRecord.my_region_number).and_return(MiqServer.where(:id => server.id))
+    # end
```
abellotti: @imtayadeway could you remove the commented code above ( & lines 45-46) and repush ?  unless you needed this for reference 😉

chrisarcand:
@abellotti I think your comment above is on old changes; hit refresh 😄

abellotti:
LGTM!! Thanks @chrisarcand and @imtayadeway for this greatly needed enhancement !!! 😍

Fryguy:
Really nice work @chrisarcand !!

simaishi:

```
$ git log -1
commit acb952b9e4bcedde51b085bcf2edb1d4f588376b
Date:   Mon Jan 15 17:08:07 2018 -0500

    Merge pull request #275 from chrisarcand/settings-crud
    
    Add Settings API for servers, zones, and regions
    (cherry picked from commit 3f407638b474b1ff8b2945e9a13666d1fb061201)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1535602
```

simaishi:
Fine backport (to manageiq repo) details:

```
$ git log -1
commit e4138807e933225b3e414aab90eb8b2594c082b9
Date:   Mon Jan 15 17:08:07 2018 -0500

    Merge pull request #275 from chrisarcand/settings-crud
    
    Add Settings API for servers, zones, and regions
    (cherry picked from commit 3f407638b474b1ff8b2945e9a13666d1fb061201)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1535605
```

PR #278, opened 2018-01-10
Adding support to apply_config_pattern operation via the REST API

igorrran:
This PR is able to:
* Add a new operation to apply a config pattern to a physical server via REST API.

Depends on: https://github.com/ManageIQ/manageiq/pull/16796
Depends on: https://github.com/ManageIQ/manageiq-api/pull/300

### DIFF
--- a/app/controllers/api/physical_servers_controller.rb
+++ b/app/controllers/api/physical_servers_controller.rb
@@ -59,18 +59,11 @@ module Api
     private
 
     def change_resource_state(state, type, id, data = {})
-      raise BadRequestError, "Must specify an id for changing a #{type} resource" unless id
-      is_apply_config_action = state == :apply_config_pattern
-
-      if single_resource?
-        ensure_resource_exists(type, id)
-        ensure_customization_script_exists(data["pattern_id"]) if is_apply_config_action
-      end
-
+      validate_resource(state, type, id, data)
       api_action(type, id) do |klass|
         begin
           server = resource_search(id, type, klass)
-          ensure_customization_script_exists(data["pattern_id"]) if is_apply_config_action
+          ensure_customization_script_exists(data["pattern_id"]) if apply_config_action(state)
           desc = "Requested server state #{state} for #{server_ident(server)}"
           api_log_info(desc)
           task_id = queue_object_action(server, desc, :method_name => state, :role => :ems_operations)
@@ -79,6 +72,19 @@ module Api
           action_result(false, err.to_s)
         end
       end
+    end
+
+    def validate_resource(state, type, id, data = {})
+      raise BadRequestError, "Must specify an id for changing a #{type} resource" unless id
+      if single_resource?
+        ensure_resource_exists(type, id)
+        # Verify the configuration pattern resource if the operation is apply_config_pattern
+        ensure_customization_script_exists(data["pattern_id"]) if apply_config_action(state)
+      end
+    end
+
+    def apply_config_action(state)
+      state == :apply_config_pattern
     end
 
     def ensure_customization_script_exists(id)
--- a/app/controllers/api/physical_servers_controller.rb
+++ b/app/controllers/api/physical_servers_controller.rb
@@ -52,24 +52,39 @@ module Api
       end
     end
 
+    def apply_config_pattern_resource(type, id, data)
+      change_resource_state(:apply_config_pattern, type, id, data)
+    end
+
     private
 
-    def change_resource_state(state, type, id)
+    def change_resource_state(state, type, id, data = {})
       raise BadRequestError, "Must specify an id for changing a #{type} resource" unless id
+      is_apply_config_action = state == :apply_config_pattern
 
-      ensure_resource_exists(type, id) if single_resource?
+      if single_resource?
+        ensure_resource_exists(type, id)
+        ensure_customization_script_exists(data["pattern_id"]) if is_apply_config_action
+      end
 
       api_action(type, id) do |klass|
         begin
           server = resource_search(id, type, klass)
+          ensure_customization_script_exists(data["pattern_id"]) if is_apply_config_action
           desc = "Requested server state #{state} for #{server_ident(server)}"
           api_log_info(desc)
           task_id = queue_object_action(server, desc, :method_name => state, :role => :ems_operations)
           action_result(true, desc, :task_id => task_id)
-        rescue => err
+        rescue StandardError => err
           action_result(false, err.to_s)
         end
       end
+    end
+
+    def ensure_customization_script_exists(id)
+      ensure_resource_exists(:customization_scripts, id)
+    rescue NotFoundError
+      raise NotFoundError, "Customization script not found"
     end
 
     def ensure_resource_exists(type, id)
--- a/app/controllers/api/physical_servers_controller.rb
+++ b/app/controllers/api/physical_servers_controller.rb
@@ -52,49 +52,39 @@ module Api
       end
     end
 
+    # Process apply config pattern operation for single and multi-resources
+    #
+    # Even the request for multi-resources isn't completed successfully, return a HTTP Status 200
     def apply_config_pattern_resource(type, id, data)
-      change_resource_state(:apply_config_pattern, type, id, data)
+      ensure_resource_exists(:customization_scripts, data["pattern_id"])
+      change_resource_state(:apply_config_pattern, type, id, [data["pattern_id"]])
+    rescue => err
+      raise err if single_resource?
+      action_result(false, err.to_s)
     end
 
     private
 
-    def change_resource_state(state, type, id, data = {})
-      validate_resource(state, type, id, data)
+    def change_resource_state(state, type, id, data = [])
+      raise BadRequestError, "Must specify an id for changing a #{type} resource" unless id
+
+      ensure_resource_exists(type, id) if single_resource?
+
       api_action(type, id) do |klass|
         begin
           server = resource_search(id, type, klass)
-          ensure_customization_script_exists(data["pattern_id"]) if apply_config_action(state)
           desc = "Requested server state #{state} for #{server_ident(server)}"
           api_log_info(desc)
-          task_id = queue_object_action(server, desc, :method_name => state, :role => :ems_operations)
+          task_id = queue_object_action(server, desc, :method_name => state, :role => :ems_operations, :args => data)
           action_result(true, desc, :task_id => task_id)
-        rescue StandardError => err
+        rescue => err
           action_result(false, err.to_s)
         end
       end
     end
 
-    def validate_resource(state, type, id, data = {})
-      raise BadRequestError, "Must specify an id for changing a #{type} resource" unless id
-      if single_resource?
-        ensure_resource_exists(type, id)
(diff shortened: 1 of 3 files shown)


config/api.yml
```diff
@@ -922,6 +922,22 @@
...
+    :klass: CustomizationScript
+    :subcollections:
+    :collection_actions:
+      :get:
+      - :name: read
+        :identifier: customization_script_show_list
+    :resource_actions:
+      :get:
+      - :name: read
+        :identifier: physical_server_show
```
abellotti: shouldn't this be customization_script_show ?

app/controllers/api/physical_servers_controller.rb
```diff
@@ -41,37 +41,77 @@ def restart_mgmt_controller_resource(type, id, _data)
...
+      api_action(type, id) do |klass|
+        begin
+          physical_server = resource_search(id, type, klass)
+          desc = "#{physical_server_ident(physical_server)} refreshing"
+          do_action(state, desc, physical_server)
+        rescue StandardError => err
+          action_result(false, err.to_s)
+        end
+      end
+    end
```
abellotti: got confused reviewing, the refresh change seems to be a refactor, should have been a separate PR or at least commit.

app/controllers/api/physical_servers_controller.rb
```diff
@@ -41,37 +41,77 @@ def restart_mgmt_controller_resource(type, id, _data)
     end
 
     def refresh_resource(type, id, _data = nil)
-      raise BadRequestError, "Must specify an id for refreshing a #{type} resource" unless id
+      refresh_physical_server("refresh_ems", type, id)
```
jntullo: I don't believe I see `refresh_physical_server` anywhere else? I am not sure I see the need for a separate method. I also agree that perhaps this could be a separate PR/commit to make it more clear

app/controllers/api/physical_servers_controller.rb
```diff
@@ -41,37 +41,77 @@ def restart_mgmt_controller_resource(type, id, _data)
...
+    end
 
-      ensure_resource_exists(type, id) if single_resource?
+    def apply_config_pattern_resource(type, id, data)
+      apply_config(:apply_config_pattern, type, id, data)
```
jntullo: Is this method going to be used in a similar way to `change_resource_state` where it will be reused with multiple states?

app/controllers/api/physical_servers_controller.rb
igorrran: Yes. The refresh change is a refactor. I'll separate this.

config/api.yml
```diff
@@ -922,6 +922,22 @@
       :delete:
       - :name: delete
         :identifier: ab_group_delete
+  :customization_scripts:
+    :description: Customization Script
+    :identifier: customization_script
```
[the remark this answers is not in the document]
igorrran: Yes, make senses

config/api.yml
igorrran: Yes, It should be. I'll fix this, thanks

igorrran:
@abellotti @jntullo Could you take a look in this now that the Pull Request https://github.com/ManageIQ/manageiq-api/pull/300 was merged?

app/controllers/api/physical_servers_controller.rb
```diff
@@ -52,26 +52,47 @@ def refresh_resource(type, id, _data = nil)
...
         end
       end
     end
 
+    def validate_resource(state, type, id, data = {})
+      raise BadRequestError, "Must specify an id for changing a #{type} resource" unless id
+      if single_resource?
+        ensure_resource_exists(type, id)
+        # Verify the configuration pattern resource if the operation is apply_config_pattern
+        ensure_customization_script_exists(data["pattern_id"]) if apply_config_action(state)
```
jntullo: I see that this is being called twice now within `change_resource_state`, both here and in the method above (`change_resource_state`). WDYT about moving this check to the `apply_config_pattern_resource` method above, since it seems to be specific to this method? That will remove the need to check `if apply_config_action`

app/controllers/api/physical_servers_controller.rb
```diff
@@ -52,26 +52,47 @@ def refresh_resource(type, id, _data = nil)
...
+        rescue StandardError => err
           action_result(false, err.to_s)
         end
       end
     end
 
+    def validate_resource(state, type, id, data = {})
+      raise BadRequestError, "Must specify an id for changing a #{type} resource" unless id
+      if single_resource?
+        ensure_resource_exists(type, id)
```
jntullo: I do not believe this call is necessary, because when you do a `resource_search`, it will raise the appropriate message if the resource does not exist

igorrran: Yes. I agree. I'll make the changes

app/controllers/api/physical_servers_controller.rb
igorrran: But, the problem is that moving this check to apply_config_pattern_resource method It won't works if not single resource. For the operation in multiple resources is necessary to have a check inside the api_action iteration.

igorrran:
Hey @jntullo I made the changes. I removed the validate_resource and moved the customization_script check from change resource state method to apply config pattern method. WDYT

app/controllers/api/physical_servers_controller.rb
```diff
@@ -52,9 +52,20 @@ def refresh_resource(type, id, _data = nil)
       end
     end
 
+    # Process apply config pattern operation for single and multi-resources
+    #
+    # Even the request for multi-resources isn't completed successfully, return a HTTP Status 200
+    def apply_config_pattern_resource(type, id, data)
+      ensure_customization_script_exists(data["pattern_id"])
+      change_resource_state(:apply_config_pattern, type, id, data)
+    rescue StandardError => err
```
jntullo: This should not only be catching `StandardError`s but all errors, which is consistent with the rest of the API

app/controllers/api/physical_servers_controller.rb
```diff
@@ -64,14 +75,23 @@ def change_resource_state(state, type, id)
...
           action_result(false, err.to_s)
         end
       end
     end
 
+    # Process the check of the customization script by id
+    #
+    # @raise [NotFoundError] if id is invalid
+    def ensure_customization_script_exists(id)
+      ensure_resource_exists(:customization_scripts, id)
```
jntullo: Instead of a wrapper method around `ensure_resource_exists` for this particular case, I think it may make sense to add a message such as `"#{type} not found"` in `ensure_resource_exists` when the `NotFoundError` is raised. Thoughts?

spec/requests/physical_servers_spec.rb
```diff
@@ -483,4 +483,103 @@
...
+
+        post(api_physical_server_url(nil, 999_999), :params => gen_request(:apply_config_pattern, "pattern_id" => config_pattern.id, "uuid" => 999_999))
+
+        expect(response).to have_http_status(:not_found)
+      end
+    end
+
+    context "with a valid physical server id and an invalid config pattern id" do
+      it "it responds with 404 Not Found" do
+        ps = FactoryGirl.create(:physical_server)
```
jntullo: Instead of creating a physical server in each test, how do you feel about defining it below "describe" for re-use? You may want to consider creating two, since multiple are used in many cases.

app/controllers/api/physical_servers_controller.rb
igorrran: Should I change into change_resource method too?

app/controllers/api/physical_servers_controller.rb
igorrran: Yes. It makes sense

app/controllers/api/physical_servers_controller.rb
jntullo: @igorrran Yes that would be great 👍

igorrran: @jntullo I have changed the code to be catching all errors, but the codeclimate have been warning about this. Could you take a look please?

app/controllers/api/physical_servers_controller.rb
```diff
@@ -64,16 +75,16 @@ def change_resource_state(state, type, id)
           server = resource_search(id, type, klass)
           desc = "Requested server state #{state} for #{server_ident(server)}"
           api_log_info(desc)
-          task_id = queue_object_action(server, desc, :method_name => state, :role => :ems_operations)
+          task_id = queue_object_action(server, desc, :method_name => state, :role => :ems_operations, :args => data)
           action_result(true, desc, :task_id => task_id)
-        rescue => err
+        rescue Exception => err
```
jntullo: if you do `rescue => err`, it should make code climate happy 😄

igorrran: But rescue => err, ruby assumes that you mean StandardError

jntullo: @igorrran that should be sufficient here, as far as I can tel

app/controllers/api/physical_servers_controller.rb
```diff
@@ -64,7 +75,7 @@ def change_resource_state(state, type, id)
           server = resource_search(id, type, klass)
           desc = "Requested server state #{state} for #{server_ident(server)}"
           api_log_info(desc)
-          task_id = queue_object_action(server, desc, :method_name => state, :role => :ems_operations)
+          task_id = queue_object_action(server, desc, :method_name => state, :role => :ems_operations, :args => data)
```
abellotti: does :args default to a hash or an array ?  I see hash in 68 but MiqTask.generic_action_with_callback wants an array.

app/controllers/api/physical_servers_controller.rb
```diff
@@ -52,9 +52,20 @@ def refresh_resource(type, id, _data = nil)
       end
     end
 
+    # Process apply config pattern operation for single and multi-resources
+    #
+    # Even the request for multi-resources isn't completed successfully, return a HTTP Status 200
+    def apply_config_pattern_resource(type, id, data)
+      ensure_resource_exists(:customization_scripts, data["pattern_id"])
```
abellotti: do we need this method ?  can't we just call resource_search instead ?

igorrran: Yeah, We can just call resource_search, but I think it would make more sense if we used the result of the method to do something. Since we just need to do the check I guess that ensure_resource_script makes sense. What do you think?

igorrran: I can remove it in change_resource_state, but I'll have to raise an exception when the operation is for single resource, like I did in apply_config_pattern_resource in line 62. But I can't remove the ensure_resource_exists method in line 59 because I have to ensure that the config_pattern resource is valid also, and the resource_search just check the physical_server resource.

app/controllers/api/physical_servers_controller.rb
igorrran: Yes. Should be an array. You're right

abellotti:
Thanks @igorrran for the API enhancement. LGTM!! 👍

PR #279, opened 2018-01-11
Allow additional provider parameters to be specified on create

jntullo:
This update will allow individual providers to have more control over the parameters that are allowed when creating a new manager. This will also be a good transition starting point to move a lot of the validation inside of the providers themselves, rather than having it all within the API controller.

As an example, the Azure provider would like users to provide `azure_tenant_id`, which is not possible under the current API validation. By updating `.api_allowed_attributes` on the `Azure::CloudManager` to return `azure_tenant_id`, users will be allowed to specify that attribute for Azure Cloud Managers.

@miq-bot add_label wip, enhancement 
cc: @juliancheal @bronaghs 

Dependent on https://github.com/ManageIQ/manageiq/pull/16802

### DIFF
--- a/spec/requests/providers_spec.rb
+++ b/spec/requests/providers_spec.rb
@@ -455,6 +455,25 @@ describe "Providers API" do
         .and_return([OvirtSDK4::ProbeResult.new(:version => '3')])
     end
 
+    it 'allows provider specific attributes to be specified' do
+      allow(ManageIQ::Providers::Azure::CloudManager).to receive(:api_allowed_attributes).and_return(%w(azure_tenant_id))
+      tenant = FactoryGirl.create(:cloud_tenant)
+      api_basic_authorize collection_action_identifier(:providers, :create)
+
+      post(api_providers_url, :params => { "type"            => "ManageIQ::Providers::Azure::CloudManager",
+                                           "name"            => "sample azure provider",
+                                           "hostname"        => "hostname",
+                                           "zone"            => @zone,
+                                           "azure_tenant_id" => tenant.id,
+                                           "credentials"     => {}})
+
+      expected = {
+        "results" => [a_hash_including("uid_ems" => tenant.id.to_s, "name" => "sample azure provider")]
+      }
+      expect(response).to have_http_status(:ok)
+      expect(response.parsed_body).to include(expected)
+    end
+
     it "rejects creation without appropriate role" do
       api_basic_authorize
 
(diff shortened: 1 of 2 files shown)


juliancheal:
Is this ready to be un-wipped?

jntullo:
@juliancheal will unwip when https://github.com/ManageIQ/manageiq/pull/16802 is merged

juliancheal:
Good point @cben I'll check it's read as `azure_tenant_id` too.

The field that the user would know from Azure documentation is `tenant_id`, but as that clashes with ManageIQ tenants we've had to change what value we store it in.

bronaghs:
@cben Why do you believe the user cannot read back the `azure_tenant_id`? Do you mean like this? :

`ManageIQ::Providers::Azure::CloudManager.first.azure_tenant_id`

spec/requests/providers_spec.rb
```diff
@@ -421,6 +421,25 @@ def have_endpoint_attributes(expected_hash)
...
+      post(api_providers_url, :params => { "type"            => "ManageIQ::Providers::Azure::CloudManager",
+                                           "name"            => "sample azure provider",
+                                           "hostname"        => "hostname",
+                                           "zone"            => @zone,
+                                           "azure_tenant_id" => tenant.id,
+                                           "credentials"     => {}})
+
+      expected = {
+        "results" => [a_hash_including("uid_ems" => tenant.id.to_s, "name" => "sample azure provider")]
+      }
```
bronaghs: @jntullo - does the `results` hash mirror the database columns? Is that why the first hash key is `uid_ems` and not `azure_tenant_id`

cben:
I meant via API, by GET /API/providers/1

bronaghs:
Ah @cben, thanks for clarifying - I touched on the same point in my review comment [above](https://github.com/ManageIQ/manageiq-api/pull/279#pullrequestreview-88919126)
I, too, think we should read back `azure_tenant_id`.

spec/requests/providers_spec.rb
jntullo: @bronaghs yeah, it is all of the physical attributes

jntullo:
@bronaghs I did some thinking on this and now that I see this is a similar issue in openstack (where `keystone_v3_domain_id` is an alias of `:uid_ems`), I am thinking it may be a good idea to consider a serializer for these objects, so that they can get read back the same way. Thoughts? @juliancheal @cben @abellotti

abellotti:
LGTM!! Thanks @jntullo for this enhancement. 👍

PR #281, opened 2018-01-12
Remove useless expand_paths

chrisarcand:
These do literally nothing.

root.join('lib') == root.join('lib').expand_path #=> true

I'm sure it's a relic from before Rails.root was a Pathname

@bdunne Please review (and take note for all the other plugins/generator/whatever)


simaishi:
This was indirectly backported to Gaprindashvili branch by backport of https://github.com/ManageIQ/manageiq-api/pull/288

PR #287, opened 2018-01-17
Fix special characters in MIQ_GROUP header

jntullo:
Non-ASCII characters are not [permitted in headers](https://www.greenbytes.de/tech/webdav/rfc7230.html#rfc.section.A.2.p.9). In our case, we have an `X_MIQ_GROUP` header that specifies the group description. Our group descriptions allow special characters (as in the case `SR-APP-EPM-Membre-équipe`, which poses an interesting dilemma.

Headers with special characters get encoded incorrectly (ie, `SR-APP-EPM-Membre-\xE9quipe`), causing the following error, which does not allow the SUI to proceed with login if the user's current group contains a special character:
```
ActiveRecord::StatementInvalid: PG::CharacterNotInRepertoire: ERROR:  invalid byte sequence for encoding "UTF8": 0xe9 0x71 0x75
: SELECT  "miq_groups".* FROM "miq_groups" INNER JOIN "miq_groups_users" ON "miq_groups"."id" = "miq_groups_users"."miq_group_id" WHERE "miq_groups_users"."user_id" = $1 AND "miq_groups"."description" = $2 LIMIT $3
```
Gif to demonstrate before behavior (pretty uneventful):
login_fail_sui

This fix properly encodes it back into UTF8, however, I am on the fence on whether we should allow this, or if group description validation should be changed to not allow special characters (although, may be difficult because groups already exist with special characters). 

The following gif demonstrates login working successfully with a user that belongs to a group with a special character:

login_success_sui

Thoughts? @imtayadeway @abellotti @gtanzillo 
cc: @AllenBW @yrudman 
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1531626#c9

@miq-bot add_label bug, blocker, gaprindashvili/yes

### DIFF
--- a/app/controllers/api/base_controller/authentication.rb
+++ b/app/controllers/api/base_controller/authentication.rb
@@ -44,6 +44,7 @@ module Api
       def authorize_user_group(user_obj)
         group_name = request.headers[HttpHeaders::MIQ_GROUP]
         if group_name.present?
+          group_name = CGI.unescape(group_name)
           group_obj = user_obj.miq_groups.find_by(:description => group_name)
           raise AuthenticationError, "Invalid Authorization Group #{group_name} specified" if group_obj.nil?
           user_obj.current_group_by_description = group_name
--- a/spec/requests/authentication_spec.rb
+++ b/spec/requests/authentication_spec.rb
@@ -106,6 +106,19 @@ describe "Authentication API" do
     end
   end
 
+  context "Group Authorization with special characters" do
+    let(:special_char_group) { FactoryGirl.create(:miq_group, :description => "équipe", :miq_user_role => @role) }
+
+    it "permits group headers to be specified with properly escaped descriptions" do
+      @user.miq_groups << special_char_group
+      api_basic_authorize
+
+      get api_entrypoint_url, :headers => {Api::HttpHeaders::MIQ_GROUP => CGI.escape(special_char_group.description)}
+
+      expect(response).to have_http_status(:ok)
+    end
+  end
+
   context "Authentication/Authorization Identity" do
     let(:group1) { FactoryGirl.create(:miq_group, :description => "Group1", :miq_user_role => @role) }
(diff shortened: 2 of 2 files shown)


gtanzillo:
@jntullo Based on your comment, do you think we still need this change to the API?

jntullo:
@gtanzillo I am not sure. If we decide to keep this header available in the API as a way to authenticate against a different group than the user's `current_group`, then it would make sense that we should fix this in the event that someone does have a special character in their group name and are using this header. 

However, I'm not sure that this header is even needed anymore. Users are now able to change their current group ("action": "set_current_group"), and it seems to make more sense that they are authenticated against their current group, not the one specified in the headers. I'll start a discussion in the API gitter room. 

I will remove the BZ reference from the commit, though, as I believe the SUI approach is the better solution to that particular bug.

abellotti:
Yes, we still need to fix this and keep supporting the header.

jntullo:
@abellotti pushed up another commit that may be a better approach. Perhaps we should require that the header be properly encoded. I tested by properly escaping the header like so:
```
req['X_MIQ_GROUP'] = CGI.escape('SR-APP-EPM-Membre-équipe')
```
and it works as expected. 

cc: @imtayadeway

imtayadeway:
> Perhaps we should require that the header be properly encoded

Agree. If we must keep compatibility with the current behavior, we shouldn't be trying to work around clients sending non-ASCII characters.

abellotti:
Thanks @jntullo for fixing this.  Will merge when 🍏

simaishi:

```
$ git log -1
commit 9cbb5d91f4db9ce0fd855c06ddf7b2dcd1f593d6
Date:   Wed Jan 17 15:28:25 2018 -0500

    Merge pull request #287 from jntullo/bz_1531626
    
    Fix special characters in MIQ_GROUP header
    (cherry picked from commit e89fc56e6a6be50447fe61c210f254843c61eb6d)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1536047
```

PR #289, opened 2018-01-18
Rename specs to match collection names

jntullo:
Both the `picture_spec` and `orchestration_template` specs should have pluralized name to match their collections, as is the convention with all other spec files

@miq-bot add_label cleanup


jntullo:
@miq-bot assign @imtayadeway

PR #291, opened 2018-01-18
AssetDetails model should be AssetDetail

bdunne:
Depends on https://github.com/ManageIQ/manageiq/pull/16833


abellotti:
probably g/yes because https://github.com/ManageIQ/manageiq/pull/16833 is marked as such.

PR #293, opened 2018-01-19
Verify against the product features file instead of seeding

imtayadeway:
Seeding makes 1000s of calls database, making this (consistently) the
slowest test in the API suite. Loading the product features fixture
file into memory and traversing it instead is orders of magnitude
faster.

Before this change, running this one test takes for me 3-5 seconds
typically. After, it takes ~0.05 seconds.

### DIFF
--- a/spec/lib/api/api_config_spec.rb
+++ b/spec/lib/api/api_config_spec.rb
@@ -11,7 +11,12 @@ describe 'API configuration (config/api.yml)' do
     end
 
     describe 'identifiers' do
-      let(:miq_product_features) { MiqProductFeature.seed.values.flatten.to_set }
+      let(:miq_product_features) do
+        identifiers = Set.new
+        each_product_feature { |f| identifiers.add(f[:identifier]) }
+        identifiers
+      end
+
       let(:api_feature_identifiers) do
         collection_settings.each_with_object(Set.new) do |(_, cfg), set|
           set.add(cfg[:identifier]) if cfg[:identifier]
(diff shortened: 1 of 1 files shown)


imtayadeway:
@abellotti I'm going to ignore the cop complaint as it's a bit pedantic - using `yield` won't obviate the need to assign to `block` since it's getting passed around in the lines that follow

spec/lib/api/api_config_spec.rb
```diff
@@ -38,6 +43,13 @@
         dangling = api_feature_identifiers - miq_product_features
         expect(dangling).to be_empty
       end
+
+      def each_product_feature(feature = YAML.load_file("#{MiqProductFeature::FIXTURE_PATH}.yml"), &block)
+        block.call(feature)
+        Array(feature[:children]).each do |child|
+          each_product_feature(child, &block)
+        end
+      end
```
abellotti: 👍  on the perf gain.  Though one concern is API having knowledge of the above FIXTURE_PATH. I wonder if the each_product_feature method or some variant of it should live in app/models/miq_product_feature.rb and we'd leverage from the let(:miq_product_features) statement as we do above.

imtayadeway: I like the idea, although I dislike the idea of increasing the surface of `MiqProductFeature`'s interface with another method - especially one that should never be used elsewhere - it's exposing the model's underlying fixture data. I think we may have to acknowledge that this is not a conventional test (it's checking configuration, not exercising behavior) and may have to do break the rules a bit to provide some quick verification that things aren't grossly misconfigured.

abellotti: We can keep as is for now, especially that this spec is the only user of this new method.  We just have to realize that possible future changes to MiqProductFeature's internals might break this test here, which we'd need to fix separately.

PR #295, opened 2018-01-21
Add api specs for prometheus endpoints

zeari:
BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1535941
spec tests for https://github.com/ManageIQ/manageiq-providers-kubernetes/pull/220

This wont catch the regression handled in https://github.com/ManageIQ/manageiq-providers-kubernetes/pull/220 but we rely on api too much for us not to have at least some prometheus spec tests here.

@elad661 @cben @moolitayer PTAL

### DIFF
--- a/spec/requests/providers_spec.rb
+++ b/spec/requests/providers_spec.rb
@@ -132,11 +132,46 @@ describe "Providers API" do
       }
     }
   end
-  let(:sample_containers_multi_end_point) do
+  let(:prometheus_connection) do
+    {
+      "endpoint"       => {
+        "role"              => "prometheus",
+        "hostname"          => "prometheus.example.com",
+        "port"              => 443,
+        "security_protocol" => "ssl-without-validation"
+      },
+      "authentication" => {
+        "role"     => "prometheus",
+        "auth_key" => SecureRandom.hex
+      }
+    }
+  end
+  let(:prometheus_alerts_connection) do
+    {
+      "endpoint"       => {
+        "role"              => "prometheus_alerts",
+        "hostname"          => "prometheus.example.com",
+        "port"              => 443,
+        "security_protocol" => "ssl-without-validation"
+      },
+      "authentication" => {
+        "role"     => "prometheus_alerts",
+        "auth_key" => SecureRandom.hex
+      }
+    }
+  end
+  let(:sample_containers_multi_end_point_with_hawkular) do
     {
       "type"                      => containers_class,
-      "name"                      => "sample containers provider with multiple endpoints",
+      "name"                      => "sample containers provider with multiple endpoints and hawkular",
       "connection_configurations" => [default_connection, hawkular_connection]
+    }
+  end
+  let(:sample_containers_multi_end_point_with_prometheus) do
+    {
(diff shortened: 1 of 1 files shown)


zeari:
@miq-bot add_label providers/containers

elad661:
I know you're not touching that part in the file, but I don't like the fact that we have `security_protocol` with values that makes no sense. I'll send a PR to our provider to reject those, so can you also change those values to something that makes sense while you're at it?

spec/requests/providers_spec.rb
```diff
@@ -596,6 +631,35 @@ def token(connection)
...
+
+          expect(provider.connection_configurations.prometheus.endpoint).to have_endpoint_attributes(
+            prometheus_connection["endpoint"]
+          )
+
+          expect(provider.connection_configurations.prometheus_alerts.endpoint).to have_endpoint_attributes(
+            prometheus_alerts_connection["endpoint"]
+          )
+          expect(provider.authentication_token(:prometheus)).to eq(token(prometheus_connection))
+          expect(provider.authentication_token(:prometheus_alerts)).to eq(token(prometheus_alerts_connection))
```
cben: I'm not sure this tested what we wanted to test.
AFAICT it only checks parent endpoints got created correctly; we mainly want to test child manager got created (and can see the endpoint it needs).

Consider also adding test for deletion of prometheus endpoint and maybe of whole provider resulting in child manager deletion.

elad661: @cben testing the child manager got created should probably not be done in the API tests. If I understand correctly, API tests are supposed to test the interface, not the implementation, and the child manager is definitely an implementation detail.

zeari: I was thinking of `it doesnt blow up` type tests

Should the sub-manager logic go away(which i feel it might) then we will need to change these spec tests so endpoint creation is enough IMO.

zeari:
> I know you're not touching that part in the file, but I don't like the fact that we have security_protocol with values that makes no sense in some of those tests. I'll send a PR to our provider to reject those, so can you also change those values to something that makes sense while you're at it?

Changed to `"ssl-without-validation"`

spec/requests/providers_spec.rb
cben: > API tests are supposed to test the interface, not the implementation, and the child manager is definitely an implementation detail.

Well, I really want an integration test that implementation reacted correctly to API *somewhere*. 
 Do you see a better place/way to test it?
OK, in principle, if we have a test in our repo that directly creating parent endpoint creates the child (do we?), and test that API creates endpoints, then *in theory* we're covered.  I have low trust for theory in this case though :-)

> I was thinking of it doesnt blow up type tests

Hmm, yes, for this specific failure it was blowing up, that's enough.

zeari:
@abellotti Please review

abellotti:
I think this is fine, testing what the inbound API request is supposed to do.  Not sure if there are additional provider wiring tests to do for this scenario, but that may be best in the provider repo.

So, I'm 👍  here.  /cc @jntullo thoughts ?

PR #296, opened 2018-01-22
Update role identifiers for tasks collection

jntullo:
When user roles are seeded, they have either the `miq_task_all_ui` feature or `miq_task_my_ui` feature ([or both](https://github.com/ManageIQ/manageiq/blob/master/db/fixtures/miq_user_roles.yml#L67-L68)), which is what the UI also uses to validate that they can view requests. However, the API uses `tasks_view`. This is causing a problem where users that are able to see tasks in the UI are unable to see tasks through the API. This updates the product features used in the API to match what is used in the UI.

Note, no changes were needed for this due to `api_user_role_allows?` method already [accepting arrays](https://github.com/ManageIQ/manageiq-api/blob/master/app/controllers/api/base_controller/renderer.rb#L496), but have updated the api config spec accordingly, since this is the first place we are doing so.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1535962

### DIFF
--- a/config/api.yml
+++ b/config/api.yml
@@ -2716,16 +2716,22 @@
     :collection_actions:
       :get:
       - :name: read
-        :identifier: tasks_view
+        :identifier:
+        - miq_task_all_ui
+        - miq_task_my_ui
       :post:
       - :name: query
-        :identifier: tasks_view
+        :identifier:
+        - miq_task_all_ui
+        - miq_task_my_ui
       - :name: delete
         :identifier: miq_task_delete
     :resource_actions:
       :get:
       - :name: read
-        :identifier: tasks_view
+        :identifier:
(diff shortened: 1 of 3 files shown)


abellotti:
This LGTM. Thanks @jntullo 👍

simaishi:

```
$ git log -1
commit e31423fb0df4650b067be4d268c7dafebfd69dd0
Date:   Tue Jan 23 09:00:07 2018 -0500

    Merge pull request #296 from jntullo/bz_1535962
    
    Update role identifiers for tasks collection
    (cherry picked from commit 79e5081b52ba259cecd7e660dbd1b09e3e31c41f)
    
    Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1552278
```

PR #297, opened 2018-01-24
Return Time in API entrypoint server_info

jntullo:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1530953

### DIFF
--- a/app/controllers/api/api_controller.rb
+++ b/app/controllers/api/api_controller.rb
@@ -63,6 +63,7 @@ module Api
         :version     => Vmdb::Appliance.VERSION,
         :build       => Vmdb::Appliance.BUILD,
         :appliance   => MiqServer.my_server.name,
+        :time        => Time.now.utc.iso8601,
         :server_href => "#{@req.api_prefix}/servers/#{MiqServer.my_server.id}",
         :zone_href   => "#{@req.api_prefix}/zones/#{MiqServer.my_server.zone.id}",
         :region_href => "#{@req.api_prefix}/regions/#{MiqRegion.my_region.id}"
--- a/spec/requests/entrypoint_spec.rb
+++ b/spec/requests/entrypoint_spec.rb
@@ -40,18 +40,21 @@ RSpec.describe "API entrypoint" do
   it "returns server_info" do
     api_basic_authorize
 
-    get api_entrypoint_url
+    Timecop.freeze(Time.utc(2018, 0o1, 0o1, 0o0, 0o0, 0o0)) do
+      get api_entrypoint_url
 
-    expect(response.parsed_body).to include(
-      "server_info" => a_hash_including(
-        "version"     => Vmdb::Appliance.VERSION,
-        "build"       => Vmdb::Appliance.BUILD,
-        "appliance"   => MiqServer.my_server.name,
-        "server_href" => api_server_url(nil, MiqServer.my_server),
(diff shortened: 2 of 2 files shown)


abellotti:
Thanks @jntullo for this enhancement.  This LGTM!!  

I think we're ok, but can you just test the manageiq-api-client that there are no issues fetching the server info.  (without the time of course, I'll create an issue in that repo).

jntullo:
@abellotti Just checked the client and there are no issues with fetching the server info ✨

spec/requests/entrypoint_spec.rb
```diff
@@ -47,6 +47,7 @@
         "version"     => Vmdb::Appliance.VERSION,
         "build"       => Vmdb::Appliance.BUILD,
         "appliance"   => MiqServer.my_server.name,
+        "time"        => a_kind_of(String),
```
imtayadeway: Do you know what format this time will be in? Perhaps you could leverage timecop around this test to make an assertion on it? I would expect that it would come back in ISO8601 format 😁

abellotti: come to think of it, it should be consistent with other time attrs the API normalizes, see normalize_time(), so Time.now.utc.iso8601 to be specifically.

app/controllers/api/api_controller.rb
```diff
@@ -63,6 +63,7 @@ def server_info
         :version     => Vmdb::Appliance.VERSION,
         :build       => Vmdb::Appliance.BUILD,
         :appliance   => MiqServer.my_server.name,
+        :time        => Time.now.utc,
```
abellotti: Time.now.utc.iso8601.    normalize_time(Time.now) would be preferred though that's currently a private method.

imtayadeway: Isn't this later passed through the normalization process anyway?

abellotti: yes, but the attribute "time" is not a time/date time declared attribute in any of the models we reference, so it won't be normalizing it as such.  (Api::Environment.time_attributes).

imtayadeway: Maybe we should add "time" to the list to avoid the double-pass?

abellotti: was thinking about that, also renaming this to "timestamp" but that has a different semantic.  If we want to add this to the list to leverage the normalizer, we may want to rename it first to be clear as to its content, maybe "current_time" ?

spec/requests/entrypoint_spec.rb
```diff
@@ -40,18 +40,21 @@
...
-        "region_href" => api_region_url(nil, MiqRegion.my_region)
+    Timecop.freeze(Time.new(2018, 0o1, 0o1, 0o0, 0o0, 0o0).utc) do
+      get api_entrypoint_url
+
+      expect(response.parsed_body).to include(
+        "server_info" => a_hash_including(
+          "version"     => Vmdb::Appliance.VERSION,
+          "build"       => Vmdb::Appliance.BUILD,
+          "appliance"   => MiqServer.my_server.name,
+          "time"        => Time.now.utc.iso8601.to_s,
```
imtayadeway: With time frozen, you can now make an assertion on a literal, i.e. `"2018-01-01T00:00:00.00Z"`

abellotti:
Thanks @jntullo for this enhancement. 👍

@psav for your testing pleasures.

PR #299, opened 2018-01-25
Allow multiple role identifiers for cloud volume

lpichler:
pulled out from https://github.com/ManageIQ/manageiq/pull/15249 - there is description
but the problem is solved only for storage managers.

@miq-bot add_label gaprindashvili/yes, blocker

# links
https://bugzilla.redhat.com/show_bug.cgi?id=1538003

cc @gtanzillo 

@miq-bot assign @abellotti

### DIFF
--- a/config/api.yml
+++ b/config/api.yml
@@ -1795,10 +1795,14 @@
     :collection_actions:
       :get:
       - :name: read
-        :identifier: ems_infra_show_list
+        :identifier:
+        - ems_infra_show_list
+        - ems_block_storage_show_list
       :post:
       - :name: query
-        :identifier: ems_infra_show_list
+        :identifier:
+        - ems_infra_show_list
+        - ems_block_storage_show_list
       - :name: create
         :identifier: ems_infra_new
       - :name: edit
@@ -1810,7 +1814,9 @@
     :resource_actions:
       :get:
       - :name: read
-        :identifier: ems_infra_show
+        :identifier:
+        - ems_infra_show
+        - ems_block_storage_show
       :post:
       - :name: edit
         :identifier: ems_infra_edit
--- a/spec/lib/api/api_config_spec.rb
+++ b/spec/lib/api/api_config_spec.rb
@@ -19,7 +19,7 @@ describe 'API configuration (config/api.yml)' do
 
       let(:api_feature_identifiers) do
         collection_settings.each_with_object(Set.new) do |(_, cfg), set|
-          set.add(cfg[:identifier]) if cfg[:identifier]
+          Array(cfg[:identifier]).each { |id| set.add(id) }
           keys = [:collection_actions, :resource_actions, :subcollection_actions, :subresource_actions]
           Array(cfg[:subcollections]).each do |s|
             keys << "#{s}_subcollection_actions" << "#{s}_subresource_actions"


spec/lib/api/api_config_spec.rb
```diff
@@ -17,9 +17,19 @@
...
+          else
+            set.add(identifier)
+          end
+        end
+      end
+
       let(:api_feature_identifiers) do
         collection_settings.each_with_object(Set.new) do |(_, cfg), set|
-          set.add(cfg[:identifier]) if cfg[:identifier]
+          add_identifier_to_set(cfg[:identifier], set)
```
abellotti: Can't this line just be changed to:

```
Array(cfg[:identifier]).each { |id| set.add(id) }
```

Without introducing the add_identifier_to_set method.

spec/lib/api/api_config_spec.rb
```diff
@@ -29,7 +39,7 @@
             cfg[action_type].each do |_, method_cfg|
               method_cfg.each do |action_cfg|
                 next unless action_cfg[:identifier]
-                Array(action_cfg[:identifier]).each { |id| set.add(id) }
+                Array(action_cfg[:identifier]).each { |id| add_identifier_to_set(action_cfg[:identifier], set) }
```
abellotti: probably don't need this as it already deals with the Array

config/api.yml
```diff
@@ -1792,10 +1792,14 @@
     :collection_actions:
       :get:
       - :name: read
-        :identifier: ems_infra_show_list
+        :identifier:
+        - ems_infra_show_list
+        - ems_block_storage_show_list
```
abellotti: Are Block Storage Managers providers ?  i.e. subclass of ExtManagementSystem or Provider ? If so then this change is fine, i.e.  User is authorized to see both variants as per the miq product feature selections and should be here since they are both accessible via /api/providers.

lpichler:
> This suggests to me that there's a problem with our configuration - we have a concept of a role that we haven't identified, that is the parent of these two other identifiers. Could this be addressed by adding a parent identifier?

Can show me an example of your idea?
Do you think for example to add children section to [miq_product_features.yml](https://github.com/ManageIQ/manageiq/blob/master/db/fixtures/miq_product_features.yml#L921) ?

if so, it will cause that the added subtree will be displayed in the tree of product role feature when the role is edited.

config/api.yml
lpichler: yes, exactly, it is:
```
irb(main):013:0> ExtManagementSystem > ManageIQ::Providers::StorageManager
true
```

spec/lib/api/api_config_spec.rb
lpichler: yes, sure, it is much better.

spec/lib/api/api_config_spec.rb
lpichler: changed 👍

abellotti:
Thanks @lpichler for updating and fixing this. LGTM!! 👍 

@imtayadeway please take a 👀 Thanks!

lpichler:
ping @imtayadeway

simaishi:

```
$ git log -1
commit 1035373aa2488d0cd94c0ca783f42b1b9eab2467
Date:   Thu Feb 1 11:00:41 2018 -0500

    Merge pull request #299 from lpichler/allow_multiple_role_identifiers_for_cloud_volume
    
    Allow multiple role identifiers for cloud volume
    (cherry picked from commit aca3a26078bfe0875327bc989561d34a109f649f)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1552821
```

PR #300, opened 2018-01-25
Adding the customization_script retrieve operation via the REST API

igorrran:
This PR is able to:
* Add a new operation to retrieving a configuration pattern from customization scripts endpoint

### DIFF
--- a/spec/requests/customization_script_spec.rb
+++ b/spec/requests/customization_script_spec.rb
@@ -13,7 +13,7 @@ RSpec.describe "Customization Scripts" do
     end
 
     context "with valid id" do
-      it "returns both id and href" do
+      it "shows all of its properties" do
         config_pattern = FactoryGirl.create(:customization_script)
 
         api_basic_authorize action_identifier(:customization_scripts, :read, :resource_actions, :get)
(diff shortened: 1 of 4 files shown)


abellotti:
Thanks @igorrran for making this separate PR!!  could you add the collection test in spec/requests/collection_specs.rb and add support for the bulk "query" action (see other collections you'd need a post for that, and a second bulk query test also in spec/requests/collection_specs.rb).

Thanks !!

abellotti:
Ping @jntullo when you get a chance to review 👀   Thanks!!

spec/requests/customization_script_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+        api_basic_authorize
+        get api_customization_script_url(nil, config_pattern)
+
+        expect(response).to have_http_status(:forbidden)
+        expect(response.parsed_body["error"]).to include("kind" => "forbidden")
+      end
+    end
+
+    context "with valid id" do
+      it "returns both id and href" do
```
jntullo: just a NIT here, but it should return the whole config pattern resource, not just id and href, so this is a little misleading

igorrran: Should return all config pattern details(href, id, name, manager_id, created_at, description...)?

jntullo: @igorrran don't worry about testing all of the details...just update this to describe it as "returns the resource" or something similar. Thanks!!

abellotti:
Thanks @igorrran for making the update and the API Enhancement!! 👍

PR #301, opened 2018-01-25
Queue chargeback reports of services

jntullo:
This allows queueing of chargeback reports on individual services:
```
POST /api/services/:id
{ "action": "queue_chargeback_report" }
```

or multiple:

```
POST /api/services
{
   "action": "queue_chargeback_report", 
  "resources": [{"id": <svc_id>}, {"href": "api/services/:id"} ]
}
```

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1530952

### DIFF
--- a/app/controllers/api/services_controller.rb
+++ b/app/controllers/api/services_controller.rb
@@ -135,6 +135,14 @@ module Api
       action_result(false, err.to_s)
     end
 
+    def queue_chargeback_report_resource(_type, id, _data)
+      service = resource_search(id, :services, Service)
+      task = service.queue_chargeback_report_generation(:userid => current_user.id)
+      action_result(true, "Queued chargeback report generation for #{service_ident(service)}", :task_id => task.id)
+    rescue StandardError => err
+      action_result(false, "Could not queue chargeback report generation - #{err}")
+    end
+
     private
 
     def validate_resource(data)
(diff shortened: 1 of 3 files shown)


abellotti:
Hi @jntullo, thanks for the enhancement!!

Good in general, the only thing that I'm not crazy about is the all.  signature wise and functionality in general.

I see the method in the model does a Service.all...  is this really what we want for the API, shouldn't it be all services that I can see (rbac) ? maybe double check use case ?

second, the signature, so many ways for doing this, parameters, separate action, ...  Maybe a hash for now, we can always add other directives.

```
{
  "resource" : {
     "targets" : "all"
  }
}
```

they can also search/query/filter services first, then do a bulk action on those  (my preference as they get filtering capabilities and rbac).

jntullo:
Yeah, i was a bit worried about the `all` as well. Pushed up another commit that allows for the specifying of individual services. 

@psav, thoughts? This will allow you to specify service resources to queue chargeback reports for.

abellotti:
Thanks @jntullo for the update.  I feel better with this change, leveraging our bulk action mechanism for the all.  👍 

@psav if you're good with approach, I can merge.  Thanks!!

abellotti:
ping @psav please let us know if you're good with this. Thanks.

jntullo:
Dependent on https://github.com/ManageIQ/manageiq/pull/17135

abellotti:
Thanks for the task_id update @jntullo, LGTM!! will merge when 🍏

PR #303, opened 2018-01-26
Allow an array to be specified on PATCH for settings

jntullo:
The rest of the API supports an array being passed to PATCH, and this will allow it to remain consistent with what is accepted elsewhere. In addition, it supports multiple updates to the settings at once.

Example:
```
PATCH /api/zone/:id/settings
[
  { "api": {"authentication_timeout": "<new_value>"} },
  { "binary_blob": {"purge_window_size": "<new_value>"} }
]
```
cc: @psav

### DIFF
--- a/app/controllers/api/base_controller.rb
+++ b/app/controllers/api/base_controller.rb
@@ -117,7 +117,13 @@ module Api
       case @req.method
       when :patch
         raise ForbiddenError, "You are not authorized to edit settings." unless super_admin?
-        resource.add_settings_for_resource(@req.json_body)
+        if @req.json_body.kind_of?(Array)
+          @req.json_body.each do |setting|
+            resource.add_settings_for_resource(setting)
+          end
+        else
+          resource.add_settings_for_resource(@req.json_body)
+        end
       when :delete
         raise ForbiddenError, "You are not authorized to remove settings." unless super_admin?
         resource.remove_settings_path_for_resource(*@req.json_body)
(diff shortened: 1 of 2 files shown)


abellotti:
Thanks @jntullo for the enhancement !!

One alternative is to handle the patch document syntax we support for API resources (action/path/value), the following signature could also work:

PATCH /api/servers/:id/settings

```
[
   { "action" : "delete", "path" : "ems/ems_amazon/event_groups/addition/critical" },
   { "action" : "delete", "path" : "workers/worker_base/event_catcher/event_catcher_openshift/poll" },
   { "action" : "edit", "path" : "log/level_fog", "value" : "debug" },
   { "action" : "edit", "path" : "server/startup_timeout", "value" : 600 }
]
```

This would allow bulk edits and deletes of settings in a single call.

chrisarcand:
> This would allow bulk edits and deletes of settings in a single call.

The current settings API allows bulk edits already. It's literally just a hash to be merged in.

jntullo:
@chrisarcand I realize that now. Missed that originally. 👍

abellotti:
With the API now supporting PATCH with standard hash signature in addition to the PATCH doc as per https://github.com/ManageIQ/manageiq-api/pull/302 and that above Hash for settings PATCH support the bulk anyways, maybe we should drop this PR and simply look into enhancing both python and ruby clients to add support for hash signature for patching settings.

jntullo:
I agree @abellotti 👍

PR #304, opened 2018-01-26
Add settings as a subcollection

jntullo:
This allows settings to be returned as a subcollection via `?expand=settings`

### DIFF
--- a/app/controllers/api/base_controller/renderer.rb
+++ b/app/controllers/api/base_controller/renderer.rb
@@ -420,6 +420,8 @@ module Api
             :expand_resources => @req.expand?(sc)
           }
           json.set! sc.to_s, collection_to_jbuilder(sc.to_sym, sctype, subresources, copts)
+        elsif subresources.kind_of?(Hash)
+          json.set!(sc, subresources)
         else
           sc_key_id = collection_config.resource_identifier(sctype)
           json.set! sc.to_s do |js|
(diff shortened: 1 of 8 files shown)


app/controllers/api/base_controller/renderer.rb
```diff
@@ -420,6 +420,8 @@ def expand_subcollection(json, sc, sctype, subresources)
             :expand_resources => @req.expand?(sc)
           }
           json.set! sc.to_s, collection_to_jbuilder(sc.to_sym, sctype, subresources, copts)
+        elsif subresources.kind_of?(Hash)
+          json.set!(sc, subresources)
```
abellotti: This will not normalize it like the regular GET /api/settings, so it will return passwords and such, we'd need json.set!(sc, normalize_hash(....))

abellotti:
Hi @jntullo PR is good, Thanks!!  Maybe just add a test showing that are normalizer is in effect, i.e. set "authentication" : { "bind_pwd" : "..." } and test that "bind_pwd" is not returned.   Thanks.

abellotti:
ping @jntullo

abellotti:
not much more than the current /api/servers/1/settings.  GET /api/servers/1?expand=settings would be returning the servers/1 object + the settings they get with /api/servers/1/settings.

chrisarcand:
You have to think of what the user *could* request though; with an expansion and multiple `&attributes=`, that's much more than just the object itself. 

¯\\\_(ツ)_/¯ Just thought I'd point out, if you're not worried, great! :shipit: 👍

abellotti:
Thanks @jntullo for going through the normalizer and adding the test. 👍

PR #306, opened 2018-01-29
Delete of search filters

jntullo:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1530948

Allows deletes by id, href or description
```
POST /api/search_filters
{
  "action": "delete",
  "resources": [
    { "id": ":id" },
    { "href": ":href" },
    { "description": ":description"}
  ]
}
```
or resource delete by ID 
```
POST /api/search_filters/:id
{ "action": "delete" } 
```

or delete via `DELETE`
```
DELETE /api/search_filters/:id
```

cc: @psav

### DIFF
--- a/app/controllers/api/search_filters_controller.rb
+++ b/app/controllers/api/search_filters_controller.rb
@@ -1,0 +1,4 @@
+module Api
+  class SearchFiltersController < BaseController
+  end
+end
--- a/config/api.yml
+++ b/config/api.yml
@@ -2236,6 +2236,33 @@
       :get:
       - :name: read
         :identifier: miq_report_view
+  :search_filters:
+    :description: Filters
+    :verbs: *gpd
+    :klass: MiqSearch
+    :identifying_attrs: description
+    :options:
+    - :collection
+    - :show
+    :collection_actions:
+      :get:
+      - :name: read
+        :identifier: my_settings_default_filters
+      :post:
+      - :name: query
+        :identifier: my_settings_default_filters
+      - :name: delete
+        :identifier: my_settings_default_filters
+    :resource_actions:
+      :get:
+      - :name: read
+        :identifier: my_settings_default_filters
+      :post:
+      - :name: delete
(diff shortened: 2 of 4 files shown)


jntullo:
I will be following this up with a create action for search filters

app/controllers/api/search_filters_controller.rb
```diff
@@ -0,0 +1,22 @@
+module Api
+  class SearchFiltersController < BaseController
+    def delete_resource(type, id, data = {})
+      id ||= data['description']
```
abellotti: we could add identifying_attrs: of description in the api.yml and this would be done for us by the manager.

app/controllers/api/search_filters_controller.rb
```diff
@@ -0,0 +1,22 @@
...
+    private
+
+    def resource_search(id, type, klass)
+      if id.to_s =~ /\A\d+\z/
+        super
+      else
+        resource = klass.find_by!(:description => id)
+        filter_resource(resource, type, klass)
+      end
+    end
```
abellotti: may not be needed with identifying_attrs defined.

config/api.yml
```diff
@@ -2211,6 +2211,24 @@
       :get:
       - :name: read
         :identifier: miq_report_view
+  :search_filters:
+    :description: Filters
+    :verbs: *gpd
```
abellotti: with get enabled, we should have a test added in collections_spec.rb to test them.

config/api.yml
```diff
@@ -2211,6 +2211,24 @@
...
+    :klass: MiqSearch
+    :options:
+    - :collection
+    - :show
+    :collection_actions:
+      :post:
```
abellotti: don't we need get and identifying attrs in collection/resource actions ?

abellotti:
ping @jntullo, this LGTM. can you look though into the travis failure ? Thanks.

abellotti:
👍

PR #307, opened 2018-01-30
Updated container images scan action to fix the action return signature

abellotti:
Updated container images scan action to fix the action return signature being returned was a hash of the task object, this is not consistent with the action result signature where we have the task_id and task_href of the task just created with the href of the targeted resource.

This is a follow-up to https://github.com/ManageIQ/manageiq-api/pull/245


enoodle:

PR #308, opened 2018-01-31
Fix error message for required params on metric rollups subcollection

jntullo:
resource_type is set by the controller when the request is made based off of the collection that has been specified. Currently the error message informs the user that all three of `resource_type`, `capture_interval`, and `start_date` are required. This returns a more appropriate error message based off of the missing parameters.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1540254


abellotti:
LGTM!! Thanks @jntullo for fixing this. 👍

PR #309, opened 2018-01-31
Implementing change password action for providers endpoint

douglasgabriel:
**This PR adds:**
* `change_password` action for `/providers` endpoint

**This PR depends on:**
~https://github.com/ManageIQ/manageiq/pull/16895~ - Merged
~https://github.com/ManageIQ/manageiq-providers-lenovo/pull/130~ - Merged
~https://github.com/ManageIQ/manageiq/pull/16883~ - Merged
~https://github.com/lenovo/xclarity_client/pull/81~ - Merged
~https://github.com/ManageIQ/manageiq/pull/17323~ - Merged

### DIFF
--- a/app/controllers/api/providers_controller.rb
+++ b/app/controllers/api/providers_controller.rb
@@ -106,7 +106,47 @@ module Api
       action_result(false, "Could not resume Provider - #{err}")
     end
 
+    # Process change_password action for a single resource or a collection of resources
+    def change_password_resource(type, id, data = {})
+      if single_resource?
+        change_password(type, id, data)
+      else
+        change_password_multiple_providers(type, id, data)
+      end
+    end
+
     private
+
+    # Process password change request for a single resource
+    #
+    # @raise [BadRequestError] if no id is passed or some required data is missing
+    #                          or some error occur when try to change password on provider client
+    # @raise [NotFoundError]   if there isn't providers with the id passed
+    def change_password(type, id, data)
+      raise BadRequestError, "Must specify an id for change password of a #{type} resource" unless id
+      api_action(type, id) do |klass|
+        provider = resource_search(id, type, klass)
+        desc = "Change password requested for Physical Provider #{provider.name}"
+        task_id = provider.change_password_queue(User.current_user.userid, data["current_password"], data["new_password"])
+        action_result(true, desc, :task_id => task_id)
+      end
+    end
+
+    # Process password change for a collection of resources
+    #
+    # Even the request isn't completed successfully, return a HTTP Status 200
+    #   with individual response for all resources.
+    #
+    # @see #change_password
+    #
+    # @return [Hash] contains details about the request proccess
+    #   :success [Boolean] indicates if the request was successfully completed
+    #   :message [String]  description of request proccess
+    def change_password_multiple_providers(type, id, data)
+      change_password(type, id, data)
+    rescue BadRequestError, ActiveRecord::RecordNotFound, MiqException::Error => exception
+      action_result(false, _(exception.message))
+    end
 
     def provider_ident(provider)
       "Provider id:#{provider.id} name:'#{provider.name}'"
--- a/spec/requests/providers_spec.rb
+++ b/spec/requests/providers_spec.rb
@@ -1156,6 +1156,139 @@ describe "Providers API" do
     end
   end
 
+  describe 'change provider password' do
+    let(:ems_physical_infra) { FactoryGirl.create(:ems_physical_infra) }
+    let(:invalid_change_password_payload) do
+      { "action"           => "change_password",
+        "current_password" => "current_password",
+        "new_password"     => "new_password" }
+    end
+
+    let(:valid_change_password_payload) do
+      { "action"           => "change_password",
+        "current_password" => "current_password",
+        "new_password"     => "new_password" }
+    end
+
+    let(:valid_change_password_payload_for_multiple_providers) do
+      {
+        "action"    => "change_password",
+        "resources" => [
+          {
+            "href"             => api_provider_url(nil, ems_physical_infra),
+            "current_password" => "current_password",
+            "new_password"     => "new_password"
+          },
+          {
+            "href"             => api_provider_url(nil, ems_physical_infra),
+            "current_password" => "current_password",
+            "new_password"     => "new_password"
+          }
+        ]
+      }
+    end
+
+    let(:invalid_change_password_payload_for_multiple_providers) do
+      {
+        "action"    => "change_password",
+        "resources" => [
+          {
+            "href"             => api_provider_url(nil, 999_999),
+            "current_password" => "current_password",
+            "new_password"     => "new_password"
+          },
+          {
+            "href"             => api_provider_url(nil, ems_physical_infra),
+            "current_password" => "current_password",
+            "new_password"     => "new_password"
+          }
+        ]
+      }
+    end
+
+    let(:single_resource_success_response) do
+      {
+        "message"   => a_string_matching(/Change password requested for Physical Provider #{ems_physical_infra.name}/i),
+        "href"      => api_provider_url(nil, ems_physical_infra),
+        "success"   => true,
+        "task_id"   => a_kind_of(String),
+        "task_href" => a_string_matching(api_tasks_url)
+      }
+    end
+
+    let(:single_resource_fail_response) do
+      {
+        "success" => false,
+        "message" => a_string_matching(/Couldn't find ExtManagementSystem with 'id'=999999/i)
+      }
+    end
+
+    context 'with a non existent provider' do
+      it 'returns a bad request error if no id is provided' do
+        api_basic_authorize collection_action_identifier(:providers, :change_password)
+
+        payload = {
+          "action" => "change_password",
+          "href"   => ""
+        }
+
+        post(api_providers_url, :params => payload)
+        expect_bad_request("Must specify an id for change password of a providers resource")
+      end
+
+      it 'returns a not found error' do
+        api_basic_authorize collection_action_identifier(:providers, :change_password)
+
+        post(api_provider_url(nil, 999_999), :params => valid_change_password_payload)
+
+        expect(response).to have_http_status(:not_found)
+      end
+
+      it 'returns a 200 code for multiple providers' do
+        api_basic_authorize collection_action_identifier(:providers, :change_password)
+
+        post(api_providers_url, :params => invalid_change_password_payload_for_multiple_providers)
+
+        expected = {
+          "results" => a_collection_containing_exactly(
+            a_hash_including(single_resource_fail_response), a_hash_including(single_resource_success_response)
+          )
+        }
+
+        expect(response).to have_http_status(:ok)
+        expect(response.parsed_body).to include(expected)
+      end
+    end
+
+    context 'with a valid request' do
+      before { allow_any_instance_of(ExtManagementSystem).to receive(:change_password) { true } }
+
+      it 'proccess the request successfully' do
+        api_basic_authorize collection_action_identifier(:providers, :change_password)
+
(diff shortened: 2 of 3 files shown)


abellotti:
I wonder if this shouldn't just be a POST edit for changing the password.

/cc @jntullo

douglasgabriel:
Maybe, but remember that this isn't just a database update, it also will trigger a call to provider's client to request a password change

jntullo:
@douglasgabriel if it is triggering a call to a provider's client, it should probably be queued. Is this an operation that takes some time?

douglasgabriel:
Hm, Got it! Not really @jntullo, is an instant operation

douglasgabriel:
What about the response from provider's client? In the UI project I have a form that uses this endpoint and print the provider's error message for user. I don't think that I can show the message for user if this operation be queued

jntullo:
@douglasgabriel when a task is queued, we return the `task_id` and `task_href` as part of the response so that a user can check on the status of the task

douglasgabriel:
Yes, I know. But a password change can fail for some reasons (authentication, password policy, password weakness and so on) because of that I think that is a good idea show to user some feedback about this failures, and we just can see which failure is if contact the provider. In the API I've already see this behavior of return a task, but I'm not sure what I can do in the UI page that use this endpoint. Did you understand my concern @jntullo ?

spec/requests/providers_spec.rb
```diff
@@ -1156,6 +1156,55 @@ def gen_import_request
...
+        api_basic_authorize collection_action_identifier(:providers, :change_password)
+
+        post(api_provider_url(nil, 999_999), :params => valid_change_password_payload)
+
+        expect(response).to have_http_status(:not_found)
+      end
+    end
+
+    context 'with a valid request' do
+      it 'proccess the request successfully' do
```
imtayadeway: Can you add a test case for bulk password change?

spec/requests/providers_spec.rb
```diff
@@ -1156,6 +1156,55 @@ def gen_import_request
...
+        allow_any_instance_of(ExtManagementSystem).to receive(:change_password) { true }
+
+        api_basic_authorize collection_action_identifier(:providers, :change_password)
+
+        post(api_provider_url(nil, ems_physical_infra), :params => valid_change_password_payload)
+
+        expect(response).to have_http_status(:ok)
```
imtayadeway: Can you add an expectation on the response body?

app/controllers/api/providers_controller.rb
```diff
@@ -230,5 +271,22 @@ def validate_provider_class
...
     end
+
+    # Checks if the change password request data is valid
+    #
+    # @raise [BadRequestError] if some required data is missing
+    #                          if new_password and confirm_password are differents
+    #
+    # @return [Boolean] true if all request data is  valid
+    def validate_change_password_request_data(data)
+      if data["current_password"].blank? ||
```
imtayadeway: This feels unfortunate as the API now has to have some domain  knowledge about what constitutes a valid password change. IMO it would be better to pass this data to the `change_password_queue` method and allow the task to reveal what went wrong. Alternatively, you could add a service to the domain model that validates password request changes, but it's important that it lives there and not in the API layer.

Some further thoughts: is it possible for the the request to fail if it passes validation? What are the cases in which it could fail? Ideally we'd have test cases illustrating all these paths.

app/controllers/api/providers_controller.rb
```diff
@@ -106,8 +106,49 @@ def resume_resource(type, id, _data)
...
+
     private
 
+    # Process password change request for a single resource
+    #
+    # @raise [BadRequestError] if no id is passed or some required data is missing
+    #                          or some error occur when try to change password on provider client
+    # @raise [NotFoundError]   if there isn't providers with the id passed
+    def change_password(type, id, data)
+      raise BadRequestError, "Must specify an id for change password of a #{type} resource" unless id
```
imtayadeway: I always have a hard time remembering how this path is reachable (IIRC we do some checking further up the stack) - can you add a test case for it, proving that it responded with this message?

app/controllers/api/providers_controller.rb
```diff
@@ -106,8 +106,49 @@ def resume_resource(type, id, _data)
...
+      validate_change_password_request_data(data)
+      api_action(type, id) do |klass|
+        provider = resource_search(id, type, klass)
+        desc = "Change password requested for Physical Provider #{provider.name}"
+        task_id = provider.change_password_queue(User.current_user.userid, data["current_password"], data["new_password"])
```
imtayadeway: Do all providers respond to this method? I can't seem to find it in the base class.

spec/requests/providers_spec.rb
```diff
@@ -1156,6 +1156,55 @@ def gen_import_request
...
+
+        post(api_provider_url(nil, 999_999), :params => valid_change_password_payload)
+
+        expect(response).to have_http_status(:not_found)
+      end
+    end
+
+    context 'with a valid request' do
+      it 'proccess the request successfully' do
+        allow_any_instance_of(ExtManagementSystem).to receive(:change_password) { true }
```
imtayadeway: Would it be possible to do this without `any_instance_of`? Have you looked into using webmock or something similar? Since this is an integration test, stubbing it at the network layer would give me greater confidence that everything is wired up correctly. And, since I had some concerns about all providers responding to this (above), that would give me greater confidence as a passer-by that we haven't stubbed a method that doesn't exist.

app/controllers/api/providers_controller.rb
douglasgabriel: Well, this validation is more about a data validation, not a domain. Here it validates if all data is provide by client and if it is well formatted (in this case, it require that the field `new_password` and `confirm_password` are equal, applying a "pattern" in change password routine, forcing the client to be sure what he is doing). The point is that this validation is more about the client/API communication and not about a domain rule, so I think its fair to have this code living here. Additionally, if you take a look on `change_password_queue` I just need to pass 2 params (`current_password` and `new_password`).

Yes, it is possible to fail if the data validation passes. To change password the data must be validate in the provide against its password policies, in failure case, the task will reveal the error.

I don't think that we can write specs for it, once it depends on provider's password rules.

app/controllers/api/providers_controller.rb
douglasgabriel: If you do a bulk request and not fill the `href` field. I don't think that it is verified previously, once I remove this verification no exception is thrown to the user. Anyway, I wrote a spec for it.

app/controllers/api/providers_controller.rb
douglasgabriel: No, for now, only Lenovo's provider respond to this method. You can see the implementation [here](https://github.com/ManageIQ/manageiq/pull/16895/files#diff-54a07cbbdbeffee51b848dc93f209776R62)

Fryguy:
@imtayadeway At a high-level does it make sense to implement this as a new action, or is it simpler to just have this work similar to other property changes with a PATCH request on the password property.  So I could see:

PATCH /api/providers/1234

```json
{
  "password": "new_password",
  "current_password": "current_password"
}
```

I realize this one is slightly different that a typical PATCH because current_password is not an existing property, which is why an action might be a better route.

app/controllers/api/providers_controller.rb
imtayadeway: This could be problematic - would it result in a 500 if a user targets the wrong provider? How do we reveal through the API what this can be used for?

spec/requests/providers_spec.rb
imtayadeway: To me, this is a big code smell - we've stubbed a method that doesn't exist on the target object. This makes me question the value of this test - is it just there to prove that the stub works?

This ties into https://github.com/ManageIQ/manageiq-api/pull/309#discussion_r182776204 - we can't smooth over those details. I want to see in another test what happens when the provider doesn't implement password change. As for this test, I would much rather that a) the specific lenovo provider is built for this test, and b) we stub at the network layer

imtayadeway:
> I realize this one is slightly different that a typical PATCH because current_password is not an existing property, which is why an action might be a better route.

@Fryguy yeah, that is my main concern with that approach, and why I think having an action is preferable. I think *ideally* this would have been modeled as a RESTful resource (i.e. you could CREATE a password change request on a provider), but since that isn't the way things have been done historically I think this is OK.

app/controllers/api/providers_controller.rb
imtayadeway: > (in this case, it require that the field new_password and confirm_password are equal, applying a "pattern" in change password routine, forcing the client to be sure what he is doing)

I disagree that this is not a domain concern - it's just one that hasn't been modeled. If some other part of the system decided to implement password changes, they'd need to duplicate this knowledge of what things are required, and how certain fields are supposed to match, etc...

There could be a number of ways to do this. One suggestion might be to wrap the password change in a service that lives outside of the API, that takes these fields and does the validation on them. Another could be that the service just does the validation, and the API makes the decision of whether to try to update the password based on the success of that. Either way, the API doesn't *know* what's required to make a password update, it's just going back and forth between the user and the domain model and responding appropriately.

douglasgabriel:
At the API level, the response will be the same of a successfully request, i.e. will return a task, to see any error, the user needs to follow the task status. I'm not sure if there is some test based on the task status. But this scenario is already covered on the `manageiq` core project, anyway.

spec/requests/providers_spec.rb
```diff
@@ -1156,6 +1156,145 @@ def gen_import_request
...
+            "href"             => api_provider_url(nil, ems_physical_infra),
+            "current_password" => "current_password",
+            "new_password"     => "new_password",
+            "confirm_password" => "new_password"
+          },
+          {
+            "href"             => api_provider_url(nil, ems_physical_infra),
+            "current_password" => "current_password",
+            "new_password"     => "new_password",
+            "confirm_password" => "new_password"
```
abellotti: why is there a confirm_password field in these examples ?  I expect this in a UI and not in an API.

douglasgabriel: Like the API should be an interface with some client, I think that is important to have this field here too, providing a well know interface to password change, but yeah, it is a good discussion, where in the previous that we have we decide to keep this field, do you think that I must remove that?

abellotti: yes please, should not be required at the API level, if the model method requires it, you can default that to new_password if not specified.

btw, where is change_password_queue method defined ?  I can only see the one in 
app/models/mixins/authentication_mixin.rb which as the following signature:

```
def change_password_queue(userid, current_password, new_password, auth_type = :default)
```

unless I'm not looking at the right thing.

douglasgabriel: You are looking at the right place, but this PR changes the method signature https://github.com/ManageIQ/manageiq/pull/17323

abellotti: but the signature of the change_password_queue method that we call in that PR has not changed. I don't see the confirm_password parameter there.

abellotti:
The one question I had regarding the signature, how we call change_password_queue seems to be different than app/models/mixins/authentication_mixin.rb's change_password_queue.

Also, if you could address Tim's change request.   Thanks.

douglasgabriel:
Oh, I understood your concern, sorry about the delay on it. I must remove the `confirm_password` param on the core project too.

About the @imtayadeway request, the result for a request against a provider that supports and one that doesn't is the same: it will return a task, and to see the failure the client must follow the task status, so I don't think that we do this kind of tests in the API (I mean follow the task status). Anyway, this scenario is already covered on [core project](https://github.com/ManageIQ/manageiq/blob/master/spec/models/mixins/authentication_mixin_spec.rb#L717)

douglasgabriel:
@abellotti could you take a look on this https://github.com/ManageIQ/manageiq/pull/17345  ?

abellotti:
Thanks @douglasgabriel for updating the method signature and the API Enhancement. LGTM !!  👍

PR #311, opened 2018-02-01
Return product features for all of a user's groups on the API entrypoint

jntullo:
This is needed so that the users can know what features they will have before changing groups. When they are a part of multiple groups, the current way of getting groups `/api/groups?attributes=miq_product_features` may not be available to them based off of their current groups permissions. This will allow them to see them.

It also refactors a bit as was necessary to use memoization and not have to create the product features for the current group multiple times.

Related: https://bugzilla.redhat.com/show_bug.cgi?id=1540733

cc: @AllenBW

### DIFF
--- a/app/controllers/api/api_controller.rb
+++ b/app/controllers/api/api_controller.rb
@@ -32,19 +32,17 @@ module Api
     end
 
     def auth_identity
-      user  = User.current_user
-      group = user.current_group
       {
-        :userid     => user.userid,
-        :name       => user.name,
-        :user_href  => "#{@req.api_prefix}/users/#{user.id}",
-        :group      => group.description,
-        :group_href => "#{@req.api_prefix}/groups/#{group.id}",
-        :role       => group.miq_user_role_name,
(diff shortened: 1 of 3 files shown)


abellotti:
the pre/post output size of the entrypoint goes from 11K to 250K or so.  This should be requested optionally.  We currently do provide the information for the user's current group via the authorization entry returned when we have a GET /api?attributes=authorization.  I think we should only return the group's product features if attributes=authorization is specified.  

One other concern is where we show this. Here we're merging "product_features" as part of the group info which is not an actual attribute/virtual, it's really group.miq_user_role.miq_product_features. might cause confusion (folks expecting to be able do GET /api/groups/:id?attributes=product_features). We can extend the "authorization" hash to include the per group info, or maybe instead of merging "product_features" in the group, we can call it "miq_product_features" and provide that in the model in groups (has_one :miq_product_features, :through => :miq_user_role), and make it virtual so folks can do a GET /api/groups/:id?attributes=miq_product_features.

abellotti:
Thanks @jntullo for the enhancement. LGTM!! 👍

PR #315, opened 2018-02-08
[WIP] Add transformation plans read and create

jntullo:
Very wip - still need to determine the payload for create of transformation plans

Dependent on https://github.com/ManageIQ/manageiq/pull/16960

cc: @bzwei @gmcculloug


jntullo:
As per conversation with @bzwei - going to shelf this for now and keep it as part of the service_templates controller.

PR #316, opened 2018-02-08
Allow ordering of service templates resource

jntullo:
For the new ServiceTemplateTransformationPlan, ordering must happen directly on the service template. Currently, the API only allows ordering through a service catalog. It was discussed with @bzwei that ordering can be allowed in general for a service template, and no special case is needed for the new `ServiceTemplateTransformationPlan` since there is validation around whether a service template is "orderable".

Sample order without required fields:
```
POST /api/service_templates/:id
{
   "action": "order"
}
```

Sample order with required fields:
```
POST /api/service_templates/:id
{
   "action": "order",
   "required_field_1": "filled in field"
}
```

### DIFF
--- a/config/api.yml
+++ b/config/api.yml
@@ -2487,6 +2487,8 @@
         :identifier: catalogitem_delete
       - :name: create
         :identifier: atomic_catalogitem_new
+      - :name: order
+        :identifier: svc_catalog_provision
     :subcollection_actions:
       :post:
       - :name: edit
@@ -2510,6 +2512,10 @@
         :identifier: catalogitem_edit
       - :name: delete
         :identifier: catalogitem_delete
+      - :name: order
+        :identifier: svc_catalog_provision
+        :options:
+        - :validate_action
       :delete:
       - :name: delete
         :identifier: catalogitem_delete
--- a/spec/requests/service_templates_spec.rb
+++ b/spec/requests/service_templates_spec.rb
@@ -445,4 +445,95 @@ describe "Service Templates API" do
       expect(response.parsed_body).to include(expected)
     end
   end
+
+  describe "Service Templates order" do
+    let(:service_template) { FactoryGirl.create(:service_template, :with_provision_resource_action_and_dialog, :orderable) }
+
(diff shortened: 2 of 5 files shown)


spec/requests/service_templates_spec.rb
```diff
@@ -445,4 +445,80 @@
...
+        api_basic_authorize action_identifier(:service_templates, :order, :resource_actions, :post)
+
+        post(api_service_templates_url, :params => { :action => "order", :resources => [{:href => api_service_template_url(nil, service_template)}] })
+
+        expected = {
+          "results" => [a_hash_including("href" => a_string_including(api_service_requests_url))]
+        }
+        expect(response).to have_http_status(:ok)
+        expect(response.parsed_body).to include(expected)
+      end
```
abellotti: let's add a test on the collection, ordering 2 templates. Thanks.

config/api.yml
```diff
@@ -2487,6 +2487,8 @@
         :identifier: catalogitem_delete
       - :name: create
         :identifier: atomic_catalogitem_new
+      - :name: order
+        :identifier: svc_catalog_provision
```
abellotti: don't we need the :validate_action option here too ?

jntullo: @abellotti it doesn't go on the collection, only on the resource, which is consistent with the other places we have it

jntullo:
Now dependent on https://github.com/ManageIQ/manageiq/pull/17023

config/api.yml
abellotti: 👍

abellotti:
Thanks @jntullo for the update and enhancement.  LGTM!! 👍

simaishi:

```
$ git log -1
commit c952f406381b13e463435946dd8a5d0f56a2ded5
Date:   Tue Feb 20 09:59:34 2018 -0500

    Merge pull request #316 from jntullo/service_template_order
    
    Allow ordering of service templates resource
    (cherry picked from commit 2de9dcc3cb04c20f01bee57cbe8de6b7a8b0f4c8)
```

PR #318, opened 2018-02-12
Do not return duplicate "edit" action entries

jntullo:
In the api.yml, some collections define the put and patch methods, while others do not although they are allowed by default due to the verb specification. In the case where the put and/or patch methods are defined, the edit actions were being duplicated. This ensures they are only present once each.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1540894


abellotti:
LGTM!! Thanks @jntullo for fixing this 👍

PR #320, opened 2018-02-14
Default section to "metadata" across custom attributes add action

jntullo:
When a new custom attribute is added, it defaults the `section` to "metadata" if it is not specified. In the same add action, it looks up to see if the custom attribute exists, but it does *not* default the section to metadata, causing it to not be found. This results in a *new* custom attribute with the same name, if the same request is made. Since the behavior is to *edit* instead of add, this is not consistent behavior. Instead of creating a new custom attribute, it should be able to find the custom attribute and edit it. 

example:
```
POST /api/providers/:id/custom_attributes
{
  "action" : "add",
  "resources" : [
    { "name" : "koko", "value" : "loko" }
  ]
}
```
on the first try, would create a new custom attribute "koko", as expected.

Making the same request:
```
POST /api/providers/:id/custom_attributes
{
  "action" : "add",
  "resources" : [
    { "name" : "koko", "value" : "moko" }
  ]
}
```
It would add *another* custom attribute "koko", but it should actually just update the value to "moko".

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1544800

### DIFF
--- a/app/controllers/api/subcollections/custom_attributes.rb
+++ b/app/controllers/api/subcollections/custom_attributes.rb
@@ -7,6 +7,7 @@ module Api
 
       def custom_attributes_add_resource(object, _type, _id, data = nil)
         raise BadRequestError, "#{object.class.name} does not support management of custom attributes" unless object.respond_to?(:custom_attributes)
+        data["section"] ||= "metadata"
         add_custom_attribute(object, data)
       rescue => err
         raise BadRequestError, "Could not add custom attributes - #{err}"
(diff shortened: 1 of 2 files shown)


abellotti:
LGTM!! Thanks @jntullo for the fix. 👍

simaishi:

```
$ git log -1
commit deb7a2bd83dee986ae77a7c268fe6a8a0fe6f117
Date:   Wed Feb 14 08:40:45 2018 -0500

    Merge pull request #320 from jntullo/custom_attrs_create
    
    Default section to "metadata" across custom attributes add action
    (cherry picked from commit 2ef6cfe981f78c56443088685c9b21411db55bec)
    
    Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1552824
```

PR #321, opened 2018-02-14
Make cloud_tenants and flavors subcollections consistent with others

jntullo:
Infra providers do not have `cloud_tenants` or `flavors`, causing a 500 error to be returned when they are requested. This makes the behavior consistent with other subcollections, checking to see if the object responds to the subcollection. 

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1545240


abellotti:
That should do it 👍   Thanks @jntullo

simaishi:
@abellotti Can this be `gaprindashvili/yes` ?

PR #324, opened 2018-02-16
Only allow one cart to be created at a time

jntullo:
This ensures that the user is assigned before saving, to ensure that only one cart is created at a time.

Intended to be fixed by https://github.com/ManageIQ/manageiq/pull/16351, but failed QA because the API is not setting a user. 

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1493788


app/controllers/api/service_orders_controller.rb
```diff
@@ -60,6 +60,15 @@ def add_request_to_cart(workflow)
       workflow.add_request_to_cart
     end
 
+    def create_service_order(data)
+      ServiceOrder.new(data).tap do |service_order|
+        service_order.assign_user
+        service_order.save!
+      end
```
abellotti: can't the above be replaced with ServiceOrder.create(data) ?  we might get assign_user with that as it seems to be a before create action.

jntullo: @abellotti that's what I was hoping for, but unfortunately the assign_user happens after the validation. I played around with the model a bit but was unable to update it without breaking anything else. I can play around with it more if you would prefer that route.

abellotti: Ahh ok, no worries. this is ok for now.

simaishi:
@abellotti  Can this be `gaprindashvili/yes` ?

abellotti:
This can go to gaprindashvili, but it may also require https://github.com/ManageIQ/manageiq/pull/16351

simaishi:
https://github.com/ManageIQ/manageiq/pull/16351 is already in gaprindashvili (PR was merged before branch was created), so we're good there.  Thanks @abellotti !

PR #334, opened 2018-03-01
Add pause and resume actions to Providers

jntullo:
Actual method names are `enable!` and `disable!`, but actions are "pause" and "resume" to be consistent with the UI

https://bugzilla.redhat.com/show_bug.cgi?id=1507812


abellotti:
Thanks @jntullo for the API Enhancement. 👍

PR #336, opened 2018-03-05
[WIP] Added an API Log request parser

abellotti:
Added an API Log request parser
 
 Parse API Requests from the log file specified

  If file is not specified, the default log/api.log is parsed.

  Usage:
    api_log_parser [options] [file]

```
  api_log_parser options are:
  -s, --sort-by=<s>    Sort by field (default: request.requested_at)
  -r, --reverse        Sort in descending order
  -f, --filter=<s+>    Filter requests with specific field values
  -o, --format=<s>     How to format output, [json | none] (default: json)
  -h, --help           Show this message
```

[skip ci]

### DIFF
--- a/exe/api_log_parser
+++ b/exe/api_log_parser
@@ -1,0 +1,138 @@
+#!/usr/bin/env bundle exec ruby
+RAILS_ROOT = File.expand_path(File.join(__dir__, "..", "..", "manageiq"))
+
+require 'manageiq-gems-pending'
+require 'active_support'
+require 'miq_logger_processor'
+require 'trollop'
+require 'json'
+require 'active_support/core_ext'   # deep_symbolize_keys
+require 'more_core_extensions/all'  # fetch_path
+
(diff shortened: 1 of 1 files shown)


abellotti:
@gtanzillo WIP, let's kick the tire and refine. Thanks!!

exe/api_log_parser
```diff
@@ -0,0 +1,143 @@
+#!/usr/bin/env bundle exec ruby
```
imtayadeway: I've never seen a shebang declared in this way...did you try using the more typical `#!/usr/bin/env ruby` and following the steps outlined in https://bundler.io/v1.16/guides/bundler_setup.html ? That way you may also be able to tag only the gems you need so that only they are required when you setup bundle - which should improve the load time significantly.

abellotti:
maybe with something like --filter=authentication.token=nil  or some variation.

PR #338, opened 2018-03-06
Folders subcollection on providers

jntullo:
Part 1 of https://bugzilla.redhat.com/show_bug.cgi?id=1472279

### DIFF
--- a/config/api.yml
+++ b/config/api.yml
@@ -1122,6 +1122,12 @@
       :get:
       - :name: read
         :identifier: floating_ip_show
+  :folders:
+    :description: Folders
+    :options:
+    - :subcollection
+    :verbs: *g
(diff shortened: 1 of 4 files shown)


config/api.yml
```diff
@@ -1122,6 +1122,13 @@
       :get:
       - :name: read
         :identifier: floating_ip_show
+  :folders:
+    :description: Datacenters
+    :identifier: datacenter_controller
```
abellotti: should description be Folders ?  Datacenter.all.collect(&:name) is a subset than EmsFolder.all.collect(&:name)

jntullo: @abellotti yup 😆 I started out doing datacenters...didn't change it to folders

PR #345, opened 2018-03-12
Allow region to be queried by the region number

jntullo:
Add in the ability to look up a region by its region number, ie `10` rather than its actual id from the database `10000000000001`

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1552899

### DIFF
--- a/app/controllers/api/regions_controller.rb
+++ b/app/controllers/api/regions_controller.rb
@@ -1,4 +1,11 @@
 module Api
   class RegionsController < BaseController
+    private
+
+    def resource_search(id, type, klass)
+      region = MiqRegion.find_by(:region => id)
+      return region unless region.nil?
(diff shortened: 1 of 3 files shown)


app/controllers/api/regions_controller.rb
```diff
@@ -1,4 +1,11 @@
 module Api
   class RegionsController < BaseController
+    private
+
+    def resource_search(id, type, klass)
+      region = MiqRegion.find_by(:region => id)
+      return region unless region.nil?
+      super(id, type, klass)
```
abellotti: Hmm, might be problematic GET /api/regions/:id is now non-deterministic.  I think we need to go all one way or the other, href's generated are using region numbers and not record id's.  

take a look at resource_identifier for automate_workspaces. might be helpful.

PR #348, opened 2018-03-15
Assign and Unassign alert definition profiles

jntullo:
Had to deviate slightly from the signature defined [here](https://github.com/ManageIQ/manageiq-api/pull/177#issuecomment-361722583) because it was not taking into account the alert definition profile resource. Tag references accept both `href` and `name` (name of the classification)

```
POST /api/alert_definition_profiles
{  
   "action":"assign",
   "resources":[  
      {  
         "id":"<alert_definition_profile_id>",
         "resources":[  
            {  
               "href":"/api/hosts",
               "tag":{  
                  "href":"/api/tags/:id"
               }
            },
            {  
               "href":"/api/hosts",
               "tag":{ 
                   "category":"<category name>",
                  "name":"<tag name>"
               }
            },
            {  
               "href":"/api/clusters/:id"
            }
         ]
      }
   ]
}
```

```
POST /api/alert_definition_profiles/:id
{  
   "action":"assign",
   "resources":[  
      {  
         "href":"/api/hosts",
         "tag":{  
            "href":"/api/tags/:id"
         }
      },
      {  
         "href":"/api/clusters/:id"
      }
   ]
}
```

```
POST /api/alert_definition_profiles
{  
   "action":"unassign",
   "resources":[  
      {  
         "id":"<alert_definition_profile_id>",
         "resources":[  
            {  
               "href":"/api/hosts",
               "tag":{  
                  "href":"/api/tags/:id"
               }
            },
            {  
               "href":"/api/hosts",
               "tag":{  
                   "category":"<category name>",
                  "name":"<tag name>"
               }
            },
            {  
               "href":"/api/clusters/:id"
            }
         ]
      }
   ]
}
```

```
POST /api/alert_definition_profiles/:id
{  
   "action":"unassign",
   "resources":[  
      {  
         "href":"/api/hosts",
         "tag":{  
            "href":"/api/tags/:id"
         }
      },
      {  
         "href":"/api/clusters/:id"
      }
   ]
}
```

BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1539379

### DIFF
--- a/app/controllers/api/alert_definition_profiles_controller.rb
+++ b/app/controllers/api/alert_definition_profiles_controller.rb
@@ -27,9 +27,7 @@ module Api
       data['resources'].each do |resource|
         href = Href.new(resource['href'])
         if resource.key?('tag')
-          tag_id = Href.new(resource['tag']['href']).subject_id
-          tag = resource_search(tag_id, :tags, collection_class(:tags))
-          profile.assign_to_tags([tag.classification], href.subject)
+          profile.assign_to_tags([fetch_tag_classification_resource(resource['tag'])], href.subject)
         else
           assignable_resource = resource_search(href.subject_id, href.subject, collection_class(href.subject))
           profile.assign_to_objects([assignable_resource])
@@ -45,9 +43,7 @@ module Api
       data['resources'].each do |resource|
         href = Href.new(resource['href'])
         if resource.key?('tag')
-          tag_id = Href.new(resource['tag']['href']).subject_id
-          tag = resource_search(tag_id, :tags, collection_class(:tags))
-          profile.unassign_tags([tag.classification], href.subject)
+          profile.unassign_tags([fetch_tag_classification_resource(resource['tag'])], href.subject)
         else
           assignable_resource = resource_search(href.subject_id, href.subject, collection_class(href.subject))
           profile.unassign_objects([assignable_resource])
@@ -62,6 +58,12 @@ module Api
 
     def alert_definition_profile_ident(profile)
       "Alert Definition Profile id:#{profile.id} name:'#{profile.name}'"
+    end
+
+    def fetch_tag_classification_resource(data)
+      tag_id = Href.new(data['href']).subject_id
+      tag = resource_search(tag_id, :tags, collection_class(:tags))
+      tag.classification
     end
   end
 end
--- a/app/controllers/api/alert_definition_profiles_controller.rb
+++ b/app/controllers/api/alert_definition_profiles_controller.rb
@@ -40,6 +40,24 @@ module Api
       action_result(false, "Could not assign Alert Definition Profile - #{err}")
     end
 
+    def unassign_resource(type, id, data)
+      profile = resource_search(id, type, collection_class(type))
+      data['resources'].each do |resource|
+        href = Href.new(resource['href'])
+        if resource.key?('tag')
+          tag_id = Href.new(resource['tag']['href']).subject_id
+          tag = resource_search(tag_id, :tags, collection_class(:tags))
+          profile.unassign_tags([tag.classification], href.subject)
+        else
+          assignable_resource = resource_search(href.subject_id, href.subject, collection_class(href.subject))
+          profile.unassign_objects([assignable_resource])
+        end
+      end
+      action_result(true, "Unassigned resources from #{alert_definition_profile_ident(profile)}")
+    rescue => err
+      action_result(false, "Could not unassign Alert Definition Profile - #{err}")
+    end
+
     private
 
     def alert_definition_profile_ident(profile)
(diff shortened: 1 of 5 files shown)


app/controllers/api/alert_definition_profiles_controller.rb
```diff
@@ -21,5 +21,55 @@ def edit_resource(type, id = nil, data = {})
...
+      "Alert Definition Profile id:#{profile.id} name:'#{profile.name}'"
+    end
+
+    # rubocop:disable Rails/DynamicFindBy
+    def fetch_tag_classification_resource(data)
+      if data.key?('href') || data.key?('id')
+        tag = resource_search(parse_id(data, :tags), :tags, collection_class(:tags))
+        return tag.classification
+      elsif data.key?('name')
+        return Classification.find_by_name(data['name'])
```
jntullo: ignoring the rubocop error for this because it is an actual class method

cben:
Does referencing tag by name use Classification#name or the Tag#name full path  `/managed/category/name`?
Seems to be just the name part.  Inconsistent with tags collection API that gives full path under "name":
http://manageiq.org/docs/reference/fine/api/reference/tagging#querying-tags

cben:
OTOH, assigning tags takes either "category" + "name" or "path":
http://manageiq.org/docs/reference/fine/api/reference/tagging#tag-specification
"path" is much better when refering to /managed/cat/name, I wish we used that terminology consistently.
Anyway now the tags API is already inconsistent with itself, need to pick one :-(

Hmm, if I read this PR correctly when referring to tag by name, you always want a category, not a single tag, right?  If so how about using `{"tag": {"category": "..."}}`?

cc @elad661

cben:
OK, Elad helped me find the core code: https://github.com/ManageIQ/manageiq/blob/d99daaac2/app/models/mixins/assignment_mixin.rb#L225-L235
=> Assignment to tags works with an single entry (Classification instance or id), not a whole category.
(This also makes sense — "assign all things tagged Foo : Bar".  The alternative "assign all things with any tag under Foo category" would be less useful.)

So I suppose this should take `{"tag": {"category": "department", "name": "finance"}}` pairs.
And ideally `{"tag": {"path" : "/managed/department/finance"}}` too.

jntullo:
@cben updated to use `{"tag": {"category": "department", "name": "finance"}}` and `{"tag": {"href": "<tag_href>"}}` or the tag id

app/controllers/api/mixins/tags.rb
```diff
@@ -0,0 +1,41 @@
...
+        parse_tag(data)
+      end
+
+      def parse_tag(data)
+        return {} if data.blank?
+
+        category = data["category"]
+        name     = data["name"]
+        return {:category => category, :name => name} if category && name
+        return tag_path_to_spec(name) if name && name[0] == '/'
```
cben: Side note [was like this, out of scope to this PR]: this differs from documentation.
http://manageiq.org/docs/reference/fine/api/reference/tagging#tag-specification says you give full path in "path" key.  This code takes full path in "name" (if "category" not given).
Dunno if right solution is to fix doc or code...

app/controllers/api/mixins/tags.rb
```diff
@@ -0,0 +1,41 @@
...
+                 klass = collection_class(:tags)
+                 klass.find(href.split('/').last)
+               end
+        tag.present? ? tag_path_to_spec(tag.name).merge(:id => tag.id) : {}
+      end
+
+      def tag_path_to_spec(path)
+        tag_path = path[0..7] == Api::BaseController::TAG_NAMESPACE ? path[8..-1] : path
+        parts    = tag_path.split('/')
+        {:category => parts[1], :name => parts[2]}
```
cben: Side note [was like this, out of scope to this PR]:
this ignores the namespace, meaning you can only specify /managed tags, even if you give full path or href.
if you specify a tag in another namespace you'll either get wrong tag or nothing/error (didn't test)...

app/controllers/api/alert_definition_profiles_controller.rb
```diff
@@ -60,10 +63,13 @@ def alert_definition_profile_ident(profile)
       "Alert Definition Profile id:#{profile.id} name:'#{profile.name}'"
     end
 
+    # rubocop:disable Rails/DynamicFindBy
     def fetch_tag_classification_resource(data)
-      tag_id = Href.new(data['href']).subject_id
-      tag = resource_search(tag_id, :tags, collection_class(:tags))
-      tag.classification
+      tag_spec = tag_specified(data["id"], data)
```
abellotti: 👍 tag_specified method as common way we parse tag specs in API.

jntullo:
@cben it does not return it by default, but they can be expanded via `GET /api/resources/alert_definition_profiles?expand=resources&attributes=taggings` or `GET /api/resources/alert_definition_profiles/:id?attributes=taggings`

PR #356, opened 2018-03-28
Enhanced API to catch Settings validation errors

abellotti:
Enhanced API to catch Settings validation errors and reporting accordingly to the client via a BadRequestError.

### DIFF
--- a/app/controllers/api/base_controller.rb
+++ b/app/controllers/api/base_controller.rb
@@ -117,7 +117,7 @@ module Api
       case @req.method
       when :patch
         raise ForbiddenError, "You are not authorized to edit settings." unless super_admin?
-        resource.add_settings_for_resource(@req.json_body)
+        resource_update_settings(resource, @req.json_body)
       when :delete
         raise ForbiddenError, "You are not authorized to remove settings." unless super_admin?
         resource.remove_settings_path_for_resource(*@req.json_body)
@@ -140,6 +140,12 @@ module Api
       else
         raise ForbiddenError, "You are not authorized to view settings."
       end
+    end
+
+    def resource_update_settings(resource, updated_settings)
+      resource.add_settings_for_resource(updated_settings)
+    rescue VMDB::Config::Validator::Error => err
+      raise BadRequestError, "Settings validation failed - #{err}"
     end
 
     def current_user
--- a/app/controllers/api/base_controller.rb
+++ b/app/controllers/api/base_controller.rb
@@ -117,7 +117,11 @@ module Api
       case @req.method
       when :patch
         raise ForbiddenError, "You are not authorized to edit settings." unless super_admin?
-        resource_update_settings(resource, @req.json_body)
+        begin
+          resource.add_settings_for_resource(@req.json_body)
+        rescue Vmdb::Settings::ConfigurationInvalid => err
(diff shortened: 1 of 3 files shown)


abellotti:
Marking as WIP for now, tests will fail until we update/sync up with @fryguy's magic settings fix.

spec/requests/settings_spec.rb
```diff
@@ -3,6 +3,25 @@
...
 describe "Settings API" do
   let(:api_settings) { Api::ApiConfig.collections[:settings][:categories] }
+  let(:server) { FactoryGirl.create(:miq_server) }
+  let(:super_admin) { FactoryGirl.create(:user, :role => 'super_administrator', :userid => 'alice', :password => 'alicepassword') }
+
+  context "Settings Update" do
+    it "updates to settings return a BadRequestError upon failed validation" do
+      api_basic_authorize(:user => super_admin.userid, :password => super_admin.password)
+
+      patch(api_server_settings_url(nil, server), :params => {:authentication => {:mode => "bogus_auth_mode"}})
```
imtayadeway: The other server settings specs are in `spec/requests/servers_spec.rb` - would it make sense for this to live there too?

app/controllers/api/base_controller.rb
```diff
@@ -117,7 +117,7 @@ def settings
       case @req.method
       when :patch
         raise ForbiddenError, "You are not authorized to edit settings." unless super_admin?
-        resource.add_settings_for_resource(@req.json_body)
+        resource_update_settings(resource, @req.json_body)
```
imtayadeway: I'm not opposed to extracting a method here, but I think its name should reflect the responsibilit(y|ies) of the new method - here's it's to try an update and raise on a validation error. I'm having a hard time trying to find a succinct way to say that, so maybe inline for now?

spec/requests/settings_spec.rb
abellotti: starting going that way, but then realizing I'm really testing a settings update handling that's generic to all and it didn't make sense to duplicate for servers, zone, region, so I put that here.

app/controllers/api/base_controller.rb
abellotti: wasn't crazy about have the begin/rescue inside the when block, (you never know when rubocop starts complaining about complexity 😉 ) so I created the separate method. I can go either way really. up to you.

imtayadeway: I think my preference is for in-line :+1:

spec/requests/settings_spec.rb
abellotti: ahh good point, I'll move it the proper file.

abellotti:
@Fryguy PR updated following https://github.com/ManageIQ/manageiq/pull/17247   Thanks!!

abellotti:
```
PATCH /api/servers/1/settings

{
  "authentication" : { "bind_timeout" : 123 }
}
```

now gives us:

```
{
  "error": {
    "kind": "bad_request",
    "message": "Settings validation failed - authentication-mode: authentication mode, nil, is invalid. Should be one of: amazon, ldap, ldaps, httpd, database, none",
  }
}
```

abellotti:
poke @imtayadeway

PR #359, opened 2018-04-05
Always challenge a user with no/bad credentials

imtayadeway:
According to RFC 2617:

> The 401 (Unauthorized) response message is used by an origin server
  to challenge the authorization of a user agent. This response MUST
  include a WWW-Authenticate header field containing at least one
  challenge applicable to the requested resource.

We currently only issue a challenge when no credentials are provided -
if a user fails authentication they are not challenged again. This
broken implementation can cause some frustration for instance when
using the web browser, which will cache a mis-typed user/password
combination and not challenge the user again.

This is somewhat problematic as we only use the `Authorization` header
for Basic Authentication. It seems like a safe compromise for the user
to be challenged to use Basic Authentication if authentication fails
for whatever reason.

### DIFF
--- a/app/controllers/api/base_controller.rb
+++ b/app/controllers/api/base_controller.rb
@@ -37,8 +37,6 @@ module Api
     rescue_from(ActiveRecord::StatementInvalid) { |e| api_error(:bad_request, e) }
     rescue_from(JSON::ParserError)              { |e| api_error(:bad_request, e) }
     rescue_from(MultiJson::LoadError)           { |e| api_error(:bad_request, e) }
-    rescue_from(MiqException::MiqEVMLoginError) { |e| api_error(:unauthorized, e) }
-    rescue_from(AuthenticationError)            { |e| api_error(:unauthorized, e) }
     rescue_from(ForbiddenError)                 { |e| api_error(:forbidden, e) }
     rescue_from(BadRequestError)                { |e| api_error(:bad_request, e) }
     rescue_from(NotFoundError)                  { |e| api_error(:not_found, e) }
--- a/app/controllers/api/base_controller/authentication.rb
+++ b/app/controllers/api/base_controller/authentication.rb
@@ -31,6 +31,7 @@ module Api
         api_log_error("AuthenticationError: #{e.message}")
         response.headers["Content-Type"] = "application/json"
         request_http_basic_authentication("Application", ErrorSerializer.new(:unauthorized, e).serialize.to_json)
+        log_api_response
       end
 
       def user_settings
--- a/app/controllers/api/base_controller/authentication.rb
+++ b/app/controllers/api/base_controller/authentication.rb
@@ -14,11 +14,8 @@ module Api
         else
           success = authenticate_with_http_basic do |u, p|
             begin
-              user = User.authenticate(
-                u, p, request,
-                :require_user => true,
-                :timeout      => ::Settings.api.authentication_timeout.to_i_with_method
-              )
+              timeout = ::Settings.api.authentication_timeout.to_i_with_method
+              user = User.authenticate(u, p, request, :require_user => true, :timeout => timeout)
               auth_user_obj = userid_to_userobj(user.userid)
               authorize_user_group(auth_user_obj)
               validate_user_identity(auth_user_obj)
--- a/app/controllers/api/base_controller/authentication.rb
+++ b/app/controllers/api/base_controller/authentication.rb
@@ -31,6 +31,7 @@ module Api
         end
         log_api_auth
       rescue AuthenticationError => e
+        api_log_error("AuthenticationError: #{e.message}")
         response.headers["Content-Type"] = "application/json"
         request_http_basic_authentication("Application", ErrorSerializer.new(:unauthorized, e).serialize.to_json)
       end
--- a/app/controllers/api/base_controller/authentication.rb
+++ b/app/controllers/api/base_controller/authentication.rb
@@ -12,22 +12,26 @@ module Api
         elsif request.headers[HttpHeaders::AUTH_TOKEN]
           authenticate_with_user_token(request.headers[HttpHeaders::AUTH_TOKEN])
         else
-          authenticate_options = {
-            :require_user => true,
-            :timeout      => ::Settings.api.authentication_timeout.to_i_with_method
-          }
-
-          if (user = authenticate_with_http_basic { |u, p| User.authenticate(u, p, request, authenticate_options) })
-            auth_user_obj = userid_to_userobj(user.userid)
-            authorize_user_group(auth_user_obj)
-            validate_user_identity(auth_user_obj)
-            User.current_user = auth_user_obj
-          else
-            request_http_basic_authentication
-            return
-          end
+          authenticate_with_http_basic do |u, p|
+            begin
+              user = User.authenticate(
+                u, p, request,
+                :require_user => true,
+                :timeout      => ::Settings.api.authentication_timeout.to_i_with_method
+              )
+              auth_user_obj = userid_to_userobj(user.userid)
+              authorize_user_group(auth_user_obj)
+              validate_user_identity(auth_user_obj)
+              User.current_user = auth_user_obj
+            rescue MiqException::MiqEVMLoginError => e
+              raise AuthenticationError, e.message
+            end
+          end or raise AuthenticationError
         end
         log_api_auth
+      rescue AuthenticationError => e
+        response.headers["Content-Type"] = "application/json"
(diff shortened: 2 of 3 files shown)


imtayadeway:
Ignoring rubocop as its suggestion is not valid Ruby

app/controllers/api/base_controller/authentication.rb
```diff
@@ -12,22 +12,26 @@ def require_api_user_or_token
...
+                :timeout      => ::Settings.api.authentication_timeout.to_i_with_method
+              )
+              auth_user_obj = userid_to_userobj(user.userid)
+              authorize_user_group(auth_user_obj)
+              validate_user_identity(auth_user_obj)
+              User.current_user = auth_user_obj
+            rescue MiqException::MiqEVMLoginError => e
+              raise AuthenticationError, e.message
+            end
+          end or raise AuthenticationError
```
Fryguy: why the `or`?  that's generally a no-no

imtayadeway: That's just a Perl-ism - `|| raise` isn't valid Ruby, so the alternative is to make this (perhaps) less readable by assigning to a variable that's only used to check that this is non-nil

Fryguy: since the "guts" of the other legs of the if clause are methods, can we make this a method as well?  Maybe `authenticate_with_basic_auth`?

app/controllers/api/base_controller/authentication.rb
```diff
@@ -12,22 +12,26 @@ def require_api_user_or_token
...
         end
         log_api_auth
```
Fryguy: Strangely this only logged successful logins, but do we log unsuccessful logins anywhere?

app/controllers/api/base_controller/authentication.rb
imtayadeway: Yeah I considered this, but then it's not clear to me at a glance how that method (or any similarly-named method following the scheme above) differs crucially from the one from Rails' Authentication library (`authenticate_with_http_basic`) so I felt that might be better dealt with in a follow up?

Fryguy: honestly, I find the variable more readable...or use tap (I do find tap less readable).

See also https://github.com/bbatsov/ruby-style-guide#no-and-or-or

`||` is valid Ruby, so not sure why you think it's not...maybe you are using it without parenthesis?

```ruby
[1] pry(main)> false || raise("blah")
RuntimeError: blah
from (pry):20:in `<main>'
[2] pry(main)> false || raise "blah"
SyntaxError: unexpected tSTRING_BEG, expecting keyword_do or '{' or '('
false || raise "blah"
                ^
```

app/controllers/api/base_controller/authentication.rb
imtayadeway: We have never logged an unsuccessful basic authentication login....it looks like we might lose some logging around the other cases. I'll try to rectify that

app/controllers/api/base_controller/authentication.rb
imtayadeway: Ah yes, I must have misinterpreted that error, but I find having to know that that `raise` requires parentheses perhaps a bit arcane. I'll change to use the variable.

FWIW I prefer Avdi's take on the no `or`/`and` rule :wink:

app/controllers/api/base_controller/authentication.rb
```diff
@@ -12,22 +12,25 @@ def require_api_user_or_token
...
+              raise AuthenticationError, e.message
+            end
           end
+          raise AuthenticationError unless success
         end
         log_api_auth
+      rescue AuthenticationError => e
+        api_log_error("AuthenticationError: #{e.message}")
+        response.headers["Content-Type"] = "application/json"
+        request_http_basic_authentication("Application", ErrorSerializer.new(:unauthorized, e).serialize.to_json)
```
abellotti: We're missing the logging of the 401,

before this PR, we get the following:

```
[----] I, [2018-04-16T17:48:16.502716 #26595:3fdeb52eb958]  INFO -- : MIQ(Api::ApiController.log_request_initiated)  
[----] I, [2018-04-16T17:48:16.502900 #26595:3fdeb52eb958]  INFO -- : MIQ(Api::ApiController.log_request) API Request:    {:requested_at=>"2018-04-16 21:48:16 UTC", :method=>"GET", :url=>"http://localhost:3000/api"}
[----] E, [2018-04-16T17:48:16.571327 #26595:3fdeb52eb958] ERROR -- : MIQ(Api::ApiController.api_error) API Error
[----] E, [2018-04-16T17:48:16.571381 #26595:3fdeb52eb958] ERROR -- : MIQ(Api::ApiController.api_error) MiqException::MiqEVMLoginError: Authentication failed
[----] I, [2018-04-16T17:48:16.571986 #26595:3fdeb52eb958]  INFO -- : MIQ(Api::ApiController.log_request) Response:       {:completed_at=>"2018-04-16 21:48:16 UTC", :size=>"0.108 KBytes", :time_taken=>"0.069 Seconds", :status=>401}
```

The last Response: entry is missing.  This stuff is handled by api_error, so instead of adding more of that logic here, maybe the above can be as follows:

```
rescue AuthenticationError => e
   api_error(:unauthorized, e)
   request_http_basic_authentication("Application", ErrorSerializer.new(:unauthorized, e).serialize.to_json)
end
```

Or possibly get api_error to return to the response generated

```
rescue AuthenticationError => e
   request_http_basic_authentication("Application", api_error(:unauthorized, e).to_json)
end
```

app/controllers/api/base_controller.rb
```diff
@@ -37,8 +37,6 @@ class BaseController < ActionController::API
     rescue_from(ActiveRecord::StatementInvalid) { |e| api_error(:bad_request, e) }
     rescue_from(JSON::ParserError)              { |e| api_error(:bad_request, e) }
     rescue_from(MultiJson::LoadError)           { |e| api_error(:bad_request, e) }
-    rescue_from(MiqException::MiqEVMLoginError) { |e| api_error(:unauthorized, e) }
-    rescue_from(AuthenticationError)            { |e| api_error(:unauthorized, e) }
```
abellotti: I see AuthenticationError raised from within this file so that's ok to remove. I'm not sure about MiqEVMLoginError in the other 2 token based auth.  Should be verified. If that error can be raised, we'd need to rescue MiqException::MiqEVMLoginError for those cases too.

app/controllers/api/base_controller/authentication.rb
imtayadeway: Interesting. It looks like `request_http_basic_authentication` is short-circuiting the `after_action` that invokes this - it's not specific to `api_error`. Will look into it.

app/controllers/api/base_controller.rb
imtayadeway: I think this is OK because:

1. I see tests that cover the happy/unhappy paths around authenticating with tokens
2. System token authentication does a catch-all rescue, and that specific error inherits from `StandardError`
3. User token authentication doesn't do any actual logging-in as such, it just talks to the token manager

app/controllers/api/base_controller/authentication.rb
imtayadeway: Addressed in https://github.com/ManageIQ/manageiq-api/pull/359/commits/9fe4247787aaa858a1cb9ddfc2413f8ca12318ee

abellotti: Thanks @imtayadeway for the update.

abellotti:
LGTM!!  Thanks @imtayadeway for fixing this.  👍

himdel:
(summarizing gitter :point_up: [April 24, 2018 3:24 PM](https://gitter.im/ManageIQ/api?at=5adf4c461130fe3d360ff59f))

Turns out the change in `app/controllers/api/base_controller/authentication.rb` changes what happes when `validate_user_identity` fails.

Before, it would not prevent the user from logging in, but now that it is being run *inside* a `authenticate_with_http_basic do ...` block, it does.

Turns out logging in as a newly created user fails with `User's Group is missing` (#366) because their `current_group` is not set yet.

himdel:
Looks like this is a problem...

`DELETE /api/auth` with `x-auth-token` will request HTTP Basic Authentication if that token already expired.
So, trying to log out with an expired token causes the popup.

There seems to be no way of disabling the popup on the browser side, so this may need to be undone.
(Or the UI needs to keep track of the token validity, which seems too error-prone.)

himdel:
Login dialog popup problem addressed in https://github.com/ManageIQ/manageiq-api/pull/488

Issue #366, opened 2018-04-24
Unable to login with a newly created user


mzazrivec:
How to reproduce:
1. In classic UI, create a new role, new group and a new user
2. try to login into the UI with the new user

You'll see the login fails.

During login, the following request is sent to API:
```
I, [2018-04-24T17:23:31.655840 #31920]  INFO -- : Started GET "/api/auth?requester_type=ui" for ::1 at 2018-04-24 17:23:31 +0200
I, [2018-04-24T17:23:31.748765 #31920]  INFO -- : Processing by Api::AuthController#show as JSON
I, [2018-04-24T17:23:31.749126 #31920]  INFO -- :   Parameters: {"requester_type"=>"ui"}
I, [2018-04-24T17:23:31.841456 #31920]  INFO -- : Filter chain halted as :require_api_user_or_token rendered or redirected
I, [2018-04-24T17:23:31.841585 #31920]  INFO -- : Completed 401 Unauthorized in 92ms (ActiveRecord: 10.2ms)
```

The json that comes back:
```
{"error":{"kind":"unauthorized","message":"Invalid User user01 specified, User's Group is missing","klass":"Api::AuthenticationError"}}
```

Inretestingly, logging in with `admin` user works fine.

imtayadeway:
@mzazrivec I know that conversations about this issue have occurred variously on gitter etc.. - can you provide a few details here to help consolidate what we know;

1. What's the expected behavior?
1. Was there a time when the API performed the expected behavior, and
1. If so, can you identify a revision/release when the product did the right thing?

If you can also clarify the relationships between your user/role/groups in setup steps above?

I don't think I have a complete picture of what is going on - I think it was speculated on gitter that the reason it fails is because the user doesn't yet have a current group. I'm convinced this isn't the issue - we have had tests around this case which haven't changed in a long time - certainly predating the change in #359 .

Thanks!

himdel:
[WIP answer, still looking for some details...]

@imtayadeway @mzazrivec is on PTO, I'll try to provide the details...


> 1. What's the expected behavior?

Ideally this should behave identically to the Rails login (TODO look up how that works).

Practically, the login just needs to not fail and needs to pick any group from `user.miq_groups` to be the `current_group`.

(In theory we could fail with a list of possible groups instead of picking one, but that potentially leaks info about user's group membership *and* would require extra UI.)


> 2. Was there a time when the API performed the expected behavior, and

[...]

PR #368, opened 2018-04-25
[WIP] If a user never logged in, default the current_group

abellotti:
If a user never logged in, the current_group could be nil and issues validating the user in API.

We now default the current_group (usually miq_groups.first) if nil

Fixes https://github.com/ManageIQ/manageiq-api/issues/366

### DIFF
--- a/app/controllers/api/base_controller/authentication.rb
+++ b/app/controllers/api/base_controller/authentication.rb
@@ -52,6 +52,8 @@ module Api
           group_obj = user_obj.miq_groups.find_by(:description => group_name)
           raise AuthenticationError, "Invalid Authorization Group #{group_name} specified" if group_obj.nil?
           user_obj.current_group_by_description = group_name
+        elsif user_obj.current_group.nil? && user_obj.miq_groups.present?
+          user_obj.change_current_group
         end
       end
 
--- a/spec/requests/authentication_spec.rb
+++ b/spec/requests/authentication_spec.rb
@@ -44,6 +44,7 @@ describe "Authentication API" do
     end
 
     it "test basic authentication with a user without a group" do
+      @user.miq_groups = []
       @user.current_group = nil
       @user.save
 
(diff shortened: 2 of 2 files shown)


abellotti:
marking as wip, need a spec.

app/controllers/api/base_controller/authentication.rb
```diff
@@ -52,6 +52,8 @@ def authorize_user_group(user_obj)
           group_obj = user_obj.miq_groups.find_by(:description => group_name)
           raise AuthenticationError, "Invalid Authorization Group #{group_name} specified" if group_obj.nil?
           user_obj.current_group_by_description = group_name
+        elsif user_obj.current_group.nil? && user_obj.miq_groups.present?
```
imtayadeway: I have a couple of concerns here: 

1. I don't think this restores the previous behavior, it seems like a workaround to prevent it from blowing up. It is sort of a new feature that is fixing a bug, and I'd prefer those two things to be separate PRs
2. I have a larger concern that extends to other things in this method, which is that the API seems to know too much about authorization-domain things. I'd prefer for us just to pass the group name to some more central authorization code, which would handle the correcting of the user's current group if that's what we're deciding is what it should do, and should just return success/failure in some way.

abellotti:
Interesting, my/the expectation from the API perspective is that the following works regardless of UI changes made:

```
userid, name, password = "api_test_user_create_1", "Aab Test 1", "smartvm"

miq = ManageIQ::API::Client.new

miq.users.find_by(:userid => userid).try(:delete)

user = miq.users.create(:userid => userid, :name => name, :password => password, :group => { :description => "EvmGroup-user" })

puts "#{user.userid} properly created"

miq2 = ManageIQ::API::Client.new(:user => user.userid, :password => password)

puts "#{miq2.identity.userid} properly authenticated"
```

which works fine with latest API/Model.

which brings back the question of, what changed in UI's user creation.

thoughts @imtayadeway ?

abellotti:
This is no longer needed.  Works with latest code base, thanks @gtanzillo for kicking the tire.

Issue #269, opened 2018-01-09
Find a better way to test posts with query params


imtayadeway:
There's an open issue for this in rake-test: https://github.com/rack-test/rack-test/issues/150. However, I don't see this moving ahead soon.

Alternatively, it may be better to respond to this test pain by figuring out a better way to implement this.

Links:
https://github.com/ManageIQ/manageiq-api/blob/47010cec9af5b80993df29af93a94cd9775ae1a5/spec/requests/providers_spec.rb#L370-L371
https://github.com/ManageIQ/manageiq-api/blob/47010cec9af5b80993df29af93a94cd9775ae1a5/spec/requests/providers_spec.rb#L1209-L1210

JPrause:
@gtanzillo @abellotti is this still a valid issue? If yes, please remove the stale label. If not can you close.
If there's no update by next week, I'll be closing this issue.

JPrause:
Closing issue. If you feel the issue needs to remain open, please either reopen or let me know and it will be reopened. 
@miq-bot close_issue

PR #374, opened 2018-05-03
Prefer QUERY_STRING rack variable to concatenating paths

imtayadeway:
Fixes https://github.com/ManageIQ/manageiq-api/issues/269

### DIFF
--- a/spec/requests/providers_spec.rb
+++ b/spec/requests/providers_spec.rb
@@ -402,7 +402,11 @@ describe "Providers API" do
     it 'creates valid foreman provider' do
       api_basic_authorize collection_action_identifier(:providers, :create)
 
-      post(api_providers_url + '?provider_class=provider', :params => gen_request(:create, sample_foreman))
+      post(
+        api_providers_url,
+        :headers => {"QUERY_STRING" => "provider_class=provider"},
+        :params  => {:action => "create", :resource => sample_foreman}
+      )
 
       expect(response).to have_http_status(:ok)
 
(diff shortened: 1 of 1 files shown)


spec/requests/providers_spec.rb
```diff
@@ -1536,19 +1548,30 @@ def gen_import_request
...
         expect_bad_request("#{generic_provider.class.name} does not support management of custom attributes")
       end
 
       it 'cannot edit custom_attribute' do
         api_basic_authorize subcollection_action_identifier(:providers, :custom_attributes, :edit, :post)
-        post(url, :params => gen_request(:edit, :href => api_provider_custom_attribute_url(nil, generic_provider, attr)))
+        post(
+          api_provider_custom_attributes_url(nil, generic_provider),
+          :headers => {"QUERY_STRING" => "provider_class=provider"},
+          :params  => {
```
abellotti: would prefer to keep using gen_request in all the above :params as to not deviate with the rest of the specs.  Thanks.

imtayadeway: Personally, I think `gen_request` should be slowly deprecated in favor of a more transparent request declaration as the above, as we have started to do elsewhere: 1) `gen_request` isn't really descriptive, 2) it obscures important details about how the request is structured. A test, as well as exercising code in the system, should also serve as a clear example, and these details are crucial to understanding what was sent in the request. As such I'd much rather deal with inconsistency while seizing opportunities such as this to slowly move toward a more clear example as we have to change the surrounding code anyway.

PR #365, opened 2018-04-23
[WIP] Add refresh parameter to ResourceActionWorkflow initialization

eclarizio:
This ties in with https://github.com/ManageIQ/manageiq/pull/17329 and will be needed in order to distinguish `ResourceActionWorkflow` initialization calls in the backend.

WIP for now, but I think this is probably how the API change will end up looking.

https://bugzilla.redhat.com/show_bug.cgi?id=1559757

PR #386, opened 2018-05-29
Ensure refresh => true option gets passed for service_template refresh

eclarizio:
Related to https://github.com/ManageIQ/manageiq-api/pull/365, but the Service UI was using this endpoint instead for its refreshes. The Service UI is going to be changed to use the same endpoint as the Classic UI in https://github.com/ManageIQ/manageiq-ui-service/pull/1438, but this will need to be updated anyway in case anyone decides they want to use the API call outside of either the SUI or CUI.

https://bugzilla.redhat.com/show_bug.cgi?id=1578080

/cc @d-m-u 

@miq-bot assign @abellotti

### DIFF
--- a/app/controllers/api/subcollections/service_templates.rb
+++ b/app/controllers/api/subcollections/service_templates.rb
@@ -97,7 +97,7 @@ module Api
         refresh_fields = data["fields"]
         return action_result(false, "Must specify fields to refresh") if refresh_fields.blank?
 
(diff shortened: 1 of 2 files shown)


JPrause:
@miq-bot add_label blocker

abellotti:
LGTM!! Thanks @eclarizio for fixing this. 👍

PR #394, opened 2018-06-12
Add .idea folder to .gitignore

felipedf:
This adds .idea folder to .gitignore


Fryguy:
@felipedf Generally, we avoid putting user-specific things into .gitignore files (e.g. IDE files and OS-specific things like `.DS_Store`).  You can use your personal global .gitignore file for this purpose.

PR #395, opened 2018-06-13
Rename dialog_edit and dialog_new to *_editor

himdel:
The `dialog_edit` and `dialog_new` features were removed in https://github.com/ManageIQ/manageiq/pull/17550.

The new features to do the same thing seem to be called `dialog_new_editor` and `dialog_edit_editor` - renaming.

Cc @skateman , @abellotti

PR #396, opened 2018-06-14
Ignore case of the userid when validating it.

jvlcek:
The authentication code ignores the case of the userid. The PR ensures the API user validation
code does also.

All userids are created by the authentication code in lowercase. Users can be manually created with the UI in mixed case. However the UI will not allow a second user to be created with a mismatching case.

If **_User_** exists and one attempts to create **_user_** the UI will report: _**Userid is not unique**_

### To test:

1. Set authentication mode to: **Database**
2. From the UI create a new user with at least one uppercase letter in **Username**. e.g.: **_User_**
3. Try to login with this user.

**Note:** The case of the value specified for **Username** on the login screen is ignored. So if **_User_** was create one could log in with **_User_** or **_user_**

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1590398

### DIFF
--- a/lib/services/api/user_token_service.rb
+++ b/lib/services/api/user_token_service.rb
@@ -75,7 +75,7 @@ module Api
     end
 
     def validate_userid(userid)
-      raise "Invalid userid #{userid} specified" unless User.where('lower(userid) = ?', userid.downcase).exists?
+      raise "Invalid userid #{userid} specified" unless User.in_my_region.where('lower(userid) = ?', userid.downcase).exists?
     end
   end
 end
--- a/lib/services/api/user_token_service.rb
+++ b/lib/services/api/user_token_service.rb
@@ -75,7 +75,7 @@ module Api
     end
 
     def validate_userid(userid)
-      raise "Invalid userid #{userid} specified" unless User.exists?(:userid => userid)
+      raise "Invalid userid #{userid} specified" unless User.where('lower(userid) = ?', userid.downcase).exists?
     end
   end
 end
--- a/config/api.yml
+++ b/config/api.yml
@@ -2615,9 +2615,9 @@
       - :name: delete
         :identifier: dialog_delete
       - :name: create
-        :identifier: dialog_new
+        :identifier: dialog_new_editor
       - :name: edit
-        :identifier: dialog_edit
+        :identifier: dialog_edit_editor
       - :name: copy
         :identifier: dialog_copy_editor
     :resource_actions:
@@ -2632,7 +2632,7 @@
       - :name: delete
         :identifier: dialog_delete
       - :name: edit
-        :identifier: dialog_edit
+        :identifier: dialog_edit_editor
       - :name: copy
         :identifier: dialog_copy_editor
       :delete:
--- a/spec/lib/services/api/user_token_service_spec.rb
+++ b/spec/lib/services/api/user_token_service_spec.rb
@@ -1,13 +1,13 @@
 RSpec.describe Api::UserTokenService do
+  before do
+    @user = FactoryGirl.create(:user_with_group)
+  end
+
+  let(:user_token_service) { described_class.new }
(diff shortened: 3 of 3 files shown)


jvlcek:
@miq-bot assign @gtanzillo

jvlcek:
@miq-bot add_label bug

jvlcek:

jvlcek:
@h-kataria Just FYI. Thank you for the help.

jvlcek:
@gtanzillo I had considered using  `User.in_my_region.where` but this Api code never limited the check to the current region and I felt that doing so might introduce failures elsewhere. When the backend authenticates it does limit the check to the current region so that case is covered.

Let me know if you feel I should use: `User.in_my_region.where`

Thank you, JoeV

lib/services/api/user_token_service.rb
```diff
@@ -75,7 +75,7 @@ def log_init(mod, name, options)
     end
 
     def validate_userid(userid)
-      raise "Invalid userid #{userid} specified" unless User.exists?(:userid => userid)
+      raise "Invalid userid #{userid} specified" unless User.where('lower(userid) = ?', userid.downcase).exists?
```
gtanzillo: @jvlcek You should add the region scoping here. I think it was missed wth the original code and is a bug waiting to happen. We should only allow users that are defined in the current region to log in.

jvlcek: @gtanzillo Wildo. Thank you.

jvlcek:
@skateman Please review to ensure the specS that I am removing align with the [PR 17550](https://github.com/ManageIQ/manageiq/pull/17550) that you made to remove dialog_edit and dialog_new. Thanks man!

jvlcek:
@gtanzillo @abellotti 

This PR makes a 1 line code change in **_lib/services/api/user_token_service.rb_** and updated
to the related spec to exercise this change.

However Travis builds failed for 2 specs due to PR: ManageIQ/manageiq#17550,
that has removed 2 product features, _dialog_new_ and _dialog_edit_

The updates to files _**config/api.yml**_ and **_spec/requests/service_dialogs_spec.rb_** were
necessary to compensate for the changes made in PR: ManageIQ/manageiq#17550

Please review. Thank you! JoeV

himdel:
The test fixes look good, if they get merged here, we can close https://github.com/ManageIQ/manageiq-api/pull/395 :)

skateman:
@jvlcek I am okay with this, however, I'm starting to feel that I deleted an API feature by dropping those featurs :disappointed: so I think this is more @abellotti's call than mine.

himdel:
Please use #395 instead to fix the tests.
(as soon as @skateman reviews)

skateman:
@romanblanco can you please verify that this doesn't break your dialog editor?

jvlcek:
I'm pretty sure we have this correctly updated to align with the changes to the dialogs in manageiq.

@himdel @skateman @gtanzillo and @abellotti can each of you please take a quick look?

Thank you! JoeV

romanblanco:
@skateman tested, Dialog Editor seems to work correctly :+1:

abellotti:
LGTM!! Thanks @jvlcek for fixing this.  👍

simaishi:
Skipped backport of spec file (which doesn't exist in G branch) as per discussion with @jvlcek 

```
$ git log -1
commit e1c46824d09267eb9241e9af019bebc8887e181f
Date:   Thu Jun 21 11:38:03 2018 -0400

    Merge pull request #396 from jvlcek/bz1590398_mixed_case_userid
    
    Ignore case of the userid when validating it.
    
    (cherry picked from commit 956c0d571a16bc36e2df271ae4cf6a8e802f316c)
    
    Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1638501
```

PR #402, opened 2018-06-19
Remove references to host provisioning

gmcculloug:
Remove references to host provisioning.

Related to https://github.com/ManageIQ/manageiq/pull/17604

### DIFF
--- a/config/api.yml
+++ b/config/api.yml
@@ -2326,8 +2326,6 @@
       - :name: create
         :identifiers:
         - :klass: AutomationRequest
-        - :klass: MiqHostProvisionRequest
-          :identifier: host_miq_request_new
         - :klass: MiqProvisionRequest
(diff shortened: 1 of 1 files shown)


gmcculloug:
@abellotti Not sure if there are any other changes required to remove host_provisioning from the repo.  Hoping you can take a quick look and advise.  Thanks.

abellotti:
that's it for this repo.  The only question would be if there were current calls doing that and are we breaking/removing capability ?

gmcculloug:
@abellotti This is old logic which is limited in functionality and I am not aware of anyone using it.  There are preferred tools, like https://www.theforeman.org for this task.

cc @Fryguy

abellotti:
Thanks @gmcculloug for the spring cleaning.  LGTM!! 👍

PR #405, opened 2018-06-25
[GAPRINDASHVILI] Fix CI failure: correctly set @role before authentication

mzazrivec:
This is so that `api_basic_authorize()` can properly set product features for given `@role`.


abellotti:
ping @gtanzillo when you get a chance to 👀   Thanks.

PR #408, opened 2018-06-27
API support for updating and deleting `service_template` schedules

gtanzillo:
**Get schedule**
`GET /api/service_templates/5?attributes=miq_schedule`

**Get scheduled time**
`GET /api/service_templates/58?attributes=schedule_time`

**Update scheduled time**
```
POST /api/service_templates/59
{
  "action" : "edit",
  "schedule_time" : "20180704T100000Z"
}
```

**Delete schedule**
```
POST /api/service_templates/59
{ 
  "action" : "edit",
  "schedule_time" : null
}
```

Depends on https://github.com/ManageIQ/manageiq/pull/17645

Paired with @abellotti

### DIFF
--- a/app/controllers/api/service_templates_controller.rb
+++ b/app/controllers/api/service_templates_controller.rb
@@ -22,7 +22,11 @@ module Api
       catalog_item = resource_search(id, type, collection_class(:service_templates))
       if data.key?("schedule_time")
         schedule_time = data&.delete("schedule_time")
-        catalog_item.update_schedule(schedule_time)
+        if schedule_time.nil?
+          catalog_item.delete_schedule
+        else
+          catalog_item.update_schedule(schedule_time)
+        end
       end
       catalog_item.update_catalog_item(data.deep_symbolize_keys, User.current_user.userid)
     rescue => err
--- a/app/controllers/api/service_templates_controller.rb
+++ b/app/controllers/api/service_templates_controller.rb
@@ -20,6 +20,10 @@ module Api
 
     def edit_resource(type, id, data)
       catalog_item = resource_search(id, type, collection_class(:service_templates))
+      if data.key?("schedule_time")
+        schedule_time = data&.delete("schedule_time")
+        catalog_item.update_schedule(schedule_time)
+      end
       catalog_item.update_catalog_item(data.deep_symbolize_keys, User.current_user.userid)
     rescue => err
       raise BadRequestError, "Could not update Service Template - #{err}"
--- a/lib/api/environment.rb
+++ b/lib/api/environment.rb
@@ -15,7 +15,7 @@ module Api
     end
 
     def self.time_attributes
-      @time_attributes ||= ApiConfig.collections.each.with_object(Set.new(%w(expires_on))) do |(_, cspec), result|
+      @time_attributes ||= ApiConfig.collections.each.with_object(Set.new(%w(expires_on schedule_time))) do |(_, cspec), result|
         next if cspec[:klass].blank?
         klass = cspec[:klass].constantize
         klass.columns_hash.each do |name, typeobj|
--- a/spec/requests/service_templates_spec.rb
+++ b/spec/requests/service_templates_spec.rb
@@ -99,6 +99,30 @@ describe "Service Templates API" do
       expect(response).to have_http_status(:ok)
       response.parsed_body['service_resources'].each { |resource| expect(resource.keys).to_not include('href') }
     end
+
+    context "with a schedule" do
+      let(:st)            { FactoryGirl.create(:service_template, :name => "st") }
+      let(:schedule_time) { Time.now.utc }
+      let!(:schedule)     { st.order(@user, {}, {}, schedule_time.to_s) }
+
+      it "gets the miq_schedule" do
+        api_basic_authorize collection_action_identifier(:service_templates, :read, :get)
+
+        get(api_service_template_url(nil, st), :params => {:attributes => 'miq_schedule'})
+
+        expect(response).to have_http_status(:ok)
+        expect(response.parsed_body["miq_schedule"]["run_at"]["start_time"]).to eq(schedule_time.iso8601)
+      end
+
+      it "gets the scheduled time" do
+        api_basic_authorize collection_action_identifier(:service_templates, :read, :get)
+
+        get(api_service_template_url(nil, st), :params => {:attributes => 'schedule_time'})
+
+        expect(response).to have_http_status(:ok)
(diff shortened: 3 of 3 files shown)


AllenBW:
😍 💘 :godmode: 

yer about to unwip a week of 💌 

🙇‍♀️ 🌮 💃

gtanzillo:
This PR will be red until ManageIQ/manageiq#17645 is merged

app/controllers/api/service_templates_controller.rb
```diff
@@ -20,6 +20,10 @@ def create_resource(_type, _id, data)
 
     def edit_resource(type, id, data)
       catalog_item = resource_search(id, type, collection_class(:service_templates))
+      if data.key?("schedule_time")
+        schedule_time = data&.delete("schedule_time")
+        catalog_item.update_schedule(schedule_time)
```
jrafanie: It feels like this code should be responsible for calling the "update or delete schedules" method if the `schedule_time` is provided or not.  See comment [here](https://github.com/ManageIQ/manageiq/pull/17645#discussion_r199217984)

app/controllers/api/service_templates_controller.rb
```diff
@@ -20,6 +20,14 @@ def create_resource(_type, _id, data)
 
     def edit_resource(type, id, data)
       catalog_item = resource_search(id, type, collection_class(:service_templates))
+      if data.key?("schedule_time")
+        schedule_time = data&.delete("schedule_time")
+        if schedule_time.nil?
```
jrafanie: Do we have a way to make the delete more explicit?  It's a little hidden that passing `schedule_times => nil` in the hash performs a deletion.  I'm good with this as a first pass if we can't do that easily.

gtanzillo: @jrafanie Since this request is on the service template we can't really do it with a `delete` action. We could make it more explicit by requiring the call to pass "delete" as the value of the schedule_time in the `edit` action request. What do you think about that?

Fryguy:
An edit action if delete would delete the service

app/controllers/api/service_templates_controller.rb
jrafanie: I think that makes more sense.  We're pretty stuck because the delete action isn't what we want to do.  I'm fine with doing this using the string "delete" or "remove"

Fryguy:
REST-wise I would expect an association to the schedule resource under the service template and you send a DELETE request to that.  Not sure what I think about devolving a schedule to an attribute.

gtanzillo:
>REST-wise I would expect an association to the schedule resource under the service template and you send a DELETE request to that

I was thinking that too. But I was also thinking that the schedule was an internal thing that we don't expose directly today. With reporting schedules, they are accessed and edited through the report.

jrafanie:
> I was thinking that too. But I was also thinking that the schedule was an internal thing that we don't expose directly today. With reporting schedules, they are accessed and edited through the report.

yeah, unless we promote the report/service_template schedules to an actual resource, we have to bury them like this.  I'm guessing we'll eventually want schedules to be a first class resource so how do we allow that to happen without breaking API in the future?

gtanzillo:
If we need to promote schedules to a first class object in the API in the future we could do so and deprecate the functionality here. That should be doable.

Fryguy:
Ah ok, I thought was @bdunne was doing was trying to promote the scheduled to their own resource.

An edit action of delete is still wrong though precisely because of this... The edit is happening in a service template, not on the schedule attribute itself.

app/controllers/api/service_templates_controller.rb
```diff
@@ -20,6 +20,14 @@ def create_resource(_type, _id, data)
 
     def edit_resource(type, id, data)
       catalog_item = resource_search(id, type, collection_class(:service_templates))
+      if data.key?("schedule_time")
+        schedule_time = data&.delete("schedule_time")
```
Fryguy: You've already checked the validity of data before this, so the &. is unnecessary.

lib/api/environment.rb
```diff
@@ -15,7 +15,7 @@ def self.encrypted_attributes
     end
 
     def self.time_attributes
-      @time_attributes ||= ApiConfig.collections.each.with_object(Set.new(%w(expires_on))) do |(_, cspec), result|
+      @time_attributes ||= ApiConfig.collections.each.with_object(Set.new(%w(expires_on schedule_time))) do |(_, cspec), result|
```
Fryguy: This feels weird to do at the generic level.

app/controllers/api/service_templates_controller.rb
```diff
@@ -20,6 +20,14 @@ def create_resource(_type, _id, data)
...
+        if schedule_time.nil?
+          catalog_item.delete_schedule
+        else
+          catalog_item.update_schedule(schedule_time)
+        end
+      end
       catalog_item.update_catalog_item(data.deep_symbolize_keys, User.current_user.userid)
```
Fryguy: This line has the user as a parameter which implies rbac is being done, but your code is bypassing the rbac.  Maybe your code should go into the this method too?

spec/requests/service_templates_spec.rb
```diff
@@ -184,6 +208,30 @@
...
+        post(api_service_template_url(nil, st), :params => { :action => "edit", :schedule_time => new_time.to_s })
+
+        st.reload
+        expect(st.schedule_time.to_s).to eq(new_time.to_s)
+      end
+
+      it "can delete a schedule" do
+        api_basic_authorize collection_action_identifier(:service_templates, :edit)
+
+        post(api_service_template_url(nil, st), :params => { :action => "edit", :schedule_time => nil })
```
Fryguy: What if you send nil and there isn't already a schedule?

Fryguy: Also what happens if the user sends garbage or times in the far past or future?

app/controllers/api/service_templates_controller.rb
gtanzillo: RBAC would have been done before this. Looking at the code, it seems like the user being saved in the workflow.

app/controllers/api/service_templates_controller.rb
bdunne: Is that how we work with CRUD actions on other subcollections?

AllenBW:
Hi friends, wanted to pop in andddddddddd say hiiiiiiiiiii (its the most polite way of asking for a status update) ❤️ 🍍 😮

gtanzillo:
Closing because we decided to go with an actual subcollection for schedules instead of an attribute.

PR #409, opened 2018-07-04
Moving the operations routine to a mixin

douglasgabriel:
__This PR is able to__
- create `Api::Mixins::Operations` to share the operations routine with others controllers;

__Main goal__
The enqueue of operations for physical components could be shared, so would be good have a mixin with those routines, as a first step, the `physical_switches_controller` was refactored to this new approach and then the `physical_chassis_controller` and `physical_servers_controller` will be refactored in upcoming PRs

### DIFF
--- a/app/controllers/api/mixins/operations.rb
+++ b/app/controllers/api/mixins/operations.rb
@@ -1,0 +1,65 @@
+module Api
+  module Mixins
+    #
+    # Contains the routines to enqueue operations for resources
+    #
+    module Operations
+      def perform_action(action, type, id)
+        if single_resource?
+          enqueue_action_single_resource(action, type, id)
+        else
+          enqueue_action_multiple_resources(action, type, id)
+        end
+      end
+
+      #
+      # Enqueues the action for a single resource.
+      #
+      # @param [symbol] action - action to be enqueued
+      # @param [symbol] type   - type of the resource
+      # @param [number] id     - id of the resource
+      #
+      def enqueue_action_single_resource(action, type, id)
+        raise BadRequestError, "Must specify an id for changing a #{type} resource" unless id
+
+        resource = resource_search(id, type, collection_class(type))
+
+        api_action(type, id) do
+          begin
+            desc = "Performing #{action} for #{resource_identify(resource)}"
+            api_log_info(desc)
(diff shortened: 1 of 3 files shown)


douglasgabriel:
@miq-bot assign @gtanzillo

Fryguy:
This looks good to me.  @agrare I think we talked about something similar the other day with respect to "pluggable API for operations". What do you think?

app/controllers/api/mixins/operations.rb
```diff
@@ -0,0 +1,65 @@
...
+      #   be built individually for each resource. Always responding with status 200.
+      #
+      # @param [symbol] action - action to be enqueued
+      # @param [symbol] type   - type of the resource
+      # @param [number] id     - id of the resource
+      #
+      def enqueue_action_multiple_resources(action, type, id)
+        enqueue_action_single_resource(action, type, id)
+      rescue ActiveRecord::RecordNotFound => err
+        action_result(false, _(err.message))
```
Fryguy: This doesn't seem to match up with the comment, since you aren't building any error message (and for that matter only have 1 resource).  I realize this is a copy-paste from the physical_switches_controller, but if it's being promoted to a common mixin, it needs to be consistent with other modules in the base.

douglasgabriel: Actually, even the request is made for multiple resources, this method is called individually for each one. With `built individually`, in the comment, I mean to call `action_result(false, _(err.message))` each time that occurs an error for some resource, which is what the method does. It is a copy-paste from physical_switches_controller, but with some adaptations to work with other resources, we can see this on this PR https://github.com/ManageIQ/manageiq-api/pull/410 where physical_chassis_controller also uses this routine.

gtanzillo:
This looks good to me as well. As far as pluggability goes, this seems like something that would still belong in the core since i the mixin seems to be completely generic. We still need to work what the pluggability for the actions in the controller will look like.

@agrare Are you good with this?

agrare:
@gtanzillo yeah I'm fine with this, I'd be curious if we can refactor some existing actions on other controllers to use this mixin.

Re pluggability I was thinking that the actions would be dynamically defined based on the api.yml, with this the provider plugins will still need to define their own action methods even if they just use this mixin which is a good step forward.

PR #418, opened 2018-07-11
Allow plugins to bring api.yml extensions

agrare:
This allows Vmdb::Plugins like providers to bring their own api.yml
extensions, enabling for example new operations that those providers
bring to be accessible through the API.

### DIFF
--- a/lib/api/api_config.rb
+++ b/lib/api/api_config.rb
@@ -1,7 +1,9 @@
 require "config"
 
 module Api
+  ::Config.overwrite_arrays = false
   ApiConfig = ::Config::Options.new.tap do |o|
+    Vmdb::Plugins.each { |plugin| o.add_source!(plugin.root.join('config/api.yml').to_s) }
     o.add_source!(ManageIQ::Api::Engine.root.join("config/api.yml").to_s)
     o.load!
   end


agrare:
cc @gtanzillo @Fryguy @Ladas @martinpovolny

agrare:
Example:
```
agrare@agrare-thinkpad:~/src/manageiq/manageiq-providers-vmware$ cat config/api.yml 
---
:base:
  :hello: world

agrare@agrare-thinkpad:~/src/manageiq/manageiq$ rails c
>> Api::ApiConfig.base.hello
=> "world"
```

lib/api/api_config.rb
```diff
@@ -3,6 +3,7 @@
 module Api
   ApiConfig = ::Config::Options.new.tap do |o|
     o.add_source!(ManageIQ::Api::Engine.root.join("config/api.yml").to_s)
+    Vmdb::Plugins.each { |plugin| o.add_source!(plugin.root.join('config/api.yml').to_s) }
```
gtanzillo: Do we want to put a rescue here in case there's a provider that introduces some malformed yaml?

Ladas:
@agrare wow, that looks really simple. What happens if there are conflicting endpoints?

agrare:
> What happens if there are conflicting endpoints?

Looks like last one wins, same as settings.yml

Ladas:
@agrare are we ok with that? Or should we encourage namespacing? Also should we check conflicts with the core? Or maybe just letting core to be the last?

lib/api/api_config.rb
agrare: It only fails on the `.load!` no on the `.add_source!` if there is malformed YAML and at that point I'm not sure how we would handle it.
Blowing up quickly has its benefits :D

agrare:
@Ladas good call on letting core be last let me push that change

Fryguy:
As discussed, we don't need namespacing.

I am ok with core yml being last, but it might be tricky if we allow them to override specific actions (like power on) and they want to bring their own rbac/product features that differs from core...they would need override capability.  That being said, if we change something in core, we don't want to have an accidental override, so I'm not sure.  This isn't end-user code, but curated code, so I'm maybe it's something we fix with tests or specs or something.

agrare:
@Fryguy hm you were right about the array merge issue, adding the actions to the collection_actions.post array overwrites the previous value instead of merging them, here is the `config/api.yml` changes from @Ladas 's [PR](https://github.com/ManageIQ/manageiq-api/pull/398) put in the manageiq-providers-amazon repo:

```
---
:collections:
  :instances:
    :collection_actions:
      :post:
      - :name: special_stop
        :identifier: instance_stop
      - :name: special_start
        :identifier: instance_start
    :resource_actions:
      :post:
      - :name: special_stop
        :identifier: instance_stop
      - :name: special_start
        :identifier: instance_start
  :security_groups:
    :subcollection_actions:
      :post:
      - :name: create
        :identifier: security_group_new

**Comment by Ladas** [comment]
@Fryguy @agrare ok, so what is the way forward? Changing the merging strategy so the arrays are merged and uniqued (by :name or :identifier). Or changing the format so the :name or :identifier becomes a key, so getting rid of the array?

**Comment by agrare** [comment]
@Ladas yeah I don't know the config gem well enough to know if we can control how arrays are merged

**Comment by agrare** [comment]
@Fryguy :+1: `::Config.overwrite_arrays = false` fixed it thanks!

```
>> Api::ApiConfig.collections.instances.collection_actions.post.map(&:name).include?("special_start")
=> true
```

PR #421, opened 2018-07-13
Request cancel

bdunne:
Based on changes in: https://github.com/ManageIQ/manageiq/pull/17687
https://bugzilla.redhat.com/show_bug.cgi?id=1564257
cc @abellotti @gtanzillo @gmcculloug @bzwei

### DIFF
--- a/app/controllers/api/mixins/resource_cancel.rb
+++ b/app/controllers/api/mixins/resource_cancel.rb
@@ -1,0 +1,14 @@
+module Api
+  module Mixins
+    module ResourceCancel
+      def cancel_resource(type, id, _data)
+        api_action(type, id) do |klass|
+          resource_search(id, type, klass).cancel
+          action_result(true, "#{klass.name} #{id} canceled")
+        end
+      rescue => err
+        action_result(false, err.to_s)
+      end
+    end
+  end
+end
--- a/spec/support/api/helpers.rb
+++ b/spec/support/api/helpers.rb
@@ -224,6 +224,12 @@ module Spec
           expect(response.headers['Access-Control-Allow-Methods']).to include('OPTIONS')
         end
 
+        def expect_unauthorized_request
+          api_basic_authorize
+          yield
+          expect(response).to have_http_status(:forbidden)
+        end
+
         def select_attributes(attrlist)
           attrlist.sort.select { |attr| !::Api.encrypted_attribute?(attr) }
         end
--- a/spec/support/api/helpers.rb
+++ b/spec/support/api/helpers.rb
@@ -172,7 +172,7 @@ module Spec
 
         def expect_single_action_result(options = {})
           expect(response).to have_http_status(:ok)
-          expected = {}
+          expected = options.slice("href", "message", "success")
           expected["success"] = options[:success] if options.key?(:success)
           expected["message"] = a_string_matching(options[:message]) if options[:message]
           expected["href"] = a_string_matching(options[:href]) if options[:href]
--- a/spec/support/api/helpers.rb
+++ b/spec/support/api/helpers.rb
@@ -224,7 +224,7 @@ module Spec
           expect(response.headers['Access-Control-Allow-Methods']).to include('OPTIONS')
         end
 
-        def expect_unauthorized_request
+        def expect_forbidden_request
           api_basic_authorize
           yield
           expect(response).to have_http_status(:forbidden)
--- a/app/controllers/api/automation_requests_controller.rb
+++ b/app/controllers/api/automation_requests_controller.rb
@@ -1,5 +1,6 @@
 module Api
   class AutomationRequestsController < BaseController
+    include Api::Mixins::ResourceCancel
     include Subcollections::RequestTasks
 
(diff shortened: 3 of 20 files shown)


bzwei:
@bdunne v2v RFE needs task level cancellation. Do we expect another PR?

gmcculloug:
@bdunne Since the back-end is really only supporting this feature for v2v can we reduce the models this is exposed on?  At the moment we return "Unsupported" for all other request/task types so I would prefer limiting the action as much as possible.

bdunne:
@gmcculloug Many thoughts..
- I was surprised to find out that we have 4 requests endpoints.  Do you know why?
- I'd like keep these 4 endpoints as similar as possible as long as it makes sense (especially in this case where 2/4 would have to implement it in my opinion)
- I think an "Unsupported" response is better than "Bad Request Error"
- It may only be used by v2v at the moment, but I don't think the API should care about implementing things for specific use cases, it should be done in a generic and consistent way.

JPrause:
@miq-bot add_label blocker

JPrause:
@bdunne if this can be backported, can you add the gaprindashvili/yes label

gmcculloug:
@gtanzillo @abellotti Any thoughts on https://github.com/ManageIQ/manageiq-api/pull/421#issuecomment-405581100

gtanzillo:
@gmcculloug Regarding
>I was not involved in building out the API so do not have any insight into why there are 4 endpoints.

>@gtanzillo @abellotti Any thoughts on #421 (comment)

I was also not involved so I'm not sure either. It seems in some places we've exposed a single endpoint for the base class and in other places we exposed separate endpoints for each subclass like was done here. 

Perhaps @abellotti or @Fryguy can comment on that. It would be good to understand which convention we should be following, if there is one.

bdunne:
Okay, I think this is ready for review.  @jrafanie can you put another set of :eyes: on the tests?

jrafanie:
LGTM.  It's a bit hard to follow at first with the indirection but the alternative is 1000 lines of duplicated tests, so :shipit: 

It would be good if @Fryguy can comment on the separate collections for each subclass.  It seems strange.  I can't imagine having separate collections for vmware  vms, redhat vms, etc.   On the fipside, these things are not the same and are actual tasks for completely different end user features. 🤷‍♂️

app/controllers/api/mixins/resource_cancel.rb
```diff
@@ -0,0 +1,14 @@
+module Api
+  module Mixins
+    module ResourceCancel
```
Fryguy: Is this ResourceCancel or TaskCancel?

Fryguy:
> It would be good if @Fryguy can comment on the separate collections for each subclass. It seems strange. I can't imagine having separate collections for vmware vms, redhat vms, etc.

Where are you seeing this?

spec/support/api/helpers.rb
```diff
@@ -224,6 +224,12 @@ def expect_options_results(type, data = {})
           expect(response.headers['Access-Control-Allow-Methods']).to include('OPTIONS')
         end
 
+        def expect_unauthorized_request
```
Fryguy: minor, but this might read better as expect_forbidden_request, since the status is `:forbidden`

app/controllers/api/mixins/resource_cancel.rb
bdunne: It's for `ServiceOrder`, `MiqRequest`X4 or `MiqRequestTask` primary collections.

bdunne:
> @jrafanie Where are you seeing this?

`/api/requests`
`/api/automation_requests`
`/api/service_requests`
`/api/provision_requests`

Talking to Alberto, it sounds like it just grew this way over time.  Hopefully we don't need to create new endpoints for the other 12 subclasses of `MiqRequest`.

AparnaKarve:
Can we get this merged today?

bthurber:
@gtanzillo @gmcculloug @blomquisg we will need this merged ASAP for 5.9.4-2 build for tomorrow.  Anything here left to finish/review?

bthurber:
@bdunne can you respond to @Fryguy last comment?  We also need to add enhancement, transformation and grapendashvili/yes labels so this get's pulled into the next build once merged.

bdunne:
@bthurber I responded in: https://github.com/ManageIQ/manageiq-api/pull/421#issuecomment-406416017

bdunne:
@miq-bot add_labels gaprindashvili/yes enhancement transformation

bdunne:
@miq-bot add_labels enhancement
@miq-bot add_labels transformation

simaishi:

```
$ git log -1
commit b7047744b21596fafe4cbfdc1064cad1fd8ba76c
Date:   Mon Jul 30 18:02:12 2018 -0400

    Merge pull request #421 from bdunne/request_cancel
    
    Request cancel
    (cherry picked from commit 62de390490c8bc333f27feedfc58455ced1e7483)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1610533
```

PR #423, opened 2018-07-17
Adding Chassis and Racks as PhysicalStorages subcollections

EsdrasVP:
This PR is able to add:
- `GET /api/physical_storages/:c_id/physical_chassis`
- `GET /api/physical_storages/:c_id/physical_chassis/:s_id`
- `GET /api/physical_storages/:c_id/physical_racks`
- `GET /api/physical_storages/:c_id/physical_racks/:s_id`

Depends on:
- ~[ManageIQ/manageiq-schema#224](https://github.com/ManageIQ/manageiq-schema/pull/224)~ [Merged]
- ~[ManageIQ/manageiq#17616](https://github.com/ManageIQ/manageiq/pull/17616)~ [Merged]
- ~[ManageIQ/manageiq-providers-lenovo#201](https://github.com/ManageIQ/manageiq-providers-lenovo/pull/201)~ [Merged]

### DIFF
--- a/app/controllers/api/subcollections/physical_chassis.rb
+++ b/app/controllers/api/subcollections/physical_chassis.rb
@@ -1,0 +1,9 @@
+module Api
+  module Subcollections
+    module PhysicalChassis
+      def physical_chassis_query_resource(object)
+        object.respond_to?(:physical_chassis) ? ManageIQ::Providers::PhysicalInfraManager::PhysicalChassis.where(:id => object.physical_chassis_id) : []
+      end
+    end
+  end
+end
--- a/app/controllers/api/subcollections/physical_racks.rb
+++ b/app/controllers/api/subcollections/physical_racks.rb
@@ -1,0 +1,9 @@
+module Api
+  module Subcollections
+    module PhysicalRacks
+      def physical_racks_query_resource(object)
+        object.respond_to?(:physical_rack) ? PhysicalRack.where(:id => object.physical_rack_id) : []
+      end
+    end
+  end
+end
--- a/app/controllers/api/physical_storages_controller.rb
+++ b/app/controllers/api/physical_storages_controller.rb
@@ -1,5 +1,8 @@
 module Api
   class PhysicalStoragesController < BaseController
+    include Subcollections::PhysicalRacks
+    include Subcollections::PhysicalChassis
+
     def refresh_resource(type, id, _data = nil)
       raise BadRequestError, "Must specify an id for refreshing a #{type} resource" if id.blank?
 
--- a/config/api.yml
+++ b/config/api.yml
@@ -1772,6 +1772,7 @@
     :identifier: physical_chassis
     :options:
     - :collection
+    - :subcollection
     :verbs: *gp
     :klass: PhysicalChassis
     :subcollections:
@@ -1789,6 +1790,10 @@
         :identifier: physical_chassis_turn_on_loc_led
       - :name: turn_off_loc_led
         :identifier: physical_chassis_turn_off_loc_led
+    :subcollection_actions:
+      :get:
+      - :name: read
+        :identifier: physical_chassis_show_list
     :resource_actions:
       :get:
       - :name: read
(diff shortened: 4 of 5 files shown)


EsdrasVP:
@abellotti Could you check this out?

app/controllers/api/subcollections/physical_racks.rb
```diff
@@ -0,0 +1,9 @@
+module Api
+  module Subcollections
+    module PhysicalRacks
+      def physical_racks_query_resource(object)
+        object.respond_to?(:physical_rack) ? PhysicalRack.where(:id => object.physical_rack_id) : []
```
abellotti: can't this be object.physical_rack  instead of the PhysicalRack.where ... ?

app/controllers/api/subcollections/physical_chassis.rb
```diff
@@ -0,0 +1,9 @@
+module Api
+  module Subcollections
+    module PhysicalChassis
+      def physical_chassis_query_resource(object)
+        object.respond_to?(:physical_chassis) ? ManageIQ::Providers::PhysicalInfraManager::PhysicalChassis.where(:id => object.physical_chassis_id) : []
```
abellotti: same question as below

app/controllers/api/subcollections/physical_racks.rb
EsdrasVP: Actually no, because PhysicalRack has many PhysicalStorages and not the inverse. So if we call `object.physical_rack` the code will break. That's why, instead, we use the reference `physical_rack_id` which denotes the PhysicalRack that the PhysicalStorage belongs to.

bdunne: @EsdrasVP What will break?  It looks like the relationships are defined in both directions...
```
app/models/physical_chassis.rb:9:  belongs_to :physical_rack, :foreign_key => :physical_rack_id, :inverse_of => :physical_chassis
app/models/physical_rack.rb:4:  belongs_to :ext_management_system, :foreign_key => :ems_id, :inverse_of => :physical_racks,
app/models/physical_rack.rb:6:  has_many :physical_chassis, :dependent => :nullify, :inverse_of => :physical_rack
app/models/physical_rack.rb:7:  has_many :physical_servers, :dependent => :nullify, :inverse_of => :physical_rack
app/models/physical_rack.rb:8:  has_many :physical_storages, :dependent => :nullify, :inverse_of => :physical_rack
app/models/physical_server.rb:22:  belongs_to :physical_rack, :foreign_key => :physical_rack_id, :inverse_of => :physical_servers
app/models/physical_storage.rb:4:  belongs_to :physical_rack, :foreign_key => :physical_rack_id, :inverse_of => :physical_storages
```

EsdrasVP: @bdunne They have, PhysicalRack and PhysicalChassis will have many PhysicalStorages, but the problem is that PhysicalStorage will only belong to one PhysicalRack/PhysicalChassis. So what happens when I use `object.physical_rack` is that a method `all` is called to return a list of PhysicalRacks/PhysicalChassis when this list doesn't exist and only a object is returned. So when I access `/api/physical_storages/:id/physical_racks`, for instance, the following error is presented:

```
{
    "error": {
        "message": "undefined method `all' for #<PhysicalRack:0x007f1bf0c52098>",
        "klass": "NoMethodError"
    }
}
```

EsdrasVP:
@abellotti Is there any other reservation regarding this?

app/controllers/api/subcollections/physical_racks.rb
abellotti: if there's one and only one, the subcollection does not make sense here, it can currently be accessed via GET /api/physical_storages/:id?attributes=physical_rack
as shown by the OPTIONS /api/physical_storages in the relationships section.

EsdrasVP: > if there's one and only one, the subcollection does not make sense here, it can currently be accessed via GET /api/physical_storages/:id?attributes=physical_rack
> as shown by the OPTIONS /api/physical_storages in the relationships section.

@abellotti @bdunne This makes sense, I spoke with @rodneyhbrown7 and we agreed to use the current method to access the physical storage relationships. So this PR can be closed.

PR #428, opened 2018-07-23
EMS Cloud Refresh is missing

juliancheal:
When a user role only allows cloud and not infra refresh, that user
is denied refreshing as we only check for infra refresh.

Fixes BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1602413

This shows the roles defined:
screen shot 2018-07-23 at 17 50 39

### DIFF
--- a/config/api.yml
+++ b/config/api.yml
@@ -2120,6 +2120,7 @@
         :identifier: ems_infra_edit
       - :name: refresh
         :identifier: ems_infra_refresh
+        :identifier: ems_cloud_refresh
       - :name: delete
         :identifier: ems_infra_delete
       - :name: pause
@@ -2139,6 +2140,7 @@
         :identifier: ems_infra_edit
       - :name: refresh
         :identifier: ems_infra_refresh
+        :identifier: ems_cloud_refresh
       - :name: delete
         :identifier: ems_infra_delete
       - :name: import_vm
--- a/config/api.yml
+++ b/config/api.yml
@@ -2119,8 +2119,9 @@
       - :name: edit
         :identifier: ems_infra_edit
       - :name: refresh
-        :identifier: ems_infra_refresh
-        :identifier: ems_cloud_refresh
+        :identifier:
+        - ems_infra_refresh
+        - ems_cloud_refresh
       - :name: delete
         :identifier: ems_infra_delete
       - :name: pause
@@ -2139,8 +2140,9 @@
       - :name: edit
         :identifier: ems_infra_edit
       - :name: refresh
-        :identifier: ems_infra_refresh
-        :identifier: ems_cloud_refresh
+        :identifier:
+        - ems_infra_refresh
+        - ems_cloud_refresh
       - :name: delete
         :identifier: ems_infra_delete
       - :name: import_vm
--- a/spec/requests/providers_spec.rb
+++ b/spec/requests/providers_spec.rb
@@ -949,7 +949,7 @@ describe "Providers API" do
     end
 
     it "supports cloud provider refresh" do
-      api_basic_authorize collection_action_identifier(:providers, :refresh)
+      api_basic_authorize 'ems_cloud_refresh'
 
       provider = FactoryGirl.create(:ext_management_system, sample_amazon.symbolize_keys.except(:type, :credentials))
       provider.update_authentication(:default => default_credentials.symbolize_keys)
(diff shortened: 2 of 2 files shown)


juliancheal:
# TODO

Even with `ems_cloud_refresh` added, we still need additional checks. As I believe just by having either `ems_cloud_refresh` or `ems_infra_refresh` in a users role, they can perform refreshes over all Provider types, which may not be the desired outcome.

juliancheal:
@miq-bot assign @gtanzillo

juliancheal:
@gtanzillo I need to add some tests to confirm/deny my comment 
https://github.com/ManageIQ/manageiq-api/pull/428#issuecomment-407122063, but apart from that, this'll fix the BZ.

juliancheal:
Also I don't like where I've added `ems_cloud_refresh`, I would prefer a separate `ems_cloud_*` section.

juliancheal:

config/api.yml
```diff
@@ -2073,6 +2073,8 @@
         :identifier: ems_infra_new
       - :name: edit
         :identifier: ems_infra_edit
+      - :name: refresh
+        :identifier: ems_cloud_refresh
```
Fryguy: I've not seen multiple "names" in the list before, so I'm not sure what happens... Is there an example of this elsewhere in the file?  @abellotti ?

juliancheal: @Fryguy Yeah exactly, I just put it there "for now" so it was easier for me tbh to comment out `ems_infra_refresh` during testing 😂

juliancheal:
@Fryguy @abellotti I've moved the cloud refresh lines now.

abellotti:
We need tests.

juliancheal:
@abellotti Good reminder thanks. I assume there are existing tests I can look at to create new ones for this. Do you know what/where those tests are?

Sorry I'm still learning the API :)

config/api.yml
```diff
@@ -2081,6 +2081,8 @@
         :identifier: ems_infra_edit
       - :name: resume
         :identifier: ems_infra_edit
+      - :name: refresh
```
abellotti: this is off, we already have a refresh action, maybe your intent was to add the identifier to the refresh above on line 2076, similar to what query does to allow for 2 identifiers. so they can so a refresh if they have the ems_infra_refresh or ems_cloud_refresh role.  In which case we have the refresh test, we could add another test to make sure refresh works with ems_cloud_refresh too.

juliancheal: I moved to here after a comment by Jason. Should I put it underneath `identifier: ems_infra_refresh`

Yes my intent was to allow a user to refresh if they have either `ems_infra_refresh or ems_cloud_refresh role`

abellotti: yes please, that's the intent of the multiple identifiers in the actions list (user must have at least one such role).

spec/requests/providers_spec.rb
```diff
@@ -935,6 +948,17 @@ def failed_auth_action(id)
       expect_single_action_result(failed_auth_action(provider.id.to_s).symbolize_keys)
     end
 
+    it "supports cloud provider refresh" do
```
juliancheal: @abellotti do you think this is ok?

spec/requests/providers_spec.rb
```diff
@@ -935,6 +948,17 @@ def failed_auth_action(id)
...
+      api_basic_authorize collection_action_identifier(:providers, :refresh)
```
abellotti: I think for this we want the specific identifier, maybe api_basic_authorize('ems_cloud_refresh')

juliancheal:
@gtanzillo The issues breaking the API repo have been fixed, so this is now green, if @abellotti agrees, should we merge this?

juliancheal:
Ping @gtanzillo

spec/requests/providers_spec.rb
```diff
@@ -935,6 +948,17 @@ def failed_auth_action(id)
       expect_single_action_result(failed_auth_action(provider.id.to_s).symbolize_keys)
     end
 
+    it "supports cloud provider refresh" do
+      api_basic_authorize 'ems_cloud_refresh', collection_action_identifier(:providers, :refresh)
```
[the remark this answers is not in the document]
abellotti: If we're just testing the cloud refresh, we only want to test the ems_cloud_refresh identifier so probably drop the , collection_action_identifier(:providers, :refresh).

juliancheal:
@abellotti @gtanzillo change made, all green ❇️

abellotti:
Thanks @juliancheal for the update, LGTM!! 👍

PR #431, opened 2018-07-25
Added verify_ssl to the list of valid attributes

mkanoor:
When we add a provider it might be using self signed certs
We need a way to disable the certificate check with verify_ssl.
The UI provides a check box to disable certifcate check we would
need something similar for REST API


mkanoor:
@bdunne @abellotti 
Please review

abellotti:
is verify_ssl for all or specific to the provider you're working with ?  note, providers can also tell us their specific attributes via the api_allowed_attributes method.

bdunne:
I restarted the tests after https://github.com/ManageIQ/manageiq-api/pull/432 was merged, but I think this will need a rebase or at least a close / reopen to get the tests passing.

PR #432, opened 2018-07-26
Fix logging tests after merging ManageIQ/manageiq #17227

bdunne:
- use a `let` to avoid the `@` variable
- use `around` rather than `before`and `after`


bdunne:
@abellotti Please review

PR #434, opened 2018-07-27
Use the new universal methods for suspending/resuming a provider

skateman:
Parent issue: https://github.com/ManageIQ/manageiq/issues/17489
Depends on: https://github.com/ManageIQ/manageiq/pull/17452

cc @slemrmartin 

https://bugzilla.redhat.com/show_bug.cgi?id=1455145

### DIFF
--- a/app/controllers/api/providers_controller.rb
+++ b/app/controllers/api/providers_controller.rb
@@ -92,7 +92,7 @@ module Api
 
     def pause_resource(type, id, _data)
       provider = resource_search(id, type, collection_class(type))
-      provider.disable!
+      provider.pause!
       action_result(true, "Paused #{provider_ident(provider)}")
     rescue => err
       action_result(false, "Could not pause Provider - #{err}")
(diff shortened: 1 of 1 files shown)


skateman:
@slemrmartin are these methods still safe to use?

slemrmartin:
@skateman yes, `pause!` and `resume!` are still entry points for this PR

jvlcek:
@skateman Since the goal is to try to have every PR include a related spec test perhaps you could update https://github.com/ManageIQ/manageiq-api/blob/master/spec/requests/providers_spec.rb to ensure the new methods are received.

It shouldn't be too hard to add.

Besides that 👍 LGTM

skateman:
@jvlcek there are already tests for this feature, I'm just changing the method to be called on the model. So I don't really understand what else should I test here, can you please clarify?

jvlcek:
@skateman Understood and agreed. I was just thinking that maybe an
`expect(X).to receive(:pause!)` could be added to the `pause a provider` test but on second thought it's not overly valuable.

I'm OK with this PR as it is.

JoeV

Issue #381, opened 2018-05-15
RUD for MiqTemplate (or just read & update)


himdel:
Recently @andyvesel created API endpoints for editing Cloud Templates - [create #337](https://github.com/ManageIQ/manageiq-api/pull/337), [read #130](https://github.com/ManageIQ/manageiq-api/pull/130), [update #341](https://github.com/ManageIQ/manageiq-api/pull/341) and [delete #328](https://github.com/ManageIQ/manageiq-api/pull/328).

However, there is no concept of editing **cloud** templates in the UI, there is just a generic MiqTemplate editor (`VmOrTemplate` really).

For a generic MiqTemplate:
  * the only fields that can be currently edited in the UI are `name` and `description`
  * it can't be created, only updated

So, we need to have API endpoint/endpoints to edit both kinds, with the same interface (for the shared fields at least), as the one introduced in #341.

(Delete would fit logically together, but the UI is currently not using the API for deletion.)

https://bugzilla.redhat.com/show_bug.cgi?id=1487114

EDIT: Related to https://github.com/ManageIQ/manageiq-ui-classic/issues/3944, see discussion there

andyvesel:
@skateman @abellotti I was trying to update a template with PUT/PATCH requests, but it only updates attributes for local MIQ database and doesn't make any calls to Fog, so the image can't be updated on the Openstack side. Do we need it to be implemented this way? 
I'm kind of stuck on this issue. So I'll appreciate any thorough and obvious details about this.

skateman:
IMO we should somehow merge the two functionalities (i.e. the one @andyvesel created and the universal one I was trying to do) and call the openstack-specific stuff when needed.

himdel:
What @skateman said :)

Just, 2 more points:

* please make sure you come up with an interface which can be generalized to any provider, no openstack-specific code in the API
* the UiWorker is not guaranteed to be able to communicate with any external provider, and the same goes for the API => the API call should create a task to talk to the provider

martinpovolny:
Ok, so this is relevant for:

https://github.com/ManageIQ/manageiq-ui-classic/issues/3944

Where we need to be able to create openstack images. Do we need an API action that can upload the image? Or just pass in a URL of it?

martinpovolny:
Anyway it seems that for this to move forward work in needed from the core and provider (openstack) side.

himdel:
[...]

PR #435, opened 2018-07-27
Allow the PUT/PATCH editing of templates and add a test for it

skateman:
This will enable to edit the name/description of `MiqTemplate` entities through the API. The final goal is to join this with the OpenStack-related `CloudTemplate` controller somehow.

Related issue: #381

@miq-bot assign @abellotti

### DIFF
--- a/spec/requests/templates_spec.rb
+++ b/spec/requests/templates_spec.rb
@@ -29,6 +29,32 @@ RSpec.describe "Templates API" do
     end
   end
 
+  context "editing a template" do
+    let!(:template) { FactoryGirl.create(:template, :name => 'foo', :description => 'bar') }
+    before { api_basic_authorize(action_identifier(:templates, :edit)) }
+    subject { send(req, api_template_url(nil, template), :params => params) }
+
+    describe "PUT /api/templates/:c_id" do
+      let(:req) { :put }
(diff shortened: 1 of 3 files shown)


abellotti:
maybe add a PATCH test, thanks.

spec/requests/templates_spec.rb
```diff
@@ -29,6 +29,30 @@
...
+        expect { subject }.to change { template.reload.name }.to('baz')
+        expect(response).to have_http_status(:ok)
+      end
+    end
+
+    describe "PATCH /api/templates/:c_id" do
+      let(:params) { {:description => 'baz'} }
+
+      it 'edits the template description' do
+        expect { subject }.to change { template.reload.description }.to('baz')
```
abellotti: subject still exercises the put above on line 35, we need a patch for this.

skateman: Right, is it okay if I do with `send`? Or should I copy some code (that I don't like) :wink: 
```ruby
describe 'PATCH' do
  let(:req) { :patch }
  ...
end
describe 'PUT' do
  let(:req) { :put }
  ...
end
subject { send(req) ... }
```

abellotti:
Thanks @skateman for the API Enhancement.  LGTM!! 👍

PR #437, opened 2018-07-30
[RFE] Use Vmdb::Appliance.PRODUCT_NAME instead of calling i18n

skateman:
This [constant](https://github.com/ManageIQ/manageiq/pull/17773/files#diff-b37e88904167347189bb2e4a9a143876R24) allows product name customization for the SUI.

@miq-bot assign @abellotti 

Backend part: https://github.com/ManageIQ/manageiq/pull/17773 (not dependent)
RFE BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1471301

### DIFF
--- a/app/controllers/api/api_controller.rb
+++ b/app/controllers/api/api_controller.rb
@@ -71,7 +71,7 @@ module Api
 
     def product_info
       {
-        :name                 => I18n.t("product.name"),
+        :name                 => Vmdb::Appliance.PRODUCT_NAME,
         :name_full            => I18n.t("product.name_full"),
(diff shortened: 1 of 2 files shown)


app/controllers/api/api_controller.rb
```diff
@@ -71,7 +71,7 @@ def server_info
 
     def product_info
       {
-        :name                 => I18n.t("product.name"),
+        :name                 => Vmdb::Appliance.PRODUCT_NAME,
```
AparnaKarve: Looks OK with relative to the ManageIQ/manageiq#17773  change

PR #438, opened 2018-07-30
[RFE] Add the /api/product_info route with product and branding info

skateman:
If we want to customize the product name and branding, it's necessary to have a request that returns the product info without authentication. Also I would like to use this for the about modal, so adding the branding to the `/api` route as well.

RFE BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1471301

### DIFF
--- a/app/controllers/api/api_controller.rb
+++ b/app/controllers/api/api_controller.rb
@@ -6,18 +6,22 @@ module Api
 
     def index
       res = {
-        :name         => ApiConfig.base.name,
-        :description  => ApiConfig.base.description,
-        :version      => ManageIQ::Api::VERSION,
-        :versions     => entrypoint_versions,
-        :settings     => user_settings,
-        :identity     => auth_identity,
-        :server_info  => server_info,
-        :product_info => product_info
+        :name          => ApiConfig.base.name,
+        :description   => ApiConfig.base.description,
+        :version       => ManageIQ::Api::VERSION,
+        :versions      => entrypoint_versions,
+        :settings      => user_settings,
+        :identity      => auth_identity,
+        :server_info   => server_info,
+        :product_info  => product_info_data
       }
       res[:authorization] = auth_authorization if attribute_selection.include?("authorization")
       res[:collections]   = entrypoint_collections
       render_resource :entrypoint, res
+    end
+
+    def product_info
+      render_resource :product_info, product_info_data
     end
 
     private
@@ -70,14 +74,29 @@ module Api
       }
     end
 
-    def product_info
+    def product_info_data
       {
         :name                 => Vmdb::Appliance.PRODUCT_NAME,
         :name_full            => I18n.t("product.name_full"),
         :copyright            => I18n.t("product.copyright"),
         :support_website      => I18n.t("product.support_website"),
         :support_website_text => I18n.t("product.support_website_text"),
+        :branding_info        => branding_info
       }
+    end
+
+    def image_path(image)
+      ActionController::Base.helpers.image_path(image)
+    rescue Sprockets::FileNotFound # UI isn't loaded, we don't want images
+      nil
+    end
+
+    def branding_info
+      {
+        :brand      => Settings.server.custom_brand || image_path('layout/brand.svg'),
+        :logo       => Settings.server.custom_logo || image_path('layout/login-screen-logo.png'),
+        :login_logo => Settings.server.custom_login_logo.presence
+      }.compact
     end
 
     def plugin_info
--- a/spec/requests/entrypoint_spec.rb
+++ b/spec/requests/entrypoint_spec.rb
@@ -69,9 +69,31 @@ RSpec.describe "API entrypoint" do
         "name_full"            => I18n.t("product.name_full"),
         "copyright"            => I18n.t("product.copyright"),
         "support_website"      => I18n.t("product.support_website"),
-        "support_website_text" => I18n.t("product.support_website_text"),
+        "support_website_text" => I18n.t("product.support_website_text")
       )
     )
+
+    # This test will fail if you have the UI checked out and built with yarn
+    expect(response.parsed_body['product_info']['branding_info']).to eq({})
+  end
+
+  context 'UI is available' do
+    it 'product_info contains branding_info' do
+      api_basic_authorize
+
+      expect(ActionController::Base.helpers).to receive(:image_path).at_least(2).times.and_return("foo")
+
+      get api_entrypoint_url
+
+      expect(response.parsed_body).to include(
+        "product_info" => a_hash_including(
+          "branding_info" => a_hash_including(
+            "brand"      => "foo",
+            "logo"       => "foo"
+          )
+        )
+      )
+    end
   end
 
   it "will squeeze consecutive slashes in the path portion of the URI" do
--- a/app/controllers/api/base_controller.rb
+++ b/app/controllers/api/base_controller.rb
@@ -20,10 +20,10 @@ module Api
     include ActionController::HttpAuthentication::Basic::ControllerMethods
 
     before_action :log_request_initiated
-    before_action :require_api_user_or_token, :except => [:options]
-    before_action :set_gettext_locale, :set_access_control_headers, :parse_api_request, :log_api_request,
-                  :validate_api_request
-    before_action :validate_api_action, :except => [:options]
+    before_action :require_api_user_or_token, :except => [:options, :product_info]
+    before_action :set_gettext_locale, :set_access_control_headers, :parse_api_request, :log_api_request
+    before_action :validate_api_request, :except => [:product_info]
+    before_action :validate_api_action, :except => [:options, :product_info]
     before_action :validate_response_format, :except => [:destroy]
     before_action :ensure_pagination, :only => :index
     after_action :log_api_response
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -6,6 +6,7 @@ Rails.application.routes.draw do
     match "/", :to => "api#options", :via => :options
 
     get "/ping" => "ping#index"
+    get "/product_info" => "api#product_info"
 
     Api::ApiConfig.collections.each do |collection_name, collection|
       # OPTIONS action for each collection


juliancheal:
So we want to return something like

```
      res = {
        :name         => ApiConfig.base.name,
        :description  => ApiConfig.base.description,
        :version      => ManageIQ::Api::VERSION,
        :versions     => entrypoint_versions,
        :settings     => user_settings,
        :identity     => auth_identity,
        :server_info  => server_info,
        :product_info => product_info
      }
```

from the `index` method, but without needing authentication

skateman:
@juliancheal do we really need all those things? Some of them aren't even available without login.

juliancheal:
@skateman Well i didn't mean all the things my mistake.

app/controllers/api/api_controller.rb
```diff
@@ -1,7 +1,7 @@
 module Api
   class ApiController < Api::BaseController
     def options
-      head(:ok)
+      render_resource :product, product_info
```
abellotti: we need this under a product_info element, maybe just render :json => { :product_info => product_info }

skateman:
@juliancheal a *"filtered"* version of this would make sense, I'd definitely need some extra stuff from settings as well. The question here is if it's okay to use the OPTIONS for this? cc @abellotti @bdunne

app/controllers/api/api_controller.rb
skateman: Makes sense, I need 3 more items `Settings.server.custom_logo`, `Settings.server.custom_login_logo` and `Settings.server.custom_brand` (all boolean values). Do you have any idea how to structure them? Or is it okay to add it into `product_info`?

juliancheal:
@skateman yes a filtered version would be perfect. I wonder if it could just be on `index`. Return one set if authorised and another if not? It seems `options` is used more for "describe the communication options" unless I'm mistaken.

skateman:
@juliancheal the authorization is a `before_action` and it's a lot harder to catch it than to work with the `OPTIONS`

app/controllers/api/api_controller.rb
abellotti: product_info is more geared toward MiQ itself, above may fit better under server_info, or maybe even a new section like brand_info.

bdunne:
Any reason that this wouldn't be under a new `/api/product_info` endpoint?  Also, I think we already allow GET on `/api/ping` without authentication.

skateman:
@bdunne sounds reasonable, and what about the [structure of the data](https://github.com/ManageIQ/manageiq-api/pull/438#discussion_r206160780)?

app/controllers/api/api_controller.rb
bdunne: Maybe we can make use of https://github.com/ManageIQ/manageiq/pull/17755 for the version info?  I'm not sure what the rest of the information is.
Why do we want to return booleans for things like `Settings.server.custom_logo`? (I don't know how it's used)  Wouldn't it make more sense to have a `logo` key with a value of the href to the image or something like that?

skateman: @bdunne we want them booleans because they are already booleans, in the OPS UI there's a checkbox for using the custom image. The image itself is on a fixed location, so the filename is static for all the three cases. So historical reasons, that's why :)

skateman: @abellotti I think I'll go with the `brand_info`, so
```ruby
{
  :product_info => product_info,
  :brand_info => ...
}
```

app/controllers/api/api_controller.rb
```diff
@@ -20,6 +20,15 @@ def index
       render_resource :entrypoint, res
     end
 
+    def product_info
```
bdunne: Won't this be overridden by https://github.com/ManageIQ/manageiq-api/pull/438/files#diff-0ef0779741ad3209b46314a66c353351R89 ?

Fryguy: Is this also implying that we already have this endpoint?

app/controllers/api/api_controller.rb
```diff
@@ -46,6 +55,14 @@ def auth_identity
       }
     end
 
+    def branding_info
+      {
+        :custom_brand      => Settings.server.custom_brand,
+        :custom_logo       => Settings.server.custom_logo,
+        :custom_login_logo => Settings.server.custom_login_logo
```
Fryguy: What are these settings and where do they come from?  The only one in the settings template is the `custom_logo`, and that one is just a boolean.

skateman: @Fryguy https://github.com/ManageIQ/manageiq/pull/17773 forgot to include, sorry

app/controllers/api/api_controller.rb
skateman: I should not make PRs when I'm tired :cry: I'll fix this tomorrow :confused: :sleeping:

bdunne: I don't think so, because the one below is private.

skateman: Yup, my bad ... fixed now.

app/controllers/api/api_controller.rb
```diff
@@ -69,13 +86,13 @@ def server_info
...
     end
 
-    def product_info
+    def product_info_data
       {
         :name                 => I18n.t("product.name"),
         :name_full            => I18n.t("product.name_full"),
         :copyright            => I18n.t("product.copyright"),
         :support_website      => I18n.t("product.support_website"),
-        :support_website_text => I18n.t("product.support_website_text"),
```
bdunne: Any reason this changed?

skateman: Nope...reverting.

skateman:
@bdunne @Fryguy @abellotti any other issues? Can this be merged?

bdunne:
Aside from the tests failing, it looks good to me

skateman:
@bdunne are the test failures relevant? I don't see it like that :(

skateman:
@bdunne it seems like we cannot call `ActionController::Base.helpers.image_path` from the API at all because the asset load_path isn't complete. Probably something from the UI classic repo isn't loaded. I think we don't want to depend on the UI repo at all if possible, so let's try something different.

What do you think about implementing a safe-lookup for the branding images? If the lookup error happens, i.e. assets aren't available, we don't have the UI and so we don't really need the branding at all, so we just don't return the fields in the response. And if there is no lookup error, the UI is intact and we return with the correct paths for the branding images.

app/controllers/api/api_controller.rb
Fryguy: Does this fix it though?  It seems this break API compatibility because the old product_info doesn't match the new one.

skateman: Oh, you're right, is it okay if I merge the `branding_info` into `product_info`?

skateman:
According to @Fryguy's comment and discussing with @bdunne I updated the `product_info` route to return a flat hash that includes the `branding_info` if the UI is available, i.e. the asset_path won't fail with a `Sprockets::FileNotFound` exception.

app/controllers/api/api_controller.rb
```diff
@@ -70,16 +75,31 @@ def server_info
...
 
+    def branding_image_path(custom, default)
+      custom || ActionController::Base.helpers.image_path(default)
+    rescue Sprockets::FileNotFound # UI isn't loaded, we don't want images
+      nil
+    end
+
+    def branding_info
+      {
+        :brand      => branding_image_path(Settings.server.custom_brand, 'layout/brand.svg'),
```
bdunne: Why pass `Settings.server.custom_brand` into `branding_image_path` if the method doesn't do anything with it?

app/controllers/api/api_controller.rb
```diff
@@ -70,16 +75,31 @@ def server_info
...
+        :logo       => branding_image_path(Settings.server.custom_logo, 'layout/login-screen-logo.png'),
+        :login_logo => Settings.server.custom_login_logo ? Settings.server.custom_login_logo : nil
```
bdunne: Do we need the ternary operator here?

app/controllers/api/api_controller.rb
```diff
@@ -6,20 +6,25 @@ def options
...
-        :product_info => product_info
+        :name          => ApiConfig.base.name,
+        :description   => ApiConfig.base.description,
+        :version       => ManageIQ::Api::VERSION,
+        :versions      => entrypoint_versions,
+        :settings      => user_settings,
+        :identity      => auth_identity,
+        :server_info   => server_info,
+        :product_info  => product_info_data,
+        :branding_info => branding_info
```
bdunne: Do we still need this if it's also nested under `product_info`?

app/controllers/api/api_controller.rb
skateman: Yes because it can also be `false` and that's not caught by `compact`

app/controllers/api/api_controller.rb
skateman: I don't really understand this, why it would not do anything with it? The `branding_image_path` returns the path to the custom or default brand depending on what's available.

app/controllers/api/api_controller.rb
skateman: Well, we need to be able to access this without a login, so the SUI login screen can consume it.

app/controllers/api/api_controller.rb
bdunne: Does the method need the `branding_` prefix?  It is not branding specific.  If we don't need to look up the path for the value from settings, I would expect something like `:brand      => Settings.server.custom_brand || image_path('layout/brand.svg')`

app/controllers/api/api_controller.rb
bdunne: Doesn't https://github.com/ManageIQ/manageiq-api/pull/438/files#diff-b4f92b4e361ffab4fbb91c68eb3f77bbR23 remove the requirement for being logged in?

app/controllers/api/api_controller.rb
bdunne: I'm fine with putting the branding info under `product_info`, but did we somehow have `/product_info` before?
image

skateman: No, it's a new thing.

app/controllers/api/api_controller.rb
skateman: No, the `/api` route isn't available without login as it contains the `identity` part that needs a logged in user.

skateman:
@bdunne just to clarify, the `:product_info` available from the root `/api` route is not accessible without a login, even though the OPTIONS is allowed. I had to expose the `/api/product_info` explicitly, otherwise the login screen would not be able to consume it. For keeping the compatibility, the `product_info` under the root `/api` route and under the `/api/product_info` route is identical.

spec/requests/entrypoint_spec.rb
```diff
@@ -74,6 +74,25 @@
     )
   end
 
+  context 'UI is available' do
+    it 'product_info contains branding_info' do
+      api_basic_authorize
+
+      expect(ActionController::Base.helpers).to receive(:image_path).at_least(2).times.and_return("foo")
```
bdunne: Should the test above be modified or a new test added where this is not stubbed?

app/controllers/api/api_controller.rb
bdunne: > Do we still need this if it's also nested under product_info?

What I meant was that we'll have this, right?
```json
get api_entrypoint_url

{
  "name" : "ManageIQ",
  "description" : "words",
  ...,
  "product_info" : {
    ...,
    "branding_info" : {...}
  },
  "branding_info" : {...}
}
```

skateman: Oooh, you're right :roll_eyes: I don't know why I left that line there :cry:

spec/requests/entrypoint_spec.rb
skateman: I'll update the test above!

spec/requests/entrypoint_spec.rb
```diff
@@ -69,9 +69,31 @@
         "name_full"            => I18n.t("product.name_full"),
         "copyright"            => I18n.t("product.copyright"),
         "support_website"      => I18n.t("product.support_website"),
-        "support_website_text" => I18n.t("product.support_website_text"),
+        "support_website_text" => I18n.t("product.support_website_text")
       )
     )
+
+    # This test will fail if you have the UI checked out and built with yarn
+    expect(response.parsed_body['product_info']).not_to include('branding_info')
```
bdunne: It will be there, but `eq({})` and the test failure reflects that

simaishi:
@skateman Can this be `hammer/yes`?

skateman:
@miq-bot add_label hammer/yes

simaishi:
Hammer backport details:

```
$ git log -1
commit 72c603072bba9201715f4ad994a5745600b39704
Date:   Fri Oct 12 17:52:21 2018 -0400

    Merge pull request #438 from skateman/product-info-api
    
    [RFE] Add the /api/product_info route with product and branding info
    
    (cherry picked from commit a40b7095909651f7a5814d8f4cfc934ac775bb74)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1693757
```

PR #443, opened 2018-08-01
[GAPRINDASHVILI] fix test failures

bdunne:
Depends on https://github.com/ManageIQ/manageiq/pull/17787


bdunne:
@simaishi Please review.  There are some logging and query failures that I'm still looking into

PR #385, opened 2018-05-25
use request_admin_user?

kbrock:
~~Blocked on:~~
- Part of https://github.com/ManageIQ/manageiq/pull/17444

Remove generic `admin_user?` check and replace with `request_admin_user?`

https://bugzilla.redhat.com/show_bug.cgi?id=1090627

PR #447, opened 2018-08-09
Change request_controller special privileges

kbrock:
**BLOCKED BY:**
- [x] https://github.com/ManageIQ/manageiq/pull/17849

To make green, also look into:
- [x] ManageIQ/manageiq-api#451 or ManageIQ/manageiq-api#452

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1608554

---

In #385 we introduced feature `miq_request_superadmin`
*no more hardcoded admin role name - **YAY***

The UI uses a different feature, `miq_request_approval`
*Inconsistencies - **BOO***

https://github.com/ManageIQ/manageiq/pull/17849 fixes main repo's identifier
*Consistencies - **YAY***

API Tests Fail -- **BOO**

This PR changes feature identifiers used - **YAY**

---
For those concerned: `miq_request_superadmin` is deprecated

### DIFF
--- a/spec/requests/automation_requests_spec.rb
+++ b/spec/requests/automation_requests_spec.rb
@@ -46,7 +46,7 @@ describe "Automation Requests API" do
     end
 
     it "lists all the automation requests if you are admin" do
-      @group.miq_user_role = @role = FactoryGirl.create(:miq_user_role, :features => %w(miq_request_superadmin))
+      @group.miq_user_role = @role = FactoryGirl.create(:miq_user_role, :features => %w(miq_request_approval))
       other_user = FactoryGirl.create(:user)
       automation_request1 = FactoryGirl.create(:automation_request, :requester => other_user)
       automation_request2 = FactoryGirl.create(:automation_request, :requester => @user)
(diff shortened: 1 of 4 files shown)


app/controllers/api/automation_requests_controller.rb
```diff
@@ -44,12 +44,12 @@ def deny_resource(type, id, data)
 
     def find_automation_requests(id)
       klass = collection_class(:requests)
-      return klass.find(id) if User.current_user.miq_user_role.request_admin_user?
+      return klass.find(id) if User.current_user.role_allows?(:identifier => "miq_request_approval")
```
gtanzillo: Perhaps this can be put in a method like you did for the other ones. Something like `MiqUserRole#request_approver_user?` This way if we want to switch this logic to looks at a different feature we wouldn't have to change all these lines.

kbrock: Hmm. Not sure that was the problem in this case.

We have so many features. Not sure if we want to introduce a `*_user?` for every one of them.

Also, in the UI and API tend to use identifiers in most cases.
Not sure if this case in particular is worthy of a special method.

To be honest, I was looking forward to getting rid of all of them except for the super user one

jrafanie: I think request admin and report admin are extremely large parts of the UI warranting methods for these types of users.  Regardless, I'd prefer to have too many methods on user role than have so many hardcoded feature identifiers.

kbrock:
@gtanzillo darn - locally these specs are green for me.
I just re-pushed and await the 🍏 (or I guess the ❌ )

kbrock:
Merges as part of #454

PR #450, opened 2018-08-15
[DEMO] Notifications for ansible operations

skateman:
Includes @Ladas's PR #398 and adds the required `user_id` as a second argument to the `security_group_create` method.


skateman:
@Ladas first I'd like to test this, can you PM me some credentials for amazon?

PR #451, opened 2018-08-18
Set the cred.manager to an EMS not a Provider

agrare:
Credentials created by EmsRefresh set the credential.manager to the
AutomationManager not a Provider.  This causes issues where
credential.manager is expected to be an ExtManagementSystem type.

### DIFF
--- a/spec/requests/authentications_spec.rb
+++ b/spec/requests/authentications_spec.rb
@@ -1,5 +1,5 @@
 RSpec.describe 'Authentications API' do
-  let(:provider) { FactoryGirl.create(:provider_ansible_tower) }
+  let(:provider) { FactoryGirl.create(:automation_manager_ansible_tower) }
   let(:auth) { FactoryGirl.create(:ansible_cloud_credential, :resource => provider) }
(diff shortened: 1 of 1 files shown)


agrare:
cc @Fryguy @kbrock @jameswnl

spec/requests/authentications_spec.rb
```diff
@@ -1,5 +1,5 @@
 RSpec.describe 'Authentications API' do
-  let(:provider) { FactoryGirl.create(:provider_ansible_tower) }
+  let(:provider) { FactoryGirl.create(:automation_manager_ansible_tower) }
   let(:auth) { FactoryGirl.create(:ansible_cloud_credential, :resource => provider) }
   let(:auth_2) { FactoryGirl.create(:ansible_cloud_credential, :resource => provider) }
```
jameswnl: 👍 so if you fix the other parts in [here](https://github.com/jameswnl/manageiq-api/commit/35172b8dcf725287376f148113e48aeac0f0d3f6), other specs should pass as well.

agrare: Thanks @kbrock !

PR #452, opened 2018-08-19
Set the cred.manager to an EMS not a Provider

kbrock:
Introduced by ManageIQ/manageiq#17531 (but failure is actually due to a bad spec)

Credentials created by EmsRefresh set the credential.manager to the
AutomationManager not a Provider. This causes issues where
credential.manager is expected to be an ExtManagementSystem type.

Fixing test failures 99% stolen from: 

-  https://github.com/jameswnl/manageiq-api/tree/credential-spec
- https://github.com/ManageIQ/manageiq-api/pull/451


kbrock:
All remaining failures are solved by https://github.com/ManageIQ/manageiq-api/pull/447

agrare:
Cc @gtanzillo

kbrock:
Merged as part of #454

PR #455, opened 2018-08-21
Add best fit API for transformations moving vms to openstack

bdunne:
Depends on: https://github.com/ManageIQ/manageiq/pull/17880
https://trello.com/c/b90ZCyNm/183-extend-cf-api-for-mapping-of-osp-flavors-and-security-groups

### DIFF
--- a/app/controllers/api/transformation_mappings_controller.rb
+++ b/app/controllers/api/transformation_mappings_controller.rb
@@ -22,6 +22,19 @@ module Api
       end
     end
 
+    def vm_flavor_fit_resource(_type, _id, data)
+      data["mappings"].collect do |mapping|
+        source      = Api::Utils.resource_search_by_href_slug(mapping["source_href_slug"])
+        destination = Api::Utils.resource_search_by_href_slug(mapping["destination_href_slug"])
+        fit         = TransformationMapping::CloudBestFit.new(source, destination)
+        {
+          :source_href_slug => mapping["source_href_slug"],
+          :best_fit         => Api::Utils.build_href_slug(Flavor, fit.best_fit_flavor.try(:id)),
+          :all_fit          => fit.available_fit_flavors.collect { |f| Api::Utils.build_href_slug(Flavor, f.id) }
+        }
+      end
+    end
+
     private
 
     def create_mapping_items(items)
--- a/config/api.yml
+++ b/config/api.yml
@@ -3366,6 +3366,8 @@
         :identifier: transformation_mapping_new
       - :name: delete
         :identifier: transformation_mapping_delete
+      - :name: vm_flavor_fit
+        :identifier: transformation_mapping_new
     :resource_actions:
       :get:
       - :name: read
--- a/spec/requests/transformation_mappings_spec.rb
+++ b/spec/requests/transformation_mappings_spec.rb
@@ -46,6 +46,46 @@ describe "Transformation Mappings" do
     end
   end
 
+  describe "POST /api/transformation_mappings/" do
+    context "with an appropriate role" do
+      def href_slug(obj)
+        Api::Utils.build_href_slug(obj.class, obj.id)
+      end
+
+      it "can map vms to openstack flavors" do
+        openstack = FactoryGirl.create(:ems_openstack)
+        _flavor1  = openstack.flavors.create!(:cpus => 1, :memory => 1.gigabytes)
+        flavor2   = openstack.flavors.create!(:cpus => 2, :memory => 2.gigabytes)
+        flavor3   = openstack.flavors.create!(:cpus => 4, :memory => 4.gigabytes)
+        vm1       = FactoryGirl.create(:vm_vmware, :hardware => FactoryGirl.create(:hardware, :cpu1x2, :ram1GB))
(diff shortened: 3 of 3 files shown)


bdunne:
@agrare Please review

app/controllers/api/transformation_mappings_controller.rb
```diff
@@ -22,6 +22,19 @@ def validate_vms_resource(type, id, data = {})
...
 
+    def vm_flavor_fit_resource(_type, _id, data)
+      data["mappings"].collect do |mapping|
+        source      = Api::Utils.resource_search_by_href_slug(mapping["source_href"])
+        destination = Api::Utils.resource_search_by_href_slug(mapping["destination_href"])
+        fit         = TransformationMapping::CloudBestFit.new(source, destination)
+        {
+          :source_href => mapping["source_href"],
+          :best_fit => Api::Utils.build_href_slug(Flavor, fit.best_fit_flavor.try(:id)),
+          :all_fit => fit.available_fit_flavors.collect { |f| Api::Utils.build_href_slug(Flavor, f.id) }
```
bdunne: ¯\\\_(ツ)_/¯ Naming is hard... if anyone has better suggestions I'm all ears :smile:

bdunne:
@abellotti Please review

abellotti:
I think this is fine, maybe take care of the rubocop warning above.

app/controllers/api/transformation_mappings_controller.rb
```diff
@@ -22,6 +22,19 @@ def validate_vms_resource(type, id, data = {})
       end
     end
 
+    def vm_flavor_fit_resource(_type, _id, data)
+      data["mappings"].collect do |mapping|
+        source      = Api::Utils.resource_search_by_href_slug(mapping["source_href"])
+        destination = Api::Utils.resource_search_by_href_slug(mapping["destination_href"])
```
abellotti: hmm, the above 2 seems to be confusing to be.  *_href implies an API href, i.e. http://localhost:3000/api/vms/1 and not the slug, which is normally named href_slug  (for vms/1).

If those are expected to be href_slugs, you can rename then accordingly, i.e. mapping["source_href_slug"] and mapping["destination_href_slug"].  

Optionally, you can take an object reference which can take either, so mapping["source"] and mapping["destination"] which identify the object as desired

``` 
{ "href" : "http://localhost:3000/api/vms/1" }
```

or

```
{ "href_slug" : "vms/1" }
```

config/api.yml
```diff
@@ -3366,6 +3366,8 @@
         :identifier: transformation_mapping_new
       - :name: delete
         :identifier: transformation_mapping_delete
+      - :name: vm_flavor_fit
+        :identifier: transformation_mapping_new
```
abellotti: only folks that can create new mappings can run this action ?

bdunne: ¯\\\_(ツ)_/¯ That's the only use case for it at the moment.

abellotti: Ahh ok 👍 we can always change later if need be.  Thanks!!

app/controllers/api/transformation_mappings_controller.rb
bdunne: Do you have an example of the last alternative you mentioned (object reference)?

abellotti: look at http://manageiq.org/docs/reference/latest/api/reference/users#creating-users for group, it's just a hash which can be one of many identifiers, id, href, description, etc.  or zone reference for creating a provider http://manageiq.org/docs/reference/latest/api/reference/providers#creating-providers, same idea.  you can look at any of the helper methods for those parse_<something>. In this case you'd add support for href_slug in addition to href for example.

abellotti:
Thanks @bdunne for the update.

@agrare could you take a 👀 when you get a chance ?  Thanks.

abellotti:
Thanks @agrare for the 👀 !!

Thanks @bdunne for the API Enhancement.  LGTM!! 👍

AparnaKarve:
@bdunne Thank you for this PR.

Recently we completed integrating this API with our frontend work.
While testing, we came across a couple edge cases where the `best_fit` array was nil and `all_fit` was an empty array.

The current UI design expects `best_fit` to always have some value.
That said, would it be possible to return the largest possible flavor in a situation where the best-fit-flavor logic is not able to determine the best fit flavor?
The largest flavor, as I understand in the context of v2v, would be the flavor with the largest `root_disk_size`

bdunne:
@AparnaKarve At the moment, we are not comparing the `root_disk_size` of the flavor, only the memory and CPU's.  It sounds like the VM in your edge case is larger in some way (memory and/or CPU) than any of the flavors in your Openstack environment.  `all_fit` will be an array of all of the flavors that match or exceed the current memory and CPU allocated to the VM.  `best_fit` will be the smallest of the flavors in `all_fit` or `nil` (when the array is empty).  I don't think it would be appropriate to return a flavor that does not at least match the currently allocated VM specs.

AparnaKarve:
Ok, I guess we would have to discuss this use case in detail (possibly during our Wednesday call)

PR #457, opened 2018-08-23
Add plugin display name

mzazrivec:
See ManageIQ/manageiq-ui-classic#4496 for primary motivation.

### DIFF
--- a/lib/manageiq/api/engine.rb
+++ b/lib/manageiq/api/engine.rb
@@ -31,6 +31,10 @@ module ManageIQ
         $api_log.info("  #{key.to_s.ljust([24, key.to_s.length].max, ' ')}: #{val}")
       end
 
+      def self.plugin_name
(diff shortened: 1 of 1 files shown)


lib/manageiq/api/engine.rb
```diff
@@ -34,6 +34,10 @@ def self.log_kv(key, val)
       def vmdb_plugin?
         true
       end
+
+      def self.plugin_name
+        _('API')
```
bdunne: Should this be more specific like `REST API`?

mzazrivec: Sure, why not. Fixed.

bdunne:
@abellotti or @gtanzillo Can you merge?

abellotti:
Thanks @mzazrivec for enhancing this.  👍

PR #464, opened 2018-08-31
Add custom_button_events subcollection for users, tenants and groups

skateman:
As we [discussed](https://github.com/ManageIQ/manageiq-ui-classic/pull/4514#issuecomment-417295590)  with @Hyperkid123 and @h-kataria, the new API-driven react/redux-based RBAC UI should include the custom button events via the API instead of using nested resources.

@miq-bot add_label enhancement, gaprindashvili/no
cc @Hyperkid123 you can start working with this in your PR

https://bugzilla.redhat.com/show_bug.cgi?id=1634053

### DIFF
--- a/config/api.yml
+++ b/config/api.yml
@@ -958,6 +958,16 @@
       - :name: delete
       :delete:
       - :name: delete
+  :custom_button_events:
+    :description: Custom Button Events
+    :options:
+    - :subcollection
+    :verbs: *g
+    :klass: CustomButtonEvent
+    :subcollection_actions:
+      :get:
+      - :name: read
+        :identifier: custom_button_events_show_list
   :custom_button_sets:
     :description: Custom Button Sets
     :identifier: ab_buttons_accord
@@ -1352,6 +1362,7 @@
     :klass: MiqGroup
     :identifying_attrs: description
     :subcollections:
+    - :custom_button_events
     - :tags
     :collection_actions:
       :get:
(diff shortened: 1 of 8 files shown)


h-kataria:
@skateman if a feature is marked as a hidden feature it is not displayed in the features tree but role_allows call returns true for the hidden features, in this case since we are not making role_allows call it should work.

skateman:
@miq-bot assign @abellotti

config/api.yml
```diff
@@ -948,6 +948,16 @@
       - :name: delete
       :delete:
       - :name: delete
+  :custom_button_events:
+    :description: Custom Button Events
+    :options:
+    - :subcollection
+    :verbs: *g
+    :klass: CustomButtonEvent
+    :subcollection_actions:
```
bdunne: Do we also need `subresource_actions` for `get`?

skateman: I'm not completely familiar what that would allow us, can you please help me?

bdunne: IIUC, `subcollection_actions` `get` allows `get /api/users/:id/custom_button_events`, but you would need `subresource_actions` `get` to allow `get /api/users/:id/custom_button_events/:id`.  Is that correct @abellotti?

skateman: @bdunne we don't need those, this feature is an admin-only debug thing

Hyperkid123:
@bdunne what is the progress with this PR?

skateman:
@Hyperkid123 unleash ~the monster~ react

PR #470, opened 2018-09-17
Expose SecurityGroups that are connected to a NetworkRouter

miha-plesko:
With this commit we let user access NetworkRouter's SecurityGroups. Example URL:

```
GET /api/network_routers/2/security_groups
```

RFE BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1574979

@miq-bot assign @abellotti

### DIFF
--- a/config/api.yml
+++ b/config/api.yml
@@ -1659,11 +1659,13 @@
     :identifier: network_router
     :options:
     - :collection
+    - :subcollection
     - :custom_actions
     :verbs: *gpd
     :klass: NetworkRouter
     :subcollections:
     - :tags
+    - :security_groups
     :collection_actions:
       :get:
       - :name: read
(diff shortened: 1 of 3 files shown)


miha-plesko:
Kicking Travis as I think the failure is not related to this patch.

config/api.yml
```diff
@@ -1659,11 +1659,13 @@
     :identifier: network_router
     :options:
     - :collection
+    - :subcollection
```
bdunne: IIRC, this means that `NetworkRouters` can be a subcollection of something else.

spec/requests/network_routers_spec.rb
```diff
@@ -138,4 +138,17 @@
...
+
+  describe 'security groups subcollection' do
+    it "can list a router's security groups" do
+      router = FactoryGirl.create(:network_router)
+      router.security_groups = [FactoryGirl.create(:security_group)]
+      api_basic_authorize(action_identifier(:network_routers, :read, :subcollection_actions, :get))
+
+      get(api_network_router_security_groups_url(nil, router))
+
+      expect(response.parsed_body).to include('subcount' => 1)
```
bdunne: this expectation should probably expect to include the `href` of the `SecurityGroup` too.

PR #472, opened 2018-09-19
Add basic CentralAdmin support to the API

bdunne:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1628658


app/controllers/api/vms_controller.rb
```diff
@@ -152,6 +159,7 @@ def scan_resource(type, id = nil, _data = nil)
         result
       end
     end
+    central_admin :scan_resource, :scan
```
carbonin: There are a few of these in here that I don't think we support from the main app.

Doing a quick look for `api_relay` in the core source I can only find `shutdown_guest`, `reboot_guest`, `reset`, `start`, `stop`, and `suspend`. We should keep the actions we support consistent between the UI and API.

bdunne: Yeah, they were simple, so I added them.

bdunne: Also, I'd like to add all future central admin features through the api and maybe even deprecate the central admin support in the models.

app/controllers/api/mixins/central_admin.rb
```diff
@@ -0,0 +1,29 @@
...
+            else
+              region_number = ApplicationRecord.id_to_region(id)
+              InterRegionApiMethodRelay.exec_api_call(region_number, @req.subject.to_sym, action, api_args, id)
+            end
+          end
+        end
+      end
+    end
+  end
+end
```
jrafanie: enddddddd

app/controllers/api/vms_controller.rb
jrafanie: I wouldn't introduce new behavior just to make it complete.  We can add the new ones later when they're needed.

jrafanie:
Wow, and this works?  Looks good.  My only concern is what @carbonin brought up above

bdunne:
> Wow, and this works?

For better or worse, it does the same thing that the central admin through the classic-ui does.  So, the actions get passed down properly, but the response returned to the user is garbage.

Valid action:
```
$ api -l https://global_region post /api/vms/2
{
  "action" : "stop"
}

{
  "error": {
    "message": "undefined method `keys' for nil:NilClass",
    "klass": "NoMethodError"
  }
}

$ api -l https://remote_region post /api/vms/2
{
  "action" : "stop"
}

{
  "success": true,
  "message": "VM id:2 name:'Apache_PHP_rh_summit_002' stopping",
  "task_id": "4",
  "task_href": "https://remote_region/api/tasks/4",
  "href": "https://remote_region/api/vms/2"
}
```

Invalid action:
```
$ api -l https://global_region post /api/vms/3
{
  "action" : "start"
}

{
  "error": {
    "message": "The VM is retired",
    "klass": "InterRegionApiMethodRelay::InterRegionApiMethodRelayError"
  }
}

$ api -l https://remote_region post /api/vms/3
{
  "action" : "start"
}

{
  "success": false,
  "message": "The VM is retired",
  "href": "https://remote_region/api/vms/3"
}
```

I'm investigating what needs to be done to make the response useful.

jrafanie:
> I'm investigating what needs to be done to make the response useful.

Interesting edge cases you're testing. It's a great idea to try vms in various states.

app/controllers/api/vms_controller.rb
carbonin: I still think we should leave out the extra methods here until they are also supported through the classic UI. If this PR is meant to solve a particular BZ we should limit the scope that issue rather than also adding new features.

bdunne: Since enabling them all is so easy, I figured I would just do it proactively rather than waiting for another high severity customer case saying Action X doesn't work.

carbonin: Right, but as I understand it this was because of missing operations in the API vs the classic UI. If you really do want to add the new ones, then please create an additional PR to add the same operations in core for consistency.

jrafanie: I can see the value in separating the bugfix from the enhancement part of this PR. 👍

bdunne: Okay, I removed the things we don't central admin in the UI.  I'll open a followup PR to add those.

bdunne:
> @bdunne and I want to create a new label and call it bughancement with the color of the existing two labels added together.

:rofl:

simaishi:
@bdunne `hammer/yes` and `gaprindashvili/yes` ?

bdunne:
@simaishi I think so.  Also related are: https://github.com/ManageIQ/manageiq/pull/17987 and https://github.com/ManageIQ/manageiq/pull/18036

simaishi:
Hammer backport details:

```
$ git log -1
commit 60cb7c906f7ec9724e5ac56f76e2558bfae1fd0b
Date:   Wed Oct 3 08:56:43 2018 -0400

    Merge pull request #472 from bdunne/central_admin
    
    Add basic CentralAdmin support to the API
    
    (cherry picked from commit 4dbf6b0bbb3cc10be69d6259c069e385ee7bc10c)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1628658
```

simaishi:

```
$ git log -1
commit 96b8096ee80d56ac45faee5fe73328443a4ff085
Date:   Wed Oct 3 08:56:43 2018 -0400

    Merge pull request #472 from bdunne/central_admin
    
    Add basic CentralAdmin support to the API
    
    (cherry picked from commit 4dbf6b0bbb3cc10be69d6259c069e385ee7bc10c)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1635764
```

PR #473, opened 2018-09-20
Display plugins in the product_info under the root route

skateman:
Would be nice to have this to fix https://github.com/ManageIQ/manageiq-ui-classic/issues/4581.
@miq-bot add_label gaprindashvili/no

*GET /api*
```json
{
   ...
    "server_info": {
    "version": "master",
     "build": "unknown_8234e5eb01",
     "appliance": null,
     "time": "2018-09-20T13:06:37Z",
     "server_href": "http://localhost:3000/api/servers/1",
     "zone_href": "http://localhost:3000/api/zones/1",
     "region_href": "http://localhost:3000/api/regions/1",
     "enterprise_href": "http://localhost:3000/api/enterprises/1",
     "plugins" : {
       "ENGINE": {
         "display_name": "PLUGIN NAME",
         "version": "PLUGIN_VERSION",
       },
       ...
     },
     ...
}
```

### DIFF
--- a/app/controllers/api/api_controller.rb
+++ b/app/controllers/api/api_controller.rb
@@ -65,7 +65,8 @@ module Api
         :server_href     => "#{@req.api_prefix}/servers/#{MiqServer.my_server.id}",
         :zone_href       => "#{@req.api_prefix}/zones/#{MiqServer.my_server.zone.id}",
         :region_href     => "#{@req.api_prefix}/regions/#{MiqRegion.my_region.id}",
-        :enterprise_href => "#{@req.api_prefix}/enterprises/#{MiqEnterprise.my_enterprise.id}"
+        :enterprise_href => "#{@req.api_prefix}/enterprises/#{MiqEnterprise.my_enterprise.id}",
+        :plugins         => plugin_info
       }
     end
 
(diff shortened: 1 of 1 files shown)


juliancheal:
LGTM, I'm not sure I know the _The Right Way_™, but this looks good.

himdel:
@skateman A `{name:version}` hash is not enough for the info we're displaying..

We would need something like...

```
{
  name: {
    version,
    plugin_name: engine.respond_to?(:plugin_name) ? engine.plugin_name : engine.to_s.gsub(/ManageIQ::|::Engine/, ''),
  }
}
```

skateman:
What about an array of hashes? :wink:

Fryguy:
I feel like this should not be a new API endpoint, but instead part of /api, but curious what @abellotti thinks

skateman:
@fryguy even better for me, much easier to implement :)

abellotti:
If we're not exposing individual resources, then yeah, this is more like info about the server itself, so maybe something as follows:

GET /api

```
{
   ...
    "server_info": {
    "version": "master",
     "build": "unknown_8234e5eb01",
     "appliance": null,
     "time": "2018-09-20T13:06:37Z",
     "server_href": "http://localhost:3000/api/servers/1",
     "zone_href": "http://localhost:3000/api/zones/1",
     "region_href": "http://localhost:3000/api/regions/1",
     "enterprise_href": "http://localhost:3000/api/enterprises/1",
     "plugins" : {
          ...
     },
     ...
}
```

skateman:
Okay, changing it...

skateman:
@abellotti updated based on your suggestion

app/controllers/api/api_controller.rb
```diff
@@ -79,6 +80,15 @@ def product_info
       }
     end
 
+    def plugin_info
+      Vmdb::Plugins.versions.each_with_object({}) do |(engine, version), hash|
+        hash[engine.to_s] = {
+          :name    => engine.respond_to?(:plugin_name) ? engine.plugin_name : engine.to_s.gsub(/ManageIQ::|::Engine/, ''),
```
bdunne: Should this be `:display_name`?

PR #475, opened 2018-09-21
[GAPRINDASHVILI] Set zone name to "default" when creating with factory so that it matches settings.yml

gtanzillo:
This fixes tests failures like

```
Zones /api/zones/:id/settings with an existing settings change does not allow an authenticated non-super-admin user to delete settings
     Failure/Error: raise "configuration invalid" unless VMDB::Config::Validator.new(new_settings).valid?
     RuntimeError:
       configuration invalid
     # ./spec/manageiq/lib/vmdb/settings.rb:50:in `save!'
     # ./spec/manageiq/app/models/mixins/configuration_management_mixin.rb:13:in `add_settings_for_resource'
     # ./spec/requests/zones_spec.rb:60:in `block (4 levels) in <top (required)>'
```
Because zone name does not match default value in settings.yml

Note - on master https://github.com/ManageIQ/manageiq/pull/17140 removed `:zone: default` from settings.yml so the failure does not exist there.

I was considering backporting that PR but figured a surgical fix to the failing test was lower risk.

### DIFF
--- a/spec/requests/regions_spec.rb
+++ b/spec/requests/regions_spec.rb
@@ -42,10 +42,10 @@ describe "Regions API" do
   end
 
   describe "/api/regions/:id/settings" do
-    let(:region_number) { ApplicationRecord.my_region_number + 1 }
+    let(:region_number) { ApplicationRecord.my_region_number }
     let(:id) { ApplicationRecord.id_in_region(1, region_number) }
     let(:region) { FactoryGirl.create(:miq_region, :id => id, :region => region_number) }
-    let(:zone) { FactoryGirl.create(:zone, :id => id) }
+    let(:zone) { FactoryGirl.create(:zone, :id => id, :name => "default") }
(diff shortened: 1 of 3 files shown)


Fryguy:
@bdunne Please review.  Not sure if this is the best way to do this as opposed to using Zone.seed, create_miq_server_zone, or traits.

gtanzillo:
Thanks @bdunne, there are a couple of other places where I need to do this. Working on that now.

spec/requests/regions_spec.rb
```diff
@@ -42,10 +42,10 @@
   end
 
   describe "/api/regions/:id/settings" do
-    let(:region_number) { ApplicationRecord.my_region_number + 1 }
```
gtanzillo: Needed to make this change too because zones for the current region are created in the EVM test setup. From what I can tell, these test do not rely on having a new region. They just need a region.

gtanzillo:
Closing this PR in favor of backporting ManageIQ/manageiq#17140 because there are other places where tests are failing the same way.

PR #476, opened 2018-09-24
Deny standalone service template ordering when product setting is enabled

eclarizio:
The issue this PR fixes is that currently, you can simply POST to the service_template order endpoint with whatever data you want (ideally, there is sort of a flow, i.e. you request the dialog, you may request any dialog field refreshes, and *then* you can POST to the service_template order endpoint with the data you've gathered from the previous requests). With any random data, it won't really work, but for example if a dropdown only contains the options "A" or "B", but the API request is submitted with "C", it is accepted.

This change, along with a product setting called `allow_api_service_ordering`, that the users must set to `false` first (as it is defaulted to `true` via [this PR](https://github.com/ManageIQ/manageiq/pull/18029)), should ensure that only the API calls that come from the UIs will be accepted. Otherwise everything will behave as it does now. This is mostly just a quick fix solution until we can potentially implement RBAC for API users, or add code to the model to validate the values against expected values, but those options are both a lot more involved.

/cc @tinaafitz 

https://bugzilla.redhat.com/show_bug.cgi?id=1632416

### DIFF
--- a/app/controllers/api/mixins/service_templates.rb
+++ b/app/controllers/api/mixins/service_templates.rb
@@ -3,11 +3,8 @@ module Api
     module ServiceTemplates
       def order_service_template(id, data, scheduled_time = nil)
         service_template = resource_search(id, :service_templates, ServiceTemplate)
-        unless api_request_allowed? && service_template.orderable?
-          raise BadRequestError, "#{service_template_ident(service_template)} cannot be ordered"
-        end
-        init_defaults = !request_from_ui? && Settings.product.run_automate_methods_on_service_api_submit
-        request_result = service_template.order(User.current_user, (data || {}), {:submit_workflow => request_from_ui?, :init_defaults => init_defaults}, scheduled_time)
+        raise BadRequestError, "#{service_template_ident(service_template)} cannot be ordered" unless orderable?
+        request_result = service_template.order(User.current_user, (data || {}), order_request_options, scheduled_time)
         errors = request_result[:errors]
         if errors.present?
           raise BadRequestError, "Failed to order #{service_template_ident(service_template)} - #{errors.join(", ")}"
@@ -17,9 +14,24 @@ module Api
 
       private
 
+      def orderable?
+        api_request_allowed? && service_template.orderable?
+      end
+
+      def api_request_allowed?
+        return true if request_from_ui?
+        Settings.product.allow_api_service_ordering
+      end
+
       def request_from_ui?
         return false if request.headers["x-auth-token"].blank?
         token_info.present?
+      end
+
+      def order_request_options
+        init_defaults = !request_from_ui? && Settings.product.run_automate_methods_on_service_api_submit
+
+        {:submit_workflow => request_from_ui?, :init_defaults => init_defaults}
       end
 
       def token_info
@@ -29,15 +41,6 @@ module Api
 
       def service_template_ident(st)
         "Service Template id:#{st.id} name:'#{st.name}'"
-      end
-
-      def api_request_allowed?
-        return true if request_from_ui?
-        Settings.product.allow_api_service_ordering
-      end
-
-      def request_from_ui?
-        !request.authorization.try(:downcase).try(:starts_with?, "basic")
       end
     end
   end
--- a/app/controllers/api/mixins/service_templates.rb
+++ b/app/controllers/api/mixins/service_templates.rb
@@ -3,7 +3,9 @@ module Api
     module ServiceTemplates
       def order_service_template(id, data, scheduled_time = nil)
         service_template = resource_search(id, :service_templates, ServiceTemplate)
-        raise BadRequestError, "#{service_template_ident(service_template)} cannot be ordered" unless service_template.orderable?
+        unless api_request_allowed? && service_template.orderable?
+          raise BadRequestError, "#{service_template_ident(service_template)} cannot be ordered"
+        end
         init_defaults = !request_from_ui? && Settings.product.run_automate_methods_on_service_api_submit
         request_result = service_template.order(User.current_user, (data || {}), {:submit_workflow => request_from_ui?, :init_defaults => init_defaults}, scheduled_time)
         errors = request_result[:errors]
@@ -27,6 +29,15 @@ module Api
 
       def service_template_ident(st)
         "Service Template id:#{st.id} name:'#{st.name}'"
+      end
+
+      def api_request_allowed?
+        return true if request_from_ui?
+        Settings.product.allow_api_service_ordering
+      end
+
+      def request_from_ui?
+        !request.authorization.try(:downcase).try(:starts_with?, "basic")
       end
     end
   end
--- a/spec/requests/service_catalogs_spec.rb
+++ b/spec/requests/service_catalogs_spec.rb
@@ -376,7 +376,10 @@ describe "Service Catalogs API" do
     end
 
     before do
-      stub_settings_merge(:product => double(:allow_api_service_ordering => true))
+      stub_settings_merge(:product => {:allow_api_service_ordering => true})
+      userid = User.first.userid
+      test_token = Api::UserTokenService.new.generate_token(userid, "api")
+      request_headers["x-auth-token"] = test_token
     end
 
     def init_st(service_template, resource_action)
--- a/spec/requests/service_catalogs_spec.rb
+++ b/spec/requests/service_catalogs_spec.rb
@@ -375,6 +375,10 @@ describe "Service Catalogs API" do
       request_headers["x-auth-token"] = test_token
     end
 
+    before do
+      stub_settings_merge(:product => double(:allow_api_service_ordering => true))
+    end
+
     def init_st(service_template, resource_action)
       service_template.resource_actions = [resource_action]
       dialog1.dialog_tabs << tab1
--- a/spec/requests/service_templates_spec.rb
+++ b/spec/requests/service_templates_spec.rb
@@ -466,11 +466,13 @@ describe "Service Templates API" do
 
   describe "Service Templates order" do
     let(:service_template) { FactoryGirl.create(:service_template, :with_provision_resource_action_and_dialog, :orderable) }
-    let(:product_settings) { double(:allow_api_service_ordering => allow_api_service_ordering) }
     let(:allow_api_service_ordering) { true }
 
     before do
-      stub_settings_merge(:product => product_settings)
+      stub_settings_merge(:product => {:allow_api_service_ordering => allow_api_service_ordering})
+      userid = User.first.userid
+      test_token = Api::UserTokenService.new.generate_token(userid, "api")
+      request_headers["x-auth-token"] = test_token
     end
 
     it "is forbidden without appropriate role" do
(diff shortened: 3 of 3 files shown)


eclarizio:
@miq-bot add_reviewer @bdunne 

@bdunne Can you take a look for me, please? And/or suggest anyone else that can take a look?

app/controllers/api/mixins/service_templates.rb
```diff
@@ -17,6 +19,14 @@ def order_service_template(id, data, scheduled_time = nil)
       def service_template_ident(st)
         "Service Template id:#{st.id} name:'#{st.name}'"
       end
+
+      def api_request_allowed?
+        standalone_api_request? ? !Settings.product.deny_api_service_ordering : true
```
bdunne: Can the Setting have a positive name?

spec/requests/service_templates_spec.rb
```diff
@@ -576,6 +576,32 @@
         expect(actions).to_not include("order")
       end
     end
+
+    context "with the product setting denying standalone service template ordering" do
+      let(:template_no_display) { FactoryGirl.create(:service_template, :display => false) }
+      let(:product_settings) { double(:deny_api_service_ordering => true) }
+
+      before do
+        allow(Settings).to receive(:product).and_return(product_settings)
```
bdunne: Prefer `stub_settings_merge`

app/controllers/api/mixins/service_templates.rb
```diff
@@ -17,6 +19,14 @@ def order_service_template(id, data, scheduled_time = nil)
...
+      end
+
+      def standalone_api_request?
+        request.authorization.present? && request.authorization.to_s.split(" ", 2).first.downcase == "basic"
```
bdunne: `request.authorization.try(:starts_with?, "basic")`

bdunne: Also, this will break if we allow other authentication types to the api

eclarizio: Agreed, but do you know of another way to distinguish standalone API calls versus API calls made from the UI?

app/controllers/api/mixins/service_templates.rb
eclarizio: The only issue I have with changing it to a positive name is then the users who want it to work the way it works after this PR would be to set the value to "false", which feels weird but I don't know if we have a precedent for this or not.

bdunne: No, but I think the default should be in [here](https://github.com/ManageIQ/manageiq/blob/master/config/settings.yml) anyway.

app/controllers/api/mixins/service_templates.rb
bdunne: I can't think of anything at the moment

abellotti: maybe we can leverage the token manager namespace, api, ws,  (might need to add param or separate api/ui as 2 namespaces).

eclarizio:
@bdunne I agree, but fixing the validation to take care of ensuring that the values passed in through the API versus actual acceptable values is much more involved because of dynamic dialogs and how we don't want to call into automate when we shouldn't have to (and the potential sequencing of those calls depending on which fields are dependent on which for refreshes). If the dialog has some expensive automate calls, submitting will take way longer than it should since for the majority of the cases the API call will have proper values already.

A PR could build on this one so that instead of just flat out denying the request a product setting is no longer needed and it delegates to the model to verify that the values passed in are acceptable values.

app/controllers/api/mixins/service_templates.rb
eclarizio: Also, I pulled the "basic" check from here: https://github.com/rails/rails/blob/fc5dd0b85189811062c85520fd70de8389b55aeb/actionpack/lib/action_controller/metal/http_authentication.rb#L104.

I suppose we could do `.try(:downcase).try(:starts_with?, "basic")`? Not sure how else to handle the fact that it could be "Basic", "BASIC", or any other various capitalizations.

eclarizio:
Not sure what labels to put on this now that we've transitioned into Hammer.

@miq-bot add_label bug

app/controllers/api/mixins/service_templates.rb
```diff
@@ -17,6 +19,14 @@ def order_service_template(id, data, scheduled_time = nil)
       def service_template_ident(st)
         "Service Template id:#{st.id} name:'#{st.name}'"
       end
+
+      def api_request_allowed?
+        standalone_api_request? ? Settings.product.allow_api_service_ordering : true
+      end
+
+      def standalone_api_request?
+        request.authorization.present? && request.authorization.try(:starts_with?, "basic")
```
carbonin: If we're doing the `.present?` check, why do we need the `.try`? Do we expect `request.authorization` not to respond to `starts_with?`?

Also, is the premise here valid @abellotti @jvlcek? Does external auth not work with the API? Or am I misunderstanding this check?

app/controllers/api/mixins/service_templates.rb
```diff
@@ -17,6 +19,14 @@ def order_service_template(id, data, scheduled_time = nil)
       def service_template_ident(st)
         "Service Template id:#{st.id} name:'#{st.name}'"
       end
+
+      def api_request_allowed?
+        standalone_api_request? ? Settings.product.allow_api_service_ordering : true
+      end
+
+      def standalone_api_request?
```
carbonin: Can we name this better? I don't understand what it means for an API request to be "standalone", maybe reverse the logic and call it `request_from_ui?`

app/controllers/api/mixins/service_templates.rb
eclarizio: Good question. I mentioned before when @bdunne suggested using the try that I pulled this line from https://github.com/rails/rails/blob/fc5dd0b85189811062c85520fd70de8389b55aeb/actionpack/lib/action_controller/metal/http_authentication.rb#L104.

app/controllers/api/mixins/service_templates.rb
eclarizio: That's fair, I was struggling to come up with a good name, but reversing the logic and naming it `request_from_ui?` sounds better to me

eclarizio:
@carbonin I agree, ultimately the auth type was the only thing I could think of to determine the difference between UI API calls and API calls originating from `curl` or any other client, but if there is a better way to determine it, it's easy to replace the logic here. I thought of a special header as well, but originally I heard that this was supposed to be some kind of high priority customer issue, so I wanted to get something out relatively quickly. Tina created the BZ just so we would have something to link to this PR, but I have no insight into how high the priority is now for the customer.

abellotti:
I think it's best if we use/extend our UI assigned tokens for this.  Since API calls coming in from the UI use their tokens, we already have that information (or easy to add) for differentiation. This way this will work for basic as well as ext-auth requests from the UI.

mkanoor:
@eclarizio Isn't this a validation problem, say if you had two drop down items in a service dialog
state = NY
fruit = pick one of ['apple', 'banana', 'orange']
county = pick one of Dynamic content from Automate (e.g. get list of counties based on state)

When the caller sends in
**state = NY, fruit = apple, county = Ulster** its valid this can be called from the UI  and from a CLI making the REST API call and if we made the automate call to get the list of counties we would know that Ulster is a valid county in NY
but when some one passes in

**state = NY, fruit = banana, county = bergen** this is invalid because Bergen is not a county in NY. When the user is in the UI we will never present Bergen as an option since we would have made the dynamic call

If during  validation of the dialog if we fetched the list of the dynamic components we could have rejected that invalid county.

I think the validation is making an assumption that if the field is a drop down the user has selected a valid item so even in the static fruit list if someone passed in kiwi we would still let it thru.

eclarizio:
> If during validation of the dialog if we fetched the list of the dynamic components we could have rejected that invalid county.

We could, you're right, but that (to me) is an unnecessary automate call, because we should have already done that when the user selected NY from the UI, which (correct me if you think I'm wrong here) will be a *large* majority of the time.

Maybe in your example case, the automate call isn't very expensive, but if the dialog has multiple expensive automate calls, I don't think the trade off of always calling automate to validate the values is worth it compared to how often it would actually be useful since again, the majority of the order submit calls will be going through the UIs. Unless someone purposely mucks with data, it is impossible to actually send through values that don't exist.

mkanoor:
@eclarizio 
There are 2 distinct phases here

1. User interacting with the dialogs making selections, causing data to change, maybe some data input validation happening like checking regular expression, minimum length etc.

2. The user presses the order button at this point we have a hash of all the selected values, these values have to be validated before the request is created or as the first step after the request is created. @Fryguy has correctly pointed out in the past that we do some validation before the MiqRequest is created which ends up holding the API calls and causing them to hang. We should create a request then validate all the parameters and once validation is successful mark the request as "Ready" for approval or Reject it.

The UI today mostly focuses on (1) but its easy to do (2) by doing some of the same things that were done in (1). That code is already there. It just needs to be used during the ordering phase.

(2) cannot be done in Automate, we cannot have Automate load up a dialog and for each field call all the Automate Methods to create dynamic content, all of that has to be done from the dialog side since there is already code doing this validation.

Granted "Validation" might take a long time but if there is an option that says "Strict Dialog Validation" and this can be done after the MiqRequest is created it would be ideal. We wont he holding the Request API call while we do the validation.

If we dont do this validation all the REST API calls are going in without any validation and we run into the issues reported in this BZ.

JPrause:
@eclarizio is this will be able to be backported, can you add the hammer/yes and gaprindashvili/yes labels

JPrause:
@miq-bot add_label blocker

app/controllers/api/mixins/service_templates.rb
```diff
@@ -15,19 +17,24 @@ def order_service_template(id, data, scheduled_time = nil)
...
+
+      def api_request_allowed?
+        return true if request_from_ui?
+        Settings.product.allow_api_service_ordering
+      end
+
       def request_from_ui?
         return false if request.headers["x-auth-token"].blank?
-        token_info.present?
+        !token_info.empty?
```
bdunne: Aren't these the same?

app/controllers/api/mixins/service_templates.rb
```diff
@@ -15,19 +17,24 @@ def order_service_template(id, data, scheduled_time = nil)
 
       private
 
+      def service_template_ident(st)
```
bdunne: I don't think this move should be in the diff

spec/requests/service_catalogs_spec.rb
```diff
@@ -375,6 +375,13 @@ def sc_template_url(id, st_id = nil)
       request_headers["x-auth-token"] = test_token
     end
 
+    before do
+      stub_settings_merge(:product => double(:allow_api_service_ordering => true))
```
bdunne: You shouldn't need a double here, it can be a nested hash

spec/requests/service_templates_spec.rb
```diff
@@ -466,6 +466,15 @@
 
   describe "Service Templates order" do
     let(:service_template) { FactoryGirl.create(:service_template, :with_provision_resource_action_and_dialog, :orderable) }
+    let(:product_settings) { double(:allow_api_service_ordering => allow_api_service_ordering) }
```
bdunne: Same here

bdunne: Also, it doesn't feel like this should be in a `let`

eclarizio:
@miq-bot add_label gaprindashvili/yes, hammer/yes

simaishi:
Hammer backport details:

```
$ git log -1
commit 55732b3d20dece9cf4743fa0cd850af5c0c11e82
Date:   Mon Oct 22 11:49:56 2018 -0400

    Merge pull request #476 from eclarizio/dialog_ordering_security_issue
    
    Deny standalone service template ordering when product setting is enabled
    
    (cherry picked from commit 7343ad7cad22f24639a23ff3a9d6c5182d64172d)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1632416
```

PR #483, opened 2018-10-02
[HAMMER] Update i18n catalog for hammer

mzazrivec:


simaishi:
@himdel can you please review?

PR #503, opened 2018-10-29
[HAMMER] Add translations for hammer

mzazrivec:


locale/es/ManageIQ_Api.po
```diff
@@ -1,27 +1,28 @@
-# Spanish translations for ManageIQ_Api package.
-# Copyright (C) 2018 THE PACKAGE'S COPYRIGHT HOLDER
```
bdunne: Was this supposed to change?

mzazrivec: This is OK I suppose. These comments don't matter that much really, those removed lines were generated automatically during string collection.

simaishi:
@bdunne ping

PR #511, opened 2018-11-14
[WIP] Return ActionResult instead of Hash when retiring Service

yrudman:
BZs: https://bugzilla.redhat.com/show_bug.cgi?id=1633142 and  https://bugzilla.redhat.com/show_bug.cgi?id=1579017

@miq-bot add-label hummer/yes, gaprindashvili/yes, bug, services

### DIFF
--- a/app/controllers/api/services_controller.rb
+++ b/app/controllers/api/services_controller.rb
@@ -143,7 +143,27 @@ module Api
       action_result(false, "Could not queue chargeback report generation - #{err}")
     end
 
+    def retire_resource(type, id = nil, data = nil)
+      raise BadRequestError, "Must specify an id for retiring a #{type} resource" unless id
+      api_log_info("############## type=#{type}")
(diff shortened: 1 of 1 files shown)


yrudman:
closed in favor of https://github.com/ManageIQ/manageiq/pull/18210

app/controllers/api/services_controller.rb
```diff
@@ -143,8 +143,28 @@ def queue_chargeback_report_resource(_type, id, _data)
       action_result(false, "Could not queue chargeback report generation - #{err}")
     end
 
+    def retire_resource(type, id = nil, data = nil)
```
bdunne: We should probably fix this in `base_controller/generic.rb` and remove the methods from VmsController if possible.

yrudman: yeaa, it would be more correct, but this PR is very looks very braking for backporting

PR #518, opened 2018-12-03
Support result pagination

wyrmiyu:
My customer asked me to provide this patch to the upstream. The context here is that they use Service Now to call API to collect reports and as they have a huge environment, the API produces over 25 MB large reports.

Further, they are restricted by a corporate HTTP proxy that they cannot control, which has a limit of 25 MB set on a single request.

Thus following modification is to allow pagination of these large result sets to reduce the size of a single request.

This checks for virtual attribute "result_set" and takes parameters "offset" for the amount of results that are skipped from the beginning and "limit" for the maximum amount of results returned.

### DIFF
--- a/app/controllers/api/base_controller/renderer.rb
+++ b/app/controllers/api/base_controller/renderer.rb
@@ -284,9 +284,7 @@ module Api
       def get_result_set(value)
         offset = request.params[:offset] ? request.params[:offset].to_i : 0
         limit = request.params[:limit] ? request.params[:limit].to_i : 0
-        if limit > 0
-          return value.sort_by{|hash| hash[:id] }[offset...offset+limit]
-        end
+        return value.sort_by { |hash| hash[:id] }[offset...offset + limit] if limit.positive?
         value
       end
 
--- a/app/controllers/api/base_controller/renderer.rb
+++ b/app/controllers/api/base_controller/renderer.rb
@@ -260,7 +260,6 @@ module Api
         add_hash json, result
       end
 
-
       def fetch_direct_virtual_attribute(type, resource, attr)
         return unless attr_accessible?(resource, attr)
         virtattr_accessor = virtual_attribute_accessor(type, attr)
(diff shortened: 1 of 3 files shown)


app/controllers/api/base_controller/renderer.rb
```diff
@@ -264,12 +264,29 @@ def fetch_direct_virtual_attribute(type, resource, attr)
...
         virtattr_accessor = virtual_attribute_accessor(type, attr)
         value = virtattr_accessor ? send(virtattr_accessor, resource) : virtual_attribute_search(resource, attr)
-        value = add_custom_action_hrefs(value) if attr == "custom_actions"
+        value = handle_virtual_attributes(attr, value)
         result = {attr => normalize_attr(attr, value)}
         # set nil vtype above to "#{type}/#{resource.id}/#{attr}" to support id normalization
         [value, result]
       end
 
+      def handle_virtual_attributes(attr, value)
```
bdunne: What do you think of this?

```ruby
def handle_virtual_attributes(attr, value)
  case attr
  when "custom_actions"
    add_custom_action_hrefs(value)
  when "result_set"
    get_result_set(value)
  else
    value
  end
end
```

app/controllers/api/base_controller/renderer.rb
```diff
@@ -264,12 +264,29 @@ def fetch_direct_virtual_attribute(type, resource, attr)
...
+        when "custom_actions"
+          value = add_custom_action_hrefs(value)
+        when "result_set"
+          value = get_result_set(value)
+        end
+        value
+      end
+
+      def get_result_set(value)
+        offset = request.params[:offset] ? request.params[:offset].to_i : 0
```
bdunne: The ternary operator shouldn't be necessary on these since `nil.to_i` is `0`.

`offset = request.params[:offset].to_i`

kbrock:
This had promise.
I think the main issue here was using ruby to filter the sub-resultset instead of using a build in which is smart enough to use the database where ever possible.

Is this still of interest?

lpichler:
@wyrmiyu can you show how api call looks like ? with all newly added parameters.

lpichler:
@wyrmiyu is it ok when I will take over it ? [I have PR ](https://github.com/ManageIQ/manageiq-api/pull/547)

lpichler:
Closing as this work is done here https://github.com/ManageIQ/manageiq-api/pull/547

PR #521, opened 2018-12-05
[Gaprindashvili] Fix invalid count on report index

NickLaMuro:
Since `User#admin?` and `User#admin_user?` are not equivalent, there is a possibility when not using the `admin` user to cause an invalid query of:

```SQL
SELECT  "miq_reports".* FROM "miq_reports" WHERE () LIMIT 1000 OFFSET 0
```

Where the `WHERE ()` is caused by trying to do a:

```ruby
MiqReport.where(MiqReport.all.where_clause.ast)
```

This happens because the check in place to avoid this in `Api::ReportsController#reports_search_conditions`, `User#admin?`, which only can be a user with `userid == "admin"`, is less specific then the check in `MiqReport.for_user`, which calls `User#admin_user?` and allows any user with either of the following roles:

- `"EvmRole-super_administrator"`
- `"EvmRole-administrator"`

When a user is just a `"EvmRole-administrator"`, this will cause an error, and this error is corrected by using the same check.

Links
-----
* Related fix for `master`/`hammer`:  https://github.com/ManageIQ/manageiq-api/pull/520
* Offending PR:  https://github.com/ManageIQ/manageiq/pull/15472

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1656242

### DIFF
--- a/app/controllers/api/reports_controller.rb
+++ b/app/controllers/api/reports_controller.rb
@@ -7,7 +7,7 @@ module Api
     before_action :set_additional_attributes, :only => [:index, :show]
 
     def reports_search_conditions
-      MiqReport.for_user(User.current_user).where_clause.ast unless User.current_user.admin?
+      MiqReport.for_user(User.current_user).where_clause.ast unless User.current_user.admin_user?
     end
 
     def find_reports(id)
--- a/spec/requests/reports_spec.rb
+++ b/spec/requests/reports_spec.rb
@@ -203,6 +203,33 @@ RSpec.describe "reports API" do
   end
 
   context "with an appropriate role" do
+    # Setup in a similar was to the reproduction steps in this BZ:
+    #
+    #   https://bugzilla.redhat.com/show_bug.cgi?id=1656242
+    #
+    # But note that this is only a spec valid for gaprindashvili.  If you are
+    # seeing this elsewhere in a newer release, please do some git blaming...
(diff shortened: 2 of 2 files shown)


NickLaMuro:
@miq-bot assign @gtanzillo 

@gtanzillo I will leave it to your discretion on whether or not it makes sense to move forward with this fix (I assume it does, but I could be wrong).

djberg96:
@gtanzillo any reason not to merge? Looks like the Hammer version already was.

gtanzillo:
@djberg96 
> @gtanzillo any reason not to merge? Looks like the Hammer version already was.

No, but since this is a PR against gaprindasvili, it needs to me merged by @simaishi. I just noticed that it doesn't have the `gaprindashvili/yes` tag on it which, I believe, is what Satoe uses to find backport PRs. I'll asdd the label

djberg96:
@gtanzillo aha, thanks!

simaishi:
@djberg96 @gtanzillo Since this is a PR against `gaprindashvili` branch, the `/yes` label isn't needed. However, I can't just merge even if it's reviewed and approved. All PRs going to branches have to be approved by bug triage team (which this was never approved for).

Also, we're no longer doing regular maintenance releases for gaprindashvili nor downstream release that's based on gaprindashvili branch. So I'm closing this PR.

PR #522, opened 2018-12-07
Add conversion host resource access

djberg96:
This adds the `resource` POST action to the conversion hosts REST API. The `resource` is a polymorphic resource that is either a Vm or Host.

Originally I was going to add subcollections for Vm's and Host's as per the BZ, but as was pointed out to me, that wouldn't really make sense since it's only a single resource.

Completes https://bugzilla.redhat.com/show_bug.cgi?id=1643938

### DIFF
--- a/app/controllers/api/conversion_hosts_controller.rb
+++ b/app/controllers/api/conversion_hosts_controller.rb
@@ -7,8 +7,7 @@ module Api
     #
     # POST /api/conversion_hosts/1 {'action': 'resource'}
     #
-    def resource_resource(type, id = nil, _data = nil)
-      raise BadRequestError, "Must specify an id for fetching a #{type}.resource" unless id
+    def resource_resource(type, id, _data = nil)
       conversion_host = resource_search(id, type, collection_class(type))
       conversion_host.resource
     rescue => err
--- a/app/controllers/api/conversion_hosts_controller.rb
+++ b/app/controllers/api/conversion_hosts_controller.rb
@@ -1,5 +1,18 @@
 module Api
   class ConversionHostsController < BaseController
     include Subcollections::Tags
+
+    # Interface for the polymorphic resource, which could be either a Vm or
+    # a Host. Failure to provide a conversion host ID will raise an error.
+    #
+    # POST /api/conversion_hosts/1 {'action': 'resource'}
+    #
+    def resource_resource(type, id = nil, _data = nil)
+      raise BadRequestError, "Must specify an id for fetching a #{type}.resource" unless id
+      conversion_host = resource_search(id, type, collection_class(type))
+      conversion_host.resource
+    rescue => err
+      raise BadRequestError, "Invalid resource #{type}/#{id}: #{err}"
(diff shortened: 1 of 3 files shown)


djberg96:
@miq-bot add_reviewer @agrare @jameswnl

djberg96:
@miq-bot add_reviewer @agrare

djberg96:
@miq-bot add_reviewer @jameswnl

app/controllers/api/conversion_hosts_controller.rb
```diff
@@ -1,5 +1,18 @@
 module Api
   class ConversionHostsController < BaseController
     include Subcollections::Tags
+
+    # Interface for the polymorphic resource, which could be either a Vm or
+    # a Host. Failure to provide a conversion host ID will raise an error.
+    #
+    # POST /api/conversion_hosts/1 {'action': 'resource'}
+    #
+    def resource_resource(type, id = nil, _data = nil)
```
jameswnl: you are raising on `nil id`, do we need defaulting `id` to bi `nil`?

djberg96: I was sort of copying what some of the other code does, I wasn't sure under what circumstance that could happen. I can remove it.

app/controllers/api/conversion_hosts_controller.rb
```diff
@@ -1,5 +1,18 @@
 module Api
   class ConversionHostsController < BaseController
     include Subcollections::Tags
+
+    # Interface for the polymorphic resource, which could be either a Vm or
+    # a Host. Failure to provide a conversion host ID will raise an error.
+    #
+    # POST /api/conversion_hosts/1 {'action': 'resource'}
```
jameswnl: is this a `post`?

config/api.yml
```diff
@@ -967,6 +967,8 @@
         :identifier: conversion_host_edit
       - :name: delete
         :identifier: conversion_host_delete
+      - :name: resource
+        :identifier: conversion_host_resource
```
jameswnl: you are doing a `get`, right?

djberg96: Can't do custom actions with a `GET` in the ManageIQ rest api. Not as far as I can tell, as it failed when I tried, and the docs say that this is what you're supposed to do:

https://github.com/ManageIQ/manageiq-api/blob/master/CONTRIBUTING.md#custom-resource-actions

@abellotti That's correct, right?

app/controllers/api/conversion_hosts_controller.rb
djberg96: Yes. See my other comment. You can only do custom actions on a POST afaik, though perhaps @abellotti knows something I don't.

djberg96:
Upon further review, going to close his, as you can get at the resource with a GET using something like:

`url --user user:passwd -H "Accept: application/json" http://localhost:3000/api/conversion_hosts/1?attributes=resource&expand=resources`

PR #528, opened 2018-12-14
[WIP] Make request_from_ui use params set in frontend dialog user controller

d-m-u:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1659602, an issue with default dialog values not populated when API auth type is auth token because the mechanism we had in place for determining the requester type (https://github.com/ManageIQ/manageiq-api/commit/f2f024a2ccabefcfbd833aa8df9b04757ab3d109) apparently wasn't bulletproof enough so if we force it by setting it in the params, maybe it'll be more right. 

## depends on
https://github.com/ManageIQ/manageiq-ui-classic/pull/5090/files

### DIFF
--- a/app/controllers/api/mixins/service_templates.rb
+++ b/app/controllers/api/mixins/service_templates.rb
@@ -24,8 +24,9 @@ module Api
       end
 
       def request_from_ui?
+        $api_log.info("requester_type: #{params['requester_type']}")
         return false if request.headers["x-auth-token"].blank?
-        token_info.present?
+        token_info.present? && params['requester_type'] == "ui"
       end
 
       def order_request_options
--- a/spec/requests/service_templates_spec.rb
+++ b/spec/requests/service_templates_spec.rb
@@ -503,6 +503,27 @@ describe "Service Templates API" do
         end
       end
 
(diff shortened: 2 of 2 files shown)


d-m-u:
@miq-bot add_label bug

d-m-u:
@miq-bot add_label hammer/yes, gaprindashvili/yes

d-m-u:
@miq-bot assign @gtanzillo

app/controllers/api/mixins/service_templates.rb
```diff
@@ -24,8 +24,9 @@ def api_request_allowed?
       end
 
       def request_from_ui?
+        $api_log.info("requester_type: #{params['requester_type']}")
```
eclarizio: I only had this for testing, probably doesn't actually need to be here.

d-m-u: It's not working anyway :( The param's being stripped out somewhere.

d-m-u:
Closing in favor of an as-yet uncreated PR for this issue because this was deemed to be The Wrong Solution.

PR #529, opened 2018-12-17
Updated service template's ui token info check to use token metadata for requester_type

abellotti:
Updated service template's ui token info check to make user of a token manager enhancement
where we store the requester_type with the api token.

- Updated the user token service to include the requester_type in the
  token metadata for non-api types.
- updated the API get auth to return the requester_type if defined.

### DIFF
--- a/app/controllers/api/mixins/service_templates.rb
+++ b/app/controllers/api/mixins/service_templates.rb
@@ -24,8 +24,11 @@ module Api
       end
 
       def request_from_ui?
-        return false if request.headers["x-auth-token"].blank?
-        token_info.present?
+        api_token = request.headers["x-auth-token"]
+        return false if api_token.blank?
+        token_info = Environment.user_token_service.token_mgr("api").token_get_info(api_token)
+        $api_log.info("XXXXXXXXXXX service_templates.rb:  The Token Used #{token_info}")
+        token_info["requester_type"] == "ui"
       end
 
       def order_request_options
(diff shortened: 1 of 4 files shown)


app/controllers/api/mixins/service_templates.rb
```diff
@@ -24,8 +24,11 @@ def api_request_allowed?
       end
 
       def request_from_ui?
-        return false if request.headers["x-auth-token"].blank?
-        token_info.present?
+        api_token = request.headers["x-auth-token"]
+        return false if api_token.blank?
+        token_info = Environment.user_token_service.token_mgr("api").token_get_info(api_token)
+        $api_log.info("XXXXXXXXXXX service_templates.rb:  The Token Used #{token_info}")
```
jrafanie: This log line can be made less scary looking by removing the XXXX...

jrafanie: My page was stale when I made that comment, looks like it's been fixed already

abellotti: yep, removed that log message :)

simaishi:
Hammer backport details:

```
$ git log -1
commit ff8bf190492d687c9ceadacac3c3eed7b73e4ec8
Date:   Mon Dec 17 13:23:05 2018 -0500

    Merge pull request #529 from abellotti/service_template_ui_token_fix
    
    Updated service template's ui token info check to use token metadata for requester_type
    
    (cherry picked from commit cfcb10b912d5a8b2b01f96e956d3d790ce5a1818)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1659602
```

PR #535, opened 2019-01-02
Add ability to enable and disable conversion hosts

djberg96:
Part of the V2V effort, this effectively wraps the `enable_conversion_host_role` and `disable_conversion_host_role` methods from the `ConversionHost` model.

~~WIP for now as I need to add specs, and make sure that the `eligible?` check isn't too strict.~~

Update: specs added, and now the create and delete actions call the `enable_queue` and `disable_queue` methods, respectively, behind the scenes (which also create/delete the record). This decision was made after some discussion with James W where we decided that the "queue" methods should only be called. The non-queue methods would probably only ever be used by devs for local testing, which we can accomplish on the console by using the model directly, if necessary.

We also decided that the `delete` method should be redefined to avoid a potential database inconsistency where the record could get deleted, but the conversion host was still enabled.

A `POST` action was still left in for disablement because it returns a body, whereas the `DELETE` action doesn't, so this offers some flexibility for end users.

### DIFF
--- a/app/controllers/api/conversion_hosts_controller.rb
+++ b/app/controllers/api/conversion_hosts_controller.rb
@@ -51,7 +51,6 @@ module Api
     # You may optionally provide an 'auth_user' parameter.
     #
     # POST /api/conversion_hosts/:id { "action": "disable" }
-    # POST /api/conversion_hosts/:id { "action": "disable" }
     # POST /api/conversion_hosts/:id { "action": "disable", "auth_user": "someone" }
     #
     # This differs from the DELETE action in that it returns a response body.
@@ -76,7 +75,6 @@ module Api
     #
     # You may optionally provide an 'auth_user' parameter.
     #
-    # DELETE /api/conversion_hosts/:id
     # DELETE /api/conversion_hosts/:id
     # DELETE /api/conversion_hosts/:id { "auth_user": "someone" }
     #
--- a/app/controllers/api/conversion_hosts_controller.rb
+++ b/app/controllers/api/conversion_hosts_controller.rb
@@ -29,9 +29,19 @@ module Api
       # The 'auth_user' param must be deleted since the model will otherwise
       # pass the data hash directly as params to ConversionHost.new.
       auth_user = data.delete('auth_user')
-      data['resource'] = resource_search(data['resource_id'], resource_type, collection_class(collection_type))
+      resource = resource_search(data['resource_id'], resource_type, collection_class(collection_type))
 
-      ConversionHost.enable_queue(data, auth_user)
+      data['resource'] = resource
+
+      api_action(_type, _id) do
+        begin
+          message = "Enabling resource id:#{resource.id} type:#{resource.type}"
+          task_id = ConversionHost.enable_queue(data, auth_user)
+          action_result(true, message, :task_id => task_id)
+        rescue err
+          action_result(false, err.to_s)
+        end
+      end
     end
 
     # Disable the conversion host role by installing the conversion host module
--- a/app/controllers/api/conversion_hosts_controller.rb
+++ b/app/controllers/api/conversion_hosts_controller.rb
@@ -28,8 +28,9 @@ module Api
     def create_resource(type, id, data)
       raise BadRequestError, "resource_id must be specified" unless data['resource_id']
       raise BadRequestError, "resource_type must be specified" unless data['resource_type']
+      raise BadRequestError, "invalid resource_type #{data['resource_type']}" unless VALID_TYPES[data['resource_type']]
 
-      resource_type = VALID_TYPES.fetch(data['resource_type']){ |resource_type| "invalid resource type: #{resource_type}" }
+      resource_type = VALID_TYPES[data['resource_type']]
       collection_type = resource_type.table_name
 
       # The 'auth_user' param must be deleted since the model will otherwise
--- a/app/controllers/api/conversion_hosts_controller.rb
+++ b/app/controllers/api/conversion_hosts_controller.rb
@@ -1,5 +1,61 @@
 module Api
   class ConversionHostsController < BaseController
     include Subcollections::Tags
+
+    # Create a conversion host and enable it. This operation will run as an
+    # MiqTask.
+    #
+    # Both the 'resource_type' and 'resource_id' are mandatory arguments,
+    # and the 'resource_type' must be either 'Host' or 'VmOrTemplate'.
+    #
+    # You may optionally pass in 'param_v2v_vddk_package_url' or 'auth_user'
+    # arguments as well.
+    #
+    # POST /api/conversion_hosts {
+    #   "name": "some_name",
+    #   "resource_type": "Host",
+    #   "resource_id": "7"
+    #   "param_v2v_vddk_package_url": "some_url"
+    #   "auth_user": "some_user"
+    # }
+    #
+    def create_resource(_type, _id, data)
+      raise BadRequestError, "resource_id must be specified" unless data['resource_id']
+      raise BadRequestError, "resource_type must be specified" unless data['resource_type']
+
+      resource_type = data['resource_type']
+      collection_type = resource_type.include?('Host') ? 'hosts' : 'vms'
+
+      # The 'auth_user' param must be deleted since the model will otherwise
+      # pass the data hash directly as params to ConversionHost.new.
+      auth_user = data.delete('auth_user')
+      data['resource'] = resource_search(data['resource_id'], resource_type, collection_class(collection_type))
+
+      ConversionHost.enable_queue(data, auth_user)
+    end
+
+    # Disable the conversion host role by installing the conversion host module
+    # and running the conversion host playbook that disables it. This operation
+    # run as an MiqTask.
+    #
+    # You may optionally provide an 'auth_user' parameter.
+    #
+    # POST /api/conversion_hosts/:id { "action": "disable" }
+    # POST /api/conversion_hosts/:id { "action": "disable" }
+    # POST /api/conversion_hosts/:id { "action": "disable", "auth_user": "someone" }
+    #
+    def disable_resource(type, id, data)
+      conversion_host = resource_search(id, type, collection_class(type))
+
+      api_action(type, id) do
+        message = "Disabling ConversionHost id:#{conversion_host.id} name:#{conversion_host.name}"
+        begin
+          task_id = conversion_host.disable_queue(data['auth_user']) # Ok if nil
+          action_result(true, message, :task_id => task_id)
+        rescue err
+          action_result(false, err.to_s)
+        end
+      end
+    end
   end
 end
--- a/app/controllers/api/conversion_hosts_controller.rb
+++ b/app/controllers/api/conversion_hosts_controller.rb
@@ -24,7 +24,7 @@ module Api
       raise BadRequestError, "resource_type must be specified" unless data['resource_type']
 
       resource_type = data['resource_type']
-      collection_type = resource_type.include?('Host') ? 'hosts' : 'vms'
+      collection_type = resource_type.classify.constantize.table_name
 
       # The 'auth_user' param must be deleted since the model will otherwise
       # pass the data hash directly as params to ConversionHost.new.
--- a/app/controllers/api/conversion_hosts_controller.rb
+++ b/app/controllers/api/conversion_hosts_controller.rb
@@ -52,45 +52,21 @@ module Api
     end
 
     # Disable the conversion host role by installing the conversion host module
-    # and running the conversion host playbook that disables it. This operation
-    # run as an MiqTask.
-    #
-    # You may optionally provide an 'auth_user' parameter.
-    #
-    # POST /api/conversion_hosts/:id { "action": "disable" }
-    # POST /api/conversion_hosts/:id { "action": "disable", "auth_user": "someone" }
-    #
-    # This differs from the DELETE action in that it returns a response body.
-    #
-    def disable_resource(type, id, data)
-      conversion_host = resource_search(id, type, collection_class(type))
-
-      api_action(type, id) do
-        message = "Disabling ConversionHost id:#{conversion_host.id} name:#{conversion_host.name}"
-        begin
-          task_id = conversion_host.disable_queue(data['auth_user']) # Ok if nil
-          action_result(true, message, :task_id => task_id)
-        rescue => err
-          action_result(false, err.to_s)
-        end
-      end
-    end
-
-    # Disable the conversion host role by installing the conversion host module
-    # and running the conversion host playbook that disables it. This operation
-    # run as an MiqTask.
+    # and running the conversion host playbook that disables it, then delete
+    # the conversion host record. This operation run as an MiqTask.
     #
     # You may optionally provide an 'auth_user' parameter.
     #
     # DELETE /api/conversion_hosts/:id
     # DELETE /api/conversion_hosts/:id { "auth_user": "someone" }
     #
-    # This differs from the POST action in that it does not return a response body.
+    # Note that you can also delete via a POST action using "action: delete" as
+    # a parameter, which will include a response body.
     #
     def delete_resource(type, id, data = {})
       delete_action_handler do
         conversion_host = resource_search(id, type, collection_class(type))
-        message = "Disabling ConversionHost id:#{conversion_host.id} name:#{conversion_host.name}"
+        message = "Disabling and deleting ConversionHost id:#{conversion_host.id} name:#{conversion_host.name}"
         begin
           task_id = conversion_host.disable_queue(data['auth_user']) # Ok if nil
           action_result(true, message, :task_id => task_id)
--- a/app/controllers/api/conversion_hosts_controller.rb
+++ b/app/controllers/api/conversion_hosts_controller.rb
@@ -2,6 +2,12 @@ module Api
   class ConversionHostsController < BaseController
     include Subcollections::Tags
 
+    # Whitelist of valid resource types
+    VALID_TYPES = {
+      'ManageIQ::Providers::Openstack::CloudManager::Vm' => ManageIQ::Providers::Openstack::CloudManager::Vm,
+      'ManageIQ::Providers::Redhat::InfraManager::Host'  => ManageIQ::Providers::Redhat::InfraManager::Host
+    }
+
     # Create a conversion host and enable it. This operation will run as an
     # MiqTask.
     #
@@ -23,13 +29,13 @@ module Api
       raise BadRequestError, "resource_id must be specified" unless data['resource_id']
       raise BadRequestError, "resource_type must be specified" unless data['resource_type']
 
-      resource_type = data['resource_type']
-      collection_type = resource_type.classify.constantize.table_name
+      resource_type = VALID_TYPES.fetch(data['resource_type']){ |resource_type| "invalid resource type: #{resource_type}" }
+      collection_type = resource_type.table_name
 
       # The 'auth_user' param must be deleted since the model will otherwise
       # pass the data hash directly as params to ConversionHost.new.
       auth_user = data.delete('auth_user')
-      resource = resource_search(data['resource_id'], resource_type, collection_class(collection_type))
+      resource = resource_search(data['resource_id'], resource_type.to_s, collection_class(collection_type))
 
       data['resource'] = resource
 
--- a/app/controllers/api/conversion_hosts_controller.rb
+++ b/app/controllers/api/conversion_hosts_controller.rb
@@ -19,7 +19,7 @@ module Api
     #
     # POST /api/conversion_hosts {
     #   "name": "some_name",
-    #   "resource_type": "Host",
+    #   "resource_type": "ManageIQ::Providers::Redhat::InfraManager::Host",
     #   "resource_id": "7"
     #   "param_v2v_vddk_package_url": "some_url"
     #   "auth_user": "some_user"
--- a/app/controllers/api/conversion_hosts_controller.rb
+++ b/app/controllers/api/conversion_hosts_controller.rb
@@ -6,7 +6,7 @@ module Api
     VALID_TYPES = {
       'ManageIQ::Providers::Openstack::CloudManager::Vm' => ManageIQ::Providers::Openstack::CloudManager::Vm,
       'ManageIQ::Providers::Redhat::InfraManager::Host'  => ManageIQ::Providers::Redhat::InfraManager::Host
-    }
+    }.freeze
 
     # Create a conversion host and enable it. This operation will run as an
     # MiqTask.
--- a/app/controllers/api/conversion_hosts_controller.rb
+++ b/app/controllers/api/conversion_hosts_controller.rb
@@ -33,9 +33,6 @@ module Api
       resource_type = VALID_TYPES[data['resource_type']]
       collection_type = resource_type.table_name
 
-      # The 'auth_user' param must be deleted since the model will otherwise
-      # pass the data hash directly as params to ConversionHost.new.
-      auth_user = data.delete('auth_user')
       resource = resource_search(data['resource_id'], resource_type.to_s, collection_class(collection_type))
 
       data['resource'] = resource
@@ -43,7 +40,7 @@ module Api
       api_action(type, id) do
         begin
           message = "Enabling resource id:#{resource.id} type:#{resource.type}"
-          task_id = ConversionHost.enable_queue(data, auth_user)
+          task_id = ConversionHost.enable_queue(data.except('auth_user'), data['auth_user'])
           action_result(true, message, :task_id => task_id)
         rescue => err
           action_result(false, err.to_s)
--- a/app/controllers/api/conversion_hosts_controller.rb
+++ b/app/controllers/api/conversion_hosts_controller.rb
@@ -38,7 +38,7 @@ module Api
           message = "Enabling resource id:#{resource.id} type:#{resource.type}"
           task_id = ConversionHost.enable_queue(data, auth_user)
           action_result(true, message, :task_id => task_id)
-        rescue err
+        rescue => err
           action_result(false, err.to_s)
         end
       end
@@ -54,6 +54,8 @@ module Api
     # POST /api/conversion_hosts/:id { "action": "disable" }
     # POST /api/conversion_hosts/:id { "action": "disable", "auth_user": "someone" }
     #
+    # This differs from the DELETE action in that it returns a response body.
+    #
     def disable_resource(type, id, data)
       conversion_host = resource_search(id, type, collection_class(type))
 
@@ -62,7 +64,32 @@ module Api
         begin
           task_id = conversion_host.disable_queue(data['auth_user']) # Ok if nil
           action_result(true, message, :task_id => task_id)
-        rescue err
+        rescue => err
+          action_result(false, err.to_s)
+        end
+      end
+    end
+
+    # Disable the conversion host role by installing the conversion host module
+    # and running the conversion host playbook that disables it. This operation
+    # run as an MiqTask.
+    #
+    # You may optionally provide an 'auth_user' parameter.
+    #
+    # DELETE /api/conversion_hosts/:id
+    # DELETE /api/conversion_hosts/:id
+    # DELETE /api/conversion_hosts/:id { "auth_user": "someone" }
+    #
+    # This differs from the POST action in that it does not return a response body.
+    #
+    def delete_resource(type, id, data = {})
+      delete_action_handler do
+        conversion_host = resource_search(id, type, collection_class(type))
+        message = "Disabling ConversionHost id:#{conversion_host.id} name:#{conversion_host.name}"
+        begin
+          task_id = conversion_host.disable_queue(data['auth_user']) # Ok if nil
+          action_result(true, message, :task_id => task_id)
+        rescue => err
           action_result(false, err.to_s)
         end
       end
--- a/app/controllers/api/conversion_hosts_controller.rb
+++ b/app/controllers/api/conversion_hosts_controller.rb
@@ -19,7 +19,7 @@ module Api
     #   "auth_user": "some_user"
     # }
     #
-    def create_resource(_type, _id, data)
+    def create_resource(type, id, data)
       raise BadRequestError, "resource_id must be specified" unless data['resource_id']
       raise BadRequestError, "resource_type must be specified" unless data['resource_type']
 
@@ -33,7 +33,7 @@ module Api
 
       data['resource'] = resource
 
-      api_action(_type, _id) do
+      api_action(type, id) do
         begin
           message = "Enabling resource id:#{resource.id} type:#{resource.type}"
           task_id = ConversionHost.enable_queue(data, auth_user)
--- a/app/controllers/api/conversion_hosts_controller.rb
+++ b/app/controllers/api/conversion_hosts_controller.rb
@@ -2,12 +2,6 @@ module Api
   class ConversionHostsController < BaseController
     include Subcollections::Tags
 
-    # Whitelist of valid resource types
-    VALID_TYPES = {
-      'ManageIQ::Providers::Openstack::CloudManager::Vm' => ManageIQ::Providers::Openstack::CloudManager::Vm,
-      'ManageIQ::Providers::Redhat::InfraManager::Host'  => ManageIQ::Providers::Redhat::InfraManager::Host
-    }.freeze
-
     # Create a conversion host and enable it. This operation will run as an
     # MiqTask.
     #
@@ -28,9 +22,18 @@ module Api
     def create_resource(type, id, data)
       raise BadRequestError, "resource_id must be specified" unless data['resource_id']
       raise BadRequestError, "resource_type must be specified" unless data['resource_type']
-      raise BadRequestError, "invalid resource_type #{data['resource_type']}" unless VALID_TYPES[data['resource_type']]
 
-      resource_type = VALID_TYPES[data['resource_type']]
+      # The scary constantize call below is mitigated by the fact that it won't get
+      # past the following checks we make before passing along the params.
+
+      resource_type = data['resource_type'].classify.safe_constantize
+
+      raise BadRequestError, "invalid resource_type #{data['resource_type']}" unless resource_type
+
+      unless resource_type.respond_to?(:supports_conversion_host?) && resource_type.supports_conversion_host?
+        raise BadRequestError, "unsupported resource_type #{resource_type}"
+      end
+
       collection_type = resource_type.table_name
 
       resource = resource_search(data['resource_id'], resource_type.to_s, collection_class(collection_type))
--- a/spec/requests/conversion_hosts_spec.rb
+++ b/spec/requests/conversion_hosts_spec.rb
@@ -87,9 +87,11 @@ describe "ConversionHosts API" do
 
       expect(response).to have_http_status(:ok)
 
-      results = response.parsed_body["results"]
-      expect(results).to be_kind_of(Array)
-      expect(results.first).to be_kind_of(Integer)
+      results = response.parsed_body["results"].first
+
+      expect(results['success']).to be_truthy
+      expect(results['href']).to eql('http://www.example.com/api/conversion_hosts/')
+      expect(results['message']).to eql("Enabling resource id:#{vm.id} type:Vm")
     end
 
     it "supports multiple conversion host creation" do
@@ -99,9 +101,13 @@ describe "ConversionHosts API" do
       post(api_conversion_hosts_url, :params => gen_request(:create, conversion_hosts))
 
       expect(response).to have_http_status(:ok)
+
       results = response.parsed_body["results"]
-      expect(results).to be_kind_of(Array)
-      expect(results.first).to be_kind_of(Integer)
+
+      expect(results).to match_array([
+        a_hash_including("message" => "Enabling resource id:#{vm.id} type:Vm"),
+        a_hash_including("message" => "Enabling resource id:#{host.id} type:Host"),
+      ])
     end
   end
 
--- a/spec/requests/conversion_hosts_spec.rb
+++ b/spec/requests/conversion_hosts_spec.rb
@@ -207,18 +207,26 @@ describe "ConversionHosts API" do
       expect_multiple_action_result(2)
 
       results = response.parsed_body['results']
+      task_one_id = results.first['task_id']
+      task_two_id = results.last['task_id']
 
-      expect(results.first['success']).to be_truthy
-      expect(results.first['message']).to eql("Disabling ConversionHost id:#{chost1.id} name:#{chost1.name}")
-      expect(results.first['task_id']).to match(/\d+/)
-      expect(results.first['task_href']).to eql("http://www.example.com/api/tasks/#{results.first['task_id']}")
-      expect(MiqTask.exists?(results.first['task_id'].to_i)).to be_truthy
+      expect(MiqTask.exists?(task_one_id.to_i)).to be_truthy
+      expect(MiqTask.exists?(task_two_id.to_i)).to be_truthy
 
-      expect(results.last['success']).to be_truthy
-      expect(results.last['message']).to eql("Disabling ConversionHost id:#{chost2.id} name:#{chost2.name}")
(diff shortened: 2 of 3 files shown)


app/controllers/api/conversion_hosts_controller.rb
```diff
@@ -1,5 +1,69 @@
...
+
+    # Enable the conversion host role by installing the conversion host module
+    # and running the conversion host playbook that enables it.
+    #
+    # POST /api/conversion_hosts/:id { "action": "enable" }
+    #
+    def enable_resource(type, id, data)
+      conversion_host = resource_search(id, type, collection_class(type))
+
+      if conversion_host.eligible?
```
jameswnl: Probably not suitable to check `eligible?` here
1) API server could be running in a different zone (firewall) from where the provider is 
2) that method check how many migration task the `conversion_host` already has. sounds like an chicken and egg

djberg96: Ok, I will skip it for now. Perhaps we'll come back to it.

djberg96:
Note that this PR will be red until https://github.com/ManageIQ/manageiq/pull/18388 is merged.

app/controllers/api/conversion_hosts_controller.rb
```diff
@@ -1,5 +1,58 @@
...
+
+      # Both the 'queue' and 'auth_user' params must be deleted since the model will
+      # otherwise pass the data hash directly as params to ConversionHost.new.
+      queue = data.delete('queue')
+
+      if queue
+        auth_user = data.delete('auth_user')
+        ConversionHost.enable_queue(data, auth_user)
+      else
+        ConversionHost.enable(data)
```
jameswnl: Why do we need a non-queue option? This operation involves playbook execution and doesn't sound like a good idea to hold back the anyone in the pipeline.
It's nice to provide options, but since it's not practical for production, we might as well not provide it.

app/controllers/api/conversion_hosts_controller.rb
```diff
@@ -1,5 +1,58 @@
...
+    # POST /api/conversion_hosts/:id { "action": "disable" }
+    # POST /api/conversion_hosts/:id { "action": "disable", "queue": true }
+    # POST /api/conversion_hosts/:id { "action": "disable", "queue": true, "auth_user": "someone" }
+    #
+    def disable_resource(type, id, data)
+      conversion_host = resource_search(id, type, collection_class(type))
+      if data['queue'].present?
+        conversion_host.disable_queue(data['auth_user'])
+      else
+        conversion_host.disable
```
jameswnl: similar to above

app/controllers/api/conversion_hosts_controller.rb
djberg96: I'm just going with what was there when I started. :)

I can alter it so that `enable_queue` is the default, and it could be disabled via `"queue":"false"`. That might be useful for devs, and it's not hard to wrap.

app/controllers/api/conversion_hosts_controller.rb
djberg96: Ok, I'll switch it so that `disable_queue` is the default.

app/controllers/api/conversion_hosts_controller.rb
```diff
@@ -1,5 +1,61 @@
...
+
+      resource_type = data['resource_type']
+      collection_type = resource_type.include?('Host') ? 'hosts' : 'vms'
+
+      # The 'auth_user' param must be deleted since the model will otherwise
+      # pass the data hash directly as params to ConversionHost.new.
+      auth_user = data.delete('auth_user')
+      data['resource'] = resource_search(data['resource_id'], resource_type, collection_class(collection_type))
+
+      ConversionHost.enable_queue(data, auth_user)
```
jameswnl: I don't know much about `action_result`, but not here?

djberg96: Near as I could tell, you want to use it in conjunction with api_action. It worked either way, but the logging was better in this case.

djberg96: Although, hm, maybe I should wrap it.

djberg96: @jameswnl Ok, updated it to return an `api_action` result. However, I think I need a 2nd pair of eyes to figure out the test failure. It looks fine, except that for some reason the word "Host" is being left out of the message, and I'm not sure why.

djberg96: @jameswnl Ok, figured it out, the factorybot for hosts doesn't automatically set the type, so I added in the tests. In practice that would always be set for STI.

spec/requests/conversion_hosts_spec.rb
```diff
@@ -70,45 +76,60 @@
...
+      api_basic_authorize(action_identifier(:conversion_hosts, :disable, :resource_actions))
+      allow(conversion_host).to receive(:check_conversion_host_role).and_return('disabled')
 
-      results = response.parsed_body["results"]
+      post(conversion_host_url, :params => {"action" => "disable"})
 
-      expect(ConversionHost.exists?(results.first["id"])).to be_truthy
-      expect(ConversionHost.exists?(results.last["id"])).to be_truthy
-      expect(results).to match_array([a_hash_including("resource_id" => vm.id.to_s), a_hash_including("resource_id" => host.id.to_s)])
+      expect(response).to have_http_status(:ok)
```
jameswnl: same, the expectation of `task_id`

spec/requests/conversion_hosts_spec.rb
```diff
@@ -70,45 +76,60 @@
...
+      expect(results).to match_array([
+        a_hash_including("message" => "Enabling resource id:#{vm.id} type:Vm"),
+        a_hash_including("message" => "Enabling resource id:#{host.id} type:Host"),
+      ])
     end
+  end
 
-    it "supports multiple conversion host creation" do
-      api_basic_authorize(collection_action_identifier(:conversion_hosts, :create))
+  context "disable" do
```
jameswnl: how about the existing `context "delete" do"`?

djberg96: Ok, done.

djberg96:
@jameswnl Ok, updated it so that the delete action also calls `disable_queue` behind the scenes, and updated the specs. For the specs, I also added some MiqTask checks.

app/controllers/api/conversion_hosts_controller.rb
```diff
@@ -1,5 +1,98 @@
...
+    #   "param_v2v_vddk_package_url": "some_url"
+    #   "auth_user": "some_user"
+    # }
+    #
+    def create_resource(type, id, data)
+      raise BadRequestError, "resource_id must be specified" unless data['resource_id']
+      raise BadRequestError, "resource_type must be specified" unless data['resource_type']
+
+      resource_type = data['resource_type']
+      collection_type = resource_type.include?('Host') ? 'hosts' : 'vms'
```
jameswnl: do we know what will UI send to us? The base class? or specific class e.g. with specific provider?
Also, does the `collection_class` handle only a subset of classes? or any class?

(I am trying to see if this can be generic as well)

app/controllers/api/conversion_hosts_controller.rb
```diff
@@ -1,5 +1,98 @@
...
+
+      # The 'auth_user' param must be deleted since the model will otherwise
+      # pass the data hash directly as params to ConversionHost.new.
+      auth_user = data.delete('auth_user')
+      resource = resource_search(data['resource_id'], resource_type, collection_class(collection_type))
```
jameswnl: what will happen when this search return nil? what happen if it raise error?
(Just want to know the behavior of api)

app/controllers/api/conversion_hosts_controller.rb
```diff
@@ -1,5 +1,98 @@
...
+    #   "resource_id": "7"
+    #   "param_v2v_vddk_package_url": "some_url"
+    #   "auth_user": "some_user"
+    # }
+    #
+    def create_resource(type, id, data)
+      raise BadRequestError, "resource_id must be specified" unless data['resource_id']
+      raise BadRequestError, "resource_type must be specified" unless data['resource_type']
+
+      resource_type = data['resource_type']
```
jameswnl: do you want to do a `delete` on this? plus the `resource_id`, since we are dropping in a `resource` object now?

app/controllers/api/conversion_hosts_controller.rb
```diff
@@ -1,5 +1,98 @@
...
+    end
+
+    # Disable the conversion host role by installing the conversion host module
+    # and running the conversion host playbook that disables it. This operation
+    # run as an MiqTask.
+    #
+    # You may optionally provide an 'auth_user' parameter.
+    #
+    # POST /api/conversion_hosts/:id { "action": "disable" }
+    # POST /api/conversion_hosts/:id { "action": "disable" }
```
jameswnl: duplicated line

app/controllers/api/conversion_hosts_controller.rb
```diff
@@ -1,5 +1,98 @@
...
+    end
+
+    # Disable the conversion host role by installing the conversion host module
+    # and running the conversion host playbook that disables it. This operation
+    # run as an MiqTask.
+    #
+    # You may optionally provide an 'auth_user' parameter.
+    #
+    # DELETE /api/conversion_hosts/:id
+    # DELETE /api/conversion_hosts/:id
```
jameswnl: duplicated

spec/requests/conversion_hosts_spec.rb
```diff
@@ -70,50 +76,79 @@
...
+    let(:ems) { FactoryBot.create(:ems_vmware, :zone => zone) }
+    let(:vm) { FactoryBot.create(:vm, :ems_id => ems.id) }
+    let(:conversion_host) { FactoryBot.create(:conversion_host, :resource => vm) }
+    let(:conversion_host_url) { api_conversion_host_url(nil, conversion_host) }
 
-      conversion_hosts = [sample_conversion_host_from_vm, sample_conversion_host_from_host]
-      post(api_conversion_hosts_url, :params => gen_request(:create, conversion_hosts))
+    before do
+      allow(conversion_host).to receive(:install_conversion_host_module).and_return(true)
+      allow_any_instance_of(ConversionHost).to receive(:ansible_playbook).and_return({})
```
jameswnl: how about mocking the `resource_search` to return the `conversion_host` and so you can mock that particular instance, instead of `allow_any_instance_of`?

app/controllers/api/conversion_hosts_controller.rb
djberg96: Eh, I'd like to leave it for now. It doesn't hurt anything to leave it.

app/controllers/api/conversion_hosts_controller.rb
djberg96: @jameswnl My expectation is that it will send the full class name, e.g. ManageIQ::Providers::Redhat::InfraManager::Host, but this handles it either way.

Pretty sure collection_class handles any class that's had a collection defined for it. Not entirely sure what you mean here by generic.

app/controllers/api/conversion_hosts_controller.rb
djberg96: The api will come back with e.g. `{"error":{"kind":"not_found","message":"Couldn't find Host with 'id'=9999","klass":"ActiveRecord::RecordNotFound"}}`

app/controllers/api/conversion_hosts_controller.rb
djberg96: fixed.

app/controllers/api/conversion_hosts_controller.rb
djberg96: fixed, thanks.

app/controllers/api/conversion_hosts_controller.rb
jameswnl: Automatically assuming it's `vms` if it is not `hosts` seems potentially fragile to me.

jameswnl: we should probably let it fail if we can't recognize the incoming type. we can't just assume it

djberg96: It will automatically fail once the updated validation is merged.

spec/requests/conversion_hosts_spec.rb
djberg96: Done.

app/controllers/api/conversion_hosts_controller.rb
djberg96: Hm, for a more generic approach, how about `resource_type.split('::').last.downcase.pluralize`

djberg96: Or we could approach it from the other direction and do `resource_type.classify.constantize.table_name`.

djberg96: @jameswnl Ok, updated.

djberg96:
@roliveri Can you merge here?

roliveri:
@djberg96 Sorry, I don't have merge privileges for this repo.

roliveri:
@Fryguy Can you merge this?

djberg96:
@miq-bot assign @bdunne

app/controllers/api/conversion_hosts_controller.rb
```diff
@@ -1,5 +1,96 @@
...
+    #   "param_v2v_vddk_package_url": "some_url"
+    #   "auth_user": "some_user"
+    # }
+    #
+    def create_resource(type, id, data)
+      raise BadRequestError, "resource_id must be specified" unless data['resource_id']
+      raise BadRequestError, "resource_type must be specified" unless data['resource_type']
+
+      resource_type = data['resource_type']
+      collection_type = resource_type.classify.constantize.table_name
```
bdunne: Should we work against a whitelist rather than constantizing whatever they send in?

app/controllers/api/conversion_hosts_controller.rb
```diff
@@ -1,5 +1,96 @@
...
+    #   "resource_id": "7"
+    #   "param_v2v_vddk_package_url": "some_url"
+    #   "auth_user": "some_user"
+    # }
+    #
+    def create_resource(type, id, data)
+      raise BadRequestError, "resource_id must be specified" unless data['resource_id']
+      raise BadRequestError, "resource_type must be specified" unless data['resource_type']
+
+      resource_type = data['resource_type']
```
bdunne: Should this also be deleted from the data array for the same reasons? https://github.com/ManageIQ/manageiq-api/pull/535/files#diff-2da9489e8f3f5aa77ab1c89559de7604R29

djberg96: No because you can't pass a resource object directly via the api. The `resource_type` is part of the table.

app/controllers/api/conversion_hosts_controller.rb
djberg96: Hm, is `safe_constantize` good enough? Otherwise it looks like the intertubes agree that `constantize` is potentially unsafe.

app/controllers/api/conversion_hosts_controller.rb
```diff
@@ -1,5 +1,96 @@
...
+    # run as an MiqTask.
+    #
+    # You may optionally provide an 'auth_user' parameter.
+    #
+    # DELETE /api/conversion_hosts/:id
+    # DELETE /api/conversion_hosts/:id { "auth_user": "someone" }
+    #
+    # This differs from the POST action in that it does not return a response body.
+    #
+    def delete_resource(type, id, data = {})
```
bdunne: Shouldn't this do the delete as well?

app/controllers/api/conversion_hosts_controller.rb
```diff
@@ -1,5 +1,96 @@
...
+        begin
+          task_id = conversion_host.disable_queue(data['auth_user']) # Ok if nil
+          action_result(true, message, :task_id => task_id)
+        rescue => err
+          action_result(false, err.to_s)
+        end
+      end
+    end
+
+    # Disable the conversion host role by installing the conversion host module
```
bdunne: Does this comment need to be updated?  It looks like a copy / paste of the method above

spec/requests/conversion_hosts_spec.rb
```diff
@@ -39,16 +43,18 @@
...
       get(api_conversion_host_url(nil, conversion_host))
 
       expect(response).to have_http_status(:forbidden)
     end
   end
 
   context "create" do
-    let(:vm) { FactoryBot.create(:vm) }
-    let(:host) { FactoryBot.create(:host) }
+    let(:zone) { FactoryBot.create(:zone, :name => "api_zone") }
```
bdunne: `name` shouldn't be necessary since it is sequenced

spec/requests/conversion_hosts_spec.rb
```diff
@@ -39,16 +43,18 @@
...
+    let(:ems) { FactoryBot.create(:ems_openstack, :zone => zone) }
+    let(:vm) { FactoryBot.create(:vm_openstack, :ems_id => ems.id) }
+    let(:host) { FactoryBot.create(:host, :ems_id => ems.id, :type => "Host") }
```
bdunne: Do you need to pass `:type => "Host"`?  It looks like this would be the default

spec/requests/conversion_hosts_spec.rb
```diff
@@ -76,44 +82,70 @@
       post(api_conversion_hosts_url, :params => sample_conversion_host_from_vm)
 
       expect(response).to have_http_status(:ok)
-      expect_result_resources_to_include_keys("results", expected_attributes)
 
-      conversion_host_id = response.parsed_body["results"].first["id"]
+      results = response.parsed_body["results"].first
 
-      expect(ConversionHost.find_by(:resource_id => vm.id).id).to eql(conversion_host_id.to_i)
+      expect(results['success']).to be_truthy
```
bdunne: This could be handled by a single expectation with `have_attributes`

spec/requests/conversion_hosts_spec.rb
```diff
@@ -76,44 +82,70 @@
...
+        a_hash_including("message" => "Enabling resource id:#{vm.id} type:#{vm.class}", "task_id" => a_kind_of(String)),
+        a_hash_including("message" => "Enabling resource id:#{host.id} type:#{host.class}", "task_id" => a_kind_of(String)),
+      ])
     end
+  end
 
-    it "supports multiple conversion host creation" do
-      api_basic_authorize(collection_action_identifier(:conversion_hosts, :create))
+  context "disable" do
+    let(:zone) { FactoryBot.create(:zone, :name => "api_zone") }
```
bdunne: same here

app/controllers/api/conversion_hosts_controller.rb
djberg96: They do the same thing, only one is on a DELETE, the other is on a POST.

spec/requests/conversion_hosts_spec.rb
```diff
@@ -76,44 +82,70 @@
...
       expect(response).to have_http_status(:ok)
-      expect_result_resources_to_include_keys("results", expected_attributes)
 
-      results = response.parsed_body["results"]
+      results = response.parsed_body
 
-      expect(ConversionHost.exists?(results.first["id"])).to be_truthy
-      expect(ConversionHost.exists?(results.last["id"])).to be_truthy
-      expect(results).to match_array([a_hash_including("resource_id" => vm.id.to_s), a_hash_including("resource_id" => host.id.to_s)])
+      expect(results['success']).to be_truthy
```
bdunne: same here

spec/requests/conversion_hosts_spec.rb
```diff
@@ -76,44 +82,70 @@
...
+      expect(results['href']).to eql("http://www.example.com/api/conversion_hosts/#{conversion_host.id}")
+      expect(results['message']).to eql("Disabling ConversionHost id:#{conversion_host.id} name:#{conversion_host.name}")
+      expect(results['task_id']).to match(/\d+/)
+      expect(MiqTask.exists?(results['task_id'].to_i)).to be_truthy
     end
   end
 
   context "delete" do
-    let(:conversion_host)             { FactoryBot.create(:conversion_host, :resource => FactoryBot.create(:vm)) }
+    let(:zone)                        { FactoryBot.create(:zone, :name => "api_zone") }
```
bdunne: same here

app/controllers/api/conversion_hosts_controller.rb
djberg96: Not sure what you mean, it does do a delete.

spec/requests/conversion_hosts_spec.rb
```diff
@@ -76,44 +82,70 @@
...
+    let(:ems)                         { FactoryBot.create(:ems_openstack, :zone => zone) }
+    let(:vm)                          { FactoryBot.create(:vm_openstack, :ems_id => ems.id) }
```
bdunne: Prefer `:ems => ems` rather than setting id's

spec/requests/conversion_hosts_spec.rb
```diff
@@ -143,32 +180,40 @@
...
-      array = []
-      array << api_conversion_host_url(nil, chost1)
-      array << api_conversion_host_url(nil, chost2)
+      results = response.parsed_body['results']
 
-      expect_multiple_action_result(2)
-      expect_result_resources_to_include_hrefs("results", array)
-      expect(ConversionHost.exists?(chost1.id)).to be_falsey
-      expect(ConversionHost.exists?(chost2.id)).to be_falsey
+      expect(results.first['success']).to be_truthy
```
bdunne: Should these use match_array instead?
```
expect(results).to match_array([a_hash_including(...), a_hash_including(…)])
```

app/controllers/api/conversion_hosts_controller.rb
bdunne: It looks like the `enable_queue` method is expecting `:resource` and not `:resource_type`

app/controllers/api/conversion_hosts_controller.rb
bdunne: Right now nothing is preventing them from passing in `"resource_type": "MiqServer"` which we probably don't want.

app/controllers/api/conversion_hosts_controller.rb
bdunne: Oh, wow I wasn't expecting that from a method named `disable`

app/controllers/api/conversion_hosts_controller.rb
djberg96: Oh, I think I see what you're saying. For `resource_id` and `resource_type` it doesn't hurt to leave them since they're still valid params to `ConversionHost.new`, whereas auth_user is not. Maybe it's inefficient, though? I guess I try not to mangle params if I can help it.

app/controllers/api/conversion_hosts_controller.rb
djberg96: The name is based on the `ConversionHost::Configurations#disable` method `disable`. Originally I was going to have separate `disable` and `disable_queue` methods but was later told we should only use the `queue` methods. I didn't use `delete` because I thought that would confuse the API.

spec/requests/conversion_hosts_spec.rb
djberg96: I was getting confused by all the different helper methods, which all behave slightly differently, so I just went with what I know.

spec/requests/conversion_hosts_spec.rb
djberg96: Tried it, doesn't work, you have to specify `ext_management_system => ems` if you want to do it that way, so the current approach is actually the shortest. I had wondered about that myself when I first worked on these. It probably *should* work, though.

bdunne: Sorry, I meant `:ext_management_system => ems`

app/controllers/api/conversion_hosts_controller.rb
bdunne: As for queuing, anything that doesn't return immediately should be queued, so I agree with that part.

I was surprised that "disabling" a ConversionHost also destroys the record.  So, I can't come back later and re-enable the same ConversionHost.

app/controllers/api/conversion_hosts_controller.rb
bdunne: `:resource_type` happens to be overridden in the core method, my concern is that the raw parameter is not constrained.

app/controllers/api/conversion_hosts_controller.rb
jameswnl: @djberg96 now I realize that if an unsupported resource is passed in, API won't fail it. It will fail only until queue_worker comes to work on it. So a whitelist is a better solution here.

(Somehow I had this impression that the `ConversionHost.enable_queue` will fail it if it is an unsupported resource, could be one of the prior implementations, i don't remember)

bdunne: I'm fine with adding the validation to the model if you prefer to do it there, but I think it needs to be done somewhere.

spec/requests/conversion_hosts_spec.rb
djberg96: Updated.

spec/requests/conversion_hosts_spec.rb
djberg96: Updated.

spec/requests/conversion_hosts_spec.rb
djberg96: Updated.

spec/requests/conversion_hosts_spec.rb
djberg96: Updated.

spec/requests/conversion_hosts_spec.rb
djberg96: Updated.

app/controllers/api/conversion_hosts_controller.rb
djberg96: Your approach was what I originally did, but as per direction from @jameswnl it was altered because we didn't want to have a situation where we have a conversion host record that isn't enabled. So, creating a conversion host necessarily enables it, and disabling a converison host necessarily deletes it.

spec/requests/conversion_hosts_spec.rb
djberg96: I can switch it to `include` but not `have_attributes` since the latter is looking for actual attributes i.e. methods, not keys. Anyway, updated.

spec/requests/conversion_hosts_spec.rb
djberg96: Updated.

spec/requests/conversion_hosts_spec.rb
djberg96: Updated.

app/controllers/api/conversion_hosts_controller.rb
djberg96: Ok, I've added a whitelist to the controller, and it will now raise a bad request error if the resource type is invalid. I've added a spec to confirm as well.

app/controllers/api/conversion_hosts_controller.rb
djberg96: This is now constrained by the whitelist.

djberg96:
@bdunne @jameswnl Ok, I've updated the PR so that the controller now uses a whitelist, and updated the specs accordingly. I also updated some rspec related issues that @bdunne brought up.

app/controllers/api/conversion_hosts_controller.rb
```diff
@@ -1,5 +1,103 @@
 module Api
   class ConversionHostsController < BaseController
     include Subcollections::Tags
+
+    # Whitelist of valid resource types
+    VALID_TYPES = {
```
bdunne: Wow, this is much more restrictive than I thought.  I'm glad we added this.

app/controllers/api/conversion_hosts_controller.rb
```diff
@@ -1,5 +1,103 @@
...
+    # run as an MiqTask.
+    #
+    # You may optionally provide an 'auth_user' parameter.
+    #
+    # POST /api/conversion_hosts/:id { "action": "disable" }
+    # POST /api/conversion_hosts/:id { "action": "disable", "auth_user": "someone" }
+    #
+    # This differs from the DELETE action in that it returns a response body.
+    #
+    def disable_resource(type, id, data)
```
bdunne: I still think this is a misleading API

bdunne:
Overall, looks great.

app/controllers/api/conversion_hosts_controller.rb
djberg96: Yep, I read an article or two that convinced me. Thanks for bringing it up.

app/controllers/api/conversion_hosts_controller.rb
djberg96: I understand, but I'm just following the core methods really.

djberg96:
@bdunne Thanks! Can you approve & merge? Or are there other issues you have yet?

app/controllers/api/conversion_hosts_controller.rb
```diff
@@ -1,5 +1,103 @@
...
+    #
+    # Both the 'resource_type' and 'resource_id' are mandatory arguments,
+    # and the 'resource_type' must be either 'Host' or 'VmOrTemplate'.
+    #
+    # You may optionally pass in 'param_v2v_vddk_package_url' or 'auth_user'
+    # arguments as well.
+    #
+    # POST /api/conversion_hosts {
+    #   "name": "some_name",
+    #   "resource_type": "Host",
```
mzazrivec: The `"Host"` in this example is not correct, right? It needs to be either of the types listed in VALID_TYPES.

app/controllers/api/conversion_hosts_controller.rb
Fryguy: My only concern is that core methods are not exposed to end users, per se, and APIs are.  And once the API is out there, you have to do the backwards compatibility dance to undo it.  IMO, if there is an issue with the name, we should choose a name that doesn't confuse end users, use that in the API, then possibly refactor the backend.

Fryguy: If I'm reading this all correctly, does "disable" actually do a "destroy" under the covers?  If so, calling it "disable" is just wrong, as disable implies you can eventually "enable" the object with the same id later on.

> This differs from the DELETE action in that it returns a response body.
> ...
> This differs from the POST action in that it does not return a response body.

Why do we care about the distinction?  My first thought is that the POST with an "action" of "delete" (as opposed to  "disable") should be equivalent to the DELETE action and both should return the response body

app/controllers/api/conversion_hosts_controller.rb
```diff
@@ -1,5 +1,103 @@
...
+    #
+    # DELETE /api/conversion_hosts/:id
+    # DELETE /api/conversion_hosts/:id { "auth_user": "someone" }
+    #
+    # This differs from the POST action in that it does not return a response body.
+    #
+    def delete_resource(type, id, data = {})
+      delete_action_handler do
+        conversion_host = resource_search(id, type, collection_class(type))
+        message = "Disabling ConversionHost id:#{conversion_host.id} name:#{conversion_host.name}"
```
Fryguy: Delete says "disable"...is that right?

spec/requests/conversion_hosts_spec.rb
```diff
@@ -1,7 +1,11 @@
 describe "ConversionHosts API" do
+  before(:all) do
```
Fryguy: @bdunne Do we still discourage `before(:all)`

app/controllers/api/conversion_hosts_controller.rb
djberg96: Oops, thought I fixed that. Thanks, updated.

spec/requests/conversion_hosts_spec.rb
djberg96: Only needs to be run once, right? Otherwise it's a peformance hit, isn't it?

app/controllers/api/conversion_hosts_controller.rb
djberg96: Yes, it's implicit that disabling a conversion host deletes it.

app/controllers/api/conversion_hosts_controller.rb
jameswnl: yeah, agree that enable/disable is confusing here. we can use create/destroy here and refactor the backend name.

app/controllers/api/conversion_hosts_controller.rb
jameswnl: sounds like renaming backend methods `enable/disable` to `create/destroy` would help clear this out?

app/controllers/api/conversion_hosts_controller.rb
djberg96: Alright, so really then it's just a matter of deleting the `disable` controller method altogether and letting the `delete_resource` work as for both.

spec/requests/conversion_hosts_spec.rb
Fryguy: `before(:all)` has side effects because of how it behaves with respect to database rollbacks after each test.  We generally avoid it because it can be a source of tests leaking into each other.

app/controllers/api/conversion_hosts_controller.rb
Fryguy: I'm not even sure we need to refactor the backend name.  enable/disable is the backend terminology and create/delete is the API, and that seems pretty straightforward.  As long as we don't add a 3rd option to the API (which is what's happening in this PR - create/delete/disable), then I think we are ok.

djberg96: Ok updated.

app/controllers/api/conversion_hosts_controller.rb
```diff
@@ -1,5 +1,79 @@
 module Api
   class ConversionHostsController < BaseController
     include Subcollections::Tags
+
+    # Whitelist of valid resource types
+    VALID_TYPES = {
+      'ManageIQ::Providers::Openstack::CloudManager::Vm' => ManageIQ::Providers::Openstack::CloudManager::Vm,
+      'ManageIQ::Providers::Redhat::InfraManager::Host'  => ManageIQ::Providers::Redhat::InfraManager::Host
```
Fryguy: I really don't like putting the provider constants in this code for 2 reasons.. 1) pluggability...perhaps this can be done with a registrations of some sort and 2) This is defined at the class level, so will force resolutions of this const at boot time.

app/controllers/api/conversion_hosts_controller.rb
```diff
@@ -1,5 +1,79 @@
...
+      raise BadRequestError, "resource_id must be specified" unless data['resource_id']
+      raise BadRequestError, "resource_type must be specified" unless data['resource_type']
+      raise BadRequestError, "invalid resource_type #{data['resource_type']}" unless VALID_TYPES[data['resource_type']]
+
+      resource_type = VALID_TYPES[data['resource_type']]
+      collection_type = resource_type.table_name
+
+      # The 'auth_user' param must be deleted since the model will otherwise
+      # pass the data hash directly as params to ConversionHost.new.
+      auth_user = data.delete('auth_user')
```
Fryguy: Prefer `data.except('auth_user')`.  This way you are no modifying the incoming data payload, should that auth_user be needed elsewhere.

spec/requests/conversion_hosts_spec.rb
```diff
@@ -1,7 +1,11 @@
 describe "ConversionHosts API" do
+  before(:each) do
```
Fryguy: Minor, but the :each is the default, and thus not necessary.

spec/requests/conversion_hosts_spec.rb
```diff
@@ -19,7 +23,7 @@
 
   context "resources" do
     it 'will show a conversion host with an appropriate role' do
-      conversion_host = FactoryBot.create(:conversion_host, :resource => FactoryBot.create(:vm))
+      conversion_host = FactoryBot.create(:conversion_host, :resource => FactoryBot.create(:vm_openstack))
```
Fryguy: Where does the vm_openstack factory come from?  If it's in core, that's probably ok, otherwise this comes back to my pluggability question.

app/controllers/api/conversion_hosts_controller.rb
djberg96: I could use the `.supports?` method, but that of course means I have to constantize first.

spec/requests/conversion_hosts_spec.rb
djberg96: Ok, removed.

app/controllers/api/conversion_hosts_controller.rb
djberg96: Ok, updated.

spec/requests/conversion_hosts_spec.rb
djberg96: It's from core: https://github.com/ManageIQ/manageiq/blob/master/spec/factories/vm_or_template.rb#L121

jameswnl: Just saying: That factory should have been in the provider's repo. We should have all those factories in their own repo. Core is about concepts that transcend specific implementations (providers) and should operates only on core objects.

app/controllers/api/conversion_hosts_controller.rb
Fryguy: as long as the constantize happens at execution time and not boot time, I'm fine with that.

Fryguy: Alternately, you can traverse all provider plugins with `Vmdb::Plugins.provider_plugins.each`, so no need to call constantize.

djberg96: I think scanning the entire class hierarchy for every provider is simply not practical. Instead, I've used a combination of `safe_constantize` and the baked in `supports_conversion_host?` logic to validate the params. I think this is as good, safe, and flexible as we can be.

app/controllers/api/conversion_hosts_controller.rb
```diff
@@ -28,9 +22,18 @@ class ConversionHostsController < BaseController
...
 
-      resource_type = VALID_TYPES[data['resource_type']]
+      # The scary constantize call below is mitigated by the fact that it won't get
+      # past the following checks we make before passing along the params.
+
+      resource_type = data['resource_type'].classify.safe_constantize
+
+      raise BadRequestError, "invalid resource_type #{data['resource_type']}" unless resource_type
+
+      unless resource_type.respond_to?(:supports_conversion_host?) && resource_type.supports_conversion_host?
```
Fryguy: Do we want to show the unsupported_reason when it respond_to, but returns false?

Fryguy: Not a show-stopper for merge.

mturley:
@djberg96 @bdunne shouldn't this have a `hammer/yes` label, since it's part of the conversion host enablement feature we planned for a 5.10.z release?

bdunne:
I marked it as `hammer/no` since I didn't see a BZ anywhere.

mturley:
I think this is part of https://bugzilla.redhat.com/show_bug.cgi?id=1622728.

djberg96:
@bdunne Let's backport to hammer if possible.

mturley:
This needs to be `hammer/yes` for https://bugzilla.redhat.com/show_bug.cgi?id=1622728, right @djberg96?

djberg96:
@miq-bot remove_label hammer/no

djberg96:
@miq-bot add_label hammer/yes

simaishi:
Hammer backport details:

```
$ git log -1
commit a911458e8cb39c7bda4fbbe9a6976b092f483fc5
Date:   Tue Feb 26 15:35:45 2019 -0500

    Merge pull request #535 from djberg96/enable_disable_conversion_hosts
    
    Add ability to enable and disable conversion hosts
    
    (cherry picked from commit 05247d95d28e305fba8e9a9965c88cd8d19b2bd4)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1702019
```

PR #537, opened 2019-01-04
Remove SQL select from exception error messages.

jvlcek:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1661445

This PR will ensure SQL database SELECT statements are not returned in the result of
failed API authentication calls.

**To test:**

Exercise the API with a curl command similar to the following:

```bash
curl -X GET 'https://<IP>/api/auth?requester_type=ui' -H 'Authorization: Basic testing' -H 'Cache-Control: no-cache' -H 'Postman-Token: c791ee47-dd7d-42c9-b9f2-a2c681d5dfe5' -k
```

Ensure that the SQL SELECT is not included in the returned error message.

**Examples of an _invalid_ and a _valid_ return:**

- This is an example of an _invalid_ return that includes the SQL SELECT.

```bash
Actual results:
{"error":{"kind":"unauthorized","message":"PG::CharacterNotInRepertoire: ERROR:  invalid byte sequence for encoding \"UTF8\": 0xb5\n: SELECT  \"users\".* FROM \"users\" WHERE (\"users\".\"id\" BETWEEN $1 AND $2) AND \"users\".\"userid\" = $3 LIMIT $4","klass":"MiqException::MiqEVMLoginError"}}
```

- This is an example of a _valid_ return, with the SQL SELECT removed.

```bash
Actual results:
{"error":{"kind":"unauthorized","message":"PG::CharacterNotInRepertoire: ERROR:  invalid byte sequence for encoding \"UTF8\": 0xb5\n: "klass":"MiqException::MiqEVMLoginError"}}
```

### DIFF
--- a/lib/services/api/error_serializer.rb
+++ b/lib/services/api/error_serializer.rb
@@ -2,12 +2,14 @@ module Api
   class ErrorSerializer
     attr_reader :kind, :error
 
+    SQL_STATEMENTS = " SELECT | UPDATE | INSERT | DELETE ".freeze
+
     def initialize(kind, error)
       @kind = kind
       @error = error
     end
 
-    def serialize(remove_sql_select = false)
+    def serialize(no_sql_query = false)
       result = {
         :error => {
           :kind    => kind,
@@ -15,9 +17,16 @@ module Api
           :klass   => error.class.name
         }
       }
-      result[:error][:message] = error.message.split(/ SELECT /i)[0] if remove_sql_select
+      result[:error][:message] = remove_sql_query(error.message) if no_sql_query
       result[:error][:backtrace] = error.backtrace.join("\n") if Rails.env.test?
       result
+    end
+
+    private
+
+    def remove_sql_query(message)
+      return message unless message =~ /PG.*ERROR/i
+      message.split(/#{SQL_STATEMENTS}/i)[0]
     end
   end
 end
--- a/lib/services/api/error_serializer.rb
+++ b/lib/services/api/error_serializer.rb
@@ -7,7 +7,7 @@ module Api
       @error = error
     end
 
-    def serialize
+    def serialize(remove_sql_select = false)
       result = {
         :error => {
           :kind    => kind,
@@ -15,6 +15,7 @@ module Api
           :klass   => error.class.name
         }
       }
+      result[:error][:message] = error.message.split(/ SELECT /i)[0] if remove_sql_select
       result[:error][:backtrace] = error.backtrace.join("\n") if Rails.env.test?
       result
     end
--- a/app/controllers/api/base_controller/authentication.rb
+++ b/app/controllers/api/base_controller/authentication.rb
@@ -43,9 +43,9 @@ module Api
         response.headers["Content-Type"] = "application/json"
         case auth_mechanism
         when :system, :token
-          render :status => 401, :json => ErrorSerializer.new(:unauthorized, e).serialize.to_json
+          render :status => 401, :json => ErrorSerializer.new(:unauthorized, e).serialize(true).to_json
         when :basic, nil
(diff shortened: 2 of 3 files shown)


jvlcek:
@miq-bot add_label bug
@miq-bot add_label hammer/yes
@miq-bot assign @gtanzillo

jvlcek:
@bdunne and @gtanzillo Please review.

I tried to preserver the existing functionality, which may not be necessary but seemed to be the safer approach. If you think it is unnecessary to preserver the existing functionality ErrorSerializer.serialize could be simplified a little.

Please let me know if you think it is valuable to preserver the existing functionality.

Thank you! JoeV

jvlcek:
@juliancheal Can you please take a look?

lib/services/api/error_serializer.rb
```diff
@@ -7,14 +7,15 @@ def initialize(kind, error)
...
-    def serialize
+    def serialize(remove_sql_select = false)
       result = {
         :error => {
           :kind    => kind,
           :message => error.message,
           :klass   => error.class.name
         }
       }
+      result[:error][:message] = error.message.split(/ SELECT /i)[0] if remove_sql_select
```
gtanzillo: I think this regexp may need to be updated to include `UPDATE`, `INSERT` and `DELETE`.

That said, I wonder if theres a cleaner way detect SQL in the error message. Perhaps -
`error.kind_of?(PG::Error)`

juliancheal:
LGTM, but I agree with @gtanzillo, let's see if we can detect all SQL errors.

jvlcek:
@juliancheal and @gtanzillo Thank you and I do agree. I had meant to ask if, even though ATM the only PG query that would hit this code is a `SELECT`, should I make it more generic and handle  `UPDATE`, `INSERT` and `DELETE`.

I will do that. Thank you! JoeV

lib/services/api/error_serializer.rb
jvlcek: @gtanzillo Maybe I'm missing something because I do not understand the benefit of detecting if the error is a SQL error? We do not want to omit the fact that there was a PG ERROR in the API return. The part we don't want to return is the SQL statement and it's arguments.

Although I do understand  confirming that the error as a  PG error before attempting to remove the SQL statement, with its associated arguments, would be valuable.

That is to say

**I think a command of the form:** 
```bash
% curl -X GET 'https://miq-host/api/auth?requester_type=ui' -H 'Authorization: Bache' -H 'Postman-Token: <token>' -k -u "admin:<pw>"
```

**Should return:**
```bash
{"error":{"kind":"unauthorized","message":"PG::CharacterNotInRepertoire: ERROR:  invalid byte sequence for encoding \"UTF8\": 0xb5\n:","klass":"Api::AuthenticationError"}
```

**but should not return an error message like this example containing the SQL query:**
```bash
{"error":{"kind":"unauthorized","message":"PG::CharacterNotInRepertoire: ERROR:  invalid byte sequence for encoding \"UTF8\": 0xb5\n: SELECT  \"users\".* FROM \"someplace\" WHERE ...","klass":"Api::AuthenticationError"}}
```
Please let me know if I am missing something.

Thank you! JoeV

simaishi:
Hammer backport details:

```
$ git log -1
commit 93195015aedcfd90fc5733a8ca22aa03fc083934
Date:   Thu Jan 10 17:40:12 2019 -0500

    Merge pull request #537 from jvlcek/bz_1661445_error_w_sql
    
    Remove SQL select from exception error messages.
    
    (cherry picked from commit 325a5a105a7aefc4c6864823890f72a071c14360)
    
    Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1686021
```

Issue #539, opened 2019-01-09
Add/Remove CustomButton(s) to CustomButtonSet


ZitaNemeckova:
It's impossible to add/remove a CustomButton to/from CustomButtonSet via API but it's possible to change `set_data[:button_order]` that only says in what order buttons go. It would good to add it as well so UI can only call API and not use extra http request to do it :)

@miq-bot assign @skateman 

Related to https://github.com/ManageIQ/manageiq-ui-classic/pull/4913

skateman:
I see two possible ways how this can be added to the API:
* `custom_buttons` as a subcollection of `custom_button_sets`, IMO this doesn't really make sense because a button can be also unassigned
* passing a `custom_button_set_id` when creating/editing a `custom_button` and doing the necessary db changes in the `Api::CustomButtonsController` under `create_resource` and `edit_resource`

I vote for the 2nd option, but I might need some help with error handling.

skateman:
Another idea thanks to @himdel: having a `before_save` callback on the `CustomButtonSet` models that maps `button_order` to the `members=` method:
```ruby
  before_save :update_members
  def update_members
    self.members = CustomButton.find(set_data[:button_order])
  end
```

I really like this idea, but I don't want to introduce asymmetry with other parts of the API that I don't know :worried: so it would be really nice if I'd have some input on these options from someone who understands the codebase a little better.

djberg96:
I guess I don't see the issue with making `custom_button` a subcollection of `custom_button_sets`. If you need to create an unassigned button, don't use the custom_button_set api.

himdel:
My concern is that we currently have 2 independent places where we list the custom buttons associated with the set (one as an array of ids in `button_order`, and the other as an unordered(? - my assumption) set in `members` (or `children`)).

If we had a nice way of having our subcollection, and custom-ordering it too, without that duplication, it would be perfect :).

gtanzillo:
@abellotti and I looked at this today

We were thinking that you could have an action on `custom_button_sets` that allows for assigning one or more buttons to the set 

```json
        POST /api/custom_button_sets/:id

                actions -> "assign_custom_buttons"
                        
                        "resources" : [
                                { "href" : "/api/custom_buttons/1" },
                                { "id" : "2" },
                                { "href" : "/api/custom_buttons/3" },
                        ]

                actions -> "unassign_custom_buttons"

                        "resources" : [
                                { "href" : "/api/custom_buttons/1" },
[...]

PR #541, opened 2019-01-14
[WIP] Make custom_buttons a subcollection of custom_button_sets

djberg96:
This PR makes `custom_buttons` a subcollection of `custom_button_sets` and allows for the easy retrieval of, or deletion of, individual custom_buttons from the collection. So, now you can do this:

```
GET /api/custom_button_sets/:c_id/custom_buttons
GET /api/custom_button_sets/:c_id/custom_buttons/:s_id
POST /api/custom_button_sets/:c_id/custom_buttons/:s_id { "action": "delete" }
```

Tested via curl and seems to work but a WIP for now until I can get some specs added and/or Zita and the UI team decide this approach is acceptable.

Addresses https://github.com/ManageIQ/manageiq-api/issues/539

### DIFF
--- a/app/controllers/api/custom_button_sets_controller.rb
+++ b/app/controllers/api/custom_button_sets_controller.rb
@@ -1,5 +1,7 @@
 module Api
   class CustomButtonSetsController < BaseController
+    include Subcollections::CustomButtons
+
     def create_resource(type, id, data)
       super(type, id, data.deep_symbolize_keys)
     end
--- a/app/controllers/api/subcollections/custom_buttons.rb
+++ b/app/controllers/api/subcollections/custom_buttons.rb
@@ -1,0 +1,13 @@
+module Api
+  module Subcollections
+    module CustomButtons
+      def custom_buttons_query_resource(object)
+        object.custom_buttons
+      end
+
+      def custom_buttons_delete_resource(_object, type, id, data)
+        delete_resource(type, id, data)
+      end
+    end
+  end
+end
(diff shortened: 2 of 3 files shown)


ZitaNemeckova:
@djberg96 There's `s` missing in description :) 
```
GET /api/custom_button_sets/:c_id/custom_buttons/:s_id
POST /api/custom_button_sets/:c_id/custom_buttons/:s_id { "action": "delete" }
```

djberg96:
@ZitaNemeckova Oops, sorry for the confusion, I've updated the original comment.

himdel:
@djberg96 I don't think this approach actually helps us.

Having the buttons acessible as a subcollection is definitely useful, we could eventually use it to get the list of buttons in one request.

But.. I'd prefer not having to ever **edit** that collection in the first place. We never want the collection to have different items than `custom_button_set.set_data[:button_order]`, and the collection does not support setting the order, so I don't think it's useful to expose "add" or "delete" this way.

(Also, there is currently no "add" and "delete" actually deletes the button completely, not just from the one collection.)

himdel:
Thanks :) 

We may still need those it backend decides otherwise, but if it passes... :)

himdel:
LGTM :+1:

djberg96:
Ok, I'll add some specs and remove the wip label once that's done.

ZitaNemeckova:
@djberg96 It gives all correct buttons but unsorted. Is it possible to sort it by `set_data[:button_order]`?

djberg96:
Digging into `ActsAsMiqSet` to see what our options are...

lpichler:
@djberg96 should I reopen it ?

PR #542, opened 2019-01-15
Do not request for basic auth if the auth_requester is set to UI

skateman:
When entering a wrong username/password on the UI login page, an HTTP basic auth popup window shows up before displaying the invalid username or password error notification. This is a standard behavior of Chrome when the server asks for using basic auth.

As a solution I added the new `:ui` authentication mechanism which differs from the Basic only in the `requester_type` parameter that is set to `ui`. Later I'm sending back just a 401 error instead of asking for a new basic auth and so not displaying the popup anymore.

**Before:**
screenshot from 2019-01-15 13-44-38

**After:**
screenshot from 2019-01-15 13-45-31

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1632718

@miq-bot add_label bug, hammer/yes
@miq-bot add_reviewer @bdunne 
@miq-bot add_reviewer @djberg96 

FYI @PanSpagetka @himdel @martinpovolny

### DIFF
--- a/app/controllers/api/base_controller/authentication.rb
+++ b/app/controllers/api/base_controller/authentication.rb
@@ -8,6 +8,8 @@ module Api
           :system
         elsif request.headers[HttpHeaders::AUTH_TOKEN]
           :token
+        elsif request.headers["HTTP_AUTHORIZATION"] && params[:requester_type] == 'ui'
+          :basic_ui
         elsif request.headers["HTTP_AUTHORIZATION"]
           :basic
         else
@@ -25,7 +27,7 @@ module Api
           authenticate_with_system_token(request.headers[HttpHeaders::MIQ_TOKEN])
         when :token
           authenticate_with_user_token(request.headers[HttpHeaders::AUTH_TOKEN])
-        when :basic, nil
+        when :basic, :basic_ui, nil
           success = authenticate_with_http_basic do |u, p|
             begin
               timeout = ::Settings.api.authentication_timeout.to_i_with_method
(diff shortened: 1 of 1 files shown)


himdel:
LGTM, we're sending `requester_type=ui` for just the login request (both ops ui and service ui), so sounds good :+1:

Fryguy:
LGTM :+1: would like @abellotti's eyes as well.

skateman:
@Fryguy ideally I'd like to just verify the same cookie as we use for UI session, then this can all go away. However, this solution is fast and backportable :wink:

abellotti:
while this solution works, it's a bit misleading from an auth_mechanism perspective.  'ui' does not tell me it's basic auth or token based ui (we now store requester_type ui as part of the token metadata).  So the Authentication log showing 'ui' is unclear.  maybe add requester_type to the Authentication log line. Would like to keep auth_mechanism as system, token or basic. Isn't it possible to check the requester_type in the require_api_user_token method in addition to auth_mechanism ? 

Another option is to have finer granularity on the basic type, i.e. instead of 'ui' have 'basic_ui' and 'basic_ws'.  Though not too crazy about that either, might as well just log & reference the optional requester_type where needed as discussed above.

skateman:
@abellotti I'm okay with changing the names to `ui_basic`, but the websocket token is never requested through basic auth, but with a normal token, so I'd like to keep it.

simaishi:
Hammer backport details:

```
$ git log -1
commit af4936f75e349760111bdca03b070f3a244f7e89
Date:   Thu Jan 17 09:11:02 2019 -0500

    Merge pull request #542 from skateman/ui-login-basic
    
    Do not request for basic auth if the auth_requester is set to UI
    
    (cherry picked from commit 80ad5a0e666eb737d0639d653e084c10717a6d03)
    
    Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1767018
```

PR #543, opened 2019-01-15
Allow authentication against the _vmdb_session cookie for UI only

skateman:
We're consuming the API from our UI with token authentication, which is not the best because we have to poll the API to renew the token. This made the complexity of the login code unnecessarily high and also the token renewal floods the log with useless information every X seconds.

The API and the UI workers are both using the same memcached server to store short-living information, such as sessions or tokens. As the two workers are basically the same application, they implicitly have the same session cookie and so through memcached, sharing the access to the same session. However, the API was not using this session, for now.

I'm proposing a new authentication mechanism for the UI only (replaces https://github.com/ManageIQ/manageiq-api/pull/542) that will test if the CSRF header is set on the API request. This header will be used from the UI only and will also prevent from opening a HTTP basic auth dialog even if the session times out. This will make https://github.com/ManageIQ/manageiq-api/pull/542 completely obsolete.

I've discussed this with @Fryguy already and honestly, due to the simplicity of this feature I am afraid that this is too good to be true. Anyway, we will use this PR with @himdel to work further on the UI part and make changes if something goes wrong. The behavior of the API should not be affected by this change, however, we will be able to drop a few hunderd lines of JS from the UI.

When using external authentication, the apache would force basic auth on every unauthorized API request, but there's a fix for that here: https://github.com/ManageIQ/manageiq-appliance/pull/224

### DIFF
--- a/app/controllers/api/base_controller/authentication.rb
+++ b/app/controllers/api/base_controller/authentication.rb
@@ -8,10 +8,13 @@ module Api
           :system
         elsif request.headers[HttpHeaders::AUTH_TOKEN]
           :token
-        elsif request.headers["HTTP_AUTHORIZATION"] && params[:requester_type] == 'ui'
-          :basic_ui
         elsif request.headers["HTTP_AUTHORIZATION"]
           :basic
+        elsif request.x_csrf_token
+          # Even if the session cookie is not set, we want to consider a request
+          # as a UI authentication request. Otherwise the response would force
+          # the browser to throw an undesired HTTP basic authentication dialog.
+          :ui_session
         else
           # no attempt at authentication, usually falls back to :basic
           nil
@@ -27,7 +30,10 @@ module Api
           authenticate_with_system_token(request.headers[HttpHeaders::MIQ_TOKEN])
         when :token
           authenticate_with_user_token(request.headers[HttpHeaders::AUTH_TOKEN])
-        when :basic, :basic_ui, nil
+        when :ui_session
+          raise AuthenticationError unless valid_ui_session?
+          auth_user(session[:userid])
+        when :basic, nil
           success = authenticate_with_http_basic do |u, p|
             begin
               timeout = ::Settings.api.authentication_timeout.to_i_with_method
@@ -44,7 +50,7 @@ module Api
         api_log_error("AuthenticationError: #{e.message}")
         response.headers["Content-Type"] = "application/json"
         case auth_mechanism
-        when :system, :token, :basic_ui
+        when :system, :token, :ui_session
           render :status => 401, :json => ErrorSerializer.new(:unauthorized, e).serialize(true).to_json
         when :basic, nil
           request_http_basic_authentication("Application", ErrorSerializer.new(:unauthorized, e).serialize(true).to_json)
@@ -126,6 +132,14 @@ module Api
       def validate_system_token_timestamp(timestamp)
         raise "Missing timestamp" if timestamp.blank?
         raise "Invalid timestamp #{timestamp} specified" if SYSTEM_TOKEN_TTL.ago.utc > timestamp
+      end
+
+      def valid_ui_session?
+        [
+          valid_authenticity_token?(session, request.x_csrf_token), # CSRF token be set and valid
+          session[:userid].present?,                                # session has a userid stored
+          request.origin.nil? || request.origin == request.base_url # origin header if set matches base_url
+        ].all?
       end
     end
   end
--- a/spec/requests/authentication_spec.rb
+++ b/spec/requests/authentication_spec.rb
@@ -202,6 +202,48 @@ describe "Authentication API" do
     end
   end
 
+  context "UI authentication against the rails session" do
+    it 'succeeds with valid session and valid CSRF token' do
+      # Inspired by https://relishapp.com/rspec/rspec-mocks/docs/configuring-responses/block-implementation
+      # Store the original new method because we're going to mock it
+      original_new = ActionDispatch::Request::Session.method(:new)
+      # This is a little crazy: the session is not accessible before the HTTP request is fired,
+      # therefore, we have to mock the new method, call the original and catch its response.
+      # Then we can mock the :[] method on the response and so set session variables.
+      # In the end the block can simply return the response as if nothing happened.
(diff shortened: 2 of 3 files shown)


skateman:
@martinpovolny I've added the CSRF token verification, it was way simpler than I thought. I assume we also want to verify the origin, but please let me know if not. cc @himdel

jvlcek:
I worked with @skateman to test these changes. The testing done was with SAML via Keycloak, External Auth and SSO (kerberos) via IPA. I believe this is a good test sampling. LGTM JoeV

skateman:
@miq-bot rm_label unmergeable

app/controllers/api/base_controller/authentication.rb
```diff
@@ -127,6 +136,14 @@ def validate_system_token_timestamp(timestamp)
         raise "Missing timestamp" if timestamp.blank?
         raise "Invalid timestamp #{timestamp} specified" if SYSTEM_TOKEN_TTL.ago.utc > timestamp
       end
+
+      def valid_ui_session?
+        [
+          valid_authenticity_token?(session, request.x_csrf_token), # CSRF token be set and valid
+          session[:userid].present?,                                # session has a userid stored
```
Fryguy: Does the api worker have access to the session?

spec/requests/authentication_spec.rb
```diff
@@ -202,6 +202,48 @@
     end
   end
 
+  context "UI authentication against the rails session" do
```
Fryguy: Instead of allows here, can we change them to expect to ensure they are called, in case code changes in the future?

app/controllers/api/base_controller/authentication.rb
skateman: @Fryguy yes, first I wanted to reinvent the wheel with connecting to memcached and all the sugar around, but it actually works. Tested on a live appliance...

spec/requests/authentication_spec.rb
skateman: :+1:

skateman: Except of one, this is doable, the `:[]` mock has to stay as it is because it can be called multiple times and anywait, having an expect would not add any value into this.

Fryguy:
>  The behavior of the API should not be affected by this change, however, we will be able to drop a few hunderd lines of JS from the UI.

wow 😮

app/controllers/api/base_controller/authentication.rb
```diff
@@ -2,16 +2,22 @@ module Api
...
-          :basic_ui
         elsif request.headers["HTTP_AUTHORIZATION"]
           :basic
+        elsif request.cookies[SESSION_COOKIE] || request.x_csrf_token
+          # A successful UI authentication requires both a valid session and a CSRF token,
+          # however, the condition is a little more permissive, because if a session expires
+          # only the CSRF token is being sent and we want the API to know that the request was
+          # initiated from the UI. Otherwise the response to a such request would force the
+          # browser to throw an undesired HTTP basic authentication dialog.
+          :cookie_ui
```
Fryguy: What I'm reading in your comment is that the CSRF token will always exist, so why have the check for the session cookie here at all.  That is, it seems this code can just be

```ruby
...
elsif request.x_csrf_token
  :cookie_ui
else
...
```

Fryguy: So, that being said, where do we use the SESSION_COOKIE at all after this point?

app/controllers/api/base_controller/authentication.rb
```diff
@@ -27,7 +33,10 @@ def require_api_user_or_token
           authenticate_with_system_token(request.headers[HttpHeaders::MIQ_TOKEN])
         when :token
           authenticate_with_user_token(request.headers[HttpHeaders::AUTH_TOKEN])
-        when :basic, :basic_ui, nil
+        when :cookie_ui
+          raise AuthenticationError unless valid_ui_session?
+          auth_user(session[:userid])
```
Fryguy: I'm assuming the cookie check is in here?

Fryguy:
I have a few questions but overall I'm ok with this approach.  I would like another set of eyes though... @jrafanie ?

app/controllers/api/base_controller/authentication.rb
skateman: @Fryguy in my tests the session cookie is sent automatically, that's AFAIK default rack behavior. You'll have a session cookie in Rails after your first request to the site, even if you're not logged in. But you're right, I'll simplify this part...

app/controllers/api/base_controller/authentication.rb
skateman: It actually happens earlier in the `valid_ui_session?` when validating the CSRF token as it is also part of the session.

jrafanie:
> I have a few questions but overall I'm ok with this approach. I would like another set of eyes though... @jrafanie ?

Yeah, I looked at it.  @skateman and I chatted and I neglected to comment here.  I'm good with the changes although I'm not sure what to do about the [verifying of the origin](https://github.com/ManageIQ/manageiq-api/pull/543#issuecomment-454906343).

skateman:
@jrafanie I've added the origin verification, it doesn't hurt as it allows `nil` origin as well which is the standard Rails behavior. The browser sends the `Origin` header with `POST` requests only, so those will get validated (tested - works). We can append the header manually to other type of requests in the UI code, but that's a decision to make in the [UI PR](https://github.com/ManageIQ/manageiq-ui-classic/pull/5164).

martinpovolny:
As discussed f2f. When talking with the API from a browser there should be 2 secrets:

 * one in possession of the browser, but out of controll of the Javascript -- that is the (http-only) cookie.
 * second in possession of the Javascript code so that it does not get automatically added by the browser (unless requested by javascript code)

The 1st is vulnerable to XSRF but relatively safe from XSS, the 2nd is vulnerable to XSS but relatively safe from XSRF.

Then together with CSP we should be safe. (We still have to work on making the CSP policy stricter.)

I have to confess that I have not read this PR in details but I believe that your solution is implemented as we discussed.

:+1:

skateman:
@martinpovolny yes it is, now it hangs on the UI side here: https://github.com/ManageIQ/manageiq-ui-classic/pull/5164

PR #547, opened 2019-01-23
Add pagination and ordering to hash column MiqReporResult#result_set

lpichler:
example:
**Sorting and Pagination:**

```
curl --user admin:smartvm -X GET -H "Accept: application/json" "http://localhost:3000/api/results/1242?hash_attribute=result_set&sort_by=disks.size&sort_order=desc&limit=3&offset=3"
```

```
/api/results/1242?hash_attribute=result_set&sort_by=disks.size&sort_order=desc&limit=4&offset=3
```
`hash_attribute=value - enables hash sorting and pagination for value column`
`sort_by - column from MiqReportResult#report#col_order`
`sort_order - asc or desc`
`limit - count of records for one page`
`offset -result will start with this record from whole result set`

returns:
```
{
   <resource>
    "count": 4,
    "pages": 1,
    "subcount": 4,
    "result_set": [
        {
            "disks.size": 1073741824,
            "mem_cpu": 1024,
            "name": "testloic",
            "retires_on": "2017-11-29T12:48:07Z",
            "storages.disk_size": null,
            "virtual_custom_attribute_Name": null
        },
        {
            "disks.size": 10737418240,
            "mem_cpu": 1024,
            "name": "demo-natevm",
            "retires_on": "2018-01-18T14:25:29Z",
            "storages.disk_size": 0,
            "virtual_custom_attribute_Name": null
        },
        {
            "disks.size": 10737418240,
            "mem_cpu": 1024,
            "name": "testloic",
            "retires_on": "2017-11-29T12:48:07Z",
            "storages.disk_size": null,
            "virtual_custom_attribute_Name": null
        },
        {
            "disks.size": 21474836480,
            "mem_cpu": 2048,
            "name": "cf-cirros001",
            "retires_on": "2018-02-15T19:36:26Z",
            "storages.disk_size": null,
            "virtual_custom_attribute_Name": null
        }
    ],
```

**Filtering:**
```
curl --user admin:smartvm -X GET -H "Accept: application/json" "http://localhost:3000/api/results/1242?hash_attribute=result_set&filter_column=disks.size&filter_string=55"
```
```
/api/results/1242?hash_attribute=result_set&filter_column=disks.size&filter_string=55
```

**filter_column** - chosen column for filtering
**filter_string** - this string value is used for filtering on  `:filter_column `, record of report result will be returned in API if  `:filter_string` is contained(case sensitive - is this ok ?) in value of   `:filter_column `.

Note: You can mix together Sorting and Pagination and Filtering

cc @martinpovolny 

- [x] https://github.com/ManageIQ/more_core_extensions/pull/67
- [x] https://github.com/ManageIQ/manageiq/pull/18532
- [x]  https://github.com/ManageIQ/manageiq/pull/18426
- [x] specs

# Links
https://bugzilla.redhat.com/show_bug.cgi?id=1678150

### DIFF
--- a/app/controllers/api/results_controller.rb
+++ b/app/controllers/api/results_controller.rb
@@ -1,29 +1,11 @@
 module Api
   class ResultsController < BaseController
+    include Parameters::ResultsController
+
     before_action :set_additional_attributes, :only => [:index, :show]
 
     def results_search_conditions
       MiqReportResult.for_user(User.current_user).where_clause.ast
-    end
-
-    def sort_order
-      params['sort_order'] == 'desc' ? :descending : :ascending
-    end
-
-    def param_result_set?
-      params.key?(:hash_attribute) && params[:hash_attribute] == "result_set"
-    end
-
-    def report_options
-      params.merge(:sort_by => params['sort_by'], :sort_order => sort_order).merge(filter_options)
-    end
-
-    def filter_options
-      filtering_enabled? ? {:filter_string => params[:filter_string], :filter_column => params[:filter_column]} : {}
-    end
-
-    def filtering_enabled?
-      params.key?(:filter_column) && params.key?(:filter_string) && params[:filter_string]
     end
 
     def result_set
--- a/app/controllers/api/results_controller.rb
+++ b/app/controllers/api/results_controller.rb
@@ -6,6 +6,74 @@ module Api
       MiqReportResult.for_user(User.current_user).where_clause.ast
     end
 
+    def apply_limit_and_offset(results, options)
+      results.slice(options['offset'].to_i, options['limit'].to_i) || []
+    end
+
+    def sort_order
+      params['sort_order'] == 'desc' ? :descending : :ascending
+    end
+
+    def sort_by(report_result)
+      sort_by(report_result).split(",").collect do |attr|
+        if report(report_result).col_order&.include?(attr)
+          attr
+        else
+          raise BadRequestError, "#{attr} is not a valid attribute for #{report_result.name}"
+        end
+      end.compact
+    end
+
+    def param_result_set?
+      params.key?(:hash_attribute) && params[:hash_attribute] == "result_set"
+    end
+
+    def format_result_set(miq_report, result_set)
+      tz = miq_report.get_time_zone(Time.zone)
+
+      col_format_hash = miq_report.col_order.zip(miq_report.col_formats).to_h
+
+      result_set.map! do |row|
+        row.map do |key, _|
+          [key, miq_report.format_column(key, row, tz, col_format_hash[key])]
+        end.to_h
+      end
+    end
+
+    def report(report_result)
+      @report ||= report_result.report || report_result.miq_report
+    end
+
+    def sort_by(report_result)
+      params['sort_by'] || report(report_result)&.sortby || report(report_result)&.col_order
+    end
+
+    def result_set
+      ensure_pagination
+
+      report_result = MiqReportResult.for_user(User.current_user).find(@req.collection_id)
+      result_set = report_result.result_set
+
+      if result_set.present? && report(report_result)
+        result_set = result_set.stable_sort_by(sort_by(report_result), sort_order)
+        result_set = apply_limit_and_offset(result_set, params)
+        result_set.map! { |x| x.slice(*report(report_result).col_order) }
+
+        result_set = format_result_set(report(report_result), result_set)
+      end
+
+      hash = {:result_set => result_set,
+              :count      => report_result.result_set.count,
+              :subcount   => result_set.count,
+              :pages      => (report_result.result_set.count / params['limit'].to_f).ceil}
+
+      report_result.attributes.merge(hash)
+    end
+
+    def show
+      param_result_set? ? render_resource(:results, result_set) : super
+    end
+
     def find_results(id)
       MiqReportResult.for_user(User.current_user).find(id)
     end
--- a/app/controllers/api/results_controller.rb
+++ b/app/controllers/api/results_controller.rb
@@ -6,66 +6,36 @@ module Api
       MiqReportResult.for_user(User.current_user).where_clause.ast
     end
 
-    def apply_limit_and_offset(results, options)
-      results.slice(options['offset'].to_i, options['limit'].to_i) || []
-    end
-
     def sort_order
       params['sort_order'] == 'desc' ? :descending : :ascending
-    end
-
-    def sort_by(report_result)
-      sort_by(report_result).split(",").collect do |attr|
-        if report(report_result).col_order&.include?(attr)
-          attr
-        else
-          raise BadRequestError, "#{attr} is not a valid attribute for #{report_result.name}"
-        end
-      end.compact
     end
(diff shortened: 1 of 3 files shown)


martinpovolny:
@lpichler : first of all: Thank you for starting the effort! :+1: 

Now the not so good part: You need to sort according to column types as defined in the report. Afaik it includes things such as "10 MB" vs "100 kB".

Anyway. This is better than nothing. So I'd even merge this and then have a better sorting implemented as a follow up PR.

Are you also going to fix the problem of missing column type information in the report data? (Mentioned here: https://github.com/ManageIQ/manageiq-ui-classic/pull/5171).

Fryguy:
Before you introduce a bumped more_core_extensions, I'd like to a) talk about the PR to more_core, because I don't necessarily agree with it and b) The bump to more_core also brings in an updated Benchmark.realtime_store that we have to remove from manageiq-gems-pending first.

app/controllers/api/results_controller.rb
```diff
@@ -1,3 +1,33 @@
+class Array
```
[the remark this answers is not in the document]
lpichler: [I pulled out it](https://github.com/ManageIQ/more_core_extensions/pull/67) because it is used in report generator. In this PR is just for usage reason for @martinpovolny UI work.

app/controllers/api/results_controller.rb
```diff
@@ -6,6 +36,50 @@ def results_search_conditions
       MiqReportResult.for_user(User.current_user).where_clause.ast
     end
 
+    def apply_limit_and_offset(results, options)
+      results.slice(options['offset'].to_i, options['limit'].to_i) || []
```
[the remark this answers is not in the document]
lpichler: it is not, it came form `MiqReportResult#BinaryBlob#data` and also we html_details (MiqReportResult#MiqReportResultDetails) - but we are storing only html. It would be nice for API reasons to store also unformatted data - but on other hand it would be third place where would be report results stored - so I am not sure about this solution

kbrock:
@libor is this related to https://github.com/ManageIQ/manageiq-api/pull/518 ?

lpichler:
@kbrock Thanks! it looks like that it is doing same - which approach do you  prefer ?

kbrock:
Lets do this, You have momentum, lets keep running with it.

I didn't realize that this data was not `offset`/`limit` friendly. Makes me sad but that is what it is.

app/controllers/api/results_controller.rb
```diff
@@ -6,6 +6,77 @@ def results_search_conditions
...
+          [key, miq_report.format_column(key, row, tz, col_format_hash[key])]
+        end.to_h
+      end
+    end
+
+    def report(report_result)
+      @report ||= report_result.report || report_result.miq_report
+    end
+
+    def sort_by(report_result)
```
kbrock: is there a reason to join here?
seems we just `split` it once it comes out

app/controllers/api/results_controller.rb
```diff
@@ -6,6 +6,77 @@ def results_search_conditions
...
+
+      sort_colums = sort_by(report_result).split(",").collect do |attr|
+        if report(report_result).col_order&.include?(attr)
+          attr
+        else
+          raise BadRequestError, "#{attr} is not a valid attribute for #{report_result.name}"
+        end
+      end.compact
+
+      {:sort_by => sort_colums, :sort_order => (params['sort_order'] || :ascending).to_sym }
```
kbrock: up to you, but I was thinking you could just return the values themselves

It would be easier for me to read on the other end

```ruby
sort_by, sort_order = sort_params_from(report_result)
```

```suggestion
      [sort_colums, (params['sort_order'] || :ascending).to_sym]
```

app/controllers/api/results_controller.rb
```diff
@@ -6,6 +6,77 @@ def results_search_conditions
       MiqReportResult.for_user(User.current_user).where_clause.ast
     end
 
+    def apply_limit_and_offset(results, options)
+      results.slice(options['offset'].to_i, options['limit'].to_i) || []
+    end
+
+    def sort_params_from(report_result)
+      params['sort_order'] ||= :ascending
```
kbrock: is `params` an external variable?

are you doing it this way to ensure the side effect of modifying the `params`?

If not, could you just use `sort_order` local variable?

lpichler:
@kbrock thanks for feedback

I updated PR according to all your comments.
I moved methods to core to models and update processing of params (sort_by and sort_params)

app/controllers/api/results_controller.rb
lpichler: no, we need to pass array, thanks!

kbrock:
This turned out beautifully @lpichler

lpichler:
@miq-bot assign @abellotti

lpichler:

app/controllers/api/results_controller.rb
```diff
@@ -6,6 +6,44 @@ def results_search_conditions
...
+      params.merge(:sort_by => params['sort_by'], :sort_order => sort_order).merge(filter_options)
+    end
+
+    def filter_options
+      filtering_enabled? ? {:filter_string => params[:filter_string], :filter_column => params[:filter_column]} : {}
+    end
+
+    def filtering_enabled?
+      params.key?(:filter_column) && params.key?(:filter_string) && params[:filter_string]
+    end
```
abellotti: can you move sort_order, param_result_set?, report_options, filter_options and filtering_enabled? to app/controllers/api/base_controller/parameter.rb ?

Also, can't we leverage the sort_params method from that file ?

lpichler: > can you move sort_order, param_result_set?, report_options, filter_options and filtering_enabled? to app/controllers/api/base_controller/parameter.rb ?

Yes, done 👍 and I put it to special module.

> 
> Also, can't we leverage the sort_params method from that file ?

it is not possible (if we will not change sort_params somehow), especially because we don't to return from [AREL](https://github.com/manageiq/manageiq-api/blob/713b792ca06baf0be0c610b91c47940539717a81/app/controllers/api/base_controller/parameters.rb) for direction or sorting - sorting in this PR is on array. 

 @abellotti  otherwise thanks for review,
 is it ok now ?

abellotti: Ahh, that makes sense.  Thanks! !

abellotti:
This LGTM!! Thanks @lpichler for enhancing this API. 👍 

@kbrock anything else needed from your side ?

jrafanie:
@kbrock ping ^

abellotti:
Hmm, can't figure out why it says WIP! You can't merge! but not in title or label. Anyone has an idea ?

jrafanie:
@abellotti do you have a plugin that can see this in the PR description, if so, amazing:

image

lpichler:
@abellotti I removed "WIP" string from description, is it possible now ? 🙏

abellotti:
yep, all good. Thanks @lpichler for this great API enhancement. 👍

lpichler:
@abellotti thanks also for review and merging 👍

PR #548, opened 2019-01-25
Add delete automate domain support

jvlcek:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1645041

This PR adds support to allow removal of automate domains via the API
This change only supports deletion of unlocked user domains. System and
GIT based domains will not be deleted.

**To test:**

- Create automated domains in the MiQ Database
- Attempt to delete one of the automated domains in the MiQ Database using the API with the _DELETE_ verb

_e.g.:_ 

`curl --insecure --user <user>:<password> -X DELETE <appliance URL>/api/automate_domains/<domain name>`

- Attempt to delete one of the automated domains in the MiQ Database using the API with the _POST_ verb

_e.g.:_

`curl --insecure --user " <user>:<password>" -H "Content-Type: application/json" --data '{"action":"delete"}' -X POST <URL>/api/automate_domains/<domain name>  `

### DIFF
--- a/app/controllers/api/automate_domains_controller.rb
+++ b/app/controllers/api/automate_domains_controller.rb
@@ -1,5 +1,11 @@
 module Api
   class AutomateDomainsController < BaseController
+    def delete_resource(type, id = nil, _data = {})
+      raise BadRequestError, "Must specify an id for deleting a #{type} resource" unless id
+
+      delete_resource_action(type, id)
+    end
+
     def refresh_from_source_resource(type, id = nil, data = nil)
       raise BadRequestError, "Must specify an id for refreshing a #{type} resource from source" unless id
 
@@ -28,6 +34,23 @@ module Api
     end
 
     private
+
+    def delete_resource_action(type, id)
+      api_action(type, id) do |klass|
+        domain = resource_search(id, type, klass)
+        api_log_info("Delete will be queued for #{automate_domain_ident(domain)}")
+
+        begin
+          # Only delete unlocked user domains. System or GIT based domains will not be deleted.
+          MiqAeDomain.where(:name => domain.name).each { |d| raise "Not deleting. Domain is locked." if d.contents_locked? }
+
+          MiqAeDomain.where(:name => domain.name).each(&:destroy_queue)
+          action_result(true, "Delete queued for #{automate_domain_ident(domain)}")
+        rescue => err
+          action_result(false, err.to_s)
+        end
+      end
+    end
 
     def automate_domain_ident(domain)
       "Automate Domain id:#{domain.id} name:'#{domain.name}'"
--- a/config/api.yml
+++ b/config/api.yml
@@ -242,7 +242,7 @@
     :description: Automate Domains
     :options:
     - :collection
-    :verbs: *gp
+    :verbs: *gpd
     :klass: MiqAeDomain
     :collection_actions:
       :get:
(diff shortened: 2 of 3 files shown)


jvlcek:

jvlcek:
@miq-bot assign

jvlcek:
@miq-bot add_label hammer/no

jvlcek:
@miq-bot assign @juliancheal

jvlcek:
@mkanoor Please review

jvlcek:
@gtanzillo Please review

jvlcek:
rerunning build. Failures look unrelated.

app/controllers/api/automate_domains_controller.rb
```diff
@@ -1,5 +1,21 @@
...
   class AutomateDomainsController < BaseController
+    def delete_resource(type, id = nil, _data = {})
+      raise BadRequestError, "Must specify an id for deleting a #{type} resource" unless id
+
+      api_action(type, id) do |klass|
+        domain = resource_search(id, type, klass)
+        api_log_info("Deleting #{automate_domain_ident(domain)}")
+
+        begin
+          MiqAeDomain.where(:name => domain.name).each(&:destroy)
```
mkanoor: @jvlcek I think we have something called readonly domains like ManageIQ and  Redhat should the user be allowed to delete these domains. Or should this be allowed only for user domains?

mkanoor: @jvlcek is this going to use some RBAC check to see if a user is allowed to delete a domain.

mkanoor: @jvlcek I think instead of destroy it might be better to call destroy_queue since we would want to delete the git directories if the domain is a git based domain.
https://github.com/ManageIQ/manageiq-automation_engine/blob/e136b74e9b2005d050149d9c7019784a531d82e1/app/models/miq_ae_domain.rb#L162

jvlcek: Thank you for your help @mkanoor! It is very much appreciated.
I have pushed a change to address two of your three comments.
I will push the resolution the the RBAC check issue and an updated spec shortly.

JoeV

jvlcek: @mkanoor RBAC is already being handled by the change to the `config/api.yml` file.
e.g.:
https://github.com/ManageIQ/manageiq-api/pull/548/files#diff-15502b585d5033aa8043cc26237a06baR257

`:identifier: miq_ae_domain_delete`

So the user must be associated with a role that has the `miq_ae_domain_delete` product feature enabled.

So RBAC `is` being check. 

Thank you.

app/controllers/api/automate_domains_controller.rb
```diff
@@ -1,5 +1,24 @@
...
+    def delete_resource(type, id = nil, _data = {})
+      raise BadRequestError, "Must specify an id for deleting a #{type} resource" unless id
+
+      api_action(type, id) do |klass|
+        domain = resource_search(id, type, klass)
+        api_log_info("Deleting #{automate_domain_ident(domain)}")
+
+        begin
+          # Do not delete the domain if any are locked
+          MiqAeDomain.where(:name => domain.name).each { |d| raise "Not deleting. Domain is locked." if d.contents_locked? }
```
mkanoor: @jvlcek Looks good, you might want to mention somewhere that we won't be deleting
* System Domains
* GIT Based domains

The only ones you can delete are user domains which are unlocked.

jvlcek: Will do. Thank you.

jvlcek:
I'm going to squash all the commits.
@juliancheal Please take a look.

mkanoor:
👍

juliancheal:
LGTM @jvlcek

PR #550, opened 2019-01-28
One semaphore to exclusively load them all

### LINKED ISSUE #544: Very first request sometimes causes Circular dependency error on ConversionHost
@Hyperkid123 told me about his problem that sometimes his very first API request fails with a HTTP 500 and there's no deterministic way to reproduce it. When I was playing with the API, I also experienced this issue in production and found out that there's a [traceback](https://pastebin.com/cZ9Nxa1D) in the api.log. Then I looked into my local api.log and found out that I had those tracebacks all the time :cry: :sweat_smile: :pray: 

So as far I could understand, this has to do something with [this fix](https://github.com/ManageIQ/manageiq-api/pull/416) created by @jrafanie. I tried looking at the affected `ConversionHost` from any possible angle, but I don't know how this could be a circular dependency. My theory is that there are two models trying to load `ApplicationRecord` at the same time.

@miq-bot add_label bug

jrafanie:
https://bugzilla.redhat.com/show_bug.cgi?id=1671458
Fixes #544

From MartinH's findings in #544, we had one thread at the
`cspec[:klass].constantize` line while another thread was trying to run
`klass.columns_hash`, causing a
`Circular dependency detected while autoloading...` error.

I then tried a bunch of things until I found a way to reliably recreate
this error:

```ruby
require_relative 'config/environment'

threads = []
4.times do
  threads << Thread.new { Api::Environment.time_attributes }
end
threads.collect(&:join)
```

I could then run the above script several times in my shell and get the
`circular dependency` error most of the time.

```
for x in `seq 1 10`; do; bundle exec ruby test.rb; done
```

With this test in place, I then tried a few solutions:

1) Move the `klass.columns_hash` block into the permit_concurrent_loads
block:

```
diff --git a/lib/api/environment.rb b/lib/api/environment.rb
index 87b34f99..f4b6554a 100644
--- a/lib/api/environment.rb
+++ b/lib/api/environment.rb
@@ -22,9 +22,10 @@ module Api
         # Temporary measure to avoid thread race condition which could lead to a deadlock
         ActiveSupport::Dependencies.interlock.permit_concurrent_loads do
           klass = cspec[:klass].constantize
-        end
-        klass.columns_hash.each do |name, typeobj|
-          result << name if %w(date datetime).include?(typeobj.type.to_s)
+
+          klass.columns_hash.each do |name, typeobj|
+            result << name if %w(date datetime).include?(typeobj.type.to_s)
+          end
         end
       end
     end
```

This did not fix the `circular dependency` error. Perhaps
`permit_concurrent_loads` doesn't handle arbitrarily deep nested autoloads
cross threads?

2) I tried Mutex#synchronize and this worked, but I'd rather we work
with the interlock provided by rails.

```
diff --git a/lib/api/environment.rb b/lib/api/environment.rb
index 87b34f99..f51de73f 100644
--- a/lib/api/environment.rb
+++ b/lib/api/environment.rb
@@ -1,5 +1,6 @@
 module Api
   class Environment
+    ONE_AUTOLOADER_LOCK = Mutex.new
     def self.url_attributes
       @url_attributes ||= Set.new(%w(href))
     end
@@ -19,12 +20,13 @@ module Api
         next if cspec[:klass].blank?
         klass = nil

-        # Temporary measure to avoid thread race condition which could lead to a deadlock
-        ActiveSupport::Dependencies.interlock.permit_concurrent_loads do
+        # Ensure we're the only thread trying to autoload classes and their columns
+        ONE_AUTOLOADER_LOCK.synchronize do
           klass = cspec[:klass].constantize
-        end
-        klass.columns_hash.each do |name, typeobj|
-          result << name if %w(date datetime).include?(typeobj.type.to_s)
+
+          klass.columns_hash.each do |name, typeobj|
+            result << name if %w(date datetime).include?(typeobj.type.to_s)
+          end
         end
       end
     end

```

3) I tried Sync.new with SH and EX locks instead of Mutex and this
failed with Thread killed errors.

4) I changed the `permit_concurrent_loads` on the interlock to `loading`
and this worked, but Yuri found this caused a deadlock.

5) Use `loading` from 4) and move the `columns_hash` call into
the `loading` block.  This solves the circular dependency and avoids the
deadlock encountered in 4).

I noticed, `permit_concurrent_loads` calls `yield_shares` with `compatible: [:load])`
and that method has this in the source code comment:

https://github.com/rails/rails/blob/bb22fe9d4a6102d2a28cb1adfd6fe9d38fc9bb22/activesupport/lib/active_support/concurrency/share_lock.rb#L166-L168

```
  Temporarily give up all held Share locks while executing the
  supplied block, allowing any +compatible+ exclusive lock request
  to proceed.
```

Perhaps, since we're loading code, `permit_concurrent_loads` is too
permissive of other exclusive lock requests and we really need to ensure
nothing else is trying to load.

### DIFF
--- a/lib/api/environment.rb
+++ b/lib/api/environment.rb
@@ -19,12 +19,12 @@ module Api
         next if cspec[:klass].blank?
         klass = nil
 
-        # Temporary measure to avoid thread race condition which could lead to a deadlock
-        ActiveSupport::Dependencies.interlock.permit_concurrent_loads do
+        # Ensure we're the only thread trying to autoload classes and their columns
+        ActiveSupport::Dependencies.interlock.loading do
           klass = cspec[:klass].constantize
-        end
-        klass.columns_hash.each do |name, typeobj|
-          result << name if %w(date datetime).include?(typeobj.type.to_s)
+          klass.columns_hash.each do |name, typeobj|
+            result << name if %w(date datetime).include?(typeobj.type.to_s)
+          end
         end
       end
     end


jrafanie:
Note, this whole problem goes away if we autoload this memoized method once at boot before puma worker threads are spawned.

jrafanie:
@skateman @himdel Updated to use the interlock but to request an exclusive lock while we're autoloading.  Was there a BZ associated with this bug or just #544?

skateman:
@jrafanie no BZ, but we might need one .. the bug can appear in older versions as well.

jrafanie:
@miq-bot rm_label wip
@miq-bot add_label hammer/yes
@miq-bot add_label bug

jrafanie:
Fixes #544

jrafanie:
Hold off on merging, @yrudman found that this seems to cause the first click after login to spin indefinitely, possibly a deadlock...

@miq-bot add_label wip

skateman:
I feel bad that I don't understand the whole rails code reloading process well enough to fix something like this :disappointed:

lib/api/environment.rb
```diff
@@ -19,12 +19,12 @@ def self.time_attributes
         next if cspec[:klass].blank?
         klass = nil
 
-        # Temporary measure to avoid thread race condition which could lead to a deadlock
-        ActiveSupport::Dependencies.interlock.permit_concurrent_loads do
+        # Ensure we're the only thread trying to autoload classes and their columns
+        ActiveSupport::Dependencies.interlock.loading do
```
jrafanie: Note, this also works:

```ruby
        Rails.application.executor.wrap do
...
```

jrafanie:
@miq-bot rm_label wip

Ok, we fixed the issue @yrudman found.  The call to `columns_hash` hits our ar_virtual modification to `load_schema!`  [here](https://github.com/ManageIQ/manageiq/blob/9001c8db2fa7eaa6a0677aa72fea1753347026e5/lib/extensions/ar_virtual.rb#L411), which calls `define_virtual_delegate` which loads the class [here](https://github.com/ManageIQ/manageiq/blob/9001c8db2fa7eaa6a0677aa72fea1753347026e5/lib/extensions/ar_virtual.rb#L125), so for now, we need to protect the `columns_hash` call too.

lib/api/environment.rb
jrafanie: @Fryguy do you prefer the executor wrapper?  I can research more tomorrow if that makes sense over the interlock.loading.

jrafanie:
> Ok, we fixed the issue @yrudman found. The call to `columns_hash` hits our ar_virtual modification to `load_schema!` [here](https://github.com/ManageIQ/manageiq/blob/9001c8db2fa7eaa6a0677aa72fea1753347026e5/lib/extensions/ar_virtual.rb#L411), which calls `define_virtual_delegate` which loads the class [here](https://github.com/ManageIQ/manageiq/blob/9001c8db2fa7eaa6a0677aa72fea1753347026e5/lib/extensions/ar_virtual.rb#L125), so for now, we need to protect the `columns_hash` call too.

I need to circle around to @kbrock and @NickLaMuro regarding the `load_schema!` monkey patch to see if there's another way to avoid the constantizing dynamically in `load_schema!`

lib/api/environment.rb
Fryguy: Yes, the executor wrapper seemed like a better approach.

kbrock:
Not sure if it was documented elsewhere but...

The issue is with virtual attribute's `delegate_attribute`.
A delegate defines an attribute based upon an attribute in another model.
So when you are loading one schema, it calls another model's schema_load.
Thus entering race condition.

It is the `type_for_attribute` [[here]](https://github.com/ManageIQ/manageiq/blob/master/lib/extensions/ar_virtual.rb#L125) that calls `load_schema!` on another class

kbrock:
`load_schema` does not happen during the class load. It happens because we `class_hash`.
The issue is the `class_hash` in one model is causing a `class_hash` in another model.

A work around could be for us to explicitly state the type of the attribute delegated. That way loading one class is not dependent upon other classes.

One interesting note, `has_one` does not load the referencing class.

jrafanie:
Still testing this...

jrafanie:
It turns out `Rails.application.executor.wrap do` still deadlocks.  Oh well, 🤷‍♂️

lib/api/environment.rb
jrafanie: haha, executor.wrap deadlocks if I hit compute -> cloud after login  🤣 ... back to `.loading`

simaishi:
Hammer backport details:

```
$ git log -1
commit b96fac3780976f91b26f286f14cfc308660b0632
Date:   Tue Feb 5 15:33:17 2019 -0500

    Merge pull request #550 from jrafanie/ensure_one_autoloader_thread
    
    One semaphore to exclusively load them all
    
    (cherry picked from commit e75c9e08e7ccfe9dbd5c2b687303b6d981d4fd03)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1673039
```

PR #552, opened 2019-01-30
[WIP] Deprecate category

kbrock:
Part of https://github.com/ManageIQ/manageiq/pull/17210

Removing `Category` from app/models.rb

It is just a child of `Classification` with a `default scope { where(:parent_id => 0) }`
So there is not much lost.

Also, in anticipation of removing `Tag` class (merging with `Classification`), I changed tests to reference `Classification` instead of `Tag` where possible.


kbrock:
wip: deleting category from models and ensuring this still works. will be non-wip soon

kbrock:
Category was created to make API easier - going to let it live on

PR #555, opened 2019-02-01
Remove references to Tag where possible

kbrock:
Test only changes.

Tag is based upon classifications, so just use classifications where possible.
Mostly replaced the `Tag.find()` with `Classification.find_by(:tag_id)`

@miq-bot add_label test

### DIFF
--- a/spec/requests/tags_spec.rb
+++ b/spec/requests/tags_spec.rb
@@ -17,20 +17,20 @@ describe "Tags API" do
 
       get api_tags_url
 
-      expect_query_result(:tags, Tag.count)
+      expect_query_result(:tags, Classification.count)
     end
 
     context "with an appropriate role" do
       it "can create a tag with category by href" do
         api_basic_authorize collection_action_identifier(:tags, :create)
-        category = FactoryBot.create(:category)
+        category = FactoryBot.create(:classification)
         options = {:name => "test_tag", :description => "Test Tag", :category => {:href => api_category_url(nil, category)}}
 
-        expect { post api_tags_url, :params => options }.to change(Tag, :count).by(1)
+        expect { post api_tags_url, :params => options }.to change(Classification, :count).by(1)
 
         result = response.parsed_body["results"].first
-        tag = Tag.find(result["id"])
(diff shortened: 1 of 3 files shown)


kbrock:
These are test only changes, I would still like to decease our reliance on tags.

jrafanie:
@kbrock is this still good to go?  Maybe rebase and ping Alberto when ready?  I can't merge here

kbrock:
re cops: I used the variables that were in the existing code. I do not want to change them

kbrock:
@jrafanie this is good to go. it only changes the spec by using the `Category` specs. We really should only be using the classifications factory since that creates tags in the proper way.

spec/requests/tags_spec.rb
```diff
@@ -210,36 +210,37 @@
...
-        end.not_to change(Tag, :count)
+        end.not_to change(Classification, :count)
 
         expect(response).to have_http_status(:forbidden)
       end
 
       it "cannot update a tag" do
         api_basic_authorize
-        tag = Tag.create(:name => "Old name")
+        classification = Classification.is_category.create(:name => "test_tag", :description => "Test Tag")
```
jrafanie: How is this different from `FactoryBot.create(:classification, :name => "test_tag", :description => "Test Tag")` ?

spec/requests/tags_spec.rb
```diff
@@ -210,36 +210,37 @@
...
+          post api_tag_url(nil, tag), :params => gen_request(:edit, :name => "new_name")
+        end.not_to change { classification.reload.name }
 
         expect(response).to have_http_status(:forbidden)
       end
 
       it "cannot delete a tag through POST" do
         api_basic_authorize
-        tag = Tag.create(:name => "Test tag")
+        tag = Classification.is_category.create(:name => "test_tag", :description => "Test Tag").tag
```
jrafanie: Same question as above for all of these `is_category` calls

jrafanie:
I had only a few comments, looks good other than those

lpichler:
@kbrock is CI failure relevant to PR ?

PR #561, opened 2019-03-04
Fix missing job for ServiceTemplateTransformationPlanTask in ServiceRequestSpec

lpichler:
caused by https://github.com/ManageIQ/manageiq/pull/18415/files
causing CI failure: https://travis-ci.org/ManageIQ/manageiq-api/jobs/501544480

@jameswnl I am not familiar with this but it looks like that there is missing `Job` for ServiceTemplateTransformationPlanTask.

spec were failing [here](https://github.com/ManageIQ/manageiq/pull/18415/files#diff-eb7db57808b5f20ff8f0f07bfd565890R159) and when I added missing job,
the error was about that there is no method cancel on Job instance.
So I just mock the method `cancel`.

I am curious why we are calling method cancel on instance when the method is not defined but can you on this fix or is here better fix for this CI failure ?

thanks @jameswnl !

### DIFF
--- a/spec/requests/service_requests_spec.rb
+++ b/spec/requests/service_requests_spec.rb
@@ -635,6 +635,13 @@ describe "Service Requests API" do
     context "SubResource#cancel" do
       let(:resource_1_response) { {"success" => true, "message" => "RequestTask #{resource_1.id} canceled"} }
       let(:resource_2_response) { {"success" => true, "message" => "RequestTask #{resource_2.id} canceled"} }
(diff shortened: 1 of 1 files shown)


jameswnl:
Hey @lpichler, I wasn't aware that you've created this pull request (I thought it was an issue) and so I also created another PR here https://github.com/ManageIQ/manageiq/pull/18521

Let me know what you think.

lpichler:
closed thanks to the https://github.com/ManageIQ/manageiq/pull/18521

thanks @jameswnl !

PR #566, opened 2019-03-20
Filter - add a new ===, !==, =~, !~ operators to match exact strings or regexes, without mangling

himdel:
Right now, if you have a VM (or anything) with a `%` or `*` in name,
it's impossible to find via the API by that name, without using wildcards.

This adds 4 new operators:

`===` and `!==`, which do the same thing as `=` and `!=` without trying to interpret the value as a regular expression,
`=~` and `!~`, which just parse the value as a regex for MiqExpression, with no extra mangling, allowing for case insensitive matches like `name =~ /foo/i`

The filtering code still mangles whitespace, but wrapping the value in single quotes (strings) or slashes (regex) prevents that.

---

This change also affects the `<=`, `>=`, `<` and `>` operators, which will no longer attempt to parse the value as a regex either.

(Related to https://bugzilla.redhat.com/show_bug.cgi?id=1680954, but not quite there yet.)

### DIFF
--- a/lib/api/filter.rb
+++ b/lib/api/filter.rb
@@ -11,6 +11,10 @@ module Api
       # string-only matching, use quotes
       "===" => {:default => "="},
       "!==" => {:default => "!="},
+
+      # regex-only matching without mangling, use slashes and optionally /i
+      "=~" => {:default => "REGULAR EXPRESSION MATCHES"},
+      "!~" => {:default => "REGULAR EXPRESSION DOES NOT MATCH"},
     }.freeze
 
     attr_reader :filters, :model
--- a/lib/api/filter.rb
+++ b/lib/api/filter.rb
@@ -1,20 +1,20 @@
 module Api
   class Filter
     OPERATORS = {
-      "!=" => {:default => "!=", :regex => "REGULAR EXPRESSION DOES NOT MATCH", :null => "IS NOT NULL"},
-      "<=" => {:default => "<="},
-      ">=" => {:default => ">="},
-      "<"  => {:default => "<", :datetime => "BEFORE"},
-      ">"  => {:default => ">", :datetime => "AFTER"},
-      "="  => {:default => "=", :datetime => "IS", :regex => "REGULAR EXPRESSION MATCHES", :null => "IS NULL"},
+      "!="  => {:default => "!=", :regex => "REGULAR EXPRESSION DOES NOT MATCH", :null => "IS NOT NULL"},
+      "<="  => {:default => "<="},
+      ">="  => {:default => ">="},
+      "<"   => {:default => "<", :datetime => "BEFORE"},
+      ">"   => {:default => ">", :datetime => "AFTER"},
+      "="   => {:default => "=", :datetime => "IS", :regex => "REGULAR EXPRESSION MATCHES", :null => "IS NULL"},
 
       # string-only matching, use quotes
       "===" => {:default => "="},
       "!==" => {:default => "!="},
 
       # regex-only matching without mangling, use slashes and optionally /i
-      "=~" => {:default => "REGULAR EXPRESSION MATCHES"},
-      "!~" => {:default => "REGULAR EXPRESSION DOES NOT MATCH"},
+      "=~"  => {:default => "REGULAR EXPRESSION MATCHES"},
+      "!~"  => {:default => "REGULAR EXPRESSION DOES NOT MATCH"},
     }.freeze
 
     attr_reader :filters, :model
--- a/lib/api/filter.rb
+++ b/lib/api/filter.rb
@@ -9,7 +9,7 @@ module Api
       "="   => {:default => "=", :datetime => "IS", :regex => "REGULAR EXPRESSION MATCHES", :null => "IS NULL"},
 
       # string-only matching, use quotes
-      "===" => {:default => "="},
+      "=="  => {:default => "="},
       "!==" => {:default => "!="},
 
       # regex-only matching without mangling, use slashes and optionally /i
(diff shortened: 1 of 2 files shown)


himdel:
Adding the operators was easier, but perhaps this should be something like `?filter_options=exact_strings` instead of a new set of operators?

lib/api/filter.rb
```diff
@@ -6,7 +6,15 @@ class Filter
       ">=" => {:default => ">="},
       "<"  => {:default => "<", :datetime => "BEFORE"},
       ">"  => {:default => ">", :datetime => "AFTER"},
-      "="  => {:default => "=", :datetime => "IS", :regex => "REGULAR EXPRESSION MATCHES", :null => "IS NULL"}
+      "="  => {:default => "=", :datetime => "IS", :regex => "REGULAR EXPRESSION MATCHES", :null => "IS NULL"},
+
+      # string-only matching, use quotes
+      "==" => {:default => "="},
+      "!==" => {:default => "!="},
```
Fryguy: ugh that's unfortunate that we need `==` and `!==`

Fryguy:
> `?filter_options=exact_strings` and `?filter_options=force_regex`

I kind of like this better, but it might make mixing and matching difficult

Fryguy:
@abellotti Please review.

lib/api/filter.rb
himdel: maybe `===` and `!==` would be better for symmetry?

I was also thinking of leaving the operators as they are and forcing the value by using the right kind of quoting character (requiring quotes for strings and slashes for regexes), but that would break existing code.

himdel:
For case insensitive string search, this works, but looks hacky, and shouldn't be a new operator.
(Case insensitive string search will be a second PR, after we settle on the details here.)

```
diff --git a/lib/api/filter.rb b/lib/api/filter.rb
index c84f7677..bbe4ebff 100644
--- a/lib/api/filter.rb
+++ b/lib/api/filter.rb
@@ -15,6 +15,9 @@ module Api
       # regex-only matching without mangling, use slashes and optionally /i
       "=~" => {:default => "REGULAR EXPRESSION MATCHES"},
       "!~" => {:default => "REGULAR EXPRESSION DOES NOT MATCH"},
+
+      # case insensitive ==
+      'i==' => {:default => "REGULAR EXPRESSION MATCHES"},
     }.freeze
 
     attr_reader :filters, :model
@@ -106,6 +109,10 @@ module Api
         filter_value.gsub!(/%|\\\*/, ".*")
       end
 
+      if operator == 'i=='
+        filter_value = "/\\A#{MiqExpression.re_escape(filter_value)}\\z/i"
+      end
+
       {:logical_or => logical_or, :operator => method, :attr => filter_attr, :value => filter_value}
     end
 
diff --git a/spec/lib/api/filter_spec.rb b/spec/lib/api/filter_spec.rb
index 99610e2d..cb7cb693 100644
--- a/spec/lib/api/filter_spec.rb
+++ b/spec/lib/api/filter_spec.rb
@@ -346,5 +346,14 @@ RSpec.describe Api::Filter do
       expected = {"REGULAR EXPRESSION MATCHES" => {"field" => "MiqReport-name", "value" => "/foo/i"}}
       expect(actual.exp).to eq(expected)
     end
+
+    it "can handle insensitive strings" do
+      filters = ["name i== ' foo.*bar%bazz/?'"]
+
+      actual = described_class.parse(filters, MiqReport)
+
+      expected = {"REGULAR EXPRESSION MATCHES" => {"field" => "MiqReport-name", "value" => "/\\A\\ foo\\.\\*bar%bazz\\/\\?\\z/i"}}
+      expect(actual.exp).to eq(expected)
+    end
   end
 end
```

himdel:
Ping @abellotti

lib/api/filter.rb
abellotti: w.r.t. == vs. ===, not sure, maybe we can do something similar to js w/ type checking.

I guess if we are doing strict type checking requiring value to be quoted, i.e. attr==="value" or attr==='value' and attr column type must be string, the === and !== might be better. 

This way we could introduce == and != in the future which would give us attr=="5" and attr==5  both being true for integer type fields.

himdel:
@abellotti Thanks! :) Should be updated...

I renamed `==` to `===` for consistency, I do see the js type checking parallel there too, agreed :),
and added specs for the negative operators as well.

abellotti:
Thanks @himdel for the updates, looks good.   Minor, but  can you take care of the ^ hash alignment rubocop warnings.

himdel:
Sure :)

abellotti:
LGTM!! Thanks @himdel for the filtering enhancement!! 👍

abellotti:
ping @Fryguy, PR is ok by me. Can you 👀 when you get a chance ?  Thanks.

abellotti:
re-ping @Fryguy, anything else here ?

djberg96:
👍

Fryguy:
Lgtm, feel free to merge.

I guess my only comment would be why `===` over `==`?  My earliest comment regarding this was not about the choice in operator, but the fact that a new operator was even needed at all.

From what I can see, and correct me if I'm wrong, but after this PR we will have `=` and `===`, but `==` is skipped, which is kind of weird.

abellotti:
yea, the new operators are needed because the current =and != have too much smarts built-in (looking for %, enabling regex in such cases, etc prohibiting string comparisons with those characters). 

I can see the weirdness though, If we don't think we'd need to do type-indifferent checking later, I think you're right, we'd be better off with == and !==

himdel:
Changed `===` back to `==`, so now we have:
* `===` and `!==`
* `=~` and `!~`

abellotti:
Thanks @himdel for updating this 👍

himdel:
(The BZ for case-insensitive search - https://bugzilla.redhat.com/show_bug.cgi?id=1680954 - was closed, so not continuing with case insensitive this way, we should probably do proper db indexes for that anyway.)

Issue #354, opened 2018-03-26
[RFE] Add ability to create a Git backed automation domain


fdupont-redhat:
It would be great to be able to create a Git backed automation domain through the API.
An example use case is bootstraping a ManageIQ appliance with domains managed with Git.
This would also be helpful in a CI/CD approach as only the dev environment would have non Git domains.

Associated RHBZ: https://bugzilla.redhat.com/show_bug.cgi?id=1600961

Fryguy:
I thought we already had this from WAY back with automate import.

Fryguy:
I remember this statement from me: https://github.com/ManageIQ/manageiq/pull/9517#issuecomment-232982860  .  Trying to find the follow-up PRs, but I only see the "refresh git based automate domain".

fdupont-redhat:
I do agree with https://github.com/ManageIQ/manageiq/pull/9517#issuecomment-232982860, in that the Git repository should not be exposed on its own. I would expect an API call similar to:

```json
POST /api/automate_domains
{
  "name": "AwesomeDomain",
  "description": "Bob's awesome domain",
  "source": "git",
  "git_repository": {
    "url": "https://github.com/Bob/AwesomeDomain",
    "branch": "production"
  }
}
```

The Git repository object would be created by the backend because the user creates a Git backed domain. Then, in a CI/CD context, there would be developers' domains, with a merge policy. Then, a tool like Jenkins would call the `refresh_from_source` action on the repo when the content of the repo changes.

fdupont-redhat:
@mkanoor what's your opinion on this one ?

fdupont-redhat:
Who is able to do that ? @abellotti @mkanoor

ThomasBuchinger:
+1 This feature would really simplify life for us. 
Are you planning on supporting the other Import formats (zip, yaml) as well? One thing not mentioned until now, which would be crucial for us is the ability to specify the Tenant. 

One thing you guys might want to keep in mind, when implementing that feature is:
There is a potential for producing conflicts in the `__domain__.yaml` if one tries to implement a feature-branch workflow in combination with Tenants (and possibly running the export-scripts on top of it). 
`igortiunov` in the forum explained the problem [here](http://talk.manageiq.org/t/solved-git-flow-for-automate-code/3362/3).

jvlcek:
This issue should be closed as the requested feature has been implemented. Please see:
https://github.com/ManageIQ/manageiq-api/pull/571

[...]

PR #571, opened 2019-04-01
Add create Git backed automation domain

jvlcek:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1600961

This PR, along with PR [manageiq-ui-classic/5416](https://github.com/ManageIQ/manageiq-ui-classic/pull/5416) will provide  the support to allow the API to be used to create a Git backed automation domain.

### Steps for Testing/QA [Optional]

Create a GIT based automate domain repository.

Use the API to issue a POST command to `/api/automate_domains with data similar to:
{
  "git_url": "https://gitlab.exampledomain.com/gitlab_repo/AwesomeDomain",
  "ref_type": "branch",
  "ref_name": "master",
}

e.g.:

> cat action_create_from_git
> {
>   "git_url": "https://gitlab.exampledomain.com/gitlab_repo/AwesomeDomain",
>   "ref_type": "branch",
>   "ref_name": "master",
> }
> curl -k --user admin:SECRET_PASSWORD -i -X POST -H "Accept: application/json" -d  @action_create_from_git  https://my-miq-host/api/automate_domains

### DIFF
--- a/app/controllers/api/automate_domains_controller.rb
+++ b/app/controllers/api/automate_domains_controller.rb
@@ -3,7 +3,7 @@ module Api
 
     REQUIRED_FIELDS = %w(git_url ref_type ref_name)
 
-    def create_resource(type, _id, data)
+    def create_from_git_resource(type, _id, data)
       assert_all_required_fields_exists(data, type, REQUIRED_FIELDS)
       raise BadRequestError, 'ref_type must be "branch" or "tag"' unless valid_ref_type?(data)
       
@@ -17,7 +17,7 @@ module Api
                                                                            data["ref_name"],
                                                                            data["ref_type"],
                                                                            User.current_user.current_tenant.id,
-                                                                           optional_data(data))
+                                                                           prepare_optional_auth(data))
 
         action_result(true, 'Creating Automate Domain from git repository', :task_id => task_id)
       rescue => err
@@ -104,15 +104,15 @@ module Api
       false
     end
 
-    def optional_data(data)
-        optional_data = {}
-        optional_data["userid"] = data["userid"] if data.has_key?("userid")
-        optional_data["password"] = data["password"] if data.has_key?("password")
+    def prepare_optional_auth(data)
+        optional_auth = {}
+        optional_auth["userid"] = data["userid"] if data.has_key?("userid")
+        optional_auth["password"] = data["password"] if data.has_key?("password")
 
         # If data["verify_ssl"] is missing or set to false use VERIFY_NONE. If true use VERIFY_PEER
-        optional_data["verify_ssl"] = data["verify_ssl"] ? OpenSSL::SSL::VERIFY_PEER : OpenSSL::SSL::VERIFY_NONE
+        optional_auth["verify_ssl"] = data["verify_ssl"] ? OpenSSL::SSL::VERIFY_PEER : OpenSSL::SSL::VERIFY_NONE
 
-        optional_data
+        optional_auth
     end
   end
 end
--- a/app/controllers/api/automate_domains_controller.rb
+++ b/app/controllers/api/automate_domains_controller.rb
@@ -1,5 +1,30 @@
 module Api
   class AutomateDomainsController < BaseController
+
+    REQUIRED_FIELDS = %w(git_url ref_type ref_name)
+
+    def create_resource(type, _id, data)
+      assert_all_required_fields_exists(data, type, REQUIRED_FIELDS)
+      raise BadRequestError, 'ref_type must be "branch" or "tag"' unless valid_ref_type?(data)
+      
+      api_log_info("Create will be queued for automate domain from #{data["git_url"]} / #{data["ref_name"]}")
+
+      begin
+        unless GitBasedDomainImportService.available?
+          raise "Git owner role must enabled to import git repositories"
+        end
+        task_id = GitBasedDomainImportService.new.queue_refresh_and_import(data["git_url"],
+                                                                           data["ref_name"],
+                                                                           data["ref_type"],
+                                                                           User.current_user.current_tenant.id,
+                                                                           optional_data(data))
+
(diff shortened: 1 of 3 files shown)


jvlcek:
@miq-bot add_label bug
@miq-bot add_label hammer/no

jvlcek:
@mkanoor Please review

app/controllers/api/automate_domains_controller.rb
```diff
@@ -1,5 +1,30 @@
 module Api
   class AutomateDomainsController < BaseController
+
+    REQUIRED_FIELDS = %w(git_url ref_type ref_name)
+
+    def create_resource(type, _id, data)
```
Fryguy: My only concern here is if there are other ways to creates domains.  That is, I understand that the UI doesn't use the REST API now, but if it did in the future, then we would need a more pure "create" method, and it would be "taken" by this method.  If this PR doesn't preclude something like that then, I think it's fine.

Here is my writeup from ages ago from how I thought this should be designed.  https://github.com/ManageIQ/manageiq/pull/9517#issuecomment-232982860 .  I *think* the way we have it here is ok, but it might be better to have an alternate action like "import_from_git" or something, just so we don't clobber the actual "create" action.

Fryguy: I read the associated issue to this: https://github.com/ManageIQ/manageiq-api/issues/354, and there is also a request for importing yaml or zip through the API as well, which also goes to my question about whether this should be "create" or something else.

jvlcek: @Fryguy Thank you for the help. I believe I've addressed the suggestions/issues you raised.

jvlcek:
@Fryguy, @abellotti and @mkanoor Please review.

app/controllers/api/automate_domains_controller.rb
```diff
@@ -1,5 +1,29 @@
 module Api
   class AutomateDomainsController < BaseController
+    REQUIRED_FIELDS = %w[git_url ref_type ref_name].freeze
+
+    def create_from_git_resource(type, _id, data)
+      assert_all_required_fields_exists(data, type, REQUIRED_FIELDS)
+      raise BadRequestError, 'ref_type must be "branch" or "tag"' unless valid_ref_type?(data)
+
+      api_log_info("Create will be queued for automate domain from #{data["git_url"]} / #{data["ref_name"]}")
```
abellotti: can the git_url include credentials ?

app/controllers/api/automate_domains_controller.rb
```diff
@@ -1,5 +1,29 @@
...
+
+      begin
+        unless GitBasedDomainImportService.available?
+          raise "Git owner role must enabled to import git repositories"
```
abellotti: raise BadRequestError, "Git owner ...."

app/controllers/api/automate_domains_controller.rb
```diff
@@ -72,5 +96,22 @@ def resource_search(id, type, klass)
     def current_tenant
       User.current_user.current_tenant || Tenant.default_tenant
     end
+
+    def valid_ref_type?(data)
+      return false unless data.key?("ref_type")
+      return true if data["ref_type"] == "tag" || data["ref_type"] == "branch"
+      false
```
abellotti: you can replace 101-103 with:
   %w(tag branch).include?(data["ref_type"]) if data.key?("ref_type")

app/controllers/api/automate_domains_controller.rb
```diff
@@ -1,5 +1,29 @@
...
+        unless GitBasedDomainImportService.available?
+          raise "Git owner role must enabled to import git repositories"
+        end
+        task_id = GitBasedDomainImportService.new.queue_refresh_and_import(data["git_url"],
+                                                                           data["ref_name"],
+                                                                           data["ref_type"],
+                                                                           User.current_user.current_tenant.id,
+                                                                           prepare_optional_auth(data))
+
+        action_result(true, 'Creating Automate Domain from git repository', :task_id => task_id)
```
abellotti: can we add some identifying text here ?  Creating Automate Domain XYZZY from git repository,
maybe using ref_name.

jvlcek: @abellotti Great idea! Thank you. I think the `ref_name` alone might not be very helpful since it could simply be the branch name  `master`.   I think both the `git_url\ref_name`.
Please let me know if you are OK with this.

app/controllers/api/automate_domains_controller.rb
jvlcek: @abellotti Will do. Thank you.

app/controllers/api/automate_domains_controller.rb
jvlcek: @abellotti Got it.

app/controllers/api/automate_domains_controller.rb
jvlcek: @abellotti  I'm sorry but I'm not clear as to how that could be a good approach but am willing to discuss any benefits you may envision.

jvlcek:
Note: This PR relies on [manageiq-ui-classic/5416](https://github.com/ManageIQ/manageiq-ui-classic/pull/5416)

app/controllers/api/automate_domains_controller.rb
abellotti: If that's not a supported ref syntax, we're ok then.  If they specify it as such in the inbound request, it will be logged all over since it's not a password type attr, so it's a broader issue than this.

abellotti:
LGTM!! Thanks @jvlcek for the API enhancement.  @Fryguy and @mkanoor can you lend some 👀 here. Thanks.

fdupont-redhat:
Fixes https://github.com/ManageIQ/manageiq-api/issues/354

PR #584, opened 2019-05-06
[WIP] add endpoints subcollections

lpichler:
`POST http://localhost:3000/api/providers `

```
{"name":"Test azure1","type":"ManageIQ::Providers::Azure::CloudManager","credentials":{"userid":"foo","password":"bar"},"endpoints":[{"url":"www.endpoints.com"}]}
```

cc @Hyperkid123

### DIFF
--- a/app/controllers/api/endpoints_controller.rb
+++ b/app/controllers/api/endpoints_controller.rb
@@ -1,0 +1,4 @@
+module Api
+  class EndpointsController < BaseController
+  end
+end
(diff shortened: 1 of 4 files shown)


Hyperkid123:
@lpichler 

Ok so i found more sub-collections and another relations that we will need to create/edit providers via API

- connection_configurations
  - this is required for openstack and wmvare cloud providers
  - its similar to endpoints, in fact it connections are listed inside endpoints
  - according to API docs you should be able to create entity with this sub-collection but you cannot expand it
  - this should include these connections for cloud providers:
    - endpoints
    - amqp
    - ssh key pair
- authentication_userid
  - is required for editing amazon/azure connection
- keystone_v3_domain_id
  - required for adding/editing OpenStack cloud provider

lpichler:
closing as discussed with @Hyperkid123 there is work in  progress about it.

PR #587, opened 2019-05-13
Remove conditional and assume plugin_name exists

Fryguy:
Depends on https://github.com/ManageIQ/manageiq/pull/18755 which ensures that all plugins will have a plugin_name, simplifying the question here.


Fryguy:
@abellotti Please review.

PR #589, opened 2019-05-13
Change plugin common methods to class methods

Fryguy:
Change plugin common methods to class methods


himdel:
Guessing you'll want to restart the specs now that https://github.com/ManageIQ/manageiq-api/pull/591 is merged :).

PR #591, opened 2019-05-14
jbuilder 2.9.0 regressed serialization, exclude it

jrafanie:
2.9.1 fixes it. Other older versions are still ok.

https://github.com/rails/jbuilder/issues/461


djberg96:
:+1:

PR #594, opened 2019-05-20
Added PxeServer create, update and delete actions

Hyperkid123:
We've got a BZ for add/edit PxeServer in ui-classic which will require us to rewrite the UI. Part of that is removing current code from ui-classic and let the API handle these actions instead.

https://bugzilla.redhat.com/show_bug.cgi?id=1537267

```javascript
POST
/api/pxe_servers {
  uri: string,
  name: string,
  pxe_directory?: string,
  windows_images_directory?: string,
  customization_directory?: string
  pxe_menus?: [{
    file_name: string
  }],
  authentication?: { //only if uri has not nfs prefix
    userid: string,
    passworkd: string
  }
}
```

### TODO
- ~file depot credentials validation id model: https://github.com/ManageIQ/manageiq/pull/18795~

### DIFF
--- a/app/controllers/api/pxe_servers_controller.rb
+++ b/app/controllers/api/pxe_servers_controller.rb
@@ -1,21 +1,40 @@
 module Api
   class PxeServersController < BaseController
-    INVALID_PXE_SERVER_ATTRS = %w[id href].freeze # Cannot update or create these
-
     include Subcollections::PxeImages
     include Subcollections::PxeMenus
+    INVALID_ATTRIBUTES = {
+      "PxeServer" => %w[id href], # Cannot update or create these
+    }.freeze
+    PXE_ATTRIBUTES = {
+      "PxeServer"      => %w[name uri],
+      "PxeMenu"        => %w[file_name],
+      "Authentication" => %w[userid password]
+    }.freeze
 
     def create_resource(_type, _id, data = {})
-      validate_pxe_server_create_data(data)
+      authentication = data.delete('authentication')
       menus = data.delete('pxe_menus')
+      validate_data_for('PxeServer', data)
 
-      server = collection_class(:pxe_servers).create(data)
-      if server.invalid?
-        raise BadRequestError, "Failed to add a pxe server - #{server.errors.full_messages.join(', ')}"
+      PxeServer.transaction do
+        server = collection_class(:pxe_servers).new(data)
+        # generates uir_prefix which checks if server needs authentication or not
+        server.verify_uri_prefix_before_save
+
+        if server.requires_credentials? && server.missing_credentials?
+          validate_data_for('Authentication', authentication || {})
+          server.update_authentication({:default => authentication.compact}, {:save => true})
+        end
+
+        server.pxe_menus = create_pxe_menus(menus) if menus
+
+        if server.invalid?
+          raise BadRequestError, "Failed to add a pxe server - #{server.errors.full_messages.join(', ')}"
+        end
+
+        server.save
+        server
       end
-
-      server.pxe_menus = create_pxe_menus(menus) if menus
-      server
     end
 
     def delete_resource(_type, id = nil, data = nil)
@@ -27,34 +46,33 @@ module Api
     def edit_resource(type, id, data)
       server = resource_search(id, type, collection_class(:pxe_servers))
       menus = data.delete('pxe_menus')
-      if menus
-        server.pxe_menus.clear
-        data['pxe_menus'] = create_pxe_menus(menus)
+      authentication = data.delete('authentication')
+      PxeServer.transaction do
+        if menus
+          server.pxe_menus.clear
+          data['pxe_menus'] = create_pxe_menus(menus)
+        end
+        server.update!(data)
+        server.update_authentication({:default => authentication.transform_keys(&:to_sym)}, {:save => true}) if authentication && server.requires_credentials?
+        server
       end
-
-      server.update!(data)
-      server
     end
 
     private
 
     def create_pxe_menus(menus)
+      menus.each do |menu|
+        validate_data_for('PxeMenu', menu)
+      end
       menus.map do |menu|
         collection_class(:pxe_menus).create(menu)
       end
     end
 
-    def validate_pxe_server_data(data)
-      bad_attrs = data.keys.select { |k| INVALID_PXE_SERVER_ATTRS.include?(k) }.compact.join(", ")
-      raise BadRequestError, "Invalid attribute(s) #{bad_attrs} specified for a pxe server" if bad_attrs.present?
-    end
-
-    def validate_pxe_server_create_data(data)
-      validate_pxe_server_data(data)
-      req_attrs = %w[name uri]
+    def validate_data_for(klass, data)
       bad_attrs = []
-      req_attrs.each { |attr| bad_attrs << attr if data[attr].blank? }
-      raise BadRequestError, "Missing attribute(s) #{bad_attrs.join(', ')} for creating a pxe server" if bad_attrs.present?
+      PXE_ATTRIBUTES[klass].each { |attr| bad_attrs << attr if data[attr].blank? }
+      raise BadRequestError, "Missing attribute(s) #{bad_attrs.join(', ')} for creating a #{klass}" if bad_attrs.present?
     end
   end
 end
--- a/app/controllers/api/pxe_servers_controller.rb
+++ b/app/controllers/api/pxe_servers_controller.rb
@@ -1,6 +1,6 @@
 module Api
   class PxeServersController < BaseController
-    INVALID_PXE_SERVER_ATTRS = %w(id href).freeze # Cannot update or create these
+    INVALID_PXE_SERVER_ATTRS = %w[id href].freeze # Cannot update or create these
 
     include Subcollections::PxeImages
     include Subcollections::PxeMenus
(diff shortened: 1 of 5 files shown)


Hyperkid123:
cc @lpichler

Hyperkid123:
@miq-bot assign @lpichler

lpichler:
you need to call `verify_uri_prefix_before_save`.

martinpovolny:
@Hyperkid123 : please, link the BZ for visibility

martinpovolny:
Ping @lpichler

app/controllers/api/pxe_servers_controller.rb
```diff
@@ -1,5 +1,91 @@
...
+      creds[:default][:userid] = authentication['userid'] if authentication['userid']
+      creds[:default][:password] = authentication['password'] if authentication['password']
+      creds
+    end
+
+    def create_pxe_menus(menus)
+      menus.each do |menu|
+        validate_pxe_create_data(menu, REQUIRED_PXE_MENU_ATTRS, 'pxe menu')
+      end
+      menus.map do |menu|
```
lpichler: this should work also

```ruby
collection_class(:pxe_menus).create(menu)
```

lpichler:
1. Validation methods: I see 2 `validate_X` methods  -what about to do 1 method which would take care of all:
similar MISSING and INVALID as we did in https://github.com/ManageIQ/manageiq-api/pull/596

```ruby
def validate_data_for(klass, data)

  klass_attributes = PXE_ATTRIBUTES[klass]


   MISSING_ATTRIBUTES = ...

   INVALID_ATTRIBUTES = ....
end
```

app/controllers/api/pxe_servers_controller.rb
lpichler: in this case, what about to put line ^ on places where `create_pxe_menus`  is called and put validation just before it ?

```ruby
if menus
  server.pxe_menus.clear
  menus.each{|menu| validate_data_for(collection_class(:pxe_menus), menu) }
  data['pxe_menus'] = collection_class(:pxe_menus).create(menu)
end
```

app/controllers/api/pxe_servers_controller.rb
```diff
@@ -1,5 +1,91 @@
...
+
+    def delete_resource(_type, id = nil, data = nil)
+      raise BadRequestError, "Must specify an id for deleting a pxe server" unless id
+
+      super
+    end
+
+    def edit_resource(type, id, data)
+      PxeServer.transaction do
+        server = resource_search(id, type, collection_class(:pxe_servers))
```
lpichler: are other validations missing here ?

app/controllers/api/pxe_servers_controller.rb
```diff
@@ -1,5 +1,91 @@
...
+        # generates uir_prefix which is ched if server needs authentication or not
+        server.verify_uri_prefix_before_save
+
+        # assign authentication if needed
+        if server.requires_credentials? && server.missing_credentials?
+          validate_pxe_create_data(authentication || {}, REQUIRED_PXE_AUTH_ATTRS, 'pxe server authentication')
+          server.update_authentication(prepare_credentials(authentication), :save => true)
+        end
+
+        # assign pxe menus if needed
```
lpichler: I think that comment like this is not needed

app/controllers/api/pxe_servers_controller.rb
```diff
@@ -1,5 +1,91 @@
 module Api
   class PxeServersController < BaseController
+    INVALID_PXE_SERVER_ATTRS = %w[id href].freeze # Cannot update or create these
+    REQUIRED_PXE_SERVER_ATTRS = %w[name uri].freeze
```
lpichler: as I wrote bellow:
you can pust this to the hash:

```ruby
PXE_ATTRIBUTES = {
   PxeServer => %w[name uri],
  ...
}.freeze

```

app/controllers/api/pxe_servers_controller.rb
```diff
@@ -1,5 +1,91 @@
...
+    def create_resource(_type, _id, data = {})
+      PxeServer.transaction do
+        authentication = data.delete('authentication')
+        menus = data.delete('pxe_menus')
+
+        validate_pxe_server_data(data)
+        validate_pxe_create_data(data, REQUIRED_PXE_SERVER_ATTRS, 'pxe server')
+
+        server = collection_class(:pxe_servers).new(data)
+        # generates uir_prefix which is ched if server needs authentication or not
```
lpichler: this is good comment 👍 (it is not obvious from `verify_uri_prefix_before_save` why we need it)

app/controllers/api/pxe_servers_controller.rb
```diff
@@ -1,5 +1,91 @@
...
+    REQUIRED_PXE_SERVER_ATTRS = %w[name uri].freeze
+    REQUIRED_PXE_MENU_ATTRS = %w[file_name].freeze
+    REQUIRED_PXE_AUTH_ATTRS = %w[userid password].freeze
+
     include Subcollections::PxeImages
+    include Subcollections::PxeMenus
+
+    def create_resource(_type, _id, data = {})
+      PxeServer.transaction do
+        authentication = data.delete('authentication')
```
lpichler: is it possible to put code outside of `PxeServer.transaction` block which is not related to the transaction ? (same for edit)

app/controllers/api/pxe_servers_controller.rb
```diff
@@ -1,5 +1,91 @@
...
+        server.update_authentication(prepare_credentials(authentication), :save => true) if authentication && server.requires_credentials?
+        server.update!(data)
+        server
+      end
+    end
+
+    private
+
+    def prepare_credentials(authentication)
+      creds = {:default => {}}
```
lpichler: can be this just?
```
{:default => authentication.compact}
```

lpichler: not sure if we need method `prepare_credentials` in this case

lpichler:
overall looks good 👍 and I added some comments - we can discuss it 👍 👾

Hyperkid123:
> overall looks good  and I added some comments - we can discuss it

I've addressed most of your feedback but i cannot use the classes as keys in hashes because they are not auto loaded correctly. I used string constants instead.

martinpovolny:
> @lpichler I've addressed most of your feedback but i cannot use the classes as keys in hashes because they are not auto loaded correctly. I used string constants instead.

Can that be done in a separate PR?

@Hyperkid123 : can you, please, add some API call examples to the Description?

lpichler:
@djberg96 can you review also please ? thanks!

martinpovolny:
ping @djberg96 , @lpichler can we move forward with this one, please?

spec/requests/pxe_servers_spec.rb
```diff
@@ -1,7 +1,13 @@
 RSpec.describe 'PxeServers API' do
-  let!(:pxe_server) { FactoryBot.create(:pxe_server) }
+  let(:pxe_server) do
```
lpichler: @Hyperkid123 you can do this:
```ruby
  let(:pxe_server) { FactoryBot.build(:pxe_server).tap { |x| x.update_authentication(:default => {:userid => 'foo', :password => 'bar'}) } }
```

lpichler:
@Hyperkid123 we need to improve authorisation tied with product features.

- we need to add for each action which is new also test case for non authorised user.
as I understand correctly it is about those actions: `pxe_server_view `, `pxe_server_create `, `pxe_server_view`, subcollections: `pxe_server_view, pxe_server_create`

and eventually you need to add it related blocks in api.yml to the `XXXX_actions:` as I did it here for edit action:

```ruby
diff --git a/config/api.yml b/config/api.yml
index 6d6e9cc3..97c12862 100644
--- a/config/api.yml
+++ b/config/api.yml
@@ -2445,6 +2445,9 @@
       :get:
       - :name: read
         :identifier: pxe_server_view
+      :patch:
+      - :name: edit
+        :identifier: pxe_server_edit
   :quotas:
     :description: TenantQuotas
     :options:
diff --git a/spec/requests/pxe_servers_spec.rb b/spec/requests/pxe_servers_spec.rb
index 5166427a..c39b9c6e 100644
--- a/spec/requests/pxe_servers_spec.rb
+++ b/spec/requests/pxe_servers_spec.rb
@@ -1,10 +1,6 @@
 RSpec.describe 'PxeServers API' do
-  let(:pxe_server) do
-    pxe_server = FactoryBot.build(:pxe_server)
-    pxe_server.update_authentication(:default => {:userid => 'foo', :password => 'bar'})
-    pxe_server.save
-    pxe_server
-  end
+  let(:pxe_server) { FactoryBot.build(:pxe_server).tap { |x| x.update_authentication(:default => {:userid => 'foo', :password => 'bar'}) } }
+
   let!(:pxe_image_1) { FactoryBot.create(:pxe_image, :pxe_server => pxe_server) }
   let!(:pxe_image_2) { FactoryBot.create(:pxe_image, :pxe_server => pxe_server) }
   let!(:pxe_menu_1) { FactoryBot.create(:pxe_menu, :pxe_server => pxe_server) }
@@ -187,6 +183,14 @@ RSpec.describe 'PxeServers API' do
       post(url, :params => {:name => 'test server', :uri => 'smb://quax', :authentication => {:userid => 'user'}})
     end
+
+    it 'will not ... ' do
+      api_basic_authorize
+      post(url, :params => {:name => 'foo', :uri => 'bar/quax',
+                            :pxe_menus      => [{:file_name => 'menu_1'}, {:file_name => 'menu_2'}],
+                            :authentication => {:userid => 'foo', :password => 'bar' }})
+    end
   end
 
   describe 'patch /api/pxe_servers/:id' do
@@ -200,6 +204,12 @@ RSpec.describe 'PxeServers API' do
       expect(response.parsed_body['uri']).to eq('updated/url')
       expect(PxeServer.find(response.parsed_body['id']).pxe_menus.first[:file_name]).to eq('updated menu')
     end
+
+    it 'will not .. ' do
+      api_basic_authorize
+      patch(url, :params => {:name => 'updated name', :uri => 'updated/url', :pxe_menus => [{:file_name => 'updated menu'}]})
+    end
   end
 
   describe 'delete /api/pxe_servers/:id' do
@@ -211,5 +221,11 @@ RSpec.describe 'PxeServers API' do
       expect(response).to have_http_status(:no_content)
       expect(PxeServer.exists?(pxe_server.id)).to be_falsey
     end
+
+    it 'will not show ...' do
+      api_basic_authorize
+      delete(url)
+    end
   end
 end

```

PR #596, opened 2019-05-24
Added create orchestration template service dialog action.

Hyperkid123:
Action for creating service dialog from orchestration template. Part of an effort of refactoring orchestration dialog forms. Based on current implementation in ui classic which this should eventually replace.
https://github.com/ManageIQ/manageiq-ui-classic/blob/master/app/controllers/catalog_controller.rb#L1108

```
POST /api/service_dialogs
{
  action: 'orchestration_template_service_dialog',
  resource: {
    label: 'dialog label'
     ot_id: '569' // orchestration template ID
  }
}
```

### DIFF
--- a/app/controllers/api/service_dialogs_controller.rb
+++ b/app/controllers/api/service_dialogs_controller.rb
@@ -3,6 +3,7 @@ module Api
     before_action :set_additional_attributes, :only => [:index, :show]
 
     CONTENT_PARAMS = %w[target_type target_id resource_action_id].freeze
+    ORCHESTRATION_TEMPLATE_DIALOG_ATTRS = %w[ot_id label].freeze
 
     def refresh_dialog_fields_resource(type, id = nil, data = nil)
       raise BadRequestError, "Must specify an id for Reconfiguring a #{type} resource" unless id
@@ -46,6 +47,14 @@ module Api
       service_dialog.deep_copy(attributes).tap(&:save!)
     rescue => err
       raise BadRequestError, "Failed to copy service dialog - #{err}"
+    end
+
+    def orchestration_template_service_dialog_resource(_type, _id, data)
+      validate_orchestration_template_dialog_create_data(data)
+      orchestration_template = OrchestrationTemplate.find(data['ot_id'])
+      raise BadRequestError, "Failed to create service dialog from orchestration template. Orchestration template with id: #{data['ot_id']} does not exist." unless orchestration_template
+
+      Dialog::OrchestrationTemplateServiceDialog.create_dialog(data['label'], orchestration_template)
     end
 
     private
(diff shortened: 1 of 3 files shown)


Hyperkid123:
@miq-bot assign @lpichler 
@miq-bot add-label hammer/no

martinpovolny:
Ping @lpichler

app/controllers/api/service_dialogs_controller.rb
```diff
@@ -98,5 +105,11 @@ def define_service_dialog(dialog_fields, data, options = {})
     def service_dialog_ident(service_dialog)
       "Service Dialog id:#{service_dialog.id} label:'#{service_dialog.label}'"
     end
+
+    def validate_orchestration_template_dialog_create_data(data)
+      bad_attrs = []
+      ORCHESTRATION_TEMPLATE_DIALOG_REQUIRED_ATTRS.each { |attr| bad_attrs << attr if data[attr].blank? }
```
lpichler: maybe rename bad_attrs to missing_attributes ? and then do:
```ruby
missing_attributes  = ORCHESTRATION_TEMPLATE_DIALOG_REQUIRED_ATTRS - data.keys
```
what do you think ?

lpichler: and think you need to check that in data are just `ORCHESTRATION_TEMPLATE_DIALOG_REQUIRED_ATTRS ` keys, right ?

so you can do in addition 
```
invalid_attributes  =  data.keys - ORCHESTRATION_TEMPLATE_DIALOG_REQUIRED_ATTRS
raise BadRequestError if invalid_attributes.present?
```

app/controllers/api/service_dialogs_controller.rb
```diff
@@ -3,6 +3,7 @@ class ServiceDialogsController < BaseController
     before_action :set_additional_attributes, :only => [:index, :show]
 
     CONTENT_PARAMS = %w[target_type target_id resource_action_id].freeze
+    ORCHESTRATION_TEMPLATE_DIALOG_REQUIRED_ATTRS = %w[ot_id label].freeze
```
lpichler: probably enough is `ORCHESTRATION_TEMPLATE_DIALOG_ATTRIBUTES` (without REQUIRED)

app/controllers/api/service_dialogs_controller.rb
```diff
@@ -48,6 +49,12 @@ def copy_resource(type, id, data)
       raise BadRequestError, "Failed to copy service dialog - #{err}"
     end
 
+    def orchestration_template_service_dialog_resource(_type, _id, data)
+      validate_orchestration_template_dialog_create_data(data)
+      ot = OrchestrationTemplate.find(data['ot_id'])
```
lpichler: what  if `ot` is nil (not exist in DB)
I think we need to cover it by some error message.

lpichler: `ot` it would be nice to use `orchestration_template` variable.

lpichler:
thanks @Hyperkid123 👍

PR #598, opened 2019-05-28
Add a subcollection under VMs for displaying disks

skateman:
This is required for the API-driven reconfigure VM screen where the user has to display the related disks. Any change is going through the `VmReconfigureRequest` model that is already supported by the API, therefore, there's no need for anything else than `GET`.

@miq-bot add_label hammer/no, changelog/yes

### DIFF
--- a/config/api.yml
+++ b/config/api.yml
@@ -1144,6 +1144,12 @@
         :identifier: storage_tag
       - :name: unassign
         :identifier: storage_tag
+  :disks:
+    :description: Disks
+    :options:
+    - :subcollection
+    :verbs: *g
+    :klass: Disk
   :enterprises:
     :description: Enterprises
     :options:
@@ -3741,6 +3747,7 @@
     :verbs: *gpd
     :klass: Vm
     :subcollections:
+    - :disks
     - :tags
     - :policies
     - :policy_profiles
@@ -3892,6 +3899,16 @@
       :delete:
       - :name: delete
         :identifier: vm_delete
+    :disks_subcollection_actions:
+      :get:
+      - :name: read
+        :identifier:
+        - vm_show
+    :disks_subresource_actions:
+      :get:
+      - :name: read
+        :identifier:
+        - vm_show
     :tags_subcollection_actions:
       :post:
       - :name: assign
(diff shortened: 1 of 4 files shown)


spec/requests/disks_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+
+        expected = {
+          "href"              => api_vm_disk_url(nil, vm, disk),
+          "id"                => disk.id.to_s,
+        }
+        expect(response.parsed_body).to include(expected)
+        expect(response).to have_http_status(:ok)
+      end
+
+      it "will not show a disk unless authorized" do
```
skateman: @lpichler @abellotti I'm not sure how RBAC works here, maybe I'm missing something but this test is failing on it. Could you please help me out?

config/api.yml
```diff
@@ -3892,6 +3899,11 @@
       :delete:
       - :name: delete
         :identifier: vm_delete
+    :disks_subcollection_actions:
```
lpichler: @skateman you define this as `disks_subresource_actions` because as I understand you are not fetching collection of disk - then also specs will pass.

thanks

spec/requests/disks_spec.rb
lpichler: @skateman you need to define also `disks_subresource_actions` because as I understand you are not fetching collection of disk - then also specs will pass.

thanks


```
diff --git a/config/api.yml b/config/api.yml
index 0269ba6b..f650396a 100644
--- a/config/api.yml
+++ b/config/api.yml
@@ -3908,6 +3908,11 @@
       - :name: read
         :identifier:
         - vm_show
+    :disks_subresource_actions:
+      :get:
+        - :name: read
+          :identifier:
+            - vm_show
     :tags_subcollection_actions:
       :post:
       - :name: assign
```

spec/requests/disks_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+
+        expect(response.parsed_body).to include(expected)
+        expect(response).to have_http_status(:ok)
+      end
+    end
+
+    describe "GET /api/vms/:c_id/disks/:s_id" do
+      it "can show a VM's disk" do
+        api_basic_authorize(subcollection_action_identifier(:vms, :disks, :read, :get))
+        hw = FactoryBot.create(:hardware)
```
lpichler: Isn't it better to put `hw, disk, vm` into `lets' to top and use for all `it` blocks ?

spec/requests/disks_spec.rb
skateman: Thanks, except the indentation in your code is wrong, but it seems working after I fixed it.

lpichler:
Merging, but if you can add API request, it will be cool 👍  (for sprint review, ... )

PR #501, opened 2018-10-25
Add support for sui product features

jvlcek:
https://bugzilla.redhat.com/show_bug.cgi?id=1637552

PR [16212](https://github.com/ManageIQ/manageiq/pull/16212) added support for
Service UI permissions.

This PR adds support for the API to honor the SUI permissions.

Chris Hale and I tested on a live appliance.

PR #599, opened 2019-05-28
Fix API retirement identifiers and role checking

d-m-u:
https://bugzilla.redhat.com/show_bug.cgi?id=1699943

GM and Tina and I discussed the fact that this call probably shouldn't have an approval check on it because the intent was to match the request system for provisioning.  This also fixes a bug where the identifiers in the api.yml file were wrong for retirement -- they should both be using retire_now, and the identifier to set a retirement date (instead it should just be service retirement in general) shouldn't show up in the post section for service retirement. 

edit: thanks AB and GM for the help

### DIFF
--- a/app/controllers/api/base_controller/generic.rb
+++ b/app/controllers/api/base_controller/generic.rb
@@ -62,18 +62,14 @@ module Api
       end
 
       def request_retire_resource(type, id, _data = nil)
-        if api_user_role_allows?('miq_request_approval')
-          klass = collection_class(type)
-          if id
-            msg = "#{User.current_user.userid} retiring #{type} id #{id} immediately as a request."
-            resource = resource_search(id, type, klass)
-            api_log_info(msg)
-            klass.make_retire_request(resource.id, User.current_user)
-          else
-            raise BadRequestError, "Must specify an id for retiring a #{type} resource"
-          end
+        klass = collection_class(type)
+        if id
+          msg = "#{User.current_user.userid} retiring #{type} id #{id} immediately as a request."
+          resource = resource_search(id, type, klass)
+          api_log_info(msg)
+          klass.make_retire_request(resource.id, User.current_user)
         else
-          raise ForbiddenError, "User lacking correct permissions to approve a #{type} retire as a request."
+          raise BadRequestError, "Must specify an id for retiring a #{type} resource"
         end
       end
 
--- a/config/api.yml
+++ b/config/api.yml
@@ -3234,12 +3234,12 @@
         - sui_services_edit
       - :name: retire
         :identifier:
-        - service_retire
+        - service_retire_now
         - sui_services_retire
       - :name: request_retire
         :identifier:
-        - service_retire
-        - sui_services_retire_date
+        - service_retire_now
+        - sui_services_retire
       - :name: set_ownership
         :identifier:
         - service_ownership
--- a/spec/requests/services_spec.rb
+++ b/spec/requests/services_spec.rb
@@ -484,8 +484,10 @@ describe "Services API" do
 
           expect(response).to have_http_status(:forbidden)
         end
+      end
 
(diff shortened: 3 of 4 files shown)


app/controllers/api/base_controller/generic.rb
```diff
@@ -62,18 +62,14 @@ def delete_resource(type, id = nil, _data = nil)
       end
 
       def request_retire_resource(type, id, _data = nil)
-        if api_user_role_allows?('miq_request_approval')
```
gmcculloug: In talking to @gtanzillo about this issue it seems like this should actually be checking for the product feature that allows the user to retire the object.  There are different product features that would need to be checked based on the `type` value and the SUI now has its own set of product features which would need to be considered as well.  (See https://github.com/ManageIQ/manageiq-api/pull/501)

But my first question is if this check actually belongs here in the first place or should it be done elsewhere.  I do not see an other calls to `api_user_role_allows?` in this file and wonder if the role check is happening before we get here.

cc @abellotti

lpichler: I am just thinking (retiring is not my area):

and other question is whether method `request_retire_resource` is used just for services or also for other objects(Request, Orchestration Stack,,,,  ), maybe we should check "X_retire" permission for each type of object (X is. Service, Request, Orchestration Stack ) and check how retirement is done in UI classic (if it is present in UI and in case there is no API used yet)

abellotti: api_user_role_allows? is called in base_controller/parser.rb as we validate the action for the caller and decide on returning the 403's or not.

gmcculloug: The logic @abellotti is referring to is the call to `api_user_role_allows?` here https://github.com/ManageIQ/manageiq-api/blob/master/app/controllers/api/base_controller/parser.rb#L155.

So I believe we are doing the correct thing by removing this check.

I do however think we have a bug that was introduced with https://github.com/ManageIQ/manageiq-api/pull/501 where the service `retire` and `request_retire` product feature lookups are checking the wrong values.

Based on Product Features for
My Services:
https://github.com/ManageIQ/manageiq/blob/master/db/fixtures/miq_product_features.yml#L438-L441

and Service UI:
https://github.com/ManageIQ/manageiq/blob/master/db/fixtures/miq_product_features.yml#L6475-L6478

I would expect the following configuration in `config/api.yml`
```diff
--- a/config/api.yml
+++ b/config/api.yml
@@ -3234,12 +3234,12 @@
         - sui_services_edit
       - :name: retire
         :identifier:
-        - service_retire
+        - service_retire_now
         - sui_services_retire
       - :name: request_retire
         :identifier:
-        - service_retire
-        - sui_services_retire_date
+        - service_retire_now
+        - sui_services_retire
       - :name: set_ownership
         :identifier:
         - service_ownership
```

@gtanzillo @tinaafitz Does this look like the proper configuration change?

tinaafitz: @gmcculloug Yes, that looks like the proper configuration change.

gtanzillo: Looks correct to me @gmcculloug

d-m-u:
@miq-bot add_label hammer/yes

simaishi:
Hammer backport details:

```
$ git log -1
commit 258310204c7541f1a64c6c298fbe63b2401480f6
Author: Gregg Tanzillo <gtanzillo@gmail.com>
Date:   Fri May 31 15:59:25 2019 -0400

    Merge pull request #599 from d-m-u/removing_approval_check_from_retire_request_call
    
    Fix API retirement identifiers and role checking
    
    (cherry picked from commit 1e1279045843c90448f5ac205beaa4b591eed494)
    
    Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1763860
```

PR #600, opened 2019-05-28
Adjust ConversionHost support check so that it uses resource instead of singleton

djberg96:
This PR updates the ConversionHostController#create method so that the supports check happens at the instance level instead of the singleton level. I also updated the comments slightly.

Effectively replaces https://github.com/ManageIQ/manageiq/pull/18799, since fiddling with the resource_type won't matter any more.

https://bugzilla.redhat.com/show_bug.cgi?id=1713394

### DIFF
--- a/app/controllers/api/conversion_hosts_controller.rb
+++ b/app/controllers/api/conversion_hosts_controller.rb
@@ -42,9 +42,7 @@ module Api
 
       resource = resource_search(data['resource_id'], resource_type.to_s, collection_class(collection_type))
 
-      unless resource.respond_to?(:supports_conversion_host?) && resource.supports_conversion_host?
-        raise BadRequestError, "unsupported resource_type #{resource_type}"
-      end
+      raise BadRequestError, "unsupported resource_type #{resource_type}" unless resource.supports_conversion_host?
 
       data['resource'] = resource
 
--- a/app/controllers/api/conversion_hosts_controller.rb
+++ b/app/controllers/api/conversion_hosts_controller.rb
@@ -37,13 +37,13 @@ module Api
 
       raise BadRequestError, "invalid resource_type #{data['resource_type']}" unless resource_type
 
-      unless resource_type.respond_to?(:supports_conversion_host?) && resource_type.supports_conversion_host?
-        raise BadRequestError, "unsupported resource_type #{resource_type}"
-      end
-
       collection_type = resource_type.table_name
 
       resource = resource_search(data['resource_id'], resource_type.to_s, collection_class(collection_type))
+
+      unless resource.respond_to?(:supports_conversion_host?) && resource.supports_conversion_host?
+        raise BadRequestError, "unsupported resource_type #{resource_type}"
+      end
 
       data['resource'] = resource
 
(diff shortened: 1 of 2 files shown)


app/controllers/api/conversion_hosts_controller.rb
```diff
@@ -37,14 +38,14 @@ def create_resource(type, id, data)
...
 
-      unless resource_type.respond_to?(:supports_conversion_host?) && resource_type.supports_conversion_host?
-        raise BadRequestError, "unsupported resource_type #{resource_type}"
-      end
-
       collection_type = resource_type.table_name
 
       resource = resource_search(data['resource_id'], resource_type.to_s, collection_class(collection_type))
 
+      unless resource.respond_to?(:supports_conversion_host?) && resource.supports_conversion_host?
```
agrare: The `.respond_to?` check shouldn't be needed as long as the model includes `SupportsFeatureMixin` since every possible feature is defined as unsupported by default.

app/controllers/api/conversion_hosts_controller.rb
```diff
@@ -14,6 +14,7 @@ class ConversionHostsController < BaseController
     #   * vmware_vddk_package_url
     #   * vmware_ssh_private_key
     #   * conversion_host_ssh_private_key
+    #   * openstack_tls_ca_certs
```
agrare: Is this related or was it just missing before?

djberg96: It's not related, just thought I would add it while I was here. :)

app/controllers/api/conversion_hosts_controller.rb
djberg96: Ok, updated.

djberg96:
@miq-bot add_labels transformation, hammer/yes

spec/requests/conversion_hosts_spec.rb
```diff
@@ -53,9 +53,11 @@
   context "create" do
     let(:zone) { FactoryBot.create(:zone) }
     let(:ems_openstack) { FactoryBot.create(:ems_openstack, :zone => zone) }
-    let(:ems_redhat) { FactoryBot.create(:ems_redhat, :zone => zone) }
+    let(:ems_redhat) { FactoryBot.create(:ems_redhat, :zone => zone, :api_version => '4.2.4') }
```
bdunne: There's an `:ems_redhat_v4` factory, is `4.2.4` important?

djberg96: @bdunne We have a minimum api version for redhat conversion hosts of 4.2.4, and the factory is 4.0.

djberg96: I imagine the factory could probably be updated, but didn't want to touch it for now since it could affect tests outside of v2v.

bdunne: > @bdunne We have a minimum api version for redhat conversion hosts of 4.2.4, and the factory is 4.0.

Okay.

> I imagine the factory could probably be updated, but didn't want to touch it for now since it could affect tests outside of v2v.

No, I think that makes it even less obvious that this api version is important here.

simaishi:
Hammer backport details:

```
$ git log -1
commit dc1165c75763e22e508877062c29166c1bc3b6d5
Date:   Wed May 29 11:53:35 2019 -0400

    Merge pull request #600 from djberg96/conversion_host_supports_feature
    
    Adjust ConversionHost support check so that it uses resource instead of singleton
    
    (cherry picked from commit a8dc1913f64950fb99b2994ceb21459bc149c566)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1717023
```

PR #603, opened 2019-06-03
Support filtering for multiple IP

jvlcek:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1684681

This PR adds filtering support for virtual attribute array element matching, which is required
for the associated RFE to work.

The associated RFE requests API filtering for VMs that are assigned multiple IP addresses.
The ipaddresses virtual element is an array or ipaddresses so the "includes all" miq expression
operator is what should have been used to filter VM ipaddresses.

This PR changes the operator for ipaddresses filtering from "equals" to  "includes all".

To test:
-------

Configure MiQ with a provider that has VMs with multiple IP Addresses.

Save the below ruby code example to a file, e.g. get_vms_w_filter_ipaddr.rb Note: change the ipaddress 192.0.2.0 in the example ruby code to be a valid IP address for one of your VMs
that has multiple IP addresses.

```ruby
#!/usr/bin/env ruby

require "net/http"
require "uri"
require "json"

uri = URI.parse("https://#{ENV['MIQ']}/api/vms/?expand=resources&attributes=ipaddresses,name&filter[]=ipaddresses='192.0.2.0'")

http = Net::HTTP.new(uri.host, uri.port)
http.use_ssl = true
http.verify_mode = OpenSSL::SSL::VERIFY_NONE

request = Net::HTTP::Get.new(uri.request_uri)
request.basic_auth("admin", "smartvm")

response = http.request(request)

puts "Reply:\n " + JSON.pretty_generate(JSON.parse(response.body.strip))
```

Then run the test:

```bash
export MIQ="<the ipaddress of your MiQ appliance>"
./get_vms_w_filter_ipaddr.rb
```

### DIFF
--- a/lib/api/filter.rb
+++ b/lib/api/filter.rb
@@ -81,10 +81,9 @@ module Api
       str_method = is_regex ? methods[:regex] : methods[:default]
 
       filter_value, method = case filter_value
-                             when /^'.*'$/
-                               [filter_value.gsub(/^'|'$/, ''), str_method]
-                             when /^".*"$/
-                               [filter_value.gsub(/^"|"$/, ''), str_method]
+                             when /^['"](.*)['"]$/
+                               unquoted_filter_value = $1
+                               [unquoted_filter_value, str_method]
                              when /^(NULL|nil)$/i
                                [nil, methods[:null] || methods[:default]]
                              else
--- a/lib/api/filter.rb
+++ b/lib/api/filter.rb
@@ -6,7 +6,7 @@ module Api
       ">="  => {:default => ">="},
       "<"   => {:default => "<", :datetime => "BEFORE"},
       ">"   => {:default => ">", :datetime => "AFTER"},
-      "="   => {:default => "=", :datetime => "IS", :regex => "REGULAR EXPRESSION MATCHES", :null => "IS NULL"},
+      "="   => {:default => "=", :datetime => "IS", :regex => "REGULAR EXPRESSION MATCHES", :string_set => "includes all", :null => "IS NULL"},
 
       # string-only matching, use quotes
       "=="  => {:default => "="},
@@ -41,6 +41,7 @@ module Api
         unless virtual_or_physical_attribute?(target_class(model, associations), attr)
           raise BadRequestError, "attribute #{attr} does not exist"
         end
+
         associations.map! { |assoc| ".#{assoc}" }
         field = "#{model.name}#{associations.join}-#{attr}"
         target = parsed_filter[:logical_or] ? or_expressions : and_expressions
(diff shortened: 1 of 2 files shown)


jvlcek:
@miq-bot add_label wip

jvlcek:
@miq-bot remove_label wip      
@miq-bot add_label bug         
@miq-bot add_label hammer/no
@miq-bot assign @abellotti

jvlcek:
Thank you to @kbrock for pairing up with me on this. Please review/comment on this PR.

lib/api/filter.rb
```diff
@@ -81,10 +82,17 @@ def parse_filter(filter)
...
       filter_value, method = case filter_value
-                             when /^'.*'$/
-                               [filter_value.gsub(/^'|'$/, ''), str_method]
-                             when /^".*"$/
-                               [filter_value.gsub(/^"|"$/, ''), str_method]
+                             when /^['"](.*)['"]$/
+                               unquoted_filter_value = $1
+                               if column_type(model, filter_attr) == :string_set
+                                 unless methods[:string_set]
+                                   raise BadRequestError, "Unsupported operator for string_set: #{operator}"
```
abellotti: are we breaking current users of the (broken) support with this raise ?

I wonder if this block should be if column_type(model, filter_attr) == :string_set && methods[:string_set]
without the raise.

I could be totally wrong.  thoughts ?

jvlcek: Thank you @kbrock and @abellotti 

@abellotti Your suggestion would be safer for sure. I will do that.

lib/api/filter.rb
```diff
@@ -81,10 +82,13 @@ def parse_filter(filter)
       str_method = is_regex ? methods[:regex] : methods[:default]
 
       filter_value, method = case filter_value
-                             when /^'.*'$/
-                               [filter_value.gsub(/^'|'$/, ''), str_method]
-                             when /^".*"$/
-                               [filter_value.gsub(/^"|"$/, ''), str_method]
+                             when /^['"](.*)['"]$/
```
abellotti: Hi Joe, thanks for the previous update.

This above was bugging me and I didn't know why.  Problem with ['"] ... ['"]   will match '\<stuff\>" which we don't want, can you update the when statement to:

```
when /^'(.*)'$/,  /^"(.*)"$/
```

Thank you.

jvlcek: @abellotti Great catch. Will do. Thank you!

abellotti:
This LGTM!!.  Thanks @jvlcek for fixing this and @kbrock for pairing/working with Joe on this 👍

kbrock:
I just stumbled upon a BZ related to filtering by ip address.
It looks like if we use `includes any` it may be faster.

Just wanted to put a link into here in case a problem comes up:
https://bugzilla.redhat.com/show_bug.cgi?id=1588082

PR #610, opened 2019-06-14
FIX CI: Parent_id should equals to nil - it is default value in request_specs

lpichler:
caused by removing this https://github.com/ManageIQ/manageiq/pull/18301/files#diff-5bb77b717064efbd95e6ab04eb5a357fL5 (in https://github.com/ManageIQ/manageiq/pull/18301)

so I update condition to match nil.

cc @kbrock @d-m-u 
@miq-bot assign @jrafanie

### DIFF
--- a/spec/requests/requests_spec.rb
+++ b/spec/requests/requests_spec.rb
@@ -258,7 +258,7 @@ RSpec.describe "Requests API" do
 
       FactoryBot.create(:classification_department_with_tags)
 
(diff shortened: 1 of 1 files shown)


kbrock:
after note:

In the future, lets use `is_category` for stuff like this:

```ruby
Classification.is_category.where(:description => 'Department').includes(:tag).first
```

I'm surprised there were any `:parent_id => ***` left anywhere in our code

PR #611, opened 2019-06-14
Improved request_spec by removing direct filtering by Classification#parent_id

yrudman:
get rid of  direct filtering by 'Classification#parent_id' in rspec

Improvement suggested by @kbrock to https://github.com/ManageIQ/manageiq-api/pull/610

@miq-bot add-label test

### DIFF
--- a/spec/requests/requests_spec.rb
+++ b/spec/requests/requests_spec.rb
@@ -258,7 +258,7 @@ RSpec.describe "Requests API" do
 
       FactoryBot.create(:classification_department_with_tags)
 
-      t = Classification.where(:description => 'Department', :parent_id => nil).includes(:tag).first
+      t = Classification.is_category.includes(:tag).find_by(:description => 'Department')
       request.add_tag(t.name, t.children.first.name)
 
       api_basic_authorize action_identifier(:requests, :read, :resource_actions, :get)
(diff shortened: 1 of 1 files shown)


yrudman:
@miq-bot add-label bug

spec/requests/requests_spec.rb
```diff
@@ -258,7 +258,7 @@
 
       FactoryBot.create(:classification_department_with_tags)
 
-      t = Classification.where(:description => 'Department', :parent_id => 0).includes(:tag).first
+      t = Classification.where(:description => 'Department', :parent_id => nil).includes(:tag).first
```
gtanzillo: This looks a bit strange because the category should have `:parent_id => 0` in Classifications. Is it possible that the factory can be fixed to ensure that the category is created that way instead of this change?

kbrock:
```suggestion
      t = Classification.is_category.includes(:tag).find_by(:description => 'Department')
```

spec/requests/requests_spec.rb
```diff
@@ -289,7 +289,7 @@
...
```
kbrock:
```suggestion
      t = Classification.is_category.includes(:tag).find_by(:description => 'Department')
```

lpichler: or `Classification.is_category` -> `Category` (it is a default scope here)

kbrock: yes. things get a little tricky with Category since a different object will be returned and specs compare and get a different value.

spec/requests/requests_spec.rb
yrudman: right, it was first thought. Second thought was - I am trying to hide possible issue

yrudman:
changed PR description and title, thank you @kbrock and @lpichler

PR #612, opened 2019-06-17
Avoid hardcoding paths to assets since it changes

jrafanie:
It's not important here what the path is to the assets.  They can change
if you're precompiling them or not.

Extracted from #606


lpichler:
@jrafanie can you fix rubocop please ?

lpichler:
@jrafanie eventually please change label if technical debt in inappropriate, thanks

Issue #595, opened 2019-05-23
[V2V] Private key information should be filtered out in log


djberg96:
When submitting private key information the private key information is being logged into the evm.log or api.log (or both). This needs to be filtered out.

At the moment this is not easy to remove because the filters are largely a series of hard coded keys set in `Vmdb::Settings::PASSWORD_FIELDS`. I think we would have to temporarily redefine `Api.encrypted_attribute` to make it work.

Fryguy:
Ideally this should be made pluggable so a plugin can bring its own set of filtered keys.

djberg96:
Moved over here: https://github.com/ManageIQ/manageiq-v2v/pull/996

PR #614, opened 2019-06-25
Add a filter_parameters setting and use it when logging api requests

djberg96:
At the moment there's no way to specify custom parameters to filter out of the api.log. It's currently fixed at whatever's specified via `Rails.application.config.filter_parameters`. Anything else gets logged in plaintext to the api.log.

This PR adjusts the `Logger#log_api_request` method so that it will also use whatever's specified in the settings.yml file's `filter_parameters` entry. For now I've added 3 entries that we would like for V2V, but others can be added as needed later.

This partially addresses https://github.com/ManageIQ/manageiq-api/issues/595 in that it will filter the api.log data, but doesn't affect the core production (evm) log. IMO that should be a separate PR because it will need to be refactored so that it's easily pluggable by all providers.

### DIFF
--- a/app/controllers/api/base_controller/logger.rb
+++ b/app/controllers/api/base_controller/logger.rb
@@ -29,7 +29,10 @@ module Api
       end
 
       def log_api_request
-        @parameter_filter ||= ActionDispatch::Http::ParameterFilter.new(Rails.application.config.filter_parameters)
+        @parameter_filter ||= ActionDispatch::Http::ParameterFilter.new(
+          Rails.application.config.filter_parameters + Settings.api.filter_parameters
+        )
+
         return unless api_log_info?
         log_request("Request", @req.to_hash)
         unfiltered_params = request.query_parameters
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -2,3 +2,7 @@
   :token_ttl: 10.minutes
   :authentication_timeout: 30.seconds
   :max_results_per_page: 1000
+  :filter_parameters:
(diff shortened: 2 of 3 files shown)


Fryguy:
I just added pluggable api, so perhaps we can use that?  I don't really like adding *more* settings into settings for things that are not user facing (settings are user facing)

Fryguy:
Actually why would we use settings at all.  I think this can be achieved with existing engine configuration, an if not, a simple class method on the engine, and then have api just enumerate all of the engines via Vmdb::Plugins.

djberg96:
> I just added pluggable api, so perhaps we can use that?

Where and how?

djberg96:
> Actually why would we use settings at all. I think this can be achieved with existing engine configuration, an if not, a simple class method on the engine, and then have api just enumerate all of the engines via Vmdb::Plugins.

A class method on the engine is still a hard coded list of values, which is what I was trying to avoid. Not sure what you mean by an engine configuration, might need some guidance there.

Fryguy:
> A class method on the engine is still a hard coded list of values, which is what I was trying to avoid.

I don't understand why.

> Where and how?

I believe in your engine.rb [here](https://github.com/ManageIQ/manageiq-v2v/blob/master/lib/manageiq/v2v/engine.rb) you can just do something like

```ruby
initializer do
  app.config.filter_parameters += %i[conversion_host_ssh_private_key openstack_tls_ca_certs vmware_ssh_private_key]
end
```

djberg96:
Cool, I'll close this.

PR #624, opened 2019-07-10
Fix the broken spec for auto approver.

lfu:
Broken after https://github.com/ManageIQ/manageiq/pull/18626.

@miq-bot add_label test


simaishi:
@kbrock @bdunne Please review. The miq_feature failure is unrelated and https://github.com/ManageIQ/manageiq-api/pull/622 will take care.

PR #631, opened 2019-07-24
keys should be symbolized before merge

AlexanderZagaynov:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1718209

### DIFF
--- a/app/controllers/api/hosts_controller.rb
+++ b/app/controllers/api/hosts_controller.rb
@@ -15,8 +15,9 @@ module Api
       resource_search(id, type, collection_class(:hosts)).tap do |host|
         all_credentials = Array.wrap(credentials).each_with_object({}) do |creds, hash|
           auth_type = creds.delete(AUTH_TYPE_ATTR) || DEFAULT_AUTH_TYPE
+          creds.symbolize_keys!
           creds.reverse_merge!(:userid => host.authentication_userid(auth_type))
-          hash[auth_type.to_sym] = creds.symbolize_keys!
+          hash[auth_type.to_sym] = creds
         end
         host.update_authentication(all_credentials) if all_credentials.present?
       end
--- a/spec/requests/hosts_spec.rb
+++ b/spec/requests/hosts_spec.rb
@@ -23,6 +23,17 @@ RSpec.describe "hosts API" do
         expect(response).to have_http_status(:ok)
       end
 
+      it "prevents duplicate string/symbol keys mess" do
+        host = FactoryBot.create(:host)
+        api_basic_authorize action_identifier(:hosts, :edit)
+        options = { :credentials => { 'userid' => "I'm", 'password' => 'abc123' } }
+
+        expect do
+          post api_host_url(nil, host), :params => gen_request(:edit, options)
+        end.to change { host.reload.authentication_password(:default) }.to('abc123')
+        expect(response).to have_http_status(:ok)
+      end
+
       it "sending non-credentials attributes will result in a bad request error" do
         host = FactoryBot.create(:host_with_authentication)
         api_basic_authorize action_identifier(:hosts, :edit)
--- a/spec/requests/hosts_spec.rb
+++ b/spec/requests/hosts_spec.rb
@@ -23,7 +23,7 @@ RSpec.describe "hosts API" do
         expect(response).to have_http_status(:ok)
       end
 
-      it "prevents duplicate string/symbol keys mess" do
+      it "can edit the password on a host without creating duplicate keys" do
         host = FactoryBot.create(:host)
         api_basic_authorize action_identifier(:hosts, :edit)
         options = { :credentials => { 'userid' => "I'm", 'password' => 'abc123' } }


juliancheal:
@miq-bot asign @abellotti

juliancheal:
👍 LGTM

juliancheal:
@miq-bot assign @abellotti

AlexanderZagaynov:
@miq-bot add_label ivanchuk/yes

lpichler:
@AlexanderZagaynov can you add some spec for this fix ? 
thanks!

lpichler:
@AlexanderZagaynov and where is the issue ?

this works:
```
{'aa' => 'ddd'}.reverse_merge(:u => 'aa').symbolize_keys!
```

AlexanderZagaynov:
@lpichler what do you mean "this works"? Try with `hsh = { 'a' => 'b' }.reverse_merge(:a => 'c')`, and then `hsh.symbolize_keys!`

lpichler:
```
 hsh = { 'a' => 'b' }.reverse_merge(:a => 'c')
hsh.symbolize_keys! 

```
returns 
```
{
    :a => "b"
}
```

I just don't understand what is the issue because from code:

in `creds`:

```
{
    "password" => "abc123"
}
```

there you do:
```
 creds.reverse_merge!(:userid => "whatever")
```
you have in `creds`

```
creds
{
    "password" => "abc123",
       :userid => "testuser"
}
```

and then finally  `creds.symbolize_keys!` returns:
 
```
{
    :password => "abc123",
      :userid => "testuser"
}
```

what looks same result as with you change.

so what is the issue or what I missed ?

AlexanderZagaynov:
@lpichler issue is that we have `{ 'userid' => 'something' }` in the hash, and then we trying to merge there `:userid`. See https://bugzilla.redhat.com/show_bug.cgi?id=1718209#c7

juliancheal:
@lpichler The problem was, the data from the api request was `"userid" => "foo"`, but in this line https://github.com/ManageIQ/manageiq-api/blob/master/app/controllers/api/hosts_controller.rb#L18 
`creds.reverse_merge!(:userid => host.authentication_userid(auth_type))`

if there was no `userid` already it would return `nil` ie. `:userid => nil`

which would then result in `"userid" => "foo", :userid => nil`

Once they were symbolised we would only get `:userid => nil` in the hash.

Therefore when we got to `update_authentications` method it would delete the new creds we were updating as the `userid` was `nil` so it would think it was a delete.

lpichler:
ah I see thanks @juliancheal for explanation!
@AlexanderZagaynov now I understand 

to be honest it is pretty hidden what is going on in code 

I would to like prefer something like this (it is pseudocode):

```
unless host.authentication_userid(auth_type)
    use userid from host.authentication_userid(auth_type)
else if there userid in creds
   use userid from params
else
... 
end
```

what do you think about it ?
can you also cover it by specs ?

thanks!

AlexanderZagaynov:
@lpichler I believe the simplest changes are better. This PR are quite obvious, and if we'll start inventions and refactoring, we're at risk to never finish that.
P.S. I would like to refactor it totally, to be honest :)

app/controllers/api/hosts_controller.rb
```diff
@@ -15,8 +15,9 @@ def edit_resource(type, id, data = {})
       resource_search(id, type, collection_class(:hosts)).tap do |host|
         all_credentials = Array.wrap(credentials).each_with_object({}) do |creds, hash|
           auth_type = creds.delete(AUTH_TYPE_ATTR) || DEFAULT_AUTH_TYPE
+          creds.symbolize_keys!
           creds.reverse_merge!(:userid => host.authentication_userid(auth_type))
-          hash[auth_type.to_sym] = creds.symbolize_keys!
+          hash[auth_type.to_sym] = creds
```
skateman: I'd rather prefer not using mutability here and drop the exclamation marks:
```suggestion
          hash[auth_type.to_sym] = creds.symbolize_keys.reverse_merge(:userid => host.authentication_userid(auth_type))
```

AlexanderZagaynov: I just following existing pattern, trying to KISS

lpichler:
> @lpichler I believe the simplest changes are better. This PR is quite obvious, and if we'll start inventions and refactoring, we're at risk to never finish that.
> P.S. I would like to refactor it totally, to be honest :)

it is hard to read from code what has bigger priority if params are comming request - who remembers behaviour of `reverse_merge` with same keys ?

but ok if you are going to refactor it - so just add some specs and I can merge it ;)

thanks

jvlcek:
@juliancheal Thank you for explaining the issue above.
@AlexanderZagaynov +1 to keeping this change simple. 
@lpichler +1 to add some specs.

AlexanderZagaynov:
@lpichler 
> who remembers behaviour of `reverse_merge` with same keys?

keys are not the same, that's the problem. they was string and symbol - it's a different objects. `reverse_merge` has no any relation to that :)

I added the spec. Here is the result without my change:
```ruby
Failures:

  1) hosts API editing a host's password with an appropriate role prevents duplicate string/symbol keys mess
     Failure/Error:
       expect do
         post api_host_url(nil, host), :params => gen_request(:edit, options)
       end.to change { host.reload.authentication_password(:default) }.to("abc123")
     
       expected `host.reload.authentication_password(:default)` to have changed to "abc123", but did not change
     # ./spec/requests/hosts_spec.rb:32:in `block (4 levels) in <top (required)>'

Finished in 3.94 seconds (files took 7.11 seconds to load)
1 example, 1 failure

Failed examples:

rspec ./spec/requests/hosts_spec.rb:26 # hosts API editing a host's password with an appropriate role prevents duplicate string/symbol keys mess
```
With this change it is green.

spec/requests/hosts_spec.rb
```diff
@@ -23,6 +23,17 @@
         expect(response).to have_http_status(:ok)
       end
 
+      it "prevents duplicate string/symbol keys mess" do
```
jvlcek: @AlexanderZagaynov Perfect. I just wrote up the same spec and tested locally! So the spec LGTM
However I'd rather the title was a bit more descriptive. Perhaps:

>  it "can edit the password on a host when keys are specified as strings" do

jvlcek: Or

it "can edit the password on a host when a new key is specified as strings" do

AlexanderZagaynov: @jvlcek I guess keys will always be a strings, because they came from request params. BTW, I tried to find what change caused this issue, but it seems pretty hard. Perhaps it is some Rails update.

AlexanderZagaynov: @jvlcek let me know if you still think to change description. I just can't say that your example is descriptive and precise, maybe we should write something else? :)

AlexanderZagaynov: @jvlcek missed your updates while was writing my response. Thing is that it is not about how keys are specified. It is about preventing situations like `{ 'userid' => 'a', :userid => 'b' }`. That's what I call "mess". Maybe this word confusing you, what about `it "prevents duplicate string/symbol keys mix" do`?

lpichler:
> keys are not the same, that's the problem. they was string and symbol - it's a different objects. `reverse_merge` has no any relation to that :)

with you change keys are also same objects.

```ruby
creds.symbolize!
=> {:userid=>'11'}
```
```ruby
creds.reverse_merge!(:userid =>'222')
=>  {:userid=>'11'} or  {:userid=>'222'} ?  it is hard to know.
```

I still think that using `.reverse_merge!` does not support readability of code when you can have same keys.

AlexanderZagaynov:
@lpichler I guess you still didn't understand the issue reason ((

spec/requests/hosts_spec.rb
AlexanderZagaynov: done

AlexanderZagaynov:
@lpichler maybe if you'll run my spec without that change, you'll get what I mean :)

spec/requests/hosts_spec.rb
lpichler: if we will stop using `.reverse_merge!` you can write nicely understandable  something like `it "updates userid  and password from params"`  ... 

 `it "updates only password from params..."`  ... 

`it "prevents duplicate string/symbol keys mess"` it looks like that it is related to some implementation detail or playing with keys in hash and there should be more real test case.

AlexanderZagaynov: @lpichler that `reverse_merge` was there before. And it is not the reason of the issue. Take a look at the `HashWithIndifferentAccess`, maybe it will help you to understand what is going on there. What about `reverse_merge` method complexity you found here - I guess that it is not normal to not understand the very basic features of the Rails framework.

jvlcek:
LGTM! Thank you @AlexanderZagaynov 
Please merge when green.

AlexanderZagaynov:
For everybody who don't understand what `reverse_merge` does: think of it as `given_options_hash.apply_defaults(...)`. If value is given - it will stay on place. If there is no such key - it will be taken from defaults.

AlexanderZagaynov:
@miq-bot add_label ivanchuk/yes

simaishi:
Ivanchuk backport details:

```
$ git log -1
commit 528cf9fedd907578bf5bc146830a0e54bc2c78d4
Date:   Thu Jul 25 17:39:52 2019 +0200

    Merge pull request #631 from AlexanderZagaynov/BZ-1718209_update_host_credentials
    
    keys should be symbolized before merge
    
    (cherry picked from commit 2f85ef4e39f14555d556bfa7e3001f8fd8f18a76)
    
    Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1718209
```

PR #645, opened 2019-08-08
Expose DDF schema for a provider when asking using OPTIONS + type

skateman:
Extracted the `providers_options` into a separate method and only returning when the `type` isn't set for the `/api/providers`. If the `type` param is set in a request, the API will try to append the data-driven-form schema if specified under the `data.provider_form_schema` key. This schema can be used by the UI to render the right form that contains all the required info for adding a provider of the given type. 

```json
// OPTIONS api/providers?type=ManageIQ::Providers::Amazon::CloudManager
{
  ...
  "data": {
    ...
    "provider_form_schema": {
      "title": "Configure AWS",
      "fields": [
        {
          "component": "text-field",
          "name": "role",
          "type": "hidden",
          "initialValue": "ec2"
        },
        {
          "component": "text-field",
          "name": "region",
          "label": "Region"
        },
        {
          "component": "text-field",
          "name": "ec2_username",
          "label": "Access Key"
        },
        {
          "component": "text-field",
          "name": "ec2_password",
          "label": "Secret Key",
          "type": "password"
        }
      ]
    }
  }
}
```

If the `type` argument cannot be safely constantized into a subclass of `ExtManagementSystem` or the provider has no DDF specified it will raise an `BadRequestError`. The request without the `type` parameter behaves like the OPTIONS request before this change.

Example of the DDF schema in the amazon provider: https://github.com/ManageIQ/manageiq-providers-amazon/pull/551

@miq-bot add_label enhancement, ivanchuk/no, hammer/no

cc @Hyperkid123, @martinpovolny, @agrare

### DIFF
--- a/app/controllers/api/providers_controller.rb
+++ b/app/controllers/api/providers_controller.rb
@@ -84,24 +84,9 @@ module Api
     end
 
     def options
-      providers_options = ManageIQ::Providers::BaseManager.leaf_subclasses.inject({}) do |po, ems|
-        po.merge(ems.ems_type => ems.options_description)
-      end
-
-      supported_providers = ExtManagementSystem.supported_types_for_create.map do |klass|
-        if klass.supports_regions?
-          regions = klass.parent::Regions.all.sort_by { |r| r[:description] }.map { |r| r.slice(:name, :description) }
-        end
-
-        {
-          :title   => klass.description,
-          :type    => klass.to_s,
-          :kind    => klass.to_s.demodulize.sub(/Manager$/, '').underscore,
-          :regions => regions
-        }.compact
-      end
-
-      render_options(:providers, "provider_settings" => providers_options, "supported_providers" => supported_providers)
+      options = providers_options
+      options['provider_form_schema'] = provider_options(params[:type]) if params[:type]
+      render_options(:providers, options)
     end
 
     def pause_resource(type, id, _data)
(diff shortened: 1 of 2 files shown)


Hyperkid123:
@skateman is there way to get `auth_type` for these fields? If so it would be great if we cound scope the fields. Something like `name: authentication.default.userid`. This would basically copy the request payload format necessary for API. It would also make it easier for providers with multiple auth types.

skateman:
@Hyperkid123 I'm just exposing the stuff @Fryguy is defining in https://github.com/ManageIQ/manageiq-providers-amazon/pull/551, if I'd do what you want here, we would end up with the same if-else-case :spaghetti: we have in ems_common.

app/controllers/api/providers_controller.rb
```diff
@@ -131,6 +114,36 @@ def change_password_resource(type, id, data = {})
...
+
+        {
+          :title   => klass.description,
+          :type    => klass.to_s,
+          :kind    => klass.to_s.demodulize.sub(/Manager$/, '').underscore,
+          :regions => regions
+        }.compact
+      end
+
+      { "provider_settings" => providers_options, "supported_providers" => supported_providers }
```
agrare: If someone asks for a specific provider type could we return all of this in both cases for consistency rather than this or [just the params_for_create](https://github.com/ManageIQ/manageiq-api/pull/645/files#diff-14a4ff3ca25095b0ebbc061c5aebf20dR123) depending on if they passed a specific provider type or not?

agrare: If you ask for a single type you should get a subset of what you would get if you asked for all of them

Fryguy:
https://github.com/ManageIQ/manageiq/pull/19120 has been merged.

skateman:
@Fryguy we don't depend on the validation backend here, this PR could go in right now without any backend modifications...but thanks, I'll take a look at it.

app/controllers/api/providers_controller.rb
skateman: I'll see what I can do :wink:

skateman: Updated!

PR #644, opened 2019-08-08
Destroy pxe_menu images upon removal

miha-plesko:
We were invoking `.delete_all` method (aliased as `.clear`) which avoids `:dependent => destroy` callbacks hence related PXE images remained in database.

Fixes https://github.com/ManageIQ/manageiq/issues/18940

@miq-bot add_label bug,ivanchuk/yes
@miq-bot assign @bdunne

PR #648, opened 2019-08-09
Protect user's work

miha-plesko:
Currently, when user updates PXE Server details and submits form, all related PxeMenus and PxeImages are discarded so she needs to run "Recreate Relationships" to inventory them again.

This approach works but unfortunately user has option to manually edit PxeImages when they are inventoried, and by discarding them we also discard her work.

With this commit we therefore go out of our way and double check which PxeMenus (and subsequent PxeImages) are *really* affected by update and only discard those.

This way, if user only updates PxeServer's name, nothing else will be recreated hence user's manual work will be preserved.

Depends on https://github.com/ManageIQ/manageiq/pull/19126
Depends on https://github.com/ManageIQ/manageiq/pull/19134
Fixes https://github.com/ManageIQ/manageiq/issues/18940

@miq-bot add_label bug,ivanchuk/yes
@miq-bot assign @bdunne 

/cc @tadeboro @matejart @gberginc

### DIFF
--- a/app/controllers/api/pxe_servers_controller.rb
+++ b/app/controllers/api/pxe_servers_controller.rb
@@ -26,7 +26,7 @@ module Api
           server.update_authentication({:default => authentication.compact}, {:save => true})
         end
 
-        server.pxe_menus = create_pxe_menus(menus) if menus
+        server.ensure_menu_list(menus.map { |el| el['file_name'] }) if menus
 
         if server.invalid?
           raise BadRequestError, "Failed to add a pxe server - #{server.errors.full_messages.join(', ')}"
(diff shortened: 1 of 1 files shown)


miha-plesko:
Sorry to bother you again @bdunne but we had an argue in our team whether the "quick and dirty" approach implemented in https://github.com/ManageIQ/manageiq-api/pull/644 is good enough for now or not. In the end I decided to fix it properly, even of we don't expect user to ever be needing this feature. Kindly ask you to review the two resulting PRs:

- https://github.com/ManageIQ/manageiq/pull/19126
- https://github.com/ManageIQ/manageiq-api/pull/648

What we basically do now is we don't just destroy all menus on server update, but we go extra mile and check which ones can actually be preserved.

miha-plesko:
Travis will be red until https://github.com/ManageIQ/manageiq/pull/19126 gets merged

app/controllers/api/pxe_servers_controller.rb
```diff
@@ -48,9 +48,18 @@ def edit_resource(type, id, data)
...
           server.pxe_menus.destroy_all
-          data['pxe_menus'] = create_pxe_menus(menus)
+          data['pxe_menus'] = create_pxe_menus(menus || [])
+        elsif menus
+          desired_menus = menus.map { |el| el['file_name'] }
+          current_menus = server.pxe_menus.map(&:file_name)
+          to_keep, to_destroy = current_menus.partition { |menu| desired_menus.include?(menu) }
+          to_create = desired_menus.reject { |menu| current_menus.include?(menu) }
+          server.pxe_menus.where(:file_name => to_destroy).destroy_all
+          data['pxe_menus'] = server.pxe_menus.where(:file_name => to_keep) + create_pxe_menus(to_create.map { |f| { 'file_name' => f } })
```
abellotti: feels a bit heavy on business logic here, I wonder if this should not be provided by a model helper method that we'd leverage.

miha-plesko: I think you're right @abellotti, here'se the utility method https://github.com/ManageIQ/manageiq/pull/19134

lpichler:
replaced by https://github.com/ManageIQ/manageiq-api/pull/707

Issue #651, opened 2019-08-16
[RFE] Add Tag Assignment for Cloud Tenants


lodgenbd:
Requesting to add tag subcollection for cloud tenant.

PR #361, opened 2018-04-19
Adding tagging support for additional collections

abellotti:
Adding tagging support for resources of the following collections:

  - /api/availability_zones
  - /api/cloud_networks
  - /api/cloud_subnets
  - /api/flavors
  - /api/network_routers
  - /api/security_groups

Standard tagging subcollection signature as follows:

  - /api/availability_zones/:id/tags
  - /api/cloud_networks/:id/tags
  - /api/cloud_subnets/:id/tags
  - /api/flavors/:id/tags
  - /api/network_routers/:id/tags
  - /api/security_groups/:id/tags

supporting regular queries as well as the assign and unassign actions.

Fixes: #77 

https://bugzilla.redhat.com/show_bug.cgi?id=1497061

PR #652, opened 2019-08-16
Add tagging for cloud tenants

lodgenbd:
Copied update wording and code from previous update that added tagging support for other cloud resources (#361)

Adding tagging support for resources of the following collections:

  /api/cloud_tenants

Standard tagging subcollection signature as follows:

  /api/cloud_tenants/:id/tags

Fixes: #651


lpichler:
@lodgenbd thanks for PR. 

Can you add some specs as we have in https://github.com/ManageIQ/manageiq-api/pull/361 ?
It will probably need also some changes in  config/api.yml. Maybe.

PR #636, opened 2019-07-30
FirmwareRegistry create/list/delete and sync_fw_binaries

miha-plesko:
With this commit we introduce following new API calls:

```
GET /api/firmware_registries
GET /api/firmware_registries/:id
DELETE /api/firmware_registries/:id
POST /api/firmware_registries/:id (action=sync_fw_binaries)
```

The POST call queues FirmwareRegistry refresh task and returns immediately.

Depends on: https://github.com/ManageIQ/manageiq/pull/19069
Needed by: https://github.com/ManageIQ/manageiq-ui-classic/pull/5929

@miq-bot add_label enhancement,ivanchuk/yes

PR #654, opened 2019-08-20
Specs for FirmwareRegistry updated

vandrlov:
Updated specs  as it was mentioned in https://github.com/ManageIQ/manageiq-api/pull/636#discussion_r312950817
Instead of `let` it uses rail helpers, so it's consistent with other specs. 
:unicorn: 

@lpichler is this ok? or should I do more changes?


lpichler:
cc @miha-plesko

miha-plesko:
Thanks @vandrlov! 👍

PR #664, opened 2019-08-27
Explicitly pass transformation_mapping when generating mapping items

djberg96:
In order to properly perform validations on a `TransformationMappingItem`, we need the parent to already be defined in some cases. This modifies the `create_mapping` method so that it accepts and uses the parent transformation mapping when creating a new transformation mapping item.

Tied to https://github.com/ManageIQ/manageiq/pull/19204


djberg96:
Closed in favor of https://github.com/ManageIQ/manageiq-api/pull/666

Issue #637, opened 2019-07-30
Get parent/children options for VMs


rvsia:
**Description**

We are converting `Virtual Machine edit form` to data driven forms and it would be nice to have an API endpoint, which returns all VMs which can be selected as a child or parent for the edited VM.

**Links to code in UI**
https://github.com/ManageIQ/manageiq-ui-classic/blob/d1822704512634446b858017adf999e560d7c850/app/controllers/vm_common.rb#L1353 https://github.com/ManageIQ/manageiq-ui-classic/blob/a459a4823a30b138f44a412b8c10a1f969b1b19a/app/controllers/vm_common.rb#L898

Cc @gtanzillo @hchrast

lpichler:
@miq-bot assign @lpichler

lpichler:
after discussion with @rvsia we need to add new endpoint for this. there is no possibility to use API filtering (no nested condition and GET request query limit)

@miq-bot unassign @lpichler 

I heard that @d-m-u is going to working on It ? 😊 if not, feel free to assign it to me 🏇

d-m-u:
@miq-bot assign @d-m-u

d-m-u:
So I would be shocked if there weren't an existing way to do this that doesn't involve the overhead of _adding a new endpoint_ for something that seems trivial.

lpichler:
@d-m-u I don't want to shock you  :D but I am afraid that we need new endpoint because are API doesn't support nested expressions 

we need to get vms  which you gen get by method [parent_choices](https://github.com/ManageIQ/manageiq-ui-classic/blob/a459a4823a30b138f44a412b8c10a1f969b1b19a/app/controllers/vm_common.rb#L898)

API request could like:

`GET /api/vms/:id/parent_choices` (maybe better name)

and you can see that in the in method is more queries
```
...
 ems_ids = VmOrTemplate.where(:id => ids).distinct.pluck(:ems_id).compact
      parent_item_scope = Rbac.filtered(VmOrTemplate.where(:ems_id => ems_ids).where.not(:id => ids).order(:name))
...
```

so It seem to me that we are not able to do it with existing possibilities in API.
Anyway if you know how let us know,

thanks!

Fryguy:
Can you explain the feature here? What is a parent/child VM?  Do you mean genealogy? If so, do we support this in the current UI, because this feels like a feature that should not exist as it could be easy for a user to create a bad connection. cc @agrare

rvsia:
[...]

PR #671, opened 2019-09-10
[WIP] Add beginning of endpoint to get parent vm choices

d-m-u:
We let people set ancestry in the UI on VMs apparently and I was assigned https://github.com/ManageIQ/manageiq-api/issues/637.

It's bad the way it is, and this makes it worse. This is horrible and we shouldn't do this but I was asked to, so. 

One of the actions we allow on edit of vms today is to update the ancestry: see https://github.com/ManageIQ/manageiq-api/commit/314e9e721b134a9e3094aac1b994d06b29dfa5c2. I'm wondering if that needs to change to accommodate the logic the UI is currently running? 

It doesn't even matter if this even does the same thing as the UI in terms of logic. We're  changing API behavior with this, because at the moment we have no existing logic to limit from the API perspective what can be picked like the UI is doing. 

Also I don't think you should be able to do this from either, frankly. 

The core identifier addition:
https://github.com/ManageIQ/manageiq/pull/19286

@lpichler @gtanzillo

### DIFF
--- a/app/controllers/api/vms_controller.rb
+++ b/app/controllers/api/vms_controller.rb
@@ -59,6 +59,14 @@ module Api
     end
     central_admin :suspend_resource, :suspend
 
+    def parent_choices_resource(ids)
+      parent_choices = {}
+      parent_choices[Digest::MD5.hexdigest(ids.inspect)] ||= begin
+        ems_ids = VmOrTemplate.where(:id => ids).distinct.pluck(:ems_id).compact
+        Rbac.filtered(VmOrTemplate.where(:ems_id => ems_ids).where.not(:id => ids).order(:name))
+      end
+    end
+
     def pause_resource(type, id = nil, _data = nil)
       raise BadRequestError, "Must specify an id for pausing a #{type} resource" unless id
 
--- a/app/controllers/api/vms_controller.rb
+++ b/app/controllers/api/vms_controller.rb
@@ -62,8 +62,8 @@ module Api
     def parent_choices_resource(ids)
       parent_choices = {}
       parent_choices[Digest::MD5.hexdigest(ids.inspect)] ||= begin
-        ems_ids = VmOrTemplate.where(:id => ids).distinct.pluck(:ems_id).compact
-        Rbac.filtered(VmOrTemplate.where(:ems_id => ems_ids).where.not(:id => ids).order(:name))
+        ems_ids = VmOrTemplate.where(:id => ids).select(:ems_id).distinct
+        Rbac.filtered(VmOrTemplate.where(:ems_id => ems_ids.where.not(:id => ids)))
       end
     end
 
--- a/config/api.yml
+++ b/config/api.yml
@@ -3992,6 +3992,9 @@
         :identifier:
         - vm_show
         - sui_vm_details_view
+      - :name: parent_choices
+        :identifier:
+        - vm_parent_choices
       :post:
       - :name: edit
         :identifier: vm_edit


d-m-u:
The API is already using ancestry -- the actual one, from the gem -- for this. I'm worried about us breaking existing implementation no matter how we do this, and I think the right way to do it is by changing this PR to not use the custom logic that the UI is today to get siblings since we have a method for that that the rest of the codebase is using.

lpichler:
@rvsia

app/controllers/api/vms_controller.rb
```diff
@@ -59,6 +59,17 @@ def suspend_resource(type, id = nil, _data = nil)
...
 
+    def parent_choices_resource(ids)
+      parent_choices = {}
+      parent_choices[Digest::MD5.hexdigest(ids.inspect)] ||= begin
+        ems_ids = VmOrTemplate.where(:id => ids).distinct.pluck(:ems_id).compact
+        Rbac.filtered(VmOrTemplate.where(:ems_id => ems_ids).where.not(:id => ids).order(:name))
+      end
+    end
+
+    def child_choices_resource()
```
lpichler: I think this is not needed, we need just `parent_choices_resource` for https://github.com/ManageIQ/manageiq-api/pull/671

d-m-u: Dude, the linked issue is called ```Get parent/children options for VMs```

config/api.yml
```diff
@@ -3992,6 +3992,12 @@
         :identifier:
         - vm_show
         - sui_vm_details_view
+      - :name: parent_choices
+        :identifier:
+        - we_don't_have_one_yet
```
lpichler: yes, it cloud be `vm_parent_choices` it is sound good to you.
we need to add it to `miq_product_fetaures.yml` in core repo - it could be hidden feature

config/api.yml
```diff
@@ -3992,6 +3992,12 @@
...
+      - :name: child_choices
```
lpichler: this is not as mentioned ^^^^. Api returns now child_resources.

d-m-u: do you mind telling me where that is?

app/controllers/api/vms_controller.rb
lpichler: yes. it was bad link, fixed, thanks 👍

config/api.yml
d-m-u: it's also one that's API specific, right?

app/controllers/api/vms_controller.rb
rvsia: In the UI form the selecting parent and childs is using the same options generated by this method https://github.com/ManageIQ/manageiq-ui-classic/blob/a459a4823a30b138f44a412b8c10a1f969b1b19a/app/controllers/vm_common.rb#L898 (so you can select a parent or childs from the same pool of vms)

I am not sure, if this is right, but it is what it is in UI.

config/api.yml
lpichler: Oh _What I wrote here ?_ so again for certitude:

we don't need `child_choices` as I mentioned it above. Api returns `child_resources` at this moment - so it is ok. (we put [together table](https://github.com/ManageIQ/manageiq-api/issues/637#issuecomment-530361764) and we found that we need only parent_choices as you have it in this PR - thanks)

 it is `virtual_has_many` child_resources - [here](https://github.com/ManageIQ/manageiq/blob/a0bf167e134e46f2b7ff827c6dc71dc9a3c42c0d/app/models/vm_or_template.rb)

and in API request you need to expose it:
```
http://localhost:3090/api/vms/1178?attributes=child_resources
```

config/api.yml
lpichler: yes, we will use it only in API and not in UI (not sure If I answered you question.)

app/controllers/api/vms_controller.rb
d-m-u: Please see what I wrote above, https://github.com/ManageIQ/manageiq-api/pull/671#issue-316007964. We're possibly changing behavior here and I'm concerned about that.

d-m-u: @gtanzillo this is still a concern.

rvsia: I am not sure we are changing anything, if this is a new action. But if so, I am not able to respond that. I would be happy to remove and simplify all constraints, but from my standpoint, I just need to get the right data for the form we are refactoring.

d-m-u: The API already uses different logic to do this, it's not exactly "a new action."

app/controllers/api/vms_controller.rb
```diff
@@ -59,6 +59,13 @@ def suspend_resource(type, id = nil, _data = nil)
     end
     central_admin :suspend_resource, :suspend
 
+    def parent_choices_resource(ids)
+      parent_choices = {}
+      parent_choices[Digest::MD5.hexdigest(ids.inspect)] ||= begin
+        Rbac.filtered(VmOrTemplate.where(:ems_id => VmOrTemplate.where(:id => ids).with_ems.group_by_ems(ids)).where.not(:id => ids))
```
kbrock:
```suggestion
        ems_ids = VmOrTemplate.where(:id => ids).select(:ems_id).distinct
        Rbac.filtered(VmOrTemplate.where(:ems_id => ems_ids.where.not(:id => ids))
```

- keeping it in its own variable is easier to read
- think the `group_by` isn't quite what we want
- is it possible to tell the api that we only want a few of the columns?

@lpichler Since this is fetching vms, does this belong as a sub-collection, or a search option on vms itself?

d-m-u: @lpichler I need an answer to Keenan's question, I'm blocked on this, and it'd be nice to get this moving.

lpichler: so OK, based on this [comment](https://github.com/ManageIQ/manageiq/pull/19287#issuecomment-531843518)  when are doing this API.

As I stated in this [comment](https://github.com/ManageIQ/manageiq/pull/19287#issuecomment-530470263) what we need:

```ruby
  def parent_choices(ids)
    @parent_choices = {}
    @parent_choices[Digest::MD5.hexdigest(ids.inspect)] ||= begin
      ems_ids = VmOrTemplate.where(:id => ids).distinct.pluck(:ems_id).compact
      parent_item_scope = Rbac.filtered(VmOrTemplate.where(:ems_id => ems_ids).where.not(:id => ids).order(:name))

      parent_item_scope.pluck(:name, :location, :id).each_with_object({}) do |vm, memo|
        memo[vm[0] + " -- #{vm[1]}"] = vm[2]
      end
    end
  end
```

(1) Parent VM select box
which is `parent_choices(vm.id)`


(2) Available VMs
`parent_choices([Vm.first.id] + Vm.first.children)`


So we need two endpoints because we need to avoid parameter `ids` as original method have 

@d-m-u  I guess that your method `parent_choices_resource`  tends to be (1).
so If you look [here](https://github.com/ManageIQ/manageiq-ui-classic/blob/3a817248039bcd87951069ab96cacc8f7c194254/app/controllers/vm_common.rb) you see that parameter `ids` are children of the `Vm` itself.


```ruby
vm_id  # id of resource (VM).
vm = VmOrTemplate.find(vm_id)
VmOrTemplate.where(:ems_id => vm.ems_id) .where.not(:id => vm_id)

#with RBAC
Rbac.filtered(VmOrTemplate.where(:ems_id => vm.ems_id) .where.not(:id => vm_id))
```

the case 2)

```ruby
vm_id  # id of resource (VM).
vm = VmOrTemplate.find(vm_id)
VmOrTemplate.where(:ems_id => vm.ems_id) .where.not(:id => vm_id)

#with RBAC
Rbac.filtered(VmOrTemplate.where(:ems_id => vm.ems_id).where.not(:id => [id] + children.pluck(:ids) )
```

then some logic can be moved to model:
`where(:ems_id => vm.ems_id)` -> `.with_ems(vm.ems_id)`
`where.not(:id => [id] + children.pluck(:ids))` -> `.without([id] + children.pluck(:ids))`


> - keeping it in its own variable is easier to read

Ok👍 


> - I think the group_by isn't quite what we want

it was `group_by_ems` (from https://github.com/ManageIQ/manageiq/pull/19287/files#diff-fb7c3f0121965870e99fe02ec498cf65R206 ) - but I prefer name  `with_ems` or something like this.
 
> is it possible to tell the api that we only want a few of the columns?

I think you need to add code to get requested column from params and validate it(if attribute exists on model), eventually use any existing method for it(take a look to [`parameters.rb`](https://github.com/ManageIQ/manageiq-api/blob/713b792ca06baf0be0c610b91c47940539717a81/app/controllers/api/base_controller/parameters.rb)), what we have in API and also pass column for ordering

and  keep existing format of api;

Example for

`?attributes=id,name&sort_by=name&sort_order=desc`

> @lpichler Since this is fetching vms, does this belong as a sub-collection, or a search option on vms itself?

we don't need isubcollection with just need it for one VM.
abd
I think it could be `search_options`, then @kbrock  I think, do  you have such API request on mind? 
 
Option 1)
`/api/vms/111?attributes=id,name&sort_by=name&sort_order=desc&search_options=parent_choices`

Option 2)
`/api/vms/111?attributes=id,name&sort_by=name&sort_order=desc&search_options=parent_choices_without_children`

NOTE: Not sure whether ` parent_choices_without_children ` is good.

app/controllers/api/vms_controller.rb
```diff
@@ -59,6 +59,14 @@ def suspend_resource(type, id = nil, _data = nil)
     end
     central_admin :suspend_resource, :suspend
 
+    def parent_choices_resource(ids)
+      parent_choices = {}
+      parent_choices[Digest::MD5.hexdigest(ids.inspect)] ||= begin
```
kbrock: you just created `parent_choices`. So the `||=` doesn't make sense to me.

According to the structure of this code, I was expecting to see a loop on `ids` and populating multiple `parent_choices`.

But I'm not totally clear what you are doing here, so don't take that as an endorsement of one way or another, just saying what it looked like you were trying to do.

Or I was expecting to see `@parent_choices ||= {}`
again, just trying to guess at your intentions

app/controllers/api/vms_controller.rb
```diff
@@ -59,6 +59,14 @@ def suspend_resource(type, id = nil, _data = nil)
...
+        ems_ids = VmOrTemplate.where(:id => ids).select(:ems_id).distinct
```
kbrock: thanks for setting `ems_id` this way

d-m-u:
Closing because @vandrlov is tasked with this.

PR #676, opened 2019-09-20
fix failing tests

d-m-u:
This addresses the recent spec failures seen from the category request. 

I don't understand why the sequencing didn't work and this clearly is the wrong fix but I wanted to prove that the sequencing has something wrong.


d-m-u:
Closing cause Keenan's got the fix for this.

PR #677, opened 2019-09-26
Remove useless param methods in ResultsController

lpichler:
it looks like leftover after https://github.com/ManageIQ/manageiq-api/pull/547

this already happens in https://github.com/ManageIQ/manageiq/blob/ed89b4df6cdc829e1413044bca48bbfe51f7ae91/app/models/miq_report_result/result_set_operations.rb#L16 
and we pull also parameters from `params` also here.

@miq-bot add_label technical_debt

@miq-bot assign @abellotti

### DIFF
--- a/app/controllers/api/base_controller/parameters.rb
+++ b/app/controllers/api/base_controller/parameters.rb
@@ -11,15 +11,7 @@ module Api
         end
 
         def report_options
-          params.merge(:sort_by => params['sort_by'], :sort_order => sort_order).merge(filter_options)
-        end
-
-        def filter_options
-          filtering_enabled? ? {:filter_string => params[:filter_string], :filter_column => params[:filter_column]} : {}
-        end
-
-        def filtering_enabled?
-          params.key?(:filter_column) && params.key?(:filter_string) && params[:filter_string]
+          params.merge(:sort_order => sort_order)
         end
       end
 


abellotti:
aren't we breaking API compatibility if we're no longer honoring those params coming in ? didn't see params referenced in https://github.com/ManageIQ/manageiq/blob/ed89b4df6cdc829e1413044bca48bbfe51f7ae91/app/models/miq_report_result/result_set_operations.rb#L16 just options. unless I'm missing something.

lpichler:
@abellotti params are passed as an second argument(`options`) to method `result_set_for_reporting(, params` [here](https://github.com/ManageIQ/manageiq-api/blob/master/app/controllers/api/results_controller.rb#L15) and
[`report_options` are basically `params`](https://github.com/ManageIQ/manageiq-api/blob/d01d1fd03ded2636de75b96094aa99ed5793685b/app/controllers/api/base_controller/parameters.rb#L14)

I also removed  `params.merge(:sort_by => params['sort_by'], ` but it looks like that it is not needed.
This works `params[:sort_by]` as well as `params['sort_by']`


 Processing params is done in [models methods]([here](https://github.com/ManageIQ/manageiq-api/blob/master/app/controllers/api/results_controller.rb#L15)) (Not sure whether it is best place for this code) and there is direct to like `options[:filter_string]`,... (which are just passed params)


@abellotti do you have any other concerns ?

abellotti:
yep, that's fine. and I see we have the specs for both sort_by and filter_string since that's the API contract even though it's handled by the model. 👍

PR #680, opened 2019-10-02
[FIX CI] Use visible attribute in services_spec

lpichler:
attribute` display` has been renamed to `visible` here https://github.com/ManageIQ/manageiq/pull/19211

ci failure https://travis-ci.org/ManageIQ/manageiq-api/jobs/592365926

@miq-bot assign @abellotti

### DIFF
--- a/spec/requests/services_spec.rb
+++ b/spec/requests/services_spec.rb
@@ -267,7 +267,7 @@ describe "Services API" do
                                                    {"action" => "remove", "path" => "description"},
                                                    {"action" => "add",    "path" => "display", "value" => true}])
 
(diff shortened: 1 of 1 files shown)


spec/requests/services_spec.rb
```diff
@@ -267,7 +267,7 @@
                                                    {"action" => "remove", "path" => "description"},
                                                    {"action" => "add",    "path" => "display", "value" => true}])
```
d-m-u: Does this one need to change too?

lpichler: it is not needed for fixing CI failure.

(but maybe it is needed for removing deprecation warning)

PR #686, opened 2019-10-17
Log exceptions to rails console

himdel:
Right now, if any UI request fails with an exception, it can be seen in rails console output and in development.log / production.log.
But, if the same problem happens with an API request, the only place where that exception is visible is api.log.

Logging API exceptions to rails output as well.
(Using the same logger invocation as ApplicationController.)

### DIFF
--- a/app/controllers/api/base_controller.rb
+++ b/app/controllers/api/base_controller.rb
@@ -193,6 +193,8 @@ module Api
       # We don't want to return the stack trace, but only log it in case of an internal error
       api_log_error("\n\n#{error.backtrace.join("\n")}") if type == :internal_server_error && !error.backtrace.empty?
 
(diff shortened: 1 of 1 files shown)


app/controllers/api/base_controller.rb
```diff
@@ -193,6 +193,8 @@ def api_error(type, error)
       # We don't want to return the stack trace, but only log it in case of an internal error
       api_log_error("\n\n#{error.backtrace.join("\n")}") if type == :internal_server_error && !error.backtrace.empty?
 
+      logger.fatal("Error caught: [#{error.class.name}] #{error.message}\n#{error.backtrace.join("\n")}")
```
abellotti: Do we really want to log a stack trace on each error, i.e. a 400 or 403, etc ? I would expect the backtrace only if we have a 500 (:internal_server_error)

abellotti: Right, I would log them all, but only add the \n#{error.backtrace.join("\n")} for internal_server_error.

PR #689, opened 2019-10-21
Remove moar extraneous factory attributes

d-m-u:
We don't need to set descriptions here and I need the github points.


djberg96:
For even more github points, you can refactor those two separate declarations into a single one, since the properties are now identical for both:

`FactoryBot.create_list(:miq_group, 2, :miq_user_role => @role)`

d-m-u:
That's for tomorrow's one-liner that poor Libor will have to deal with 😆

PR #690, opened 2019-10-21
Allow miq_regions to be edited, deleted

djberg96:
This adds edit and delete actions to the rest api for MiqRegion's. A delete can be performed using either a POST or DELETE.

A few fields have been explicitly blacklisted. Will update, if desired.

I added some rspec tags for convenience.

Partially addresses https://bugzilla.redhat.com/show_bug.cgi?id=1753806

PR #691, opened 2019-10-23
Allow Zones to be edited, deleted

djberg96:
Followup to https://github.com/ManageIQ/manageiq-api/pull/690, and very similar.

This PR allows for Zones to be edited and deleted. The only significant difference between this one and #690 is that previously some of these actions were explicitly disabled.

Partially addresses https://bugzilla.redhat.com/show_bug.cgi?id=1753806

### DIFF
--- a/app/controllers/api/zones_controller.rb
+++ b/app/controllers/api/zones_controller.rb
@@ -1,4 +1,32 @@
 module Api
   class ZonesController < BaseController
+    INVALID_ZONES_ATTRS = ID_ATTRS + %w[created_on updated_on].freeze
+
+    # Edit an existing zone. Certain fields are meant for internal use only
+    # and may not be edited. Attempting to edit one of the forbidden fields
+    # will result in a bad request error.
+    #
+    def edit_resource(type, id, data)
+      bad_attrs = data_includes_invalid_attrs(data)
+
+      if bad_attrs.present?
+        msg = "Attribute(s) '#{bad_attrs}' should not be specified for updating a zone resource"
+        raise BadRequestError, msg
+      end
+
+      super
+    end
+
+    private
+
+    # Check to see if any of the data attributes contain an invalid field.
+    # Returns a list of invalid fields as a comma separated string that you
+    # can use for error messages, or nil if the data argument is blank.
+    #
+    def data_includes_invalid_attrs(data)
+      return nil unless data
+
+      data.keys.select { |key| INVALID_ZONES_ATTRS.include?(key) }.compact.join(", ")
+    end
   end
 end
--- a/config/api.yml
+++ b/config/api.yml
@@ -4205,7 +4205,7 @@
       :post:
       - :name: query
         :identifier: zone
-      - :name: add
+      - :name: create
         :identifier: zone_new
       - :name: edit
         :identifier: zone_edit
--- a/config/api.yml
+++ b/config/api.yml
@@ -4196,7 +4196,7 @@
     - :collection
     :subcollections:
     - :settings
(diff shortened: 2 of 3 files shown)


djberg96:
@abellotti, @lpichler, @martinpovolny Please review.

config/api.yml
```diff
@@ -4207,22 +4207,19 @@
         :identifier: zone
       - :name: add
```
abellotti: this should be changed to create, add is historically there (and still supported in code), but we need to be consistent with others.

spec/requests/zones_spec.rb
```diff
@@ -1,7 +1,145 @@
...
+
+    it "forbids deletion of a zone without an appropriate role" do
+      api_basic_authorize
+      zone = FactoryBot.create(:zone, :description => "Current Region description")
+
+      delete api_zone_url(nil, zone)
+
+      expect(response).to have_http_status(:forbidden)
+    end
+  end
```
abellotti: nice test coverage !! 👍

config/api.yml
djberg96: Ok, updated.

app/controllers/api/zones_controller.rb
```diff
@@ -1,4 +1,32 @@
...
+
+      if bad_attrs.present?
+        msg = "Attribute(s) '#{bad_attrs}' should not be specified for updating a zone resource"
+        raise BadRequestError, msg
+      end
+
+      super
+    end
+
+    private
```
martinpovolny: I am a bit concerned about this being the same as this:

https://github.com/ManageIQ/manageiq-api/pull/690/files#diff-a8c5eb840e76e0a1632c7acdb2b03a43R20

Well also the part above.

I think that the behavior should be moved to a/the? parent class or a mixin.

We can agree on merging this (and the 3rd one) and then refactor it once the functionality is there. If you prefer it that way.

WDYT?

djberg96: The fields actually aren't identical - created_at vs created_on. But yes, we should eventually try to put this into some common code.

abellotti:
Hmm, I don't see where zones can be deleted in the UI unless I'm missing this.  /cc @Fryguy 

@djberg96 @martinpovolny I think we should forgo the delete, and shift to white-listing editable attributes here. gotta match the UI unless there are new UI requirements.

abellotti:
Looks like in the Settings.

image

though grayed out for current zone, what does API/Model return if the current zone is getting deleted ?

djberg96:
@abellotti There's some logic in zone.rb that prevents a zone in use from being destroyed:

https://github.com/ManageIQ/manageiq/blob/master/app/models/zone.rb#L243-L247

simaishi:
@djberg96 can this be `ivanchuk/yes`?

djberg96:
@martinpovolny or @abellotti any objections to making it ivanchuk/yes?

simaishi:
^ ping

djberg96:
Let's say yes. :)

simaishi:
Ivanchuk backport details:

```
$ git log -1
commit 97be56b07e6b6586b0921f4cf63b02464ee1b9bb
Date:   Thu Oct 24 11:39:08 2019 +0200

    Merge pull request #691 from djberg96/zones_post_actions

    Allow Zones to be edited, deleted

    (cherry picked from commit e9b1b2c93d12d09bddf8dbdead5ada5dc3af8e90)

    https://bugzilla.redhat.com/show_bug.cgi?id=1805844
```

PR #693, opened 2019-10-24
Refactor region and zone specs a bit

djberg96:
Since I just submitted PR's for updated region and zone specs, I went back and refactored them a bit. Mainly this consisted of using `expect_forbidden_request` for simple specs that use a basic authorize, as well as `expect_single_action_result` in a few places, which is a little more thorough.

There were a couple other very minor tweaks, like using a single factory where we can, and changing "forbidden" to "invalid" in some spec descriptions to avoid confusion with forbidden requests.

### DIFF
--- a/spec/requests/regions_spec.rb
+++ b/spec/requests/regions_spec.rb
@@ -129,7 +129,7 @@ RSpec.describe "Regions API", :regions do
       region = FactoryBot.create(:miq_region)
 
       expect { post api_region_url(nil, region), :params => gen_request(:delete) }.to change(MiqRegion, :count).by(-1)
-      expect(response).to have_http_status(:ok)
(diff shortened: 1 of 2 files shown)


djberg96:
@martinpovolny @lpichler Please review.

djberg96:
@miq-bot add_label cleanup

spec/requests/regions_spec.rb
```diff
@@ -9,23 +9,15 @@
 #
 
 RSpec.describe "Regions API", :regions do
+  let(:region) { FactoryBot.create(:miq_region, :region => '2') }
```
Fryguy: This is probably invalid, technically, as the id will not be part of that region's range.  However, since this was a copy/paste from below, it's ok for now.

PR #695, opened 2019-10-28
Allow Servers to be edited and deleted

djberg96:
Followup to https://github.com/ManageIQ/manageiq-api/pull/691.

Completes https://bugzilla.redhat.com/show_bug.cgi?id=1753806

~~WIP for now until I can figure out why the 3 pending specs are failing.~~

### DIFF
--- a/app/controllers/api/servers_controller.rb
+++ b/app/controllers/api/servers_controller.rb
@@ -1,20 +1,6 @@
 module Api
   class ServersController < BaseController
-    INVALID_SERVERS_ATTRS = ID_ATTRS + %w[
-      started_on
-      stopped_on
-      percent_memory
-      percent_cpu
-      cpu_time
-      memory_usage
-      memory_size
-      proportional_set_size
-      system_memory_free
-      system_memory_used
-      system_swap_free
-      system_swap_used
-      unique_set_size
-    ].freeze
+    VALID_SERVERS_ATTRS = %w[name zone]
 
     # Edit an existing server (MiqServer). Certain fields are meant for
     # internal use only and may not be edited. Attempting to edit one of
@@ -39,8 +25,7 @@ module Api
     #
     def data_includes_invalid_attrs(data)
       return nil unless data
-
-      data.keys.select { |key| INVALID_SERVERS_ATTRS.include?(key) }.compact.join(", ")
+      (data.keys - VALID_SERVERS_ATTRS).compact.join(', ')
     end
   end
 end
--- a/app/controllers/api/servers_controller.rb
+++ b/app/controllers/api/servers_controller.rb
@@ -1,4 +1,46 @@
 module Api
   class ServersController < BaseController
+    INVALID_SERVERS_ATTRS = ID_ATTRS + %w[
+      started_on
+      stopped_on
+      percent_memory
+      percent_cpu
+      cpu_time
+      memory_usage
+      memory_size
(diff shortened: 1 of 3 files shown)


djberg96:
@martinpovolny, @lpichler, @abellotti I need an extra set of eyes to see why the delete operations are failing.

Also, I see we changed the identifier for the settings operations, please let me know if what I've done here conflicts with that in any way.

Lastly, please double check the list of restricted attributes and see if you think it's ok. I was eyeballing it a bit, just looking for attributes that are (presumbably) determined through diagnostics and shouldn't be set directly.

Fryguy:
@carbonin Are there potential issues with anything exposed here?

Fryguy:
@djberg96 Are all of these attributes editable via the UI?  Any that are not editable should probably be blocked here (I see you have some, but I'm not sure on the rest).

I'm also not sure on deletes at all.  Do we allow deletes from the UI?

carbonin:
I don't see why we should allow editing anything other than name and zone.

I think almost everything else is set by the system or *really* shouldn't be edited (`guid`, `status`, `pid`, `version`...)

djberg96:
As per @himdel's notes, the MiqServer model has some restrictions on whether or not a server can be deleted, so that's why they were failing. I've updated the factories so that they pass those internal checks. I've also added some specs that explicitly check to make sure a server is not deleted if it's not considered deleteable.

In addition, as per @carbonin's comments, I've switched from a blacklist to a whitelist of valid attributes. Right now it's only name and zone.

Fryguy:
@djberg96 Did you determine if servers are deletable from the UI?  If not, then we shouldn't allow delete from the API either.

We have to be very careful with server records because they are the core of the entire worker management system.  Exposing that to end users should be more delicate than other objects.

djberg96:
@himdel Thanks.

Though, for the specs to pass, the ability to edit and/or delete a server has to be a supported product feature, which is the spec that's failing now.

Do we want to add the ability to edit and/or delete servers? If so, I'll need to add a schema migration.

djberg96:
@himdel Thanks.

I've updated the config/api.yml file to use `ops_settings` which is the identifier used for the server settings screen in general.

djberg96:

djberg96:
Note that I've updated this PR so that it uses a whitelist, so only two attributes are editable now. I've also added some specs that verify that you can't delete a server that isn't deletable.

martinpovolny:
Code looks good. Tests look great. It works.

The moving of the similar code dealing with editability of attributes is to be done in a separate PR.

@lpichler : merge?

simaishi:
@djberg96 can this be `ivanchuk/yes`?

djberg96:
@simaishi I don't see why not. @martinpovolny Any objections?

simaishi:
^ ping

djberg96:
@simaishi I say yes. :)

simaishi:
Ivanchuk backport details:

```
$ git log -1
commit 37338994e6881a205c8e5c23cbdeae9147ac3961
Date:   Wed Nov 6 13:56:43 2019 +0100

    Merge pull request #695 from djberg96/server_post_actions2

    Allow Servers to be edited and deleted

    (cherry picked from commit 57b7037eff3c3acf993613f69b80d9a03415922a)

    Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1805844
```

PR #697, opened 2019-10-29
Added some expected delete failures for zones

djberg96:
A followup PR to https://github.com/ManageIQ/manageiq-api/pull/691, this adds a couple of specs that verify that you can't delete a zone with the API that shouldn't be deleteable.


djberg96:

PR #660, opened 2019-08-25
Add endpoints for widget content generation

d-m-u:
Add ability to trigger generation of widget content from the API. 

~~@miq-bot add_label enhancement, hammer/yes, ivanchuk/yes~~ 

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1623607

@gmcculloug please review, it's your ~~party~~ funeral 
@abellotti please also review, it's less of but still kinda your party

PR #698, opened 2019-11-05
Fix call to generate_content

d-m-u:
Fixing a bug introduced originally introduced in https://github.com/ManageIQ/manageiq-api/pull/660. ```queue_object_action``` can't be used here because not all generate_content calls create tasks, and so relying on a task_id is bad. Per conversation with Libor we should just call ```queue_gen_content``` directly. 

@miq-bot add_label bug
@miq-bot assign @lpichler 
@miq-bot add label hammer/yes, ivanchuk/yes

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1761836

### DIFF
--- a/app/controllers/api/widgets_controller.rb
+++ b/app/controllers/api/widgets_controller.rb
@@ -12,8 +12,8 @@ module Api
 
     def generate_widget_content(widget)
       desc = "#{widget_ident(widget)} content generation"
-      task_id = queue_object_action(widget, desc, :method_name => "generate_content", :args => ["User", User.current_user.current_group.description, [User.current_user.userid]])
-      action_result(true, desc, :task_id => task_id)
+      widget.queue_generate_content
+      action_result(true, desc)
     rescue => err
       action_result(false, err.to_s)
     end
--- a/spec/requests/widgets_spec.rb
+++ b/spec/requests/widgets_spec.rb
@@ -32,6 +32,18 @@ describe "Widgets API" do
   end
 
   describe "Widgets generate_content action" do
+    let(:miq_widget) { FactoryBot.create(:miq_widget, :visibility => {:roles => '_ALL_'}, :resource => MiqReport.first) }
+    let(:second_miq_widget) { FactoryBot.create(:miq_widget, :visibility => {:roles => '_ALL_'}, :resource => MiqReport.first) }
+    let(:feature1) { MiqProductFeature.find_all_by_identifier("dashboard_admin") }
+    let(:user1) { FactoryBot.create(:user, :role => "role1", :features => feature1) }
+    let(:group1) { user1.current_group }
+    let(:ws) { FactoryBot.create(:miq_widget_set, :name => "Home", :userid => user1.userid, :group_id => group1.id) }
+
+    before do
+      MiqReport.seed_report("Vendor and Guest OS")
+      miq_widget.make_memberof(ws)
+    end
+
     context "with an invalid id" do
       it "it responds with 404 Not Found" do
         api_basic_authorize(action_identifier(:widgets, :generate_content, :resource_actions, :post))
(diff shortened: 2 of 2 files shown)


d-m-u:
@miq-bot add_label hammer/yes, ivanchuk/yes

d-m-u:
@miq-bot remove_label wip

d-m-u:
hey @lpichler any chance I could get a review please?

spec/requests/widgets_spec.rb
```diff
@@ -62,43 +62,83 @@
...
+      context "generate_content for a user" do
+        before do
+          MiqReport.seed_report("Vendor and Guest OS")
+          feature1 = MiqProductFeature.find_all_by_identifier("dashboard_admin")
+          @user1   = FactoryBot.create(:user, :role => "role1", :features => feature1)
+          @group1  = @user1.current_group
+        end
+
+        it "single widget" do
+          miq_widget = FactoryBot.create(:miq_widget, :visibility => {:roles => '_ALL_'}, :resource => MiqReport.first)
```
lpichler: What about to move all `FactoryBot`s to `let`s and share them in those two `it`s ?

lpichler:
@d-m-u looks good and nice specs.  I have just one thing - it looks like that you can share `FactoryBot object`s thru `let`s.

app/controllers/api/widgets_controller.rb
```diff
@@ -12,8 +12,7 @@ def generate_content_resource(type, id, _data = nil)
 
     def generate_widget_content(widget)
       desc = "#{widget_ident(widget)} content generation"
-      task_id = queue_object_action(widget, desc, :method_name => "generate_content", :args => ["User", User.current_user.current_group.description, [User.current_user.userid]])
-      action_result(true, desc, :task_id => task_id)
+      widget.queue_generate_content
```
abellotti: success action_result signature missing. should complement the failure, even without a task_id, i.e. action_result(true, desc)

d-m-u:
Anyone want to review this please?

app/controllers/api/widgets_controller.rb
```diff
@@ -12,8 +12,8 @@ def generate_content_resource(type, id, _data = nil)
...
+      action_result(true, desc)
```
lpichler: @d-m-u 
`widget.queue_generate_content` returns task_id

so can we add task id to result as well ?

  `action_result(true, desc, :task_id => task_id)`

d-m-u: No, it doesn't. It will for a user. It does not for a group.

spec/requests/widgets_spec.rb
```diff
@@ -62,43 +65,66 @@
...
-        expect(MiqTask.count).to eq(0)
-        post(api_widget_url(nil, miq_widget), :params => gen_request(:generate_content))
-
-        expect_single_action_result(:success => true, :message => /#{miq_widget.id}.* content generation/i, :href => api_widget_url(nil, miq_widget))
-        expect(MiqTask.first.message).to match(/content generation\]\ being run for user/)
+      context "generate_content for a group" do
+        before do
+          stub_settings(::Settings.to_hash.merge(:product => {:report_sync => true}))
+          MiqReport.seed_report("Vendor and Guest OS")
+          feature1 = MiqProductFeature.find_all_by_identifier("dashboard_admin")
```
lpichler: we can share some  code here.
1. We can use `let`'s for lines 72-75 and 101-104.
2. those two `before do` blocks looks same, is it possible to move them under contexts or move them on place `somehow` ?

spec/requests/widgets_spec.rb
```diff
@@ -62,43 +65,66 @@
...
+          @user1   = FactoryBot.create(:user, :role => "role1", :features => feature1)
+          @group1  = @user1.current_group
+          @ws = FactoryBot.create(:miq_widget_set, :name => "Home", :userid => @user1.userid, :group_id => @group1.id)
+          miq_widget.make_memberof(@ws)
+          api_basic_authorize('widget_generate_content')
+        end
+
+        it "single widget" do
```
lpichler: can you create a sentence what it does ? 

for example:
`it "generates single widgets "`

PR #702, opened 2019-11-06
Replace display with visible in services specs.

djberg96:
Followup to https://github.com/ManageIQ/manageiq-api/pull/680, this fixes a couple more places where `display` should be replaced with `visible` to avoid warnings.


djberg96:
@lpichler Please review.

Issue #489, opened 2018-10-10
PhysicalChassisController - duplicate controller name


himdel:
`PhysicalChassisController` already exists, provided by `manageiq-ui-classic/app/controllers/physical_chassis_controller.rb`.

But thanks to the API defining a controller by the same name, accessing `/physical_chassis/show_list` now fails:

```
[----] I, [2018-10-10T14:08:40.918265 #13280:2aeef94b9da4]  INFO -- : Started GET "/physical_chassis/show_list" for ::1 at 2018-10-10 14:08:40 +0000
[----] F, [2018-10-10T14:08:40.956717 #13280:2aeef94b9da4] FATAL -- :                                                                  
[----] F, [2018-10-10T14:08:40.956815 #13280:2aeef94b9da4] FATAL -- : LoadError (Unable to autoload constant PhysicalChassisController, expected /home/himdel/manageiq-api/app/controllers/api/physical_chassis_controller.rb to define it):
[----] F, [2018-10-10T14:08:40.956851 #13280:2aeef94b9da4] FATAL -- :   
[----] F, [2018-10-10T14:08:40.956894 #13280:2aeef94b9da4] FATAL -- : activesupport (5.0.7) lib/active_support/dependencies.rb:513:in `load_missing_constant'
activesupport (5.0.7) lib/active_support/dependencies.rb:203:in `const_missing'
activesupport (5.0.7) lib/active_support/inflector/methods.rb:268:in `const_get'
activesupport (5.0.7) lib/active_support/inflector/methods.rb:268:in `block in constantize'
activesupport (5.0.7) lib/active_support/inflector/methods.rb:266:in `each'
activesupport (5.0.7) lib/active_support/inflector/methods.rb:266:in `inject'
activesupport (5.0.7) lib/active_support/inflector/methods.rb:266:in `constantize'
activesupport (5.0.7) lib/active_support/dependencies.rb:584:in `get'
activesupport (5.0.7) lib/active_support/dependencies.rb:615:in `constantize'
actionpack (5.0.7) lib/action_dispatch/http/request.rb:81:in `controller_class'
actionpack (5.0.7) lib/action_dispatch/routing/route_set.rb:44:in `controller'
actionpack (5.0.7) lib/action_dispatch/routing/route_set.rb:30:in `serve'
actionpack (5.0.7) lib/action_dispatch/journey/router.rb:39:in `block in serve'
[...]

PR #709, opened 2019-11-11
Remove autoload_path

### LINKED ISSUE #708: controllers/api should not be on the autoload_paths
https://github.com/ManageIQ/manageiq-api/blob/9b2d09480c67eb316b350e3bc5b083a688f7f5b8/lib/manageiq/api/engine.rb#L15

This line should not exist because the controllers themselves are namespaced under `module Api`, and having it on the load path causes it to "de-namespace" during lookup.

This causes issues where similarly named top-level constants can confuse Rails' autoload search.  For example, the ui has an actual PingController defined in their app/controllers/ping_controller.rb, however if this repo's app/controller/api is earlier in the lookup chain, Rails will expect that app/controllers/api/ping_controller.rb contains that constant, and then ends up finding Api::PingController instead, causing this error:

    LoadError:  Unable to autoload constant PingController, expected /home/travis/build/ManageIQ/manageiq-ui-classic/vendor/bundle/bundler/gems/manageiq-api-9b2d09480c67/app/controllers/api/ping_controller.rb to define it

Fryguy:
This autoload_path is not necessary as all Api controllers are
namespaced.  This actually causes errors when there are similarly named
top-level classes elsewhere, as Rails wil try to autload from this path
and not find that top-level class.

Fixes #708

@abellotti @jrafanie Please review.

### DIFF
--- a/lib/manageiq/api/engine.rb
+++ b/lib/manageiq/api/engine.rb
@@ -2,7 +2,7 @@ require 'rails/engine'
 module ManageIQ
   module Api
     class Engine < ::Rails::Engine
-      isolate_namespace ManageIQ::Api
+      isolate_namespace Api
 
       # NOTE:  If you are going to make changes to autoload_paths, please make
       # sure they are all strings.  Rails will push these paths into the
@@ -12,7 +12,6 @@ module ManageIQ
       #
       #   https://bugs.ruby-lang.org/issues/14372
       #
-      config.autoload_paths << root.join('app', 'controllers', 'api').to_s
       config.autoload_paths << root.join('lib', 'services').to_s
       config.autoload_paths << root.join('lib').to_s
 


Fryguy:
https://github.com/ManageIQ/manageiq-cross_repo-tests/pull/10

jrafanie:
@Fryguy Looks good other than the typos:

`wil try to autload`  :trollface: 

Also, it looks like the cross-repo tests failed on schema

agrare:
Looks like schema's `bin/setup` doesn't do `bundle install`

himdel:
This should also fix https://github.com/ManageIQ/manageiq-api/issues/489

jrafanie:
> This should also fix #489

@himdel awesome, thanks!  I couldn't remember which issue we hit this in the past for but it really makes sense to just fix this and fix any fallout.

Fryguy:
@jrafanie @abellotti Is this good to go?  cross_repo tests are passing aside from schema, but schema is failing for unrelated reasons.

abellotti:
ok by me.  @jrafanie ?

jrafanie:
I'm going to merge with `wil try to autload` not fixed for fun.

PR #714, opened 2019-11-26
Remove load balancers from the API.

martinpovolny:
BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1754844

I am unsure about whether we want to do this. What is the correct way to deal with API endpoints for stuff that has been removed.

From the backwords compatibility POW it might be better to have an empty collection rather than just remove the collection alltogether.

WDYT: @abellotti, @lpichler, @djberg96  ?


abellotti:
if they're deprecated in the model (not removed yet), the API should still function, but we'd log a warning saying it is deprecated. When it's zapped from the model, then we'd do the same in the API.

martinpovolny:
@abellotti : to make it clear. If they are present in the model and NOT in the UI then you say keep them in the API and log a wadning. Right?

chessbyte:
Replaced with #727

PR #719, opened 2019-12-10
Fix custom_button_sets spec

agrare:
Don't match on the exact hash, just check for certain attributes being
included.

Caused by: https://github.com/ManageIQ/manageiq/pull/19573


himdel:
@miq-bot add_label hammer/yes, ivanchuk/yes

(because https://github.com/ManageIQ/manageiq/pull/19573 has the same)

PR #723, opened 2020-01-10
Extend the existing provider creation with DDF support

skateman:
@agrare [created](https://github.com/ManageIQ/manageiq-providers-amazon/pull/580/files) a new method on the amazon provider that creates a new record from the data submitted from a DDF. In order to keep backwards compatibility and also have the fancy `POST /api/providers` for creation, I added a test for a `ddf` attribute. If this attribute is set, the new creation method will be called and if not, the legacy one is executed. So it is necessary to pass a `ddf: true` attribute with the submitted values.

@miq-bot assign @martinpovolny 
cc @Hyperkid123 can you test it together with https://github.com/ManageIQ/manageiq-providers-amazon/pull/580 in your code?

### DIFF
--- a/app/controllers/api/providers_controller.rb
+++ b/app/controllers/api/providers_controller.rb
@@ -3,6 +3,7 @@ module Api
     TYPE_ATTR         = "type".freeze
     ZONE_ATTR         = "zone".freeze
     CREDENTIALS_ATTR  = "credentials".freeze
+    DDF_ATTR          = 'ddf'.freeze
     AUTH_TYPE_ATTR    = "auth_type".freeze
     DEFAULT_AUTH_TYPE = "default".freeze
     CONNECTION_ATTRS  = %w(connection_configurations).freeze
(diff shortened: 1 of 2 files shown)


Hyperkid123:
@skateman first thing on Monday

skateman:
No, I guess the zone field needs a `dataType` set to an integer and maybe its `name` should be `zone_id` instead of `zone`.

skateman:
Umm, no ... you need the name of the zone, not the ID. See here https://github.com/ManageIQ/manageiq-providers-amazon/pull/580/files#diff-235184baadf210fc9a4d037d8ffa1c37R179

skateman:
Also make sure you're loading Adam's PR mentioned above

Hyperkid123:
> Umm, no ... you need the name of the zone, not the ID. See here https://github.com/ManageIQ/manageiq-providers-amazon/pull/580/files#diff-235184baadf210fc9a4d037d8ffa1c37R179

What? Using the name instead of ID? I guess I can try it.

> Also make sure you're loading Adam's PR mentioned above

I am using all the PRs mentioned

skateman:
Okay, try it. The autoloading error is real, I sent it over to Adam. But he's on PTO this week.

skateman:
@Hyperkid123 I made some changes, if you send the zone as a string, it goes further. I guess we need the list of the regions in that dropdown to send a correct one, because that's where it's failing now...

Hyperkid123:
@skateman the regions currently are using the same values as in the original code. We can change that easily.

skateman:
After some hacking around in the UI, I was able to create a form that uses this as an endpoint and successfully created a provider, so this is good to go @lpichler 

I even added tests to make you happy :wink:

skateman:

PR #727, opened 2020-01-27
Deprecate load balancers

martinpovolny:
Add deprecation mechanism to the config file.

Replacement for https://github.com/ManageIQ/manageiq-api/pull/714 

Related BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1754844

ping @lpichler 

Do we want to check for the deprecation on more places?


abellotti:
Closing as it was done in #752

PR #729, opened 2020-01-29
Fix spec failure

ZitaNemeckova:
Check `"frame-src"` or `"child-src"`

Fixes 
```
 1) Headers Response Headers returns some headers related to security
     Failure/Error: #expect(content_security_policy_for("frame-src")).to include("'self'")
       expected nil to include "'self'", but it does not respond to `include?`
     # ./spec/requests/headers_spec.rb:85:in `block (3 levels) in <top (required)>'
```

@miq-bot add_label bug, test

### DIFF
--- a/spec/requests/headers_spec.rb
+++ b/spec/requests/headers_spec.rb
@@ -81,7 +81,7 @@ RSpec.describe "Headers" do
       expect(response.headers.to_h).to include(expected)
       expect(content_security_policy_for("default-src")).to include("'self'")
       expect(content_security_policy_for("connect-src")).to include("'self'")
-      expect(content_security_policy_for("frame-src")).to include("'self'")
+      expect(content_security_policy_for("frame-src") || content_security_policy_for("child-src")).to include("'self'")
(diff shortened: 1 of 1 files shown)


himdel:
Looks like `frame-src` went deprecated in somewhere around 2016, and browsers started reading `child-src` instead of `frame-src`, when present.

Then it went back, and `frame-src` is again a thing, with a fallback to `child-src` if not present.

Our upgrade to newer release of an ancient branch of secure_headers pushed us into the era of child-src, so this works around that, while (hopefully) staying compatible with future updates using frame-src again.

(refs: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy,  https://stackoverflow.com/questions/30023608/how-to-use-frame-src-and-child-src-in-firefox-and-other-browsers )

LGTM :)
Cc @lpichler

djberg96:
👍

djberg96:
Hm, I'm curious how coveralls calculated a reduction in test coverage. Maybe it wants each condition tested separately?

agrare:
Thanks for the fix @ZitaNemeckova

Issue #726, opened 2020-01-21
Notifications GET request returns wrong "count"


fhlavac:
When request sent to
` /api/notifications?expand=resources&attributes=details&sort_by=id&sort_order=desc&limit=100`
then in a reply (picture below)  received `count: 11468`
![image](https://user-images.githubusercontent.com/50696716/72819177-159e0200-3c6d-11ea-94ea-d41ada2beb95.png)

I would expect that this value is the total notifications count of the current user but it isn't. In this case, the number of total user's notifications is 200. To get the number of total user's notifications I have to send another request. The current value of "count" is quite useless, maybe it is a bug. Would it be possible to replace it with the total number of current user's notifications?

@martinpovolny

martinpovolny:
Investigation notes:

In `app/controllers/api/base_controller/renderer.rb` `collection_search` applies custom "search_condition" via a call to `notifications_search_conditions`.

However in `app/controllers/api/base_controller.rb` in `index` a call to `Api::QueryCounts.new` is done with `NotificationRecipient.count`.

So this is probably going to be broken everywhere where `*_search_condition` is used:

```
app/controllers/api/automation_requests_controller.rb
app/controllers/api/groups_controller.rb
app/controllers/api/notifications_controller.rb
app/controllers/api/provision_requests_controller.rb
app/controllers/api/reports_controller.rb
app/controllers/api/requests_controller.rb
app/controllers/api/results_controller.rb
app/controllers/api/service_orders_controller.rb
app/controllers/api/service_requests_controller.rb
app/controllers/api/tenant_groups_controller.rb
```

gtanzillo:
@lpichler Can you take a look at this one?

PR #731, opened 2020-01-30
[WIP] Try to apply search_conditions on counts.

martinpovolny:
Issue: https://github.com/ManageIQ/manageiq-api/issues/726

This is not the right solution to the problem, but rather an experiment.

`:count` also does not reflect RBAC filtering.


gtanzillo:
Closing as this was an experiment

PR #735, opened 2020-02-03
snapshot_create - support :snapshot_name_optional? on RHV VMs

himdel:
ovirt & redhat VMs do not require (or use) snapshot name,
both UIs use the snapshot_name_optional? logic, but only the SUI uses the API and thus fails to create snapshots for RHV VMs.

This makes the API respect parent.snapshot_name_optional? and only require snapshot if false/not present

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1698421
@miq-bot add_label bug, ivanchuk/yes


abellotti:
LGTM!! Thanks @himdel for fixing this in the API. 👍

PR #737, opened 2020-02-06
Adding support for OpenID-Connect/OAuth2 in the API

abellotti:
This Enhancement enables OpenID-Connect authentication (OAuth2) on the API. Current support for OpenID-Connect is only for the web interface and the API is not enabled as such. The only enablement needed is to configure the ManageIQ Appliance for OpenID-Connect as per:

```
https://www.manageiq.org/docs/reference/latest/auth/openid_connect
```

This enhancement leverages the appliance configuration and enables the following:

_First Use Case:_

- Authenticating with the API using a user access token (JWT) as obtained from the Identity provider. The token is simply passed to the API via the Authorization header as follows:

```
  curl -H "Authorization: Bearer <JWT>" https://<appliance>/api
```

The API performs the necessary Token introspection to validate the token and obtain the user details from the OpenID Client Claim definitions.

The API then performs the proper mapping of the OpenID Claim data to the currently supported headers as per the OIDC Assertions in the document referenced above for external authentication to work as it does today.

_Second Use Case:_

- Authenticating via Basic Auth.

  Here the standard basic authentication username and password is passed along to the API.  However, when the appliance  is configured for OpenID-Connect authentication, the API will request a token generation for the user from the IDP. If successful, It then follows the same authentication logic as the Bearer/JWT authentication mentioned above.

**NOTE:**

While this approach works fine and leverages the appliance configuration as is with no change, an alternate approach where the Apache configuration could be configured to protect the /api endpoint behind OpenID-Connect may be preferred.

Similar to how external authentication (Kerberos) is defined and wired for the API, see https://github.com/ManageIQ/manageiq-appliance/blob/236641d25b8b5d8a864893191955c8c19af1f275/TEMPLATE/etc/httpd/conf.d/manageiq-external-auth.conf.erb#L37 for the details there.

Though for OpenID-Connect/OAuth2, we would need to modify the https://github.com/ManageIQ/manageiq-appliance/blob/master/TEMPLATE/etc/httpd/conf.d/manageiq-external-auth-openidc.conf.erb file and define/protect the /api as documented here: https://github.com/zmartzone/mod_auth_openidc 

This would also allow us to have the single Apache container communicating with the IDP and having the API/WS containers only handle the trusted Apache headers coming in internally, as was done with https://github.com/ManageIQ/httpd_configmap_generator for the ManageIQ httpd container for the podified ManageIQ.

### DIFF
--- a/app/controllers/api/base_controller/authentication.rb
+++ b/app/controllers/api/base_controller/authentication.rb
@@ -52,11 +52,13 @@ module Api
                 # Basic auth, user/password but configured against OpenIDC.
                 # Let's authenticate as such and get a JWT for that user.
                 #
-                authenticate_with_jwt(get_jwt_token(u, p))
-              else
-                user = User.authenticate(u, p, request, :require_user => true, :timeout => timeout)
-                auth_user(user.userid)
+                user_jwt   = get_jwt_token(u, p)
+                token_info = validate_jwt_token(user_jwt)
+                user_data, membership = user_details_from_jwt(token_info)
+                define_jwt_request_headers(user_data, membership)
               end
+              user = User.authenticate(u, p, request, :require_user => true, :timeout => timeout)
+              auth_user(user.userid)
             rescue MiqException::MiqEVMLoginError => e
               raise AuthenticationError, e.message
             end
@@ -114,9 +116,13 @@ module Api
       end
 
       def authenticate_with_jwt(jwt_token)
-        token_data = validate_jwt_token(jwt_token)
+        token_info = validate_jwt_token(jwt_token)
+        user_data, membership = user_details_from_jwt(token_info)
+        define_jwt_request_headers(user_data, membership)
 
-        auth_user(token_data["username"])
+        timeout = ::Settings.api.authentication_timeout.to_i_with_method
+        user = User.authenticate(user_data[:username], "", request, :require_user => true, :timeout => timeout)
+        auth_user(user.userid)
       rescue => e
         raise AuthenticationError, "Failed to Authenticate with JWT - error #{e}"
       end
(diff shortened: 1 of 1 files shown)


abellotti:
Marking as [WIP] -  Actual implementation would move a lot of this code to https://github.com/ManageIQ/manageiq/blob/master/app/models/authenticator/httpd.rb

clearnight79:
@abellotti "Actual implementation" mean the alternative Apache configuration solution?

clearnight79:
@abellotti if I want to patch my current CF 5.0, is this the file I need to replace?
/opt/rh/cfme-gemset/bundler/gems/cfme-api-006d816cf2a0/app/controllers/api/base_controller/authentication.rb

abellotti:
Sorry, I meant non-prototype :)

If we wanted to go ahead and go with this implementation, some logic to move to ManageIQ authenticator and of course would need some specs here.

Alberto

abellotti:
@clearnight79 that looks about right.

running ```bundle show manageiq-api``` to get the gem's location, plus the path app/controller/api/base_controller/authentication.rb

jvlcek:
Running this on the appliance will return the path where the API code is installed:

`bundle info manageiq-api --path`

clearnight79:
I tried to enable UI OIDC following https://www.manageiq.org/docs/reference/latest/auth/openid_connect
but no success, keep getting 500 internal error after click the corporate login.
I know Redhat may have only tested with KeyCloak and the authentication module https://github.com/zmartzone/mod_auth_openidc did say it support many others like Azure, Gitlib, etc, but not MCM Pak IAM. Do you have any clue?

abellotti:
@jvlcek can we bother you to kick the tire with this PR ?  Just need to make sure it works for you too before we move forward with it.  Thanks.

/cc @Fryguy

Fryguy:
Seems @clearnight79 had some success.  So, I'm going to merge this as is, and we can iterate on cleanup, as I'd like to get this into a potential build for further testing.  I think moving forward we should...

a) Make a PR to "clean up" the code by moving this into the httpd authenticator (@jvlcek Can you take this on?)
b) Research moving this out of Ruby code and into mod_auth_openidc configuration (@jvlcek Can you also look into this after the clean up)
c) Discuss the `scope` attribute and implement as necessary (@abellotti Can you open an issue to track this?)

abellotti:
@Fryguy https://github.com/ManageIQ/manageiq/issues/19858 for the optional scope parameter.  /cc @jvlcek

abellotti:
@clearnight79 did this PR work as is for you or did you need to modify any of the OOTB configuration files 
- /etc/httpd/conf.d/manageiq-external-auth-openidc.conf
- /etc/httpd/conf.d/manageiq-remote-user-openidc.conf

And the related code here ?

simaishi:
Ivanchuk backport details:

```
$ git log -1
commit 5d482984334e606d74c72182fb07b5db33a7040b
Author: Jason Frey <jfrey@redhat.com>
Date:   Thu Feb 20 10:22:56 2020 -0500

    Merge pull request #737 from abellotti/support_oidc

    Adding support for OpenID-Connect/OAuth2 in the API

    (cherry picked from commit 95e86abb70c2a601f424e18f3d8d0f3d119c9d01)

    Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1805914
```

jvlcek:
FYI: I've created the following issues to track the above mentioned follow on work to this PR

- [Move API OpenID-Connect/OAuth2 support from API to ManageIQ authenticator.](https://github.com/ManageIQ/manageiq/issues/19865)
- [Research moving API OpenID-Connect/OAuth2 support into configuration.](https://github.com/ManageIQ/manageiq/issues/19866)

chessbyte:
Part of https://github.com/ManageIQ/manageiq/issues/19867

PR #742, opened 2020-02-19
Add compliances subcollection

d-m-u:
Per the convo in the main gitter room last night about api support of compliances [February 18, 2020 4:51 PM](https://gitter.im/ManageIQ/manageiq?at=5e4c5c6b5fd3f22ede92d754) I thought I may as well go ahead and open this.

### DIFF
--- a/app/controllers/api/vms_controller.rb
+++ b/app/controllers/api/vms_controller.rb
@@ -1,6 +1,7 @@
 module Api
   class VmsController < BaseController
     include Subcollections::Disks
+    include Subcollections::Compliances
     include Subcollections::Tags
     include Subcollections::Policies
     include Subcollections::PolicyProfiles
--- a/config/api.yml
+++ b/config/api.yml
@@ -745,6 +745,12 @@
         :identifier: ems_cluster_protect
       - :name: unassign
         :identifier: ems_cluster_protect
+  :compliances:
+    :description: Compliances
+    :options:
+    - :subcollection
+    :verbs: *g
+    :klass: Compliance
   :conditions:
     :description: Conditions
     :identifier: condition
@@ -3946,6 +3952,7 @@
     - :policies
     - :policy_profiles
     - :accounts
+    - :compliances
     - :custom_attributes
     - :security_groups
     - :software
@@ -4107,12 +4114,24 @@
       - :name: read
         :identifier:
         - vm_show
+    :compliances_subcollection_actions:
+      :get:
+      - :name: get
+        :identifier:
+        - vm_show
+        - vm_check_compliance
+        - instance_check_compliance
     :disks_subresource_actions:
       :get:
       - :name: read
         :identifier:
         - vm_show
     :cdroms_subresource_actions:
+      :get:
+      - :name: read
+        :identifier:
+        - vm_show
+    :compliances_subresource_actions:
       :get:
       - :name: read
         :identifier:
--- a/app/controllers/api/subcollections/compliances.rb
+++ b/app/controllers/api/subcollections/compliances.rb
@@ -1,0 +1,9 @@
+module Api
+  module Subcollections
+    module Compliances
+      def compliances_query_resource(object)
+        object.compliances
+      end
+    end
+  end
+end
--- a/spec/requests/compliances_spec.rb
+++ b/spec/requests/compliances_spec.rb
@@ -1,0 +1,83 @@
+RSpec.describe "Compliances API" do
+  let(:vm) { FactoryBot.create(:vm_vmware) }
+  let!(:compliance) { FactoryBot.create(:compliance, :resource => vm) }
+  let!(:compliance_detail) { FactoryBot.create(:compliance_detail, :compliance_id => compliance.id) }
+
+  describe "as a subcollection of VMs" do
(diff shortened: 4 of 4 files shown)


d-m-u:
@abellotti any chance I could get you to take a real quick look, please? thanks.

chessbyte:
@Real-Omar-Afifi want to try running with this PR to see if it gets you the info you want?

Fryguy:
I think @Real-Omar-Afifi also need compliance_details, but this can be added as a start and that as a follow up.

app/controllers/api/vms_controller.rb
```diff
@@ -1,6 +1,7 @@
 module Api
   class VmsController < BaseController
     include Subcollections::Disks
+    include Subcollections::Compliances
```
Fryguy: Minor, but can you make alphabetical?

d-m-u: yeah i was gonna do a follow-up for it

config/api.yml
```diff
@@ -4107,6 +4114,11 @@
       - :name: read
         :identifier:
         - vm_show
+    :compliances_subcollection_actions:
+      :get:
+      - :name: read
+        :identifier:
+        - vm_show
```
Fryguy: compliances are technically on a bunch of different resources.  @abellotti Is there a way to do an "OR" of identifiers?  Not sure what the right identifiers are, but I usually go with whatever the UI does... @himdel ?

abellotti: we can specify multiple identifiers (essentially the OR) via :identifiers instead of :identifier

Fryguy: Note, I'm more thinking of if this is a subcollection of more than just vms.

abellotti:
This is great @d-m-u, thanks for adding the tests that exercises the attributes=compliance_details. 👍

abellotti:
@d-m-u could you take care of the rubocop warnings ?  Thanks.

config/api.yml
himdel: There's a bunch of `*_check_compliance` features that the UI uses for the "Policy / Check Compliance of the last known configuration"  toolbar button.

So, I guess `instance_check_compliance`, `image_check_compliance`, `vm_check_compliance` and `miq_template_check_compliance` here?

config/api.yml
```diff
@@ -4107,6 +4114,14 @@
...
         :identifier:
         - vm_show
+    :compliances_subcollection_actions:
+      :get:
+      - :name: read
+        :identifiers:
+        - :klass: Vm
+          :identifier: vm_show
+          :identifier: vm_check_compliance
+          :identifier: instance_check_compliance
```
Fryguy: I don't think this is valid YAML...

Fryguy:
```ruby
s = <<~YAML
       - :klass: Vm
         :identifier: vm_show
         :identifier: vm_check_compliance
         :identifier: instance_check_compliance
YAML
YAML.load(s)
# => [{:klass=>"Vm", :identifier=>"instance_check_compliance"}]
```

Notice the keys are lost.

Fryguy: LOL...this is actually what rubocop found.

d-m-u: yeah, we don't really handle multiple identifiers all that well anyway as it turns out

abellotti:
@d-m-u try the following:

Move the subcollection action under the vms directly:

```
  :compliances_subcollection_actions:
    :get:
    - :name: get
      :identifier:
      - vm_show
      - vm_check_compliance
```

You can see such examples in the :vms: collection.

With this no need to class identify since it's per collection, and the :identifier: handles lists.

Thanks,
Alberto

config/api.yml
```diff
@@ -745,6 +745,12 @@
         :identifier: ems_cluster_protect
       - :name: unassign
         :identifier: ems_cluster_protect
+  :compliances:
+    :description: Compliances
+    :options:
+    - :subcollection
+    :verbs: *g
```
chessbyte: what does the value `*g` mean?

d-m-u: the verb set that can be used on the subcollection, in this case just the get

config/api.yml
d-m-u: thanks, I'll get to `miq_template_check_compliance` in a follow-up

config/api.yml
Fryguy: higher-level that is like a "pointer" in YAML, and it means to insert the object referenced with the name "g".  g is defined here: https://github.com/ManageIQ/manageiq-api/blob/a62b5f3bb439ce50cc009908e76737d05d122651/config/api.yml#L14-L15

simaishi:
Ivanchuk backport details:

```
$ git log -1
commit b5cecaef6fcd7d755fb76607b52f6004edfd743a
Date:   Mon Feb 24 09:22:00 2020 -0500

    Merge pull request #742 from d-m-u/adding_subcollection_for_compliance_details

    Add compliances subcollection

    (cherry picked from commit 309e5380171fd3bc8c0b3739c2d04adf6f7e46c0)

    Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1806660
```

PR #743, opened 2020-02-20
Call instance method for vm retire queueing

d-m-u:
Because of the fact that the entire retirement mechanism is built around a class method, when we introduced vm retirement queuing we should've also included an instance call to the class method. 

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1805119

Introduced in https://github.com/ManageIQ/manageiq-api/pull/662

@miq-bot assign @lpichler  

## Depends on: 
https://github.com/ManageIQ/manageiq/pull/19855

### DIFF
--- a/spec/requests/vms_spec.rb
+++ b/spec/requests/vms_spec.rb
@@ -1117,6 +1117,22 @@ describe "Vms API" do
           expect(response.parsed_body).to include(expected)
         end
 
+        it "queues retirement task" do
+          api_basic_authorize(action_identifier(:vms, :request_retire))
+          message = "VM id:#{vm.id} name:'#{vm.name}' request retire"
+          task_id = MiqTask.find_by(:name => message)&.id
+          expect(task_id).to be_nil
+          expect(MiqRequest.count).to eq(0)
+
+          post(vm_url, :params => gen_request(:request_retire))
+
+          task = MiqTask.find_by(:name => message)
+
+          MiqTask.find(task.id).miq_queue.deliver
+
+          expect(MiqQueue.count).to eq(1)
+        end
+
         it "to multiple Vms" do
           api_basic_authorize(collection_action_identifier(:vms, :request_retire))
           message_vm1 = "VM id:#{vm1.id} name:'#{vm1.name}' request retire"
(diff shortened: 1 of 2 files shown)


d-m-u:
@miq-bot add_label bug, blocker, ivanchuk/yes
hammer too, if we're still building it?

d-m-u:
Er, well, closing cause I'm an idiot and the name stay the same per Libor's review in the core PR so this doesn't need to change.

d-m-u:
"as such I'm not treating this a blocker to the current build"
@miq-bot remove_label blocker

lpichler:
this is looks good, test with backend PR 👍 

@d-m-u do you have idea how to cover it by test ?  If possible, it would be great to confirm that we run method from queue (by `MiqQueue.last.deliver`)  then the method `make_retire_request ` was called with correct parameters.
 
I think you can do it because api call for `retire_request` returns `task_id` and then you run method from queue.

`MiqTask.find(1040000000232613).miq_queue.deliver` and then check whether `make_retire_request ` was called.

Does make it sense for you ?

Thanks!

d-m-u:
https://travis-ci.com/ManageIQ/manageiq-cross_repo-tests/builds/150348522

spec/requests/vms_spec.rb
```diff
@@ -1117,6 +1117,22 @@ def update_raw_power_state(state, *vms)
...
+          expect(task_id).to be_nil
+          expect(MiqRequest.count).to eq(0)
+
+          post(vm_url, :params => gen_request(:request_retire))
+
+          task = MiqTask.find_by(:name => message)
+
+          MiqTask.find(task.id).miq_queue.deliver
+
+          expect(VmOrTemplate).to receive(:make_retire_request).once
```
kbrock: Is this actually `expect_any_instance_of(VmOrTemplate)`?

d-m-u: Yes. It needs to be calling https://github.com/ManageIQ/manageiq/blob/master/app/models/vm_or_template.rb#L351.

simaishi:
Ivanchuk backport details:

```
$ git log -1
commit e46a280ae9c0e173c9d9409367a8baa5f5cecb70
Date:   Wed Feb 26 15:02:18 2020 +0100

    Merge pull request #743 from d-m-u/change_queue_call_to_vm_retirement

    Call instance method for vm retire queueing

    (cherry picked from commit 2c70d200e88e5b06b1a4585682fe0de4a4be4806)

    https://bugzilla.redhat.com/show_bug.cgi?id=1805119
```

PR #751, opened 2020-02-24
Include provider ID in the validation task arguments if available

skateman:
This line is required to pull out the password if validating an existing provider.

### DIFF
--- a/app/controllers/api/providers_controller.rb
+++ b/app/controllers/api/providers_controller.rb
@@ -104,9 +104,10 @@ module Api
       end
     end
 
-    def verify_credentials_resource(_type, _id, data = {})
+    def verify_credentials_resource(_type, id = nil, data = {})
       klass = fetch_provider_klass(collection_class(:providers), data)
       zone_name = data.delete('zone_name')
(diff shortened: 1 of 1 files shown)


app/controllers/api/providers_controller.rb
```diff
@@ -104,9 +104,10 @@ def change_password_resource(type, id, data = {})
       end
     end
 
-    def verify_credentials_resource(_type, _id, data = {})
+    def verify_credentials_resource(_type, id = nil, data = {})
       klass = fetch_provider_klass(collection_class(:providers), data)
       zone_name = data.delete('zone_name')
+      data['id'] = id if id
```
chessbyte: should this be `data['id'] ||= id if id` to not overwrite a pre-existing value?

skateman: @chessbyte not really, the request comes in with a `data['id']` but that value is deleted and moved into the `id`. So I'm actually moving it back to the `data` for future re-use. There's no way that `data['id']` is set and even if that would be true, its value would always equal to `id`.

PR #753, opened 2020-02-25
Extend the existing provider editing endpoint with DDF support

skateman:
First of all, were not completely RESTful in this area, PUT/PATCH methods weren't supported on providers at all, so I fixed this. Then analogously with the [create endpoint](https://github.com/ManageIQ/manageiq-api/pull/723) I added a test against a `ddf` attribute. If this attribute is set, we're calling the new methods for editing a provider. If not, we're falling back to the old world and so keeping the compatibility.

### DIFF
--- a/app/controllers/api/providers_controller.rb
+++ b/app/controllers/api/providers_controller.rb
@@ -47,7 +47,12 @@ module Api
       raise BadRequestError, "Provider type cannot be updated" if data.key?(TYPE_ATTR)
 
       provider = resource_search(id, type, collection_class(:providers))
-      edit_provider(provider, data)
+
+      if data.delete(DDF_ATTR)
+        edit_provider_ddf(provider, data)
+      else
+        edit_provider(provider, data)
+      end
     end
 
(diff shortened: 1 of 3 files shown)


agrare:
Tested with an AWS provider, works great :+1:

app/controllers/api/providers_controller.rb
```diff
@@ -228,6 +233,12 @@ def create_provider_ddf(data)
       raise BadRequestError, "Could not create the new provider - #{err}"
     end
 
+    def edit_provider_ddf(provider, data)
+      provider.edit_with_params(data)
```
abellotti: does this also do what edit_provider does w.r.t. updating provider authentication and such ?

skateman: Yes, but differently. Each provider implements it on its own as the incoming data is in a different (DDF) format.

abellotti:
ping @lpichler, can we borrow your 👀 when you get a chance ?  Thanks.

PR #754, opened 2020-02-25
Create code owners for manageiq-api repo

gtanzillo:


.github/CODEOWNERS
```diff
@@ -0,0 +1,5 @@
+# Lines starting with '#' are comments.
+# Each line is a file pattern followed by one or more owners.
+
+# These owners will be the default owners for everything in the repo.
+*                                               @bdunne  @abellotti
```
chessbyte: why so many spaces before the names?

Fryguy: To keep us in suspense!

gtanzillo: It's not that exciting, I copy/pasted the top part from the one in the manageiq repo. I'll fix it.

Issue #745, opened 2020-02-20
/api/vms/{id}/policies does not return the policies in the policy profiles assigned to the VM


gtanzillo:
To reproduce:

1. Create a policy and add it to a profile
2. Assign the policy to a VM
3. Request to VM policies from the API -

http://localhost:3000/api/vms/10000000000112/policies

The result has no policy references -

```json
{
    "name": "policies",
    "count": 16,
    "subcount": 0,
    "pages": 1,
    "resources": [],
    "actions": [
        {
            "name": "assign",
            "method": "post",
            "href": "http://localhost:3000/api/vms/10000000000112/policies"
        },
        {
            "name": "unassign",
            "method": "post",
            "href": "http://localhost:3000/api/vms/10000000000112/policies"
        },
        {
            "name": "resolve",
            "method": "post",
            "href": "http://localhost:3000/api/vms/10000000000112/policies"
        }
    ],
    "links": {
        "self": "http://localhost:3000/api/vms/10000000000112/policies?offset=0",
        "first": "http://localhost:3000/api/vms/10000000000112/policies?offset=0",
        "last": "http://localhost:3000/api/vms/10000000000112/policies?offset=0"
    }
}
```

d-m-u:
@miq-bot assign @d-m-u

d-m-u:
Ugh, fine. AB's right as usual. The UI add is a policy_profile add, and 

http://localhost:3000/api/vms/#/policy_profiles shows the correct profiles. I don't have the power to close issues so maybe GT wants to verify this and then he can close it?

chessbyte:
@gtanzillo ^

gtanzillo:
I discussed this with Alberto and this would be the correct way to get the actual policies assigned to a resource -

http://localhost:3000/api/vms/10000000000229/policy_profiles?expand=resources&attributes=miq_policies

Unfortunately, this does not currently work either and needs to be fixed

chessbyte:
Part of https://github.com/ManageIQ/manageiq-api/issues/782

PR #756, opened 2020-02-26
[WIP] Change collection class to return policy sets

d-m-u:
/api/vms/{id}/policies does not return the policies in the policy profiles assigned to the VM

but it would, if we had the right collection class
Fixes https://github.com/ManageIQ/manageiq-api/issues/745

WIP because it needs a spec

### DIFF
--- a/app/controllers/api/subcollections/policies.rb
+++ b/app/controllers/api/subcollections/policies.rb
@@ -27,8 +27,8 @@ module Api
       end
 
       def object_policies(object)
-        policy_klass = collection_class(:policies)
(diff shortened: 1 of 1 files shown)


abellotti:
policies are supposed to return policies, policy_profiles is the one intended for policy profiles.

See https://www.manageiq.org/docs/reference/latest/api/reference/policies

We expose both:  /api/:collection/:id/policies and /api/:collection/:id/policy_profiles

d-m-u:
Good try though.

PR #758, opened 2020-02-27
Remove ability to delete a region

djberg96:
Apparently allowing regions to be deleted was a Bad Idea. QE realized that deleting a region could break the app:

https://bugzilla.redhat.com/show_bug.cgi?id=1805844#c3

"Deleting the Region works too, but once we delete the only region on the appliance, the appliance crashes"

As per @carbonin:

"A region generally represents a database/installation of miq. The only reason you would have more than one would be if you are in the global region and those will be deleted when the subscription to the remote is removed. There is no other use case for deleting a region."

So, this PR removes the ability to delete a region.

### DIFF
--- a/config/api.yml
+++ b/config/api.yml
@@ -2656,7 +2656,7 @@
     - :settings
     :options:
     - :collection
-    :verbs: *gpd
+    :verbs: *gp
     :klass: MiqRegion
     :collection_actions:
       :get:
@@ -2667,8 +2667,6 @@
         :identifier: region
       - :name: edit
         :identifier: region_edit
-      - :name: delete
-        :identifier: region_delete
     :resource_actions:
       :get:
       - :name: read
@@ -2676,11 +2674,6 @@
       :post:
       - :name: edit
         :identifier: region_edit
-      - :name: delete
-        :identifier: region_delete
-      :delete:
-      - :name: delete
-        :identifier: region_delete
   :reports:
     :description: Reports
     :identifier: miq_report
--- a/spec/requests/regions_spec.rb
+++ b/spec/requests/regions_spec.rb
@@ -113,40 +113,10 @@ RSpec.describe "Regions API", :regions do
   end
 
   context "delete", :delete do
-    it "can delete a region with POST" do
-      api_basic_authorize action_identifier(:regions, :delete)
-      region = FactoryBot.create(:miq_region)
-
-      expect { post api_region_url(nil, region), :params => gen_request(:delete) }.to change(MiqRegion, :count).by(-1)
-      expect_single_action_result(:success => true, :message => /#{region.id}/)
-    end
-
-    it "can delete a region with DELETE" do
-      api_basic_authorize action_identifier(:regions, :delete)
-      region = FactoryBot.create(:miq_region)
-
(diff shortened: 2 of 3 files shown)


djberg96:
@miq-bot add_label bug

config/api.yml
```diff
@@ -2667,20 +2667,13 @@
...
         :identifier: region_edit
-      - :name: delete
-        :identifier: region_delete
     :resource_actions:
       :get:
       - :name: read
         :identifier: region
       :post:
       - :name: edit
         :identifier: region_edit
```
carbonin: Just as an aside, we can't edit the region's `region_id` attribute, correct?

spec/requests/regions_spec.rb
```diff
@@ -113,40 +113,10 @@
...
-
-    it "forbids deletion of a region without an appropriate role" do
-      expect_forbidden_request do
-        region = FactoryBot.create(:miq_region, :description => "Current Region description")
-        delete(api_region_url(nil, region))
-      end
+    it "forbids deletion of a region" do
+      region = FactoryBot.create(:miq_region, :description => "Current Region description")
+      delete(api_region_url(nil, region))
+      expect(response).to have_http_status(:not_found)
```
bdunne: 404 seems like the wrong response.  404 is meant for resources not being found, I think this should probably be 405 "Method not allowed"

carbonin:
cc @simaishi We need this for the build

config/api.yml
djberg96: It can't edit the `'id` field, but currently the `region` field is editable. I don't see a `region_id` field.

carbonin: Yeah, that's the one. Really the only thing that should be editable in this table is the `description`

spec/requests/regions_spec.rb
djberg96: @bdunne, @abellotti - Is there a generic way to do this? I don't see a 405 raised anywhere else, though it makes sense.

abellotti: When an action is no longer exposed from the api.yml, the API will not create the route for it.  So the 404 is what's coming back from rails, not the api.

bdunne: Ok, not going to have that hold up this PR.

config/api.yml
carbonin: @djberg96 can you add the other attributes to `INVALID_REGIONS_ATTRS` as a part of this change?

djberg96: @carbonin ok, updated.

abellotti:
Just curious, with multiple regions, is the ability to delete one region not available in the classic ui ?

carbonin:
I don't believe the other regions are visible at all.

simaishi:
Ivanchuk backport details:

```
$ git log -1
commit 85bc2f7a60a8a306245d89f6c57aab145cf82890
Date:   Thu Feb 27 15:42:28 2020 -0500

    Merge pull request #758 from djberg96/regions_post_actions2

    Remove ability to delete a region

    (cherry picked from commit 1343e58a4654c1c27c8de2cb057ed66eac74aeff)

    Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1805844
```

PR #761, opened 2020-03-03
Differentiate between ems_cloud and infra for rbac purposes for providers collection

d-m-u:
You can add cloud providers without the rbac permissions to do so because we don't have a provider rbac check at the subclass level. 

@miq-bot add_label bug 
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1766647 

~~Requires a follow-up refactor to clean up the added duplication since this check is present both on providers and requests now. For the purposes of this bug, I'd like to vote this gets merged as is and that I do the refactoring later.~~ just kidding

### DIFF
--- a/app/controllers/api/providers_controller.rb
+++ b/app/controllers/api/providers_controller.rb
@@ -125,6 +125,16 @@ module Api
 
     private
 
+    def authorize_provider(typed_provider_klass)
+      create_action = collection_config["providers"].collection_actions.post.detect { |a| a.name == "create" }
+      provider_spec = create_action.identifiers.detect { |i| i.klass.constantize.name == typed_provider_klass.superclass.name }
+      raise BadRequestError, "Unsupported request class #{typed_provider_klass}" if provider_spec.blank?
+
+      if provider_spec.identifier && !api_user_role_allows?(provider_spec.identifier)
+        raise ForbiddenError, "Create action is forbidden for #{typed_provider_klass} requests"
+      end
+    end
+
     def provider_options(type)
       klass = type.safe_constantize
 
@@ -219,12 +229,15 @@ module Api
     def create_provider(data)
       provider_klass = fetch_provider_klass(collection_class(:providers), data)
       create_data    = fetch_provider_data(provider_klass, data, :requires_zone => true)
-      provider       = provider_klass.create!(create_data)
-      update_provider_authentication(provider, data)
-      provider
-    rescue => err
-      provider.destroy if provider
+      authorize_provider(provider_klass)
+      begin
+        provider = provider_klass.create!(create_data)
+        update_provider_authentication(provider, data)
+        provider
+      rescue => err
+        provider&.destroy
       raise BadRequestError, "Could not create the new provider - #{err}"
+      end
     end
 
     def create_provider_ddf(data)
--- a/config/api.yml
+++ b/config/api.yml
@@ -2384,7 +2384,17 @@
         - ems_infra_show_list
         - ems_block_storage_show_list
       - :name: create
-        :identifier: ems_infra_new
+        :identifiers:
+        - :klass: ManageIQ::Providers::InfraManager
+          :identifier: ems_infra_new
+        - :klass: ManageIQ::Providers::CloudManager
+          :identifier: ems_cloud_new
+        - :klass: ManageIQ::Providers::ContainerManager
+          :identifier: ems_container_new
+        - :klass: ManageIQ::Providers::NetworkManager
+          :identifier: ems_network_new
+        - :klass: Provider
+          :identifier: provider_foreman_add_provider
       - :name: edit
         :identifier: ems_infra_edit
       - :name: refresh
--- a/spec/requests/providers_spec.rb
+++ b/spec/requests/providers_spec.rb
@@ -415,7 +415,7 @@ describe "Providers API" do
     end
 
     it 'creates valid foreman provider' do
-      api_basic_authorize collection_action_identifier(:providers, :create)
+      api_basic_authorize "provider_foreman_add_provider"
 
       post(api_providers_url + '?provider_class=provider', :params => gen_request(:create, sample_foreman))
 
@@ -435,10 +435,10 @@ describe "Providers API" do
       get api_providers_url, :params => { :provider_class => 'provider' }
 
       expected = {
-        'resources' => [
-          {'href' => "#{api_provider_url(nil, provider)}?provider_class=provider"}
-        ],
-        'actions'   => [a_hash_including('href' => a_string_including('?provider_class=provider'))]
+        'resources' => [{'href' => "#{api_provider_url(nil, provider)}?provider_class=provider"}],
+        'actions'   => a_collection_including(
+          a_hash_including('href' => a_string_including('?provider_class=provider'))
+        )
       }
       expect(response).to have_http_status(:ok)
       expect(response.parsed_body).to include(expected)
@@ -500,7 +500,7 @@ describe "Providers API" do
     it 'allows provider specific attributes to be specified' do
       allow(ManageIQ::Providers::Azure::CloudManager).to receive(:api_allowed_attributes).and_return(%w(azure_tenant_id))
       tenant = FactoryBot.create(:cloud_tenant)
-      api_basic_authorize collection_action_identifier(:providers, :create)
+      api_basic_authorize "ems_cloud_new"
 
       post(api_providers_url, :params => { "type"            => "ManageIQ::Providers::Azure::CloudManager",
                                            "name"            => "sample azure provider",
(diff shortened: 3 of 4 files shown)


d-m-u:
@jrafanie please review

config/api.yml
```diff
@@ -2336,7 +2336,17 @@
         - ems_infra_show_list
         - ems_block_storage_show_list
       - :name: create
-        :identifier: ems_infra_new
+        :identifiers:
+        - :klass: EmsInfra
+          :identifier: ems_infra_new
+        - :klass: EmsCloud
+          :identifier: ems_cloud_new
+        - :klass: Provider #EmsContainer
```
d-m-u: Unlike ems_network, we don't have an alias in core for this. It'd be cool if we did, I guess I'm about to find out if reviewers hate it or not. The alternative that's here now I don't like because it doesn't keep a one-to-one mapping in core for product identifiers.

config/api.yml
```diff
@@ -2336,7 +2336,17 @@
...
+          :identifier: ems_container_new
+        - :klass: EmsNetwork
+          :identifier: ems_network_new
+        - :klass: Provider
+          identifier: provider_foreman_add_provider
```
bdunne: missing the leading `:` here

d-m-u: That's fine, since this relies on the aliases that you aren't a fan of anyway I doubt this is getting in ever.

config/api.yml
```diff
@@ -2336,7 +2336,17 @@
         - ems_infra_show_list
         - ems_block_storage_show_list
       - :name: create
-        :identifier: ems_infra_new
+        :identifiers:
+        - :klass: EmsInfra
```
bdunne: We should probably use the real class names here, not the aliases

d-m-u: You want to put in actual provider class names into this file? I hate that idea. The point is that we can't differentiate at the moment between the cloud and infra side so unless we use these... do you have a better idea?

config/api.yml
```diff
@@ -2336,7 +2336,17 @@
...
+          :identifier: ems_infra_new
+        - :klass: EmsCloud
+          :identifier: ems_cloud_new
+        - :klass: Provider #EmsContainer
+          :identifier: ems_container_new
+        - :klass: EmsNetwork
+          :identifier: ems_network_new
+        - :klass: Provider
```
bdunne: `Provider` is in the list twice and with the `detect` above, I dont think you'd ever get to this one. :disappointed:

d-m-u: I'm aware of that. But like I said, this isn't getting merged anyway, because it relies on the aliases and you nixed that.

config/api.yml
bdunne: No, I meant `ManageIQ::Providers::InfraManager` rather than `EmsInfra` and `ManageIQ::Providers::CloudManager` rather than `EmsCloud`.  These are the official classes in core.

config/api.yml
bdunne: :+1: Nice, thanks!

d-m-u: I still don't like this just being Provider

bdunne: Yeah, this one is special. https://www.manageiq.org/docs/reference/latest/api/reference/providers#foreman-support

Soon enough you'll be able to add providers of other types.  Maybe we should create a new generic "provider_add" under the API section of the product features YAML?  I'm okay with the way it is for this BZ.  @abellotti Thoughts?

abellotti: Yea, that makes sense.  Was thinking of the following for the API section, which will match closely to the api.yml but in MiqProductFeature lingo:

```
# Providers 
- :name: Providers
  :description: Everything under Providers
  :feature_type: node 
  :children:
  - :name: Create a new Provider
    :description: Create a new Provider
    :feature_type: admin
    :children:
    - :name: Create a new ManageIQ::Providers::InfraManager
      :description: Create a new ManageIQ Infrastructure Manager Provider
      :feature_type: admin
      :identifier: ems_infra_new
    - :name: Create a new ManageIQ::Providers::CloudManager
      :description: Create a new ManageIQ Cloud Manager Provider
      :feature_type: admin
      :identifier: ems_cloud_new
 ... 
```

We can iterate/refine above as needed.

spec/requests/providers_spec.rb
```diff
@@ -500,7 +500,7 @@ def have_endpoint_attributes(expected_hash)
     it 'allows provider specific attributes to be specified' do
       allow(ManageIQ::Providers::Azure::CloudManager).to receive(:api_allowed_attributes).and_return(%w(azure_tenant_id))
       tenant = FactoryBot.create(:cloud_tenant)
-      api_basic_authorize collection_action_identifier(:providers, :create)
+      api_basic_authorize "ems_cloud_new"
```
abellotti: let's add a helper in spec/support/api/helpers.rb so we don't hard code these around and sprouting dups over time.
Maybe something like:
```
collection_action_classed_identifier(type, action, method, klass)
```

with this, above can be replaced with:

```
api_basic_authorize collection_action_classed_identifier(:providers, :create, :post, "ManageIQ::Providers::CloudManager")
```

d-m-u: @abellotti any chance that could be follow-up? I'd like to close the bug and then refactor if you're okay with that.

d-m-u: Juuuuuust kidding.

spec/support/api/helpers.rb
```diff
@@ -40,6 +40,13 @@ def stub_api_action_role(collection, action_type, method, action, identifier)
           allow(::Api::ApiConfig.collections[collection][action_type]).to receive(method) { updated_method }
         end
 
+        def collection_action_classed_identifier(type, action, method, klass)
```
d-m-u: I don't believe this needs the action actually but AB said to use this signature n before this gets merged it'd be great to have someone else's eyes on it

spec/requests/providers_spec.rb
abellotti: Thanks @d-m-u for adding the helper and updating the specs. 👍

app/controllers/api/providers_controller.rb
```diff
@@ -217,12 +227,17 @@ def fetch_provider_klass(klass, data)
     def create_provider(data)
       provider_klass = fetch_provider_klass(collection_class(:providers), data)
       create_data    = fetch_provider_data(provider_klass, data, :requires_zone => true)
-      provider       = provider_klass.create!(create_data)
-      update_provider_authentication(provider, data)
-      provider
-    rescue => err
-      provider.destroy if provider
+      ##
+      authorize_provider(data["type"].constantize)
```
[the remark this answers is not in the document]
bdunne: Sorry for just noticing this, but is `provider_klass` the same as `data["type"].constantize`?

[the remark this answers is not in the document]
d-m-u: this method's called from 232: `authorize_provider(data["type"].constantize)` so yes?

spec/support/api/helpers.rb
bdunne: I think `action` will be needed in the future when we need to use this to bulk delete or run other operations that are not as simple as single instance CRUD.  Is that correct @abellotti ?

abellotti: yes, just discussed with @d-m-u, we need the action:

```
::Api::ApiConfig.collections[type][:collection_actions][method].detect { |spec| spec[:name] == action }[:identifiers].detect { |spec| spec[:klass] == klass.to_s }
```

app/controllers/api/providers_controller.rb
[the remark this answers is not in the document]
bdunne: Then we can get rid of the `constantize` and change this to `authorize_provider(provider_klass)`?

simaishi:
Jansa backport details:

```
$ git log -1
commit 062e4101a857167c3186843ffba8fa2e86dc1cde
Date:   Tue Mar 10 13:20:32 2020 -0400

    Merge pull request #761 from d-m-u/fixing_provider_actions_for_cloud_provs

    Differentiate between ems_cloud and infra for rbac purposes for providers collection

    (cherry picked from commit 011c38f6f4a1c9e2ebd233bae8d1296275c15252)

    https://bugzilla.redhat.com/show_bug.cgi?id=1766647
```

simaishi:
Ivanchuk backport details:

```
$ git log -1
commit 92ba373e8e536808a60491c06b12500bd996920a
Date:   Tue Mar 10 13:20:32 2020 -0400

    Merge pull request #761 from d-m-u/fixing_provider_actions_for_cloud_provs

    Differentiate between ems_cloud and infra for rbac purposes for providers collection

    (cherry picked from commit 011c38f6f4a1c9e2ebd233bae8d1296275c15252)

    https://bugzilla.redhat.com/show_bug.cgi?id=1766647
```

PR #767, opened 2020-03-05
[WIP] Add ARCHITECTURE.md

NickLaMuro:
This is a replacement PR for https://github.com/ManageIQ/guides/pull/318 which creates the `ARCHITECTURE.md` as discussed in the aforementioned PR.

Mostly pushing what was there previously, which still needs some updating and tweaks, but with this we can finally close that stale and rotting one on the guides repo...

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -41,7 +41,10 @@ with is the [`app/controllers/api/base_controller.rb`][] class, and included
 modules in it's associated directory.  Almost all (if not all), controllers in
 the API inherit from this controller, and the entry point for each action
 defined in the routes (more on that in a later section), are defined in this
-class
+class.
+
+(Exceptions to this are controllers like the [`Api::PingController`][], which
+just return plain text to confirm the application endpoints are responding)
 
 
 ### The Basics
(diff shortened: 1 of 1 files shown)


NickLaMuro:
FYI:  After not looking at this for over a year and trying to updated what I had started from back then, it feels like I am horribly unqualified to be commenting about the architecture and design of the ManageIQ API

So if you have anything to add or want to correct any horrible inaccuracies you have noticed, feel free to leave a comment.  Or even better and if you feel so inclined, create a PR off of this one or one entirely new that starts entirely from scratch or copies what I have with no attribution.  You don't have to worry about bruising my ego for finishing what I meekly started over a year ago and failed to complete in that time frame.

jvlcek:
Just push it! Fix it later. Better to have  _**something**_ then let this valuable data linger...

bdunne:
Sorry, I missed the WIP.  Is there anything that was left to be done?  I assume this will be a living document.

NickLaMuro:
Mostly just that I hate having a dirty git history... but no, nothing _**has**_ to be done, so it is fine for me (or others) to do followups.

PR #768, opened 2020-03-06
Remove hawkular specs

d-m-u:
Do we need these? I thought it was deprecated, maybe I'm wrong.


Fryguy:
Hawkular as a standalone middleware provider was removed, but Hawkular as a part of a container provider is still a thing.  I think these specs are for the latter and thus this should not be removed.

PR #769, opened 2020-03-06
[V2V] ConversionHost - Validate resource is not already a conversion host

fdupont-redhat:
When one tries to add a conversion host by REST API while there is an existing conversion host with the same resource, the configuration process is executed, then the ConversionHost object creation fails because the resource is already associated to another conversion host. There are two wrong consequences to that:

1. The conversion host resource is configured a second time, breaking potentially the existing configuration and blocking migration.
2. There's no error message apart from log/evm.log.

This does not affect users configuring conversion hosts through the UI, as the list of resources is filtered. So, it is limited to API.

This pull request adds a validation at the API level that returns a BadRequestError (HTTP 400) if the resource is already used by a conversion host. This fails faster and doesn't trigger the configuration phase.

RHBZ: https://bugzilla.redhat.com/show_bug.cgi?id=1810555

### DIFF
--- a/app/controllers/api/conversion_hosts_controller.rb
+++ b/app/controllers/api/conversion_hosts_controller.rb
@@ -48,6 +48,10 @@ module Api
 
       raise BadRequestError, "unsupported resource_type #{resource_type}" unless resource.supports_conversion_host?
 
+      conversion_host = ConversionHost.where(:resource => resource)
+
+      raise BadRequestError, "the resource '#{resource.name}' is already configured as a conversion host" unless conversion_host.empty?
+
       data['resource'] = resource
 
       api_action(type, id) do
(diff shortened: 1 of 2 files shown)


fdupont-redhat:
@miq-bot add-label v2v, enhancement, ivanchuk/yes

djberg96:
:+1:

app/controllers/api/conversion_hosts_controller.rb
```diff
@@ -48,6 +48,10 @@ def create_resource(type, id, data)
 
       raise BadRequestError, "unsupported resource_type #{resource_type}" unless resource.supports_conversion_host?
 
+      conversion_host = ConversionHost.where(:resource => resource)
+
+      raise BadRequestError, "the resource '#{resource.name}' is already configured as a conversion host" unless conversion_host.empty?
```
bdunne: This really feels like it should be a validation on the model.

fdupont-redhat: Like this? https://github.com/ManageIQ/manageiq/pull/19945

bdunne: :+1: Thanks!

fdupont-redhat:
Closing in favor of https://github.com/ManageIQ/manageiq/pull/19945.

PR #771, opened 2020-03-06
[WIP] Fix specs for subclassing of service orders

d-m-u:
We added ServiceOrder sub-classes and the API specs need to reflect all that...

I'm opening this as WIP cause there's a couple issues that have to get addressed first:

[create_service_order is using the base class](https://github.com/ManageIQ/manageiq-api/blob/9d146dc4207a3482a60e74e01d227ca46236bf18/app/controllers/api/service_orders_controller.rb#L64) which it can't now and I guess we have to pass the type in the call to get the class

[the api says `the default state for a service order is 'cart'`](https://github.com/ManageIQ/manageiq-api/blob/ce6ad2837d61acf9bcbc75792f7796611fc74def/spec/requests/service_orders_spec.rb#L100) so, I dunno, the [factory shouldn't maybe be creating things with the default set to ordered](https://github.com/ManageIQ/manageiq/blob/ec56ab47c4baa27b12cd00d67e49e8f2a1109321/spec/factories/service_order.rb#L11)

### DIFF
--- a/spec/requests/service_orders_spec.rb
+++ b/spec/requests/service_orders_spec.rb
@@ -1,6 +1,6 @@
 RSpec.describe "service orders API" do
   it "can list all service orders" do
-    service_order = FactoryBot.create(:shopping_cart, :user => @user)
+    service_order = FactoryBot.create(:service_order_cart, :user => @user)
     api_basic_authorize collection_action_identifier(:service_orders, :read, :get)
 
     get api_service_orders_url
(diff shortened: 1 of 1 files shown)


d-m-u:
Closed because all the things that this was fixing was covered indeed as Brandon said by https://github.com/ManageIQ/manageiq/pull/19937. That being said, the API specs are still red and for reason I said at the description at the top of this PR. So since Brandon thinks that the factory change was enough for this, maybe he can help with the other issues.

PR #773, opened 2020-03-06
ServiceOrder base class is no longer valid, prefer ServiceOrderCart

bdunne:
See https://github.com/ManageIQ/manageiq/pull/19795/files#diff-f5aafc4dc4e7284f971559aef2bb6a64R21

### DIFF
--- a/app/controllers/api/service_orders_controller.rb
+++ b/app/controllers/api/service_orders_controller.rb
@@ -61,7 +61,7 @@ module Api
     end
 
     def create_service_order(data)
-      ServiceOrder.new(data).tap do |service_order|
+      ServiceOrderCart.new(data).tap do |service_order|
         service_order.assign_user
         service_order.save!
       end
--- a/spec/requests/service_orders_spec.rb
+++ b/spec/requests/service_orders_spec.rb
@@ -44,7 +44,7 @@ RSpec.describe "service orders API" do
     expected = {
       'error' => a_hash_including(
         'kind'    => 'bad_request',
-        'message' => /Validation failed: ServiceOrder: State has already been taken/
+        'message' => /Validation failed: ServiceOrderCart: State has already been taken/
       )
     }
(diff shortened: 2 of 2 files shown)


app/controllers/api/service_orders_controller.rb
```diff
@@ -61,7 +61,7 @@ def add_request_to_cart(workflow)
     end
 
     def create_service_order(data)
-      ServiceOrder.new(data).tap do |service_order|
+      ServiceOrderCart.new(data).tap do |service_order|
```
d-m-u: We can't do this, what if this is used by v2v?

bdunne: @fdupont-redhat Does V2V use this endpoint?

gmcculloug: This method is only called when `service_requests` is blank above.  See https://github.com/ManageIQ/manageiq-api/blob/master/app/controllers/api/service_orders_controller.rb#L10-L11

The only caller for this is the SUI when it wants to create an empty cart.  Or someone calling the API directly for cart operations, it is not used by v2v.

v2v and OpsUI service ordering will create a new request and the ServiceOrder is a by-product of that operation.  So it is safe to assume that this method always wants to create a `ServiceOrderCart` object.

bdunne: Cool, thanks for the clarification @gmcculloug

simaishi:
Jansa backport details:

```
$ git log -1
commit 714b51a165ba061d2f532626fcf19ff2e96c34e2
Date:   Fri Mar 6 17:54:20 2020 -0500

    Merge pull request #773 from bdunne/service_order_cart

    ServiceOrder base class is no longer valid, prefer ServiceOrderCart

    (cherry picked from commit cd849c16925bd83dc34ed05ede096718d060be5d)
```

PR #774, opened 2020-03-10
Add helper method for multiple subcollection_action_identifiers

d-m-u:
The helper methods collection_action_identifier and subcollection_action_identifier in spec/support/api/helpers.rb do not support identifiers listed in an :identifiers block in
the config yaml.

I'll link the issue after AB looks at it and gives the 👍 or something.

### DIFF
--- a/spec/support/api/helpers.rb
+++ b/spec/support/api/helpers.rb
@@ -41,10 +41,7 @@ module Spec
         end
 
         def collection_action_classed_identifier(type, action, method, klass)
-          ::Api::ApiConfig
-            .collections[type][:collection_actions][method]
-            .detect { |spec| spec[:name] == action.to_s }[:identifiers]
-            .detect { |spec| spec[:klass] == klass.to_s }[:identifier]
+          identifier_from_multiple_class_list(::Api::ApiConfig.collections[type][:collection_actions][method], action, klass)
         end
 
         def action_identifier(type, action, selection = :resource_actions, method = :post)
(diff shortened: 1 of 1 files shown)


spec/support/api/helpers.rb
```diff
@@ -50,6 +50,13 @@ def collection_action_identifier(type, action, method = :post)
           action_identifier(type, action, :collection_actions, method)
         end
 
+        def collection_action_identifiers(type, action, method = :post, klass)
```
abellotti: dup method name as 49, shouldn't this be collection_action_classed_identifier ?

abellotti: also, wasn't that just merged ? Maybe a rebase will pick it up.

d-m-u: Erps yeah sorry don't look for a minute

spec/support/api/helpers.rb
```diff
@@ -70,6 +70,21 @@ def subcollection_action_identifier(type, subtype, action, method = :post)
...
         end
 
+        def subcollection_action_classed_identifiers(type, subtype, action, method = :post, klass)
+          subtype_actions = "#{subtype}_subcollection_actions".to_sym
+          if ::Api::ApiConfig.collections[type][subtype_actions]
+            ::Api::ApiConfig
+              .collections[type][subtype_actions][method]
+              .detect { |spec| spec[:name] == action.to_s }[:identifiers]
+              .detect { |spec| spec[:klass] == klass.to_s }[:identifier]
+          else
```
d-m-u: Okay, fine, I actually have no idea what is supposed to happen if there aren't subtype actions present for multiple ids, my initial idea is that we probably don't need the else at all but who even knows.

abellotti: you do need the else 👍  subcollection actions/identifier can be declared on the collection itself (i.e. a subcollection action's identifiers are defined per collection), or generically as part of the subcollection itself (the else case).

btw, it there  way to generalize the following pattern in a method, I see it twice here plus in the collection_action_classed_identifier.

```
              .collections[type][action_type][method]
              .detect { |spec| spec[:name] == action.to_s }[:identifiers]
              .detect { |spec| spec[:klass] == klass.to_s }[:identifier]
```

d-m-u: Yeah, I was gonna get to that in refactoring, I needed to close the bug first though.

abellotti:
Thanks @d-m-u for the new helpers!! 👍 

I'll merge once unwipped.

PR #775, opened 2020-03-11
OPTIONS hosts, clusters - remove node_type from data

himdel:
In https://github.com/ManageIQ/manageiq/pull/19953,
removing the `Host.node_types` and `EmsCluster.node_types` methods.

These would return `:non_openstack`, `:mixed_clusters`/`:mixed_hosts`, or `:openstack`.

For backward compatibility, hardcoding to the "mixed" version.


himdel:
@abellotti I don't know of anything using this, but I assume we want to keep the endpoint for compatibility?

abellotti:
What endpoint do you mean ?    the /api/clusters and /api/hosts, or you probably mean the node_types section in the OPTIONS result data:


```
  ...
  "subcollections": [
    "lans",
    "policies",
    "policy_profiles",
    "tags"
  ],
  "data": {
    "node_types": "mixed_clusters"
  }
```

I'm not sure either if someone other than UI is using OPTIONS /api/clusters or /api/hosts. 
it's ok to keep this hard coded here, I just don't know how the value is used and  if value is non_openstack, or something else what would be done differently.

I'm kind'a leaning on having it removed from here, but don't know for sure. 
/cc @Fryguy thoughts ?

abellotti:
I think OPTIONS code for clusters and node with node_type added by @AparnaKarve , Aparna, do you know the original intent/use of this option data ? Thanks.

himdel:
Updated to remove these `OPTIONS /api/hosts` and `OPTIONS /api/clusters` endpoints altogether :).

simaishi:
Jansa backport details:

```
$ git log -1
commit 69ab5c946a56c754aa846b8439e86eb3c7c0ab6d
Date:   Fri Mar 13 09:56:21 2020 -0400

    Merge pull request #775 from himdel/node-deployment-roles

    OPTIONS hosts, clusters - remove node_type from data

    (cherry picked from commit 42e8d9840e5954518b89420710640b384170ba81)
```

PR #777, opened 2020-03-17
force updated service template photo encoding

d-m-u:
the content encoding for service template pictures is thus: `"content": "<base 64 encoded image here>"` which means that before we run any backend method to create pictures we should make sure that the content is decoded properly since all main does is a `new` which won't do anything with the encoding.

otherwise getting the image back via the image_href returns not an image but base64 encoded content

also please note there's existing UI bug where if you do this with a jpg the picture won't show up correctly (we allow both png and jpg and the pngs show up okay s far as i can tell)

it's for https://bugzilla.redhat.com/show_bug.cgi?id=1770197

@miq-bot add_label bug
(hammer?)

### DIFF
--- a/app/controllers/api/service_templates_controller.rb
+++ b/app/controllers/api/service_templates_controller.rb
@@ -21,6 +21,7 @@ module Api
 
     def edit_resource(type, id, data)
       catalog_item = resource_search(id, type, collection_class(:service_templates))
+      decode_picture(data) if data["picture"]
       catalog_item.update_catalog_item(data.deep_symbolize_keys, User.current_user.userid)
     rescue => err
       raise BadRequestError, "Could not update Service Template - #{err}"
@@ -51,6 +52,10 @@ module Api
 
     def set_additional_attributes
       @additional_attributes = %w(config_info)
+    end
+
+    def decode_picture(data)
+      data["picture"]["content"] = Base64.strict_decode64(data["picture"]["content"])
     end
   end
 end
--- a/spec/requests/service_templates_spec.rb
+++ b/spec/requests/service_templates_spec.rb
@@ -134,6 +134,14 @@ describe "Service Templates API" do
         }
       }
     end
+    let(:updated_picture) do
+      {
+        :picture => {
+          :content   => "aW1hZ2U=",
+          :extension => "jpg"
+        }
+      }
+    end
 
     it "rejects requests without appropriate role" do
       api_basic_authorize
@@ -152,6 +160,17 @@ describe "Service Templates API" do
 
       expect_single_resource_query("id" => st.id.to_s, "href" => api_service_template_url(nil, st), "name" => "Updated Template Name")
       expect(st.reload.name).to eq("Updated Template Name")
+    end
+
+    it "supports picture update with base 64 decoded content" do
+      api_basic_authorize collection_action_identifier(:service_templates, :edit)
+
+      st = FactoryBot.create(:service_template)
+      post(api_service_template_url(nil, st), :params => gen_request(:edit, updated_picture))
+
+      st.reload
+      expect(st.picture.md5).to eq("78805a221a988e79ef3f42d7c5bfd418")
+      expect(st.picture.content).to eq("image")
     end
 
     it "supports edits of multiple resources" do


app/controllers/api/service_templates_controller.rb
```diff
@@ -52,5 +53,9 @@ def unarchive_resource(type, id, _data)
     def set_additional_attributes
       @additional_attributes = %w(config_info)
     end
+
+    def encode_picture(data)
```
bdunne: Should this be `decode_picture` instead?

spec/requests/service_templates_spec.rb
```diff
@@ -154,6 +162,15 @@
       expect(st.reload.name).to eq("Updated Template Name")
     end
 
+    it "supports picture update with base 64 decoded content" do
+      api_basic_authorize collection_action_identifier(:service_templates, :edit)
+
+      st = FactoryBot.create(:service_template, :name => "st1")
+      post(api_service_template_url(nil, st), :params => gen_request(:edit, updated_picture))
+
+      expect(st.reload.picture.md5).to eq("78805a221a988e79ef3f42d7c5bfd418")
```
bdunne: also we should `expect(st.reload.picture.content).to eq("image")` right?

app/controllers/api/service_templates_controller.rb
d-m-u: yeah, names n thing n stuff n whatever

app/controllers/api/service_templates_controller.rb
```diff
@@ -52,5 +53,9 @@ def unarchive_resource(type, id, _data)
     def set_additional_attributes
       @additional_attributes = %w(config_info)
     end
+
+    def decode_picture(data)
+      data["picture"]["content"] = Base64.strict_decode64(data["picture"]["content"])
```
abellotti: Can Picture.create_from_base64(...) be used here ?

d-m-u: there are three ways to do this: the UI, the options hash, and this call. the only place we have to worry about the encoding of the content is here. i think core should handle the picture create, and that side is working fine as long as you have this change.

d-m-u: but actually anyway, it can't for a better reason, doing so runs into this fun: `ActiveRecord::RecordNotSaved: You cannot call create unless the parent is saved`

bdunne: If we were strictly REST here, I think the picture upload would be an independent call to upload the image then `data` in this call would include `"picture" => /api/pictures/:id"`, not the raw content.  But I think that change is out of scope for this PR.

spec/requests/service_templates_spec.rb
```diff
@@ -154,6 +162,17 @@
       expect(st.reload.name).to eq("Updated Template Name")
     end
 
+    it "supports picture update with base 64 decoded content" do
+      api_basic_authorize collection_action_identifier(:service_templates, :edit)
+
+      st = FactoryBot.create(:service_template, :name => "st1")
```
bdunne: Sorry, just noticed this... Please don't set `:name` since the factory already sequences that.

app/controllers/api/service_templates_controller.rb
abellotti: 👍 Thanks.

simaishi:
Jansa backport details:

```
$ git log -1
commit dca8f395455a25f208e27122156e832dbea0b0ef
Date:   Tue Mar 17 12:54:35 2020 -0400

    Merge pull request #777 from d-m-u/add_spec_for_service_template_picture_update_encoding

    force updated service template photo encoding

    (cherry picked from commit 57f889ebd3503bc65e27e5c50bf42d525d209516)

    https://bugzilla.redhat.com/show_bug.cgi?id=1770197
```

simaishi:
@d-m-u can this be `ivanchuk/yes`?

d-m-u:
think so 
@miq-bot add_label ivanchuk/yes

simaishi:
Ivanchuk backport details:

```
$ git log -1
commit 7237c8a65c9414525f3f49334e141b6903ddc524
Date:   Tue Mar 17 12:54:35 2020 -0400

    Merge pull request #777 from d-m-u/add_spec_for_service_template_picture_update_encoding

    force updated service template photo encoding

    (cherry picked from commit 57f889ebd3503bc65e27e5c50bf42d525d209516)

    https://bugzilla.redhat.com/show_bug.cgi?id=1770197
```

PR #780, opened 2020-03-18
[WIP] Adding support for primary resource options

abellotti:
Prototype - NOT for merging.

Adding support for primary resource options via /api/collection/:id

```
{
  "actions": {
    "post": {
      "refresh": {
        "available": true
      },
      "shutdown_guest": {
        "available": false,
        "reason": "The VM is not powered on"
      },
      "reboot_guest": {
        "available": false,
        "reason": "The VM is not powered on"
      },
      "start": {
        "available": true
      },
      "stop": {
        "available": false,
        "reason": "The VM is not powered on"
      },
      ...
    }
  }
}
```

Prototype - proposal of general signature on OPTIONS /api/collection/:id, organized by HTTP
method, then hash of actions so callers can just access the individual action.

Implementation uses helper methods available there on vms and cloud instances, not sure if that can
be generalized for all collections otherwise we'd need to drive this capability via the api.yml config 
per collections.

Also would need to expose the custom button actions here too.

### DIFF
--- a/app/controllers/api/base_controller.rb
+++ b/app/controllers/api/base_controller.rb
@@ -112,6 +112,10 @@ module Api
       end
     end
 
+    def options_resource
+      render_resource_options(@req.collection, params[:c_id])
+    end
+
     def settings
       id       = @req.collection_id
       type     = @req.collection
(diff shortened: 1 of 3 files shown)


bdunne:
Interesting.  Is there a higher-level RFE issue that this relates to?

abellotti:
Nope, possibly in future. Just didn't want to lose the research/prototype.

abellotti:
Closing for now until we need to push this feature forward.

Issue #739, opened 2020-02-11
Dialog field data - id coming as number, not string


himdel:
1. Create a dialog with a `DialogFieldTagControl` type of input (select a category to populate the values)
1. Query the service_dialog using its ID - `GET /api/service_dialogs/<id>`
1. In `content` -> `dialog_tabs` -> `dialog_groups` -> `dialog_fields` find the `DialogFieldTagControl` type input field and check its `options` object. You will see that the id is a number.

---

`content[0].dialog_tabs[0].dialog_groups[0].dialog_fields[0]`:

```json
{
  "href": "http://localhost:3000/api/service_dialogs/6454",
  "id": "6454",
  "name": "tag_control_1",
  "description": "",
  "data_type": "string",
  "notes": null,
  "notes_display": null,
  "display": "edit",
  "display_method": null,
  "display_method_options": {},
  "required": false,
  "required_method": null,
  "required_method_options": {},
  "default_value": "",
  "values": [
    {
      "id": null,
      "name": "<None>",
      "description": "<None>"
    },
    {
      "id": 3,
      "name": "chicago",
      "description": "Chicago"
    },
    {
      "id": 4,
      "name": "london",
      "description": "London"
    },
    {
      "id": 2,
      "name": "ny",
      "description": "New York"
    },
    {
      "id": 5,
      "name": "paris",
      "description": "Paris"
    }
  ],
  "values_method": null,
  "values_method_options": {},
  "options": {
    "category_id": "1",
    "force_single_value": true,
    "sort_by": "description",
    "sort_order": "ascending",
    "category_description": "Location",
    "category_name": "location",
    "category_single_value": true
  },
  "created_at": "2020-02-11T19:25:15Z",
  "updated_at": "2020-02-11T19:25:15Z",
  "label": "Tag Control",
  "dialog_group_id": "944",
  "position": 0,
  "validator_type": null,
  "validator_rule": null,
  "reconfigurable": false,
  "dynamic": false,
  "show_refresh_button": false,
  "load_values_on_init": true,
  "read_only": false,
  "auto_refresh": false,
  "trigger_auto_refresh": false,
  "visible": true,
  "type": "DialogFieldTagControl",
  "resource_action": {
    "action": null,
    "resource_type": "DialogField",
    "ae_namespace": null,
    "ae_class": null,
    "ae_instance": null,
    "ae_message": null,
    "ae_attributes": {}
  },
  "dialog_field_responders": []
}
```

---

[...]

PR #786, opened 2020-03-24
ensure ids in dialog field values are normalized

d-m-u:
Dialog field value ids aren't normalized which means the ids come back as numbers, not strings. 

Fixes https://github.com/ManageIQ/manageiq-api/issues/739

https://bugzilla.redhat.com/show_bug.cgi?id=1817079

### DIFF
--- a/app/controllers/api/base_controller/normalizer.rb
+++ b/app/controllers/api/base_controller/normalizer.rb
@@ -38,7 +38,7 @@ module Api
           normalize_array(value)
         elsif value.respond_to?(:attributes) || value.respond_to?(:keys)
           normalize_hash(attr, value)
-        elsif attr == "id" || attr.to_s.ends_with?("_id")
+        elsif attr.to_s == "id" || attr.to_s.ends_with?("_id")
           value.to_s
         elsif Api.time_attribute?(attr)
           normalize_time(value)
--- a/spec/requests/service_dialogs_spec.rb
+++ b/spec/requests/service_dialogs_spec.rb
@@ -4,6 +4,9 @@
 # - Refresh dialog fields       /api/service_dialogs/:id "refresh_dialog_fields"
 #
 describe "Service Dialogs API" do
+  let(:dialog_tab)    { FactoryBot.create(:dialog_tab, :label => 'tab') }
+  let(:dialog_group)  { FactoryBot.create(:dialog_group, :label => 'group') }
+  let(:dialog_field) { FactoryBot.create(:dialog_field_tag_control, :label => 'test tag category', :name => 'test tag category', :options => {:id => 3}) }
   let(:dialog1)    { FactoryBot.create(:dialog, :label => "ServiceDialog1") }
   let(:dialog2)    { FactoryBot.create(:dialog, :label => "ServiceDialog2") }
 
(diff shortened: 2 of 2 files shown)


app/controllers/api/service_dialogs_controller.rb
```diff
@@ -62,6 +63,21 @@ def template_service_dialog_resource(_type, _id, data)
...
     private
 
+    def normalize_value_ids(content)
+      content["dialog_tabs"].each do |tab|
+        tab["dialog_groups"].each do |group|
+          group["dialog_fields"].each do |field|
+            field["values"].each do |values|
+              values.select { |k,v| k.to_s == "id" }.each do |id_pair|
+                id_pair.each { |k,v| values.merge!({k => normalize_attr("id",v)}) }
+                end
```
bdunne: extra `end`?

app/controllers/api/service_dialogs_controller.rb
```diff
@@ -22,6 +22,7 @@ def refresh_dialog_fields_resource(type, id = nil, data = nil)
     def fetch_service_dialogs_content(resource)
       target, resource_action = validate_dialog_content_params(params)
       resource.content(target, resource_action, true)
+      normalize_value_ids(resource.content[0])
```
abellotti: Hmm, curious, if all we're doing is normalizing the id attribute in the response, why isn't the API normalizer doing it for us ?

d-m-u: would you prefer i change the normalizer to also take symbol keys? this could be 'elsif attr.to_s == "id" || attr.to_s.ends_with?("_id")' which would solve this issue as well but it won't handle symbols at the moment

d-m-u: it's just that this only is an issue for service dialogs and in particular only dialogs with fields that have `values` so potentially dropdowns and tag controls

d-m-u: oh okay fine

abellotti: Thanks @d-m-u for the update.  This PR now fixes the underlying issue. If attr was to_s'd to check for _id names, it should have been so for checking id as well.  👍

abellotti:
Thanks @d-m-u for the update. maybe add one spec for this, that should do it.

d-m-u:
@miq-bot add_label bug

abellotti:
is this jansa/yes ?

d-m-u:
I don't know, I don't have a bug, just the github issue. I'll open a bug for it tomorrow.

chessbyte:
No need for a BZ in order to request a bug fix be backported.

d-m-u:
sure but it will need one anyway

simaishi:
Jansa backport details:

```
$ git log -1
commit 5504233f858adf9d70ad2bc262e81cfd2db23a64
Date:   Wed Mar 25 14:31:57 2020 -0400

    Merge pull request #786 from d-m-u/ensure_normalized_dialog_field_value_ids

    ensure ids in dialog field values are normalized

    (cherry picked from commit 261102c181442d22e381d962b2cb3edf5781234a)

    https://bugzilla.redhat.com/show_bug.cgi?id=1817079
```

PR #792, opened 2020-04-01
fix rails 5.2 json parse

kbrock:
JSON.parse('') was throwing error. this fixes it

```
  1) Creating responds with Bad Request if there are no resources at the top level
     Failure/Error:
Hash(JSON.parse(request.body.read)).fetch("action", "create")
                                   .in?(%w(create add))
     
     JSON::ParserError:
       765: unexpected token at ''
     # ./lib/api/create_constraint.rb:4:in `matches?'
```

@miq-bot add_label bug, rails-5-2
@miq-bot assign jrafanie

### DIFF
--- a/lib/api/create_constraint.rb
+++ b/lib/api/create_constraint.rb
@@ -1,7 +1,8 @@
 module Api
   class CreateConstraint
     def matches?(request)
-      Hash(JSON.parse(request.body.read)).fetch("action", "create").in?(%w(create add))
+      body = request.body.read
+      body.present? && Hash(JSON.parse(body)).fetch("action", "create").in?(%w[create add])
     ensure
       request.body.rewind
     end


lib/api/create_constraint.rb
```diff
@@ -1,7 +1,8 @@
 module Api
   class CreateConstraint
     def matches?(request)
-      Hash(JSON.parse(request.body.read)).fetch("action", "create").in?(%w(create add))
+      body = request.body.read
+      Hash(JSON.parse(body)).fetch("action", "create").in?(%w(create add)) unless body.empty?
```
jrafanie, line 5: So, if  `body.empty?` is true, it will return nil?  Should we return false?

jrafanie:
@miq-bot add_label jansa/yes?

lib/api/create_constraint.rb
jrafanie: @kbrock see above

bdunne: :+1: agreed, since it's a `?` method, let's return true / false

kbrock:
Thanks all - I like those changes

simaishi:
Jansa backport details:

```
$ git log -1
commit a48d36ef76b46135a32c73949201683dacbc1cf3
Date:   Thu Apr 2 23:26:44 2020 -0400

    Merge pull request #792 from kbrock/constraints_check

    fix rails 5.2 json parse

    (cherry picked from commit a48f88ad9f219bfa4d87b90c12976e20b6b61bd3)
```

PR #793, opened 2020-04-06
Expose the enablement of async notifications in /api/product_info

skateman:
We don't want to expose the whole `/api/settings` for all users, but this boolean is actually harmless to have it accessible even without a login. We will need this information quite often, so I am planning to cache this data upon login and use it further.

Required for: https://github.com/ManageIQ/manageiq-ui-classic/pull/6816

```js
// curl -XGET http://admin:smartvm@localhost:3000/api | jq '.server_info'
{
  "version": "master",
  "build": "unknown_ada003b9a3",
  "release": "Kasparov",
  "appliance": "EVM",
  "time": "2020-04-06T11:32:48Z",
  "server_href": "http://localhost:3000/api/servers/1",
  "zone_href": "http://localhost:3000/api/zones/2",
  "region_href": "http://localhost:3000/api/regions/1",
  "enterprise_href": "http://localhost:3000/api/enterprises/1",
  "plugins": {} // ...,
  "asynchronous_notifications": true
}

```

cc @himdel

### DIFF
--- a/app/controllers/api/api_controller.rb
+++ b/app/controllers/api/api_controller.rb
@@ -6,14 +6,14 @@ module Api
 
     def index
       res = {
-        :name          => ApiConfig.base.name,
-        :description   => ApiConfig.base.description,
-        :version       => ManageIQ::Api::VERSION,
-        :versions      => entrypoint_versions,
-        :settings      => user_settings,
-        :identity      => auth_identity,
-        :server_info   => server_info,
-        :product_info  => product_info_data
+        :name         => ApiConfig.base.name,
+        :description  => ApiConfig.base.description,
+        :version      => ManageIQ::Api::VERSION,
+        :versions     => entrypoint_versions,
+        :settings     => user_settings,
+        :identity     => auth_identity,
+        :server_info  => server_info,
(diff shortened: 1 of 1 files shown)


himdel:
So, IMO this is a hack, that's an actual setting, not productization related.

Instead, we need a good solution for https://github.com/ManageIQ/manageiq/issues/19110.

(But works for me as a hack, to get https://github.com/ManageIQ/manageiq-ui-classic/pull/6816 working for all users, that issue has existed for a while now, so not sure it should block this.)

skateman:
Well, I'm okay with moving this into a different endpoint. We will depend in the UI on more stuff that's under `Settings` anyway.

skateman:
@himdel what do you think about `/api/server_info`? It already exists, I just didn't notice it.

skateman:
We discussed this with @himdel, taking a different approach for now....closing

himdel:
> There's already a server_info underneath.

That's dead now, before https://github.com/ManageIQ/manageiq-ui-service/pull/604, SUI used that endpoint.

But yeah, we can just use `::Settings` from the UI for now.

PR #728, opened 2020-01-27
Fix update of CustomButton to save ae_attributes

ZitaNemeckova:
Problem: API do not save custom attributes from UI.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1793556

@miq-bot add_label wip, bug, ivanchuk/no

PR #796, opened 2020-04-09
create the resource action on a button before we try to update! it

d-m-u:
we can't `update!` on nil so this lets us create without errors

@miq-bot add_label bug

also we can't use the current error handling cause the custom button doesn't exist if it fails inside the transaction 

technically found while looking at https://bugzilla.redhat.com/show_bug.cgi?id=1754554 but while I did mention it in that ticket it does not fix that since that's opened for hammer and isn't a bug on hammer. if someone were to try what's in that ticket on master they would run into this. although it's more general than stated in that ticket since the issue fixed here blocks any custom button create on any recent version over the api and only incidentally involves dialog ids or resource actions.

### DIFF
--- a/app/controllers/api/custom_buttons_controller.rb
+++ b/app/controllers/api/custom_buttons_controller.rb
@@ -2,32 +2,31 @@ module Api
   class CustomButtonsController < BaseController
     def create_resource(_type, _id, data)
       CustomButton.transaction do
-        custom_button = CustomButton.new(data.except("resource_action", "options"))
-        custom_button.userid = User.current_user.userid
-        custom_button.options = data["options"].deep_symbolize_keys if data["options"]
-        custom_button.resource_action.update!(data["resource_action"].deep_symbolize_keys) if data.key?("resource_action")
-        custom_button.save!
-        custom_button
+        CustomButton.new(data.except("resource_action", "options")).tap do |cb|
+          cb.userid = User.current_user.userid
+          cb.options = data["options"].deep_symbolize_keys if data["options"]
+          cb.create_resource_action!(data["resource_action"].deep_symbolize_keys) if data.key?("resource_action")
+          cb.save!
+        end
       end
-    rescue
-      raise BadRequestError, "Failed to create new custom button - #{custom_button.errors.full_messages.join(", ")}"
+    rescue => err
+      raise BadRequestError, "Failed to create new custom button - #{err.message}"
     end
 
     def edit_resource(type, id, data)
       return if data.empty?
 
       CustomButton.transaction do
-        custom_button = fetch_custom_button(type, id)
-        updated_data = data['resource'] || data
-        if updated_data['resource_action'].present?
-          custom_button.resource_action = create_or_update_resource_action(updated_data['resource_action'])
+        fetch_custom_button(type, id).tap do |cb|
+          updated_data = (data['resource'] || data).dup
+          if (resource_action = updated_data.delete("resource_action")).present?
+            cb.resource_action = create_or_update_resource_action(resource_action)
+          end
+          cb.update!(updated_data.deep_symbolize_keys)
         end
-        updated_data.except!('resource_action')
-        custom_button.update!(updated_data.deep_symbolize_keys) if updated_data.present?
-        custom_button
       end
     rescue => err
-      raise BadRequestError, "Failed to update custom button - #{err}"
+      raise BadRequestError, "Failed to update custom button - #{err.message}"
     end
 
     def options
@@ -44,14 +43,11 @@ module Api
 
     def create_or_update_resource_action(data)
       return nil if data.empty?
-      id = data["id"]
+      id = data.delete("id")
       if id.present?
-        resource_action = ResourceAction.find(id)
-        data.except!("id")
-        resource_action.update!(data.deep_symbolize_keys)
-        resource_action
+        ResourceAction.find(id).tap { |ra| ra.attributes = data }
       else
-        ResourceAction.create(data)
+        ResourceAction.new(data)
       end
     end
 
--- a/spec/requests/custom_buttons_spec.rb
+++ b/spec/requests/custom_buttons_spec.rb
@@ -122,8 +122,44 @@ RSpec.describe 'CustomButtons API' do
       expect(response).to have_http_status(:ok)
       custom_button = CustomButton.find(response.parsed_body['results'].first["id"])
       expect(custom_button.options[:button_icon]).to eq("ff ff-view-expanded")
+      expect(CustomButton.count).to eq 1
+      expect(ResourceAction.count).to eq 1
       expect(custom_button.visibility['roles']).to eq(['_ALL_'])
       expect(response.parsed_body['results'].first).to include(cb_rec.except('resource_action', 'uri_attributes'))
+    end
+
+    it 'shows validation errors on failure' do
+      api_basic_authorize collection_action_identifier(:custom_buttons, :create)
+
+      cb_rec = {
+        'name'             => 'Generic Object Custom Button',
+        'description'      => 'Generic Object Custom Button description',
+        'applies_to_class' => 'GenericObjectDefinition',
+        'uri_attributes'   => {:request => "automate_method"},
+        'options'          => {
+          'button_icon'  => 'ff ff-view-expanded',
+          'button_color' => '#4727ff',
+          'display'      => true,
+        },
+        'resource_action'  => {
+          'ae_namespace' => 'SYSTEM',
+          'ae_class'     => 'PROCESS'
+        },
+        'visibility'       => {'roles' => ['_ALL_']}
+      }
+      post(api_custom_buttons_url, :params => cb_rec)
+      post(api_custom_buttons_url, :params => cb_rec)
+
+      expected = {
+        'error' => a_hash_including(
+          'kind'    => 'bad_request',
+          'message' => /Validation failed: CustomButton: Name has already been taken, CustomButton: Description has already been taken/
+        )
+      }
+      expect(response).to have_http_status(:bad_request)
+      expect(response.parsed_body).to include(expected)
+      expect(CustomButton.count).to eq 1
+      expect(ResourceAction.count).to eq 1
     end
 
     it 'can edit custom buttons by id' do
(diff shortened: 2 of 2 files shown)


d-m-u:
@ZitaNemeckova please review, I think it fixes an issue with https://github.com/ManageIQ/manageiq-api/pull/728/files

d-m-u:
because 728 got backported so this probably should as well

abellotti:
I think for the rubocop above, I prefer the

```
rescue => err
   raise BadRequestError, "... - #{err}"
end
```

at least it's consistent with all the other blocks.

app/controllers/api/custom_buttons_controller.rb
```diff
@@ -44,12 +42,8 @@ def build_custom_button_options
...
     def create_or_update_resource_action(data)
       return nil if data.empty?
-      id = data["id"]
-      if id.present?
-        resource_action = ResourceAction.find(id)
-        data.except!("id")
-        resource_action.update!(data.deep_symbolize_keys)
-        resource_action
+      if data["id"].present?
+        ResourceAction.find(id).tap { |ra| ra.update!(data.except!("id").deep_symbolize_keys) }
```
abellotti, line 46: is this right ?  doing find(id) but id is no longer defined.

kbrock: I totally did not see this

app/controllers/api/custom_buttons_controller.rb
```diff
@@ -2,29 +2,27 @@ module Api
...
+        CustomButton.new(data.except("resource_action", "options")).tap do |cb|
+          cb.userid = User.current_user.userid
+          cb.options = data["options"].deep_symbolize_keys if data["options"]
+          cb.resource_action = ResourceAction.create!(data["resource_action"].deep_symbolize_keys) if data.key?("resource_action")
+          cb.save!
+        end
       end
     rescue
-      raise BadRequestError, "Failed to create new custom button - #{custom_button.errors.full_messages.join(", ")}"
+      raise BadRequestError, "Failed to create new custom button - #{$!}"
```
bdunne: Instead of `!$` you can `rescue => error` and use `error.message` here

app/controllers/api/custom_buttons_controller.rb
d-m-u: AB with 🦅 👀

app/controllers/api/custom_buttons_controller.rb
```diff
@@ -2,29 +2,27 @@ module Api
...
-        custom_button = CustomButton.new(data.except("resource_action", "options"))
-        custom_button.userid = User.current_user.userid
-        custom_button.options = data["options"].deep_symbolize_keys if data["options"]
-        custom_button.resource_action.update!(data["resource_action"].deep_symbolize_keys) if data.key?("resource_action")
-        custom_button.save!
-        custom_button
+        CustomButton.new(data.except("resource_action", "options")).tap do |cb|
+          cb.userid = User.current_user.userid
+          cb.options = data["options"].deep_symbolize_keys if data["options"]
+          cb.resource_action = ResourceAction.create!(data["resource_action"].deep_symbolize_keys) if data.key?("resource_action")
```
bdunne, line 8: You can use the helper method `cb.create_resource_action(stuff)` here

app/controllers/api/custom_buttons_controller.rb
d-m-u: done in the last commit before you commented

app/controllers/api/custom_buttons_controller.rb
```diff
@@ -2,29 +2,27 @@ module Api
...
+          cb.userid = User.current_user.userid
+          cb.options = data["options"].deep_symbolize_keys if data["options"]
+          cb.create_resource_action(data["resource_action"].deep_symbolize_keys) if data.key?("resource_action")
+          cb.save!
+        end
       end
-    rescue
-      raise BadRequestError, "Failed to create new custom button - #{custom_button.errors.full_messages.join(", ")}"
+    rescue => err
+      raise BadRequestError, "Failed to create new custom button - #{err}"
```
bdunne, line 13: This should be `err.message` right?

app/controllers/api/custom_buttons_controller.rb
```diff
@@ -2,29 +2,27 @@ module Api
...
-        custom_button = CustomButton.new(data.except("resource_action", "options"))
-        custom_button.userid = User.current_user.userid
-        custom_button.options = data["options"].deep_symbolize_keys if data["options"]
-        custom_button.resource_action.update!(data["resource_action"].deep_symbolize_keys) if data.key?("resource_action")
-        custom_button.save!
-        custom_button
+        CustomButton.new(data.except("resource_action", "options")).tap do |cb|
+          cb.userid = User.current_user.userid
+          cb.options = data["options"].deep_symbolize_keys if data["options"]
+          cb.create_resource_action(data["resource_action"].deep_symbolize_keys) if data.key?("resource_action")
```
bdunne, line 8: Should this use the `!` method like it did before?

app/controllers/api/custom_buttons_controller.rb
```diff
@@ -46,10 +44,7 @@ def create_or_update_resource_action(data)
       return nil if data.empty?
       id = data["id"]
       if id.present?
-        resource_action = ResourceAction.find(id)
-        data.except!("id")
-        resource_action.update!(data.deep_symbolize_keys)
-        resource_action
+        data.except!("id").each_with_object(ResourceAction.find(id)) { |(k, v), ra| ra.send("#{k}=", v) }
```
bdunne, line 47: Why `each_with_object` instead of the `update!` that was here before?

bdunne: Also, this is no longer saving any changes, right?

d-m-u: doesn't need to, as Keenan pointed out, the update here isn't necessary since it's handled by line 24

bdunne: Line 24 is updating the CustomButton, not the ResourceAction

d-m-u: the button that gets saved with the resource action returned on 22, the update isn't needed twice

kbrock: @d-m-u can you point us to where the update happens?

It is a little tricky to see how this code reaches line 24

app/controllers/api/custom_buttons_controller.rb
```diff
@@ -2,32 +2,30 @@ module Api
...
       CustomButton.transaction do
-        custom_button = fetch_custom_button(type, id)
-        updated_data = data['resource'] || data
-        if updated_data['resource_action'].present?
-          custom_button.resource_action = create_or_update_resource_action(updated_data['resource_action'])
+        fetch_custom_button(type, id).tap do |cb|
+          updated_data = data['resource'] || data
+          cb.resource_action = create_or_update_resource_action(updated_data['resource_action']) if updated_data['resource_action'].present?
+          updated_data.except!('resource_action')
+          cb.update!(updated_data.deep_symbolize_keys) if updated_data.present?
```
kbrock, line 24:
```suggestion
          cb.update!(updated_data.deep_symbolize_keys)
```

I think we always want to call update
And `if !updated_data.present?` then updated_data will be empty and the attributes will be a no-op but we want to save for the create_or_update action.

Are we sure that a save to this button also saves resource_action? (I always get confused between `has_one` and `belongs_to`. One of them saves and the other does not.

d-m-u: Yeah, I agree with this suggestion, and yes, we're pretty definite this does also save the resource action.

app/controllers/api/custom_buttons_controller.rb
d-m-u: Don't think this works.

app/controllers/api/custom_buttons_controller.rb
```diff
@@ -2,32 +2,32 @@ module Api
...
       return if data.empty?
 
       CustomButton.transaction do
-        custom_button = fetch_custom_button(type, id)
-        updated_data = data['resource'] || data
-        if updated_data['resource_action'].present?
-          custom_button.resource_action = create_or_update_resource_action(updated_data['resource_action'])
+        fetch_custom_button(type, id).tap do |cb|
+          updated_data = data['resource'] || data
+          if (resource_action = updated_data.delete("resource_action")).present?
```
chessbyte, line 22: should this be `updated_data = (data['resource'] || data).dup` or `updated_data = data['resource'] || data.dup` since `updated_data.delete` is being called on the next line?

app/controllers/api/custom_buttons_controller.rb
```diff
@@ -2,32 +2,32 @@ module Api
...
+            cb.resource_action = create_or_update_resource_action(resource_action)
+          end
+          updated_data.except!('resource_action')
```
kbrock, line 25: this line is no longer necessary (the delete will take care of it

app/controllers/api/custom_buttons_controller.rb
kbrock: you could dup or you could use an except() in deep_symbolize_keys.

Before, we had updated_data.present? so we needed to modify the hash to get present correct.
Now, I think we may be able to avoid this

kbrock: looks like we need updated_data to not have 'resource_action' in it (aka drew was right)
So lets follow oleg's suggestion to dup the hash

d-m-u:
anyone else want to look at this, please?

spec/requests/custom_buttons_spec.rb
```diff
@@ -147,6 +183,44 @@
       expect(cb.reload.visibility[:roles]).to eq(['_ALL_'])
       expect(cb.reload.resource_action.ae_attributes).to include("attribute" => "is present")
     end
+
+    it 'updates the existing resource action' do
+      cb.resource_action = ResourceAction.create!(:dialog_id => 4)
```
bdunne, line 188: Should this be using the `:resource_action` factory and its `:with_dialog` trait? I'm not a fan of using hard-coded ids in test since they're unlikely to be in the same region that the test is running in.

spec/requests/custom_buttons_spec.rb
```diff
@@ -147,6 +183,44 @@
...
+
+        ]
+      }
+
+      expect(cb.resource_action.dialog_id).to eq(4)
+
+      post(api_custom_buttons_url, :params => request)
+      cb.reload
+
+      expect(ResourceAction.first.dialog_id).to eq(2)
```
bdunne, line 204: same here

app/controllers/api/custom_buttons_controller.rb
```diff
@@ -44,14 +43,11 @@ def build_custom_button_options
...
     def create_or_update_resource_action(data)
       return nil if data.empty?
-      id = data["id"]
+      id = data.delete("id")
       if id.present?
-        resource_action = ResourceAction.find(id)
-        data.except!("id")
-        resource_action.update!(data.deep_symbolize_keys)
-        resource_action
+        ResourceAction.find(id).tap { |ra| ra.attributes = data }
```
bdunne, line 48: Still feels awkward to me that we're relying on `has_one :resource_action, :autosave => true` to save the resource action when `cb.update!` is called.

spec/requests/custom_buttons_spec.rb
d-m-u: You mean line 199, right? This one should be fine.

bdunne: Both?  This is so strange, why would you be able to update that attribute in this way?

d-m-u: Because I have a ticket in which someone expressly requested this functionality?

d-m-u: It's not really that strange, being able to update or create a button and specify the dialog seems perfectly reasonable to me really.

bdunne: I guess the part I'm having trouble with it that we're updating attributes of a `ResourceAction` through the `/api/custom_buttons/:id` URL rather than `/api/custom_buttons/:id/resource_action` (doesn't exist) or `/api/resource_actions/:id` (only supports GET).  So it's hard to validate the attributes that are allowed to be changed throught the API.

In theory you could update `resource_action["resource_id"]` to the ID of a different CustomButton that already has a ResourceAction and totally break the `has_one` relation.  (We should probably have a model validation for that)

d-m-u: Okay ... but can that be done in later work, please?

PR #810, opened 2020-04-23
[BaseController::Render] Cache role allows

NickLaMuro:
When rendering a collection, the `User.role_allows?` call is hit for every single record for every single action for a given collection item.

Since this is only a "role specific" check, and not item specific, there is no reason to repeat this work for each record, and so this cache can save a noticeable amount of time on larger datasets.

Before
------

| 3205 |      14 |         26 | 1732 |
| 1241 |      11 |        7.0 |  286 |
| 1609 |      11 |        6.6 |  286 |
| 1277 |      11 |        6.5 |  286 |
| 1192 |      11 |        6.4 |  286 |

After
-----

| 2803 |      14 |         26 | 1732 |
| 1010 |      11 |        6.8 |  286 |
|  999 |      11 |        6.4 |  286 |
| 1004 |      11 |        6.7 |  286 |
| 1007 |      11 |        6.6 |  286 |

Uses the following query:

    curl -XGET "http://user:pass@example.com/api/vms?expand=resources&filter[]=active=true"

### DIFF
--- a/app/controllers/api/base_controller/renderer.rb
+++ b/app/controllers/api/base_controller/renderer.rb
@@ -514,7 +514,13 @@ module Api
 
         return custom_api_user_role_allows?(action_identifier) if custom_api_user_role_allows_method?(action_identifier)
 
-        Array(action_identifier).any? { |identifier| User.current_user.role_allows?(:identifier => identifier) }
+        @role_allows_cache ||= {}
+        Array(action_identifier).any? do |identifier|
+          unless @role_allows_cache.key?(identifier)
+            @role_allows_cache[identifier] = User.current_user.role_allows?(:identifier => identifier)
+          end
+          @role_allows_cache[identifier]
+        end
       end
 
       def render_actions(physical_attrs)


jrafanie:
LGTM, but why is it failing?

abellotti:
all PRs failing,  we need to update the conversion_hosts_spec to define a network/ipaddress on the vm's to succeed due to recent changes to v2v.

NickLaMuro:
@abellotti Thanks!  I totally missed the ping for this earlier, but thanks for restarting the specs.

Fryguy:
Is this cache reset on each session?  Otherwise, what happens if the user's roles are changed/removed?

NickLaMuro:
> Is this cache reset on each session? Otherwise, what happens if the user's roles are changed/removed?

@Fryguy That was the intent, but there is always a possibility I have completely mis-understood `ActionController` all these years... (I wish this was less of a possibility than it actually is...)

Anyway, "cache" is probably a poor word, but it is a "per request cache", or at least that is the intent.  Shouldn't live longer than the instance of the class, which I believe is a single request (for a single user).

abellotti:
yep, per request 👍

app/controllers/api/base_controller/renderer.rb
```diff
@@ -514,7 +514,10 @@ def api_user_role_allows?(action_identifier)
 
         return custom_api_user_role_allows?(action_identifier) if custom_api_user_role_allows_method?(action_identifier)
 
-        Array(action_identifier).any? { |identifier| User.current_user.role_allows?(:identifier => identifier) }
+        Array(action_identifier).any? do |identifier|
+          @role_allows_cache ||= {}
+          @role_allows_cache[identifier] ||= User.current_user.role_allows?(:identifier => identifier)
```
chessbyte, line 519: Won’t it re-evaluate all the ones that resulted in `false`?  The any? loop will stop on the first one returning `true`, I believe.  So it may be better to just remember all the identifiers, because if it seen again, it must mean that they evaluated to `false` before.

chessbyte: Thinking further, is the following any better?
`Array(action_identifier).uniq.any? { |identifier| User.current_user.role_allows?(:identifier => identifier) }`

NickLaMuro: > Won’t it re-evaluate all the ones that resulted in `false`?

Good catch, I can refactor to not use `||=` and instead use something like:

```ruby
unless @role_allows_cache.key?(identifier)
            @role_allows_cache[identifier] ||= User.current_user.role_allows?(:identifier => identifier)
end
@role_allows_cache[identifier]
```

I should also add some specs for this as well.

> Thinking further, is the following any better?

The problem is that this is done inside an N+1, or for each record, so the `uniq` serves no purpose here since they are all `unique` to begin with.

Effectively, we are doing this:

```ruby
# heavily simplifying to showcase the problem
@vms        = Vms.all
@reponse = []
@vms.each do |vm|
  @response << vm.to_json
  @response[:actions] = []
  fetch_typed_subcollection_actions.each do |action_identifier|
    if Array(action_identifier).any? { |identifier| User.current_user.role_allows?(:identifier => identifier) }
      @response[:actions] << action_identifier
    end
  end
end
```

So the performance problem lies that we have to go through that `.any?` a bunch of times and calculate the same work each time.  `api_user_role_allows?` just happens to accept an array, but for this case, it is most likely coming from here:

https://github.com/ManageIQ/manageiq-api/blob/2506100b/app/controllers/api/base_controller/renderer.rb#L451-L469

Where it actually passes a single value and does the looping itself (like I showed above).  What this solves is trying to check the same set of role permissions multiple times per request for each object brought back, when the result from that check is always the same.

chessbyte: @NickLaMuro I think the refactoring you are suggesting would be a better approach.

NickLaMuro: Woops, copy pasta mistake on my end.  I agree, that was my intent to do that, but thanks for catching.

NickLaMuro: This is fixed now.

> I should also add some specs for this as well.

I am actually going to pass on this, since I did some `grep`ing for "role" and it seems like this is pretty well exercised indirectly already, and I would have to add a entirely new spec folder to "unit" test the caching of this method this since our specs are all request spec based.

Unless that is desired, I will probably mark this as resolved.

app/controllers/api/base_controller/renderer.rb
```diff
@@ -514,7 +514,13 @@ def api_user_role_allows?(action_identifier)
 
         return custom_api_user_role_allows?(action_identifier) if custom_api_user_role_allows_method?(action_identifier)
 
-        Array(action_identifier).any? { |identifier| User.current_user.role_allows?(:identifier => identifier) }
+        @role_allows_cache ||= {}
+        Array(action_identifier).any? do |identifier|
+          if @role_allows_cache.key?(identifier)
```
jrafanie, line 519: Is the `if` reversed?  Am I missing something?  How will we ever set the `@role_allows_cache` on line 520?

NickLaMuro: Yes... yes it is.  I didn't even copy what I had above.

Welp, we know the specs fail as a result.  Fixing...

jrafanie:
LGTM 👍, I believe @chessbyte's concerns were addressed.

abellotti:
@jrafanie anything else needed here or can it be merged ?

jrafanie:
@NickLaMuro is this worthy of jansa/yes?

NickLaMuro:
Sorry, missed this last ping.  😞 

> @NickLaMuro is this worthy of jansa/yes?

@jrafanie I don't think it needs `jansa/yes?` myself, but will leave that final call up to you.  My risk-averse nature when it comes to this says to wait since I don't see it as a game changer, but I have no frame of reference besides my few examples to go off of.

simaishi:
Jansa backport details:

```
$ git log -1
commit b9d535c2be5d5b9c292a9da3ca24591b13c5458d
Date:   Mon May 4 16:32:06 2020 -0400

    Merge pull request #810 from NickLaMuro/faster-role-allow-checks

    [BaseController::Render] Cache role allows

    (cherry picked from commit 485e39ed3211d2d5d742bcc333c452a28f6ce2ac)
```

PR #813, opened 2020-04-24
add functionality to retire vms in future

d-m-u:
`make_retire_request` is the wrong thing to queue for future retirement, the method that handles things in the future is just `retire` so we should be queuing it with the args set properly 

yes yes, i completely forgot we do this on the api too, i'm sorry 

fixes https://bugzilla.redhat.com/show_bug.cgi?id=1827787

@miq-bot add_label bug

### DIFF
--- a/spec/requests/vms_spec.rb
+++ b/spec/requests/vms_spec.rb
@@ -1117,6 +1117,14 @@ describe "Vms API" do
           expect(response.parsed_body).to include(expected)
         end
 
+        it "in the future" do
+          api_basic_authorize action_identifier(:vms, :request_retire)
+          date = 2.weeks.from_now
+          post(vm_url, :params => gen_request(:request_retire, :date => date.iso8601))
+
+          expect_single_action_result(:success => true, :message => /#{vm.id}.* request retire/i, :href => api_vm_url(nil, vm))
+        end
+
         it "queues retirement task" do
           api_basic_authorize(action_identifier(:vms, :request_retire))
           message = "VM id:#{vm.id} name:'#{vm.name}' request retire"
--- a/app/controllers/api/vms_controller.rb
+++ b/app/controllers/api/vms_controller.rb
@@ -274,11 +274,22 @@ module Api
       action_result(false, "Failed to set miq_server - #{err}")
     end
 
-    def request_retire_resource(type, id, _data = nil)
+    def request_retire_resource(type, id, data = nil)
       api_action(type, id) do |klass|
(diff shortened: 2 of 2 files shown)


d-m-u:

abellotti:
should we put that logic / generalize in app/controllers/api/base_controller/generic.rb so it's available for all request_retire actions (i.e. for services too) ?

d-m-u:
i mean, it is: https://github.com/ManageIQ/manageiq-api/blob/master/app/controllers/api/base_controller/generic.rb#L81 and it's not what runs when you retire a vm from the api. will you let me get this in and then i can figure out what's wrong with that, please?

abellotti:
that line https://github.com/ManageIQ/manageiq-api/blob/d8d45da6dc18fbd13376b69a9fdf204792ac5b28/app/controllers/api/base_controller/generic.rb#L81 is for the retire action.   Didn't see the date/warn for request retire action. which means we're not supporting it for service either and we advertise it https://www.manageiq.org/docs/reference/latest/api/reference/services#service-retiring-future.

Should be fixed for Jansa.  

Follow-up is fine, we need this capability for both service and vms, so handling it in generic would be great.  Probably need a small PR for https://www.manageiq.org/docs/reference/latest/api/reference/vms#retire-vm to mention the date/warn options.

abellotti:
https://github.com/ManageIQ/manageiq-api/issues/817

spec/requests/vms_spec.rb
```diff
@@ -1117,6 +1117,14 @@ def update_raw_power_state(state, *vms)
           expect(response.parsed_body).to include(expected)
         end
 
+        it "in the future" do
```
bdunne, line 1120: Can you add a test for an invalid date and tests for warn as well?

d-m-u: and test what? warn gets sets in the mixin and is tested in core. as with the invalid date. these don't get set until something gets queued.

bdunne: ok

simaishi:
Jansa backport details:

```
$ git log -1
commit a3fc02aeaa7e76b943b1fcad62b86a1573df500d
Date:   Fri May 1 17:01:29 2020 -0400

    Merge pull request #813 from d-m-u/vm_future_retirement

    add functionality to retire vms in future

    (cherry picked from commit 414f9aa3970857d84712c72cfcb4fd9b96c7127b)

    https://bugzilla.redhat.com/show_bug.cgi?id=1827787
```

PR #814, opened 2020-04-24
fix the custom button create

d-m-u:
currently is broken with the error handling and updating a nil

surely this is small enough to get approval?

@miq-bot add_label bug 
fixes https://bugzilla.redhat.com/show_bug.cgi?id=1827818

PR #821, opened 2020-04-30
create the resource action after save for custom buttons

d-m-u:
otherwise this fails with `Failed to create new custom button - You cannot call create unless the parent is saved` on rails 5.2 🤮 

Keenan caught it in https://travis-ci.com/github/ManageIQ/manageiq-cross_repo-tests/jobs/325865143#L2435 

@miq-bot add_reviewer @jrafanie 
@miq-bot add_label bug

Followup to #814
Needed for Rails 5.2: https://github.com/ManageIQ/manageiq/issues/20032

### DIFF
--- a/app/controllers/api/custom_buttons_controller.rb
+++ b/app/controllers/api/custom_buttons_controller.rb
@@ -6,8 +6,8 @@ module Api
         custom_button = CustomButton.new(data.except("resource_action", "options"))
         custom_button.userid = User.current_user.userid
         custom_button.options = data["options"].deep_symbolize_keys if data["options"]
-        custom_button.create_resource_action!(data["resource_action"].deep_symbolize_keys) if data.key?("resource_action")
         custom_button.save!
+        custom_button.create_resource_action!(data["resource_action"].deep_symbolize_keys) if data.key?("resource_action")
         custom_button
       end
     rescue => err


d-m-u:
https://github.com/ManageIQ/manageiq-api/pull/822 or whatever

app/controllers/api/custom_buttons_controller.rb
```diff
@@ -6,8 +6,8 @@ def create_resource(_type, _id, data)
         custom_button = CustomButton.new(data.except("resource_action", "options"))
         custom_button.userid = User.current_user.userid
         custom_button.options = data["options"].deep_symbolize_keys if data["options"]
-        custom_button.create_resource_action!(data["resource_action"].deep_symbolize_keys) if data.key?("resource_action")
         custom_button.save!
+        custom_button.create_resource_action!(data["resource_action"].deep_symbolize_keys) if data.key?("resource_action")
```
abellotti, line 10: just curious, can the create_resource_action! method somehow fail, and we end up with a created Object, but never used/returned ?  do we need a transaction for the above two so the request is atomic ?

d-m-u: I don't really think so. resource action create 1) has no validations and 2) are artifacts that link dialogs, resources, and automate, there's not much of a vector for failure

jrafanie: yeah, probably similar to this one: https://github.com/ManageIQ/manageiq/pull/20013

@d-m-u can you review if this is required and do a followup PR?  This is great to fix the issue on rails 5.2

jrafanie: We cross our streams at the same time and didn't use a mutex

simaishi:
Jansa backport details:

```
$ git log -1
commit bf293ed28e335200f98e02f4bf37678645e1f1d0
Date:   Thu Apr 30 14:00:36 2020 -0400

    Merge pull request #821 from d-m-u/fixing_silly_create_before_parent

    create the resource action after save for custom buttons

    (cherry picked from commit da3529ca5a85a3c4aa8858f8ba7d94c9a110cc9e)
```

PR #822, opened 2020-04-30
Save the parent object before creating a child association

jrafanie:
Rails 5.2 raises this error if you try to persist the child before creating the
parent: `Failed to create new custom button - You cannot call create unless the
parent is saved`

Followup to #814
Needed for Rails 5.2: https://github.com/ManageIQ/manageiq/issues/20032


d-m-u:
fine, as long as

jrafanie:
Closing, duplicate of #821

PR #825, opened 2020-05-01
Added parts for the NSX-T provider

gubbe505:
Added network_services, security_policies and security_policy_rules controllers and added some additional api methods to cloud_networks and providers


gubbe505:
The build fails probably because of this change:
https://github.com/ManageIQ/manageiq-api/blob/e0a50fd21251880e8067bce4a8d8fa1622b9c5c6/app/controllers/api/subcollections/security_groups.rb#L5

In the previous version the security_groups would be converted to an Array and wouldn't be filtered by RBAC. I have looked at other methods and there is no consistent way of returning if the value is nil.

Please advise.

agrare:
@gtanzillo @bdunne are you able to help with the rbac failure here?

gtanzillo:
>The build fails probably because of this change:

The subquery collection needs to return an array. So this is correct -
`object.respond_to?(:security_groups) ? Array(object.security_groups) : []`

All results go through Rbac here https://github.com/ManageIQ/manageiq-api/blob/1ad20d24d20efdbd454c70c4f1a23c73a642e701/app/controllers/api/base_controller/renderer.rb#L192-L199

Looking at https://github.com/ManageIQ/manageiq/blob/048c86151c95786875f00f233704af01b20aec41/lib/rbac/filterer.rb#L60 security groups should get filtered

gubbe505:
Sorry, my mistake. Had to look up the problem way we changed it. The reason is that mapping to a Array does not support filter as query parameter.

Nevertheless there no consistent way of returning the xxx_query_resource within the api.

agrare:
Hey @gubbe505 any progress on this?

agrare:
Hm I can't replicate the travis error locally, it seems like it can't find one of the tags in the region

agrare:
Spec failure looks unrelated to this change, https://github.com/ManageIQ/manageiq/pull/20197

abellotti:
@gubbe505 we seem to be missing the specs here

PR #848, opened 2020-05-26
WIP - Check compliance for only specified compliance profile

gtanzillo:


gtanzillo:
Closing because If we allow checking compliance one profile at a time, the compliance status may flip back and forth depending on the outcome of each profile.

The use case is really about at reporting the compliance status within the context of a single profile and that can be "compliant" correctly while the overall compliance status can be "not compliant" because some other profile failed.

The better approach is to filter the list of compliance statuses and display the one for the specific profile.

PR #849, opened 2020-05-27
update specs to reflect memory_mb rename and resize

d-m-u:
at the moment memory_mb is an int which is problematic with aggregation and the byte conversion.

we cast it in hardware, and i'd like to get rid of the cast 

see https://github.com/ManageIQ/manageiq-schema/pull/485
and https://github.com/ManageIQ/manageiq/pull/20213

these will all need to be merged concurrently


d-m-u:
I guess not.

PR #858, opened 2020-06-19
Make travis happy by not trying to build a dead route for reports

skateman:
The `url_for` doesn't work because the `show_report` route is dead. Its method has been deleted years ago, the route has been cleaned up just now. There's an issue to track the missing feature: https://github.com/ManageIQ/manageiq-ui-classic/issues/7126


Fryguy:
Merging to get master green

PR #864, opened 2020-07-13
Return an empty collection, instead of 500 server error, for providers that don't support configured_systems or configuration_profiles subcollections

gtanzillo:
Before:
`curl --user admin:smartvm http://localhost:3000/api/providers/1000000000004/configuration_profiles -k`
```json
{"error":{"kind":"internal_server_error","message":"undefined method `configuration_profiles' for #\u003cManageIQ::Providers::Google::CloudManager:0x000055ba49eb1168\u003e","klass":"NoMethodError"}}  
```
`curl --user admin:smartvm http://localhost:3000/api/providers/1000000000004/configured_systems -k `
```json
{"error":{"kind":"internal_server_error","message":"undefined method `configured_systems' for #\u003cManageIQ::Providers::Google::CloudManager:0x000055b866e0e178\u003e","klass":"NoMethodError"}} 
```
After:
```json
curl --user admin:smartvm http://localhost:3000/api/providers/1000000000004/configuration_profiles
{
    "name": "configuration_profiles",
    "count": 0,
    "subcount": 0,
    "pages": 0,
    "resources": [],
    "links": {
        "self": "http://localhost:3000/api/providers/1000000000004/configuration_profiles?offset=0",
        "first": "http://localhost:3000/api/providers/1000000000004/configuration_profiles?offset=0",
        "last": "http://localhost:3000/api/providers/1000000000004/configuration_profiles?offset=0"
    }
}
```
```json
curl --user admin:smartvm http://localhost:3000/api/providers/1000000000004/configured_systems
{
    "name": "configured_systems",
    "count": 143,
    "subcount": 0,
    "pages": 1,
    "resources": [],
    "links": {
        "self": "http://localhost:3000/api/providers/1000000000004/configured_systems?offset=0",
        "first": "http://localhost:3000/api/providers/1000000000004/configured_systems?offset=0",
        "last": "http://localhost:3000/api/providers/1000000000004/configured_systems?offset=0"
    }
}
```
/cc @lpichler

### DIFF
--- a/spec/requests/providers_spec.rb
+++ b/spec/requests/providers_spec.rb
@@ -1808,6 +1808,7 @@ describe "Providers API" do
   context 'configuration_profiles subcollection' do
     let(:configuration_profile) { FactoryBot.create(:configuration_profile, :manager => ems) }
     let(:ems) { FactoryBot.create(:configuration_manager) }
+    let(:ems_vmware) { FactoryBot.create(:ems_vmware) }
 
     context 'GET /api/providers/:id/configuration_profiles' do
       it 'returns the configuration_profiles with an appropriate role' do
@@ -1828,6 +1829,15 @@ describe "Providers API" do
         get(api_provider_configuration_profiles_url(nil, ems))
 
         expect(response).to have_http_status(:forbidden)
+      end
+
+      it "returns a 404 for a provider that does not support configuration_profiles" do
+        api_basic_authorize subcollection_action_identifier(:providers, :configuration_profiles, :read, :get)
+
+        ems_vmware
+        get(api_provider_configuration_profiles_url(nil, ems_vmware))
+
+        expect(response).to have_http_status(:not_found)
       end
     end
 
@@ -1854,6 +1864,7 @@ describe "Providers API" do
   context 'configured_systems subcollection' do
     let(:configured_system) { FactoryBot.create(:configured_system, :manager => ems) }
     let(:ems) { FactoryBot.create(:configuration_manager) }
+    let(:ems_vmware) { FactoryBot.create(:ems_vmware) }
 
     context 'GET /api/providers/:id/configured_systems' do
       it 'returns the configured_systems with an appropriate role' do
@@ -1874,6 +1885,15 @@ describe "Providers API" do
         get(api_provider_configured_systems_url(nil, ems))
 
         expect(response).to have_http_status(:forbidden)
+      end
+
+      it "returns a 404 for a provider that does not support configured_systems" do
+        api_basic_authorize subcollection_action_identifier(:providers, :configured_systems, :read, :get)
+
+        ems_vmware
+        get(api_provider_configured_systems_url(nil, ems_vmware))
+
+        expect(response).to have_http_status(:not_found)
       end
     end
 
--- a/spec/requests/providers_spec.rb
+++ b/spec/requests/providers_spec.rb
@@ -1831,13 +1831,14 @@ describe "Providers API" do
         expect(response).to have_http_status(:forbidden)
       end
 
-      it "returns a 404 for a provider that does not support configuration_profiles" do
+      it "returns an empty collection for a provider that does not support configuration_profiles" do
         api_basic_authorize subcollection_action_identifier(:providers, :configuration_profiles, :read, :get)
 
         ems_vmware
         get(api_provider_configuration_profiles_url(nil, ems_vmware))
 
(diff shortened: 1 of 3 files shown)


lpichler:
Relations `configured_systems` and `configuration_profiles` belong to just AutomationManagers and ConfigurationManagers and so it can be reason why relations should stay here. I guess such type of issue can happen only with `/providers` endpoint.

Question:

How to tackle queries on subcollections:

```
GET /api/providers/?expand=resources&attributes=configuration_profiles

{
    "error": {
        "kind": "bad_request",
        "message": "Invalid attributes specified: configuration_profiles",
    }
}
```

but 

```
GET /api/providers/?expand=resources&attributes=configuration_profiles&filter[]=id=1

{
    "name": "providers",
    "count": 4,
    "subcount": 1,
    "subquery_count": 1,
    "pages": 1,
    "resources": [
        {
            "href": "http://localhost:3090/api/providers/1",
            "id": "1",
            "name": "Embedded Ansible",
            "created_on": "2020-07-10T13:20:27Z",
            "updated_on": "2020-07-10T13:27:33Z",
            "guid": "e9b5f3fd-c5f0-4d2d-a2cd-0843e656be1f",
            "zone_id": "2",
            "type": "ManageIQ::Providers::EmbeddedAnsible::AutomationManager",
            "api_version": null,
            "uid_ems": null,
            "host_default_vnc_port_start": null,
            "host_default_vnc_port_end": null,
            "provider_region": null,
            "last_refresh_error": "uninitialized constant ManageIQ::Providers::EmbeddedAnsible::Inventory",
            "last_refresh_date": "2020-07-10T13:27:33Z",
            "provider_id": "1",
            "realm": null,
            "tenant_id": "1",
            "project": null,
            "parent_ems_id": null,
            "subscription": null,
            "last_metrics_error": null,
            "last_metrics_update_date": null,
            "last_metrics_success_date": null,
            "tenant_mapping_enabled": null,
            "enabled": true,
            "options": null,
            "zone_before_pause_id": null,
            "last_inventory_date": null,
            "configuration_profiles": [],
            ...
```


so I am thinking whether is better to return just `[]` or `nil`. What do think about it?

bdunne:
I think it's important to maintain a consistent interface.  So, I think it would be better to return `[]` rather than `nil`.  I think we could do that by moving the relation to the base class.

agrare:
That would fix this subcollection, but there are _lots_ of associations that only exist on certain types of providers.  For example asking for any of [these](https://github.com/ManageIQ/manageiq/blob/master/app/models/manageiq/providers/cloud_manager.rb#L17-L29) off of anything other than a CloudManager will fail in the same way.  I don't know that it is feasible to move every association from every provider sub-type into the base class

agrare:
> For the other case, where this endpoint returns Providers rather than ExtManagementSystems, I don't know what to do with that. At some point we need to fix this interface.

Yeah the `/providers` endpoint is a mess

gtanzillo:
Based on @agrare comments, it seems like the safest change is to the subcollections. But I'll change to return an empty list instead of a 404, as suggested.

abellotti:
empty list is the right way, as the subcollection endpoint does exist,   other subcollections do the same,

i.e.  ```object.try(:cloud_networks) || []```. in the *_query_resource method.

spec/requests/providers_spec.rb
```diff
@@ -1875,6 +1887,16 @@ class DummyProvider; end
 
         expect(response).to have_http_status(:forbidden)
       end
+
+      it "returns a 404 for a provider that does not support configured_systems" do
```
abellotti, line 1891: minor, maybe update this and the PR description.

gtanzillo: Definitely, I meant to update the test case descriptions and forgot. Though I updated the PR description but I must have forgotten to save. I'll take care of both. Thanks!

abellotti:
LGTM !!  Thanks @gtanzillo  for fixing this. 👍

simaishi:
Jansa backport details:

```
$ git log -1
commit 30bcf0c130d9f87274aae113056bda3c70dff529
Date:   Wed Jul 15 08:11:48 2020 -0400

    Merge pull request #864 from gtanzillo/return-404

    Return an empty collection, instead of 500 server error, for providers that don't support configured_systems or configuration_profiles subcollections

    (cherry picked from commit e8db32cb0995bd733888f45f36f41a00dc80ae8b)
```

PR #873, opened 2020-07-24
Allow set_ownership action for service templates

lpichler:
When set ownership screen is saved on catalog item, there is error message:

Screenshot 2020-07-24 at 21 03 03

Custom actions for service template were not added in api.yml.

@miq-bot assign @gtanzillo 
@miq-bot add_label bug


abellotti:
can you add specs for set_ownership for service templates ?   Thanks.

PR #557, opened 2019-02-07
Performance updates for virtual_attribute_search

NickLaMuro:
Fixes two simple performance issues:

- Avoids trying to Rbac records that don't need to be "Rbac'd" (first commit)
- Avoids repeating Rbac on records that have already been checked in the initial `collection_search` (second commit)

Benchmarks
----------

**Before**

```
|   ms | queries | query (ms) | rows |
| ---: |    ---: |       ---: | ---: |
| 2197 |     322 |      194.4 | 1627 |
|  953 |     321 |      182.7 |  196 |
|  996 |     321 |      187.1 |  196 |
```

**After** (first commit)

```
|   ms | queries | query (ms) | rows |
| ---: |    ---: |       ---: | ---: |
| 2274 |     282 |      166.2 | 1612 |
|  858 |     281 |      147.9 |  181 |
|  960 |     281 |      164.2 |  181 |
```

**After** (second commit)

```
|   ms | queries | query (ms) | rows |
| ---: |    ---: |       ---: | ---: |
| 2004 |     263 |      156.8 | 1593 |
|  756 |     261 |      142.9 |  161 |
|  769 |     261 |      138.9 |  161 |
```

TODO
----

* [x] Make sure existing tests pass...
* [x] Add Tests

Links
-----

* Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1656371
* Related to changes in https://github.com/ManageIQ/manageiq/pull/15145

Issue #869, opened 2020-07-24
Slow but small response for virtual columns seemingly due to N+1 queries


jrafanie:
From a customer's DB, a request such as:

`/api/vms?expand=resources&attributes=num_cpu,name`

is only a 64KB response but takes 1.6 seconds.

```
joerafaniello@Joes-MBP-2 manageiq % bundle exec miqperf benchmark -ac 5 -d "/api/vms?expand=resources&attributes=num_cpu,name"
D, [2020-07-24T14:03:52.849756 #29451] DEBUG -- : --> logging in...
D, [2020-07-24T14:03:53.124855 #29451] DEBUG -- : --> making GET request: /api/vms?expand=resources&attributes=num_cpu,name
D, [2020-07-24T14:03:54.854853 #29451] DEBUG -- : --> making GET request: /api/vms?expand=resources&attributes=num_cpu,name
D, [2020-07-24T14:03:56.618859 #29451] DEBUG -- : --> making GET request: /api/vms?expand=resources&attributes=num_cpu,name
D, [2020-07-24T14:03:58.366955 #29451] DEBUG -- : --> making GET request: /api/vms?expand=resources&attributes=num_cpu,name
D, [2020-07-24T14:04:00.306892 #29451] DEBUG -- : --> making GET request: /api/vms?expand=resources&attributes=num_cpu,name
joerafaniello@Joes-MBP-2 manageiq % bundle exec miqperf report -l
/api/vms
|   ms | queries | query (ms) | rows |
| ---: |    ---: |       ---: | ---: |
| 1624 |    1117 |      640.8 | 1666 |
| 1650 |    1116 |      653.0 | 1665 |
| 1638 |    1116 |      638.1 | 1665 |
| 1699 |    1116 |      688.8 | 1665 |
| 1681 |    1116 |      637.1 | 1665 |
```

related to https://github.com/ManageIQ/manageiq-api/issues/805

Fryguy:
https://github.com/ManageIQ/manageiq/blob/cca2bd8e9c8839deff16ff62a2be49558885f8f4/app/models/vm_or_template.rb#L178

Feels like the code is either
- not preloading at all
- not detecting that this delegate goes through hardwares and thus not preloading it
- not supporting a `:uses => :hardwares` for preloading

NickLaMuro:
@miq-bot assign @NickLaMuro

PR #874, opened 2020-07-24
[renderer.rb] Don't re-Rbac: part deux

NickLaMuro:
This is an update 1ad20d24 where it better supports skipping Rbac when the model has already been filteerd in the initial .collection_search.

By using `base_model`, we ensure that any subclass will still work, specifically with our STI classes like `Vm`.

Links
-----

* Partially addresses: https://github.com/ManageIQ/manageiq-api/issues/869
* Original Fix:  https://github.com/ManageIQ/manageiq-api/pull/557

### DIFF
--- a/spec/requests/vms_spec.rb
+++ b/spec/requests/vms_spec.rb
@@ -39,6 +39,51 @@ describe "Vms API" do
     end
   end
 
+  context 'Vm index' do
+    before do
+      # Will not be included in the result (base model is Vm)
+      FactoryBot.create(:template)
+      FactoryBot.create(:template)
+
+      # Preload records
+      _vms = [vm, vm1, vm2, vm_openstack, vm_openstack1, vm_openstack2]
+
+      # Only once for the main query
+      expect(Rbac).to receive(:filtered).exactly(1).times.and_call_original
+      expect(Rbac).to receive(:filtered_object).never
+    end
+
+    it "lists all of the vms" do
+      api_basic_authorize action_identifier(:vms, :read, :resource_actions, :get)
+      get api_vms_url, :params => {:expand => "resources", :attributes => "num_cpu,name"}
+
+      expect(response.parsed_body["subcount"]).to eq 6
+    end
+
+    it "properly filters with Rbac" do
+      # Add tenant permissions
+      tenant = FactoryBot.create(:tenant)
+      role   = FactoryBot.create(:miq_user_role)
+      group  = FactoryBot.create(:miq_group, :tenant => tenant, :miq_user_role => role)
+
+      # Update user permissions
+      @user.update(:miq_groups => [group])
+      @role = role
+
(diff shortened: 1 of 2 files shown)


spec/requests/vms_spec.rb
```diff
@@ -39,6 +39,25 @@ def update_raw_power_state(state, *vms)
     end
   end
 
+  context 'Vm index' do
+    # Preload Vm recordss
```
Fryguy, line 43: typo `recordss` -> `records`

NickLaMuro: But two `s`s makes it cool!

jrafanie: you can leave off the second s for "savings"

spec/requests/vms_spec.rb
```diff
@@ -39,6 +39,25 @@ def update_raw_power_state(state, *vms)
     end
   end
 
+  context 'Vm index' do
+    # Preload Vm records
+    before do
+      @_vms = [vm, vm1, vm2, vm_openstack, vm_openstack1, vm_openstack2]
```
Fryguy, line 45: Can you add a preload here for other records from the same table that _won't_ be in the query results?  For example, VMs that aren't in RBAC and perhaps a template? (Same table, different model).  Asking about teplates specifically because you chose `base_model` over `base_class`, which I assume was for a specific reason.

jrafanie: > teplates

so close to tempates

Fryguy: Fixed 😉

NickLaMuro: > Asking about teplates specifically because you chose base_model over base_class, which I assume was for a specific reason.

I tested using both `base_model` and `base_class`, and there was still an N+1 with `base_class`.  But I can try again with this spec now, and in doing so, it fails to prevent `Rbac.filtered_object` from being called.

NickLaMuro: > Fixed 😉

Not before I could quote it above.  Muhahahahaha!

Fryguy: @NickLaMuro Ok, that's good..still think we should have specs that cover bonus vms as well as bonus templates. WDYT?

NickLaMuro: @Fryguy I think I have this covered, but let me know what you think.

simaishi:
Jansa backport details:

```
$ git log -1
commit 8e471ba671b126070b2c162b795193504e0b9b0a
Date:   Mon Aug 3 10:51:06 2020 -0400

    Merge pull request #874 from NickLaMuro/better-rbac-skip-filter

    [renderer.rb] Don't re-Rbac: part deux

    (cherry picked from commit 0293d947fdb6c39f77aa2523cef0ac69740dedcf)
```

Issue #870, opened 2020-07-24
Add support for returning nil/null in associations: 


jrafanie:
This was done for direct resource attributes in #253.  It needs to be done for associations such as those used in virtual attributes.

For example:
`/api/providers?expand=resources&attributes=last_compliance_status`

will return `last_compliance_status` only for providers that have a compliance and not be included in the result for those without it, resulting in inconsistent results for each resource.

Another example using an association's attribute that's sometimes null:
`/api/vms?&expand=resources&attributes=ext_management_system.last_refresh_date,name`

```
        {
            "href": "http://localhost:3000/api/vms/755",
            "name": "REDACTED",
            "id": "755",
            "ext_management_system": {
                "last_refresh_date": "2020-07-21T19:02:23Z"
            }
        },
        {
            "href": "http://localhost:3000/api/vms/759",
            "name": "REDACTED2",
            "id": "759",
            "ext_management_system": {
                "last_refresh_date": "2020-07-21T19:02:23Z"
            }
        },
        {
            "href": "http://localhost:3000/api/vms/375",
            "name": "REDACTED3",
            "id": "375"
        },
        {
            "href": "http://localhost:3000/api/vms/376",
            "name": "REDACTED4,
            "id": "376"
        },
        {
            "href": "http://localhost:3000/api/vms/377",
            "name": "REDACTED5",
            "id": "377"
        },
```

bdunne:
Fixed by #875

PR #875, opened 2020-07-24
Updated the API to return nil virtual attributes and associations

abellotti:
- Updated the API so virtual attributes and associations (relationships) are returned even if nil.

Solves: https://github.com/ManageIQ/manageiq-api/issues/870

### DIFF
--- a/app/controllers/api/base_controller/renderer.rb
+++ b/app/controllers/api/base_controller/renderer.rb
@@ -268,7 +268,7 @@ module Api
                                 else
                                   fetch_indirect_virtual_attribute(type, resource, attr_base, attr_name, object_hash)
                                 end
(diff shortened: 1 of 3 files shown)


jrafanie:
I think this should go back to jansa considering how small a change this is.

simaishi:
Jansa backport details:

```
$ git log -1
commit 4013d35a4c256376a9690c0b1413a71f92266d94
Date:   Mon Jul 27 16:05:13 2020 -0400

    Merge pull request #875 from abellotti/return_nil_associations

    Updated the API to return nil virtual attributes and associations

    (cherry picked from commit 8665c2d62d42cceed33d665b9be60b2bc41580a6)
```

PR #876, opened 2020-07-27
Use shared example in set ownership screen spec

lpichler:
this is useful for specs in https://github.com/ManageIQ/manageiq-api/pull/873


abellotti:
Nice!! LGTM @lpichler   thanks for the test updates 👍

PR #881, opened 2020-08-11
fix alert definitions specs

d-m-u:
recent policy changes mean that we need to update these tests, broken in https://github.com/ManageIQ/manageiq/pull/20381

cross repo: https://github.com/ManageIQ/manageiq-cross_repo-tests/pull/159 

~~related to https://github.com/ManageIQ/manageiq/pull/20430 but doesn't need to be merged alongside (the repo will be red till both are merged though)~~ as Keenan pointed out, they're totally unrelated, but they broke in the same PR as what 20430 is tryin' to fix (the totally unrelated related changes)


d-m-u:
Stems from a different problem, you're right, are totally unrelated.

Issue #867, opened 2020-07-22
Support "ascending" and "descending" aliases for sort_by


Fryguy:
When querying if you spell out "ascending" or "descending" they are not honored, and it can be confusing.  Seems simple enough to support these aliases for sorting.  I'd even be ok with the first "n" characters matching (i.e. "describe" would happen to work too because it starts with "desc")

PR #883, opened 2020-08-18
allow specification of sort_order with abbreviation or full form

d-m-u:
allow ascending and descending in full as sort options

there's an open Jason issue about it and I need a commit for the day, sorry for going for the 🍒 

fixes https://github.com/ManageIQ/manageiq-api/issues/867

### DIFF
--- a/app/controllers/api/base_controller/parameters.rb
+++ b/app/controllers/api/base_controller/parameters.rb
@@ -3,7 +3,7 @@ module Api
     module Parameters
       module ResultsController
         def sort_order
-          params['sort_order'] == 'desc' ? :descending : :ascending
+          %w[desc descending].include?(params['sort_order']) ? :descending : :ascending
         end
 
         def param_result_set?
@@ -93,8 +93,7 @@ module Api
         arel = klass.arel_attribute(attr)
         if order
           arel = arel.lower if options.map(&:downcase).include?("ignore_case")
-          arel = arel.desc if order.downcase == "desc"
-          arel = arel.asc if order.downcase == "asc"
+          arel = %w[desc descending].include?(order.downcase) ? arel.desc : arel.asc
         else
           arel = arel.asc
         end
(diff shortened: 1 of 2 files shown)


d-m-u:
I guess

app/controllers/api/base_controller/parameters.rb
```diff
@@ -93,8 +93,8 @@ def sort_directive(klass, attr, order, options)
         arel = klass.arel_attribute(attr)
         if order
           arel = arel.lower if options.map(&:downcase).include?("ignore_case")
-          arel = arel.desc if order.downcase == "desc"
-          arel = arel.asc if order.downcase == "asc"
+          arel = arel.desc if %w[desc descending].include?(order.downcase)
+          arel = arel.asc if %w[asc ascending].include?(order.downcase)
```
kbrock, line 97:
```suggestion
          arel = (sort_order == :ascending ? arel.asc : arel.desc) if order.present?
```

maybe the `if` needs to be a block

kbrock:
can the user not specify the sort order?
is it set either way?

if we do add `asc` when the sort order is not passed, does that change the order?
wondering if this will raise an error if the user is sorting by a virtual attribute

app/controllers/api/base_controller/parameters.rb
kbrock: since we are only going here if order is specified, I think we can get rid of the if on this line (and then obviously we don't need if in a block because it is going away)

kbrock:
pipe dream: can we get rid of that whole block?
just calling ascend or descend

will need to add the case insensitivity back in
so this may be a total and utterly useful tangent

app/controllers/api/base_controller/parameters.rb
d-m-u: i don't believe you about using `sort_order` here, sorry

d-m-u:
does someone want to merge this, please?

PR #886, opened 2020-08-27
Plugin Generator Updates

Fryguy:
Plugin Generator Updates

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,17 +1,12 @@
-source 'https://rubygems.org'
-
+# Declare your gem's dependencies in manageiq-api.gemspec.
+# Bundler will treat runtime dependencies like base dependencies, and
+# development dependencies will be added by default to the :development group.
 gemspec
 
-# Load Gemfile with dependencies from manageiq
-manageiq_gemfile = File.expand_path("spec/manageiq/Gemfile", __dir__)
-if File.exist?(manageiq_gemfile)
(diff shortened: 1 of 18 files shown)


Gemfile
```diff
@@ -1,17 +1,12 @@
-source 'https://rubygems.org'
-
+# Declare your gem's dependencies in manageiq-api.gemspec.
+# Bundler will treat runtime dependencies like base dependencies, and
+# development dependencies will be added by default to the :development group.
 gemspec
 
-# Load Gemfile with dependencies from manageiq
```
bdunne, line 5: Should this have changed?

Fryguy: Yes...the plugin generator no longer has this.  A lot of it is ceremony that is now handled by bin/setup and bin/update

Fryguy: Note we still load it (see line https://github.com/ManageIQ/manageiq-api/pull/886/files/390224f129d78f202a5da097a968d7ac42bcc0aa#diff-8b7db4d5cc4b8f6dc8feb7030baa2478R11-R12 )

Fryguy:
Note that the README changes are removed because they are all in the plugin developer doc.

agrare (approved):

Gemfile
agrare: I think this is a helpful error message, but we should move it to the generator and keep it consistent

Issue #872, opened 2020-07-24
Querying for `&attributes=hardware.cpu_sockets` is 0.5 seconds faster than `num_cpu`, WAT


jrafanie:
```
joerafaniello@Joes-MBP-2 manageiq % bundle exec miqperf benchmark -ac 5 -d "/api/vms?expand=resources&attributes=name,hardware.cpu_sockets"
D, [2020-07-24T14:55:14.826645 #30092] DEBUG -- : --> logging in...
D, [2020-07-24T14:55:15.048047 #30092] DEBUG -- : --> making GET request: /api/vms?expand=resources&attributes=name,hardware.cpu_sockets
D, [2020-07-24T14:55:16.068500 #30092] DEBUG -- : --> making GET request: /api/vms?expand=resources&attributes=name,hardware.cpu_sockets
D, [2020-07-24T14:55:17.008414 #30092] DEBUG -- : --> making GET request: /api/vms?expand=resources&attributes=name,hardware.cpu_sockets
D, [2020-07-24T14:55:18.018111 #30092] DEBUG -- : --> making GET request: /api/vms?expand=resources&attributes=name,hardware.cpu_sockets
D, [2020-07-24T14:55:18.906843 #30092] DEBUG -- : --> making GET request: /api/vms?expand=resources&attributes=name,hardware.cpu_sockets
joerafaniello@Joes-MBP-2 manageiq % bundle exec miqperf report -l
/api/vms
|   ms | queries | query (ms) | rows |
| ---: |    ---: |       ---: | ---: |
|  952 |     563 |      405.8 | 1112 |
|  875 |     563 |      392.4 | 1112 |
|  884 |     563 |      393.9 | 1112 |
|  824 |     563 |      371.7 | 1112 |
|  890 |     563 |      392.8 | 1112 |
```

```
joerafaniello@Joes-MBP-2 manageiq % bundle exec miqperf benchmark -ac 5 -d "/api/vms?expand=resources&attributes=name,num_cpu"
D, [2020-07-24T14:55:46.390251 #30111] DEBUG -- : --> logging in...
D, [2020-07-24T14:55:46.600145 #30111] DEBUG -- : --> making GET request: /api/vms?expand=resources&attributes=name,num_cpu
D, [2020-07-24T14:55:48.762894 #30111] DEBUG -- : --> making GET request: /api/vms?expand=resources&attributes=name,num_cpu
[...]

PR #887, opened 2020-08-29
AR VirtualAttribute queries + introspection

### LINKED ISSUE #880: API Performance
Issues related to improving the performance of the ManageIQ Rest API

- [x] [Slow but small response for virtual columns seemingly due to N+1 queries](https://github.com/ManageIQ/manageiq-api/issues/869)
  - [x] _[renderer.rb] Don't re-Rbac: part deux:_  #874
  - [x] _AR VirtualAttribute queries + introspection:_ ~~#877~~ https://github.com/ManageIQ/manageiq-api/pull/887 https://github.com/ManageIQ/manageiq-api/pull/890
- [x] Improve user login timings
  - [x] ~~_Use~~ ~~`.update_attribute`~~ ~~over~~ ~~`.save!`~~ ~~for User login:_ ManageIQ/manageiq#20471~~
  - [x] Reduce queries for `User#save` validations:  https://github.com/ManageIQ/manageiq/pull/20590
- [x] [Add support for returning nil/null in associations](https://github.com/ManageIQ/manageiq-api/issues/870)
- [x] [Querying for `&attributes=hardware.cpu_sockets` is 0.5 seconds faster than `num_cpu`, WAT](https://github.com/ManageIQ/manageiq-api/issues/872)

NickLaMuro:
Alternative to https://github.com/ManageIQ/manageiq-api/pull/877

Uses introspection of the url params to determine which attributes can be either preloaded via `:include_for_find` or `:extra_cols`.

TODO
----

Follow PR most likely for these

- [x] Handle nested relations
- [ ] Handle `:uses` for `virtual_columns` https://github.com/ManageIQ/manageiq-api/pull/890
- [x] Ensure this works with multiple columns on the same relation (testing this in https://github.com/ManageIQ/manageiq-api/pull/890 as well)

Performance Numbers
-------------------

Before
------

```
$ bundle exec miqperf benchmark -ac 5 "/api/vms?expand=resources&attributes=name,hardware.cpu_sockets"
| 2341 |     565 |      179.0 | 2594 |
|  792 |     563 |      165.0 | 1112 |
|  897 |     563 |      171.5 | 1112 |
|  746 |     563 |      153.9 | 1112 |
|  812 |     563 |      154.7 | 1112 |
```

After
-----

Address both slow scenarios in https://github.com/ManageIQ/manageiq-api/issues/872

(Note:  Both of these results were taken within the same server process, so the second set of results doesn't have a slower first request due to autoloading)

```
$ bundle exec miqperf benchmark -ac 5 "/api/vms?expand=resources,hardware&attributes=num_cpu,name"
| 1516 |      12 |         74 | 2041 |
|  359 |      10 |        7.7 |  559 |
|  197 |      10 |        7.3 |  559 |
|  259 |      10 |        7.9 |  559 |
|  198 |      10 |        7.3 |  559 |
```

```
$ bundle exec miqperf benchmark -ac 5 "/api/vms?expand=resources&attributes=name,hardware.cpu_sockets"
|  369 |      10 |         14 |  559 |
|  309 |      10 |        9.1 |  559 |
|  267 |      10 |        8.8 |  559 |
|  308 |      10 |        9.1 |  559 |
|  262 |      10 |        8.4 |  559 |
```

Links
-----

- Partially addresses https://github.com/ManageIQ/manageiq-api/issues/880 and https://github.com/ManageIQ/manageiq-api/issues/869
- Should fix https://github.com/ManageIQ/manageiq-api/issues/872 as well

### DIFF
--- a/app/controllers/api/base_controller/renderer.rb
+++ b/app/controllers/api/base_controller/renderer.rb
@@ -185,6 +185,8 @@ module Api
         options[:filter] = miq_expression if miq_expression
         options[:offset] = params['offset'] if params['offset']
         options[:limit] = params['limit'] if params['limit']
+        options[:extra_cols] = determine_extra_cols(klass)
+        options[:include_for_find] = determine_include_for_find(klass)
 
         filter_results(miq_expression, res, options)
       end
@@ -543,6 +545,54 @@ module Api
       def render_options(resource, data = {})
         klass = collection_class(resource)
         render :json => OptionsSerializer.new(klass, data).serialize
+      end
+
+      # This is a helper method used by both .determine_include_for_find and
+      # .determine_extra_cols to collect and filter virtual_attributes for the
+      # :include_for_find and :extra_cols options that are passed to Rbac. The
+      # intent is to reduce a large amount of shared code between those two
+      # shared methods by combining them into this one.
+      #
+      # The required block used by each of aforementioned methods is used to do
+      # custom filtering that pertains to each of those methods.
+      #
+      def virtual_attributes_for(klass)
+        return nil unless klass.respond_to?(:reflect_on_association)
+
+        type    = @req.subject
+        results = []
+
+        validate_attr_selection(klass).last.each do |vattr|
+          next if vattr == "href_slug"
+
+          attr_name, attr_base = split_virtual_attribute(vattr)
+          filtered_attr        = yield type, attr_name, attr_base
+
+          results << filtered_attr if filtered_attr
+        end
+
+        results.empty? ? nil : results
+      end
+
+      def determine_include_for_find(klass)
+        virtual_attributes_for(klass) do |type, attr_name, attr_base|
+          next if attr_base.blank?
+          next if attr_base.include?(".") # FIXME: Allow nested relations
+          next if virtual_attribute_accessor(type, attr_name)
+          next if Rbac::Filterer::CLASSES_THAT_PARTICIPATE_IN_RBAC.include?(attr_base)
+
+          attr_base
+        end
+      end
+
+      def determine_extra_cols(klass)
+        virtual_attributes_for(klass) do |type, attr_name, attr_base|
+          return if attr_base.present?
+          return if virtual_attribute_accessor(type, attr_name)
+          return unless klass.attribute_supported_by_sql?(attr_name)
+
+          attr_name.to_sym
+        end
       end
     end
   end
--- a/app/controllers/api/base_controller/renderer.rb
+++ b/app/controllers/api/base_controller/renderer.rb
@@ -575,13 +575,20 @@ module Api
       end
 
       def determine_include_for_find(klass)
-        virtual_attributes_for(klass) do |type, attr_name, attr_base|
+        attrs = virtual_attributes_for(klass) do |type, attr_name, attr_base|
           next if attr_base.blank?
-          next if attr_base.include?(".") # FIXME: Allow nested relations
           next if virtual_attribute_accessor(type, attr_name)
           next if Rbac::Filterer::CLASSES_THAT_PARTICIPATE_IN_RBAC.include?(attr_base)
 
           attr_base
+        end
+
+        # Handle nested relationships and convert to a hash
+        if attrs
+          attrs.each_with_object({}) do |key, include_for_find|
+            nested = include_for_find
+            key.split(".").each { |k| nested = nested[k] ||= {} }
+          end
         end
       end
 
--- a/app/controllers/api/base_controller/renderer.rb
+++ b/app/controllers/api/base_controller/renderer.rb
@@ -350,15 +350,19 @@ module Api
       def attr_virtual?(object, attr)
         return false if ID_ATTRS.include?(attr)
         primary = attr_split(attr).first
-        (object.class.respond_to?(:reflect_on_association) && object.class.reflect_on_association(primary)) ||
-          (object.class.respond_to?(:virtual_attribute?) && object.class.virtual_attribute?(primary)) ||
-          (object.class.respond_to?(:virtual_reflection?) && object.class.virtual_reflection?(primary))
+        klass   = object
+        klass   = object.class if object.kind_of?(ActiveRecord::Base)
+        (klass.respond_to?(:reflect_on_association) && klass.reflect_on_association(primary)) ||
+          (klass.respond_to?(:virtual_attribute?) && klass.virtual_attribute?(primary)) ||
+          (klass.respond_to?(:virtual_reflection?) && klass.virtual_reflection?(primary))
       end
 
       def attr_physical?(object, attr)
         return true if ID_ATTRS.include?(attr)
-        (object.class.respond_to?(:has_attribute?) && object.class.has_attribute?(attr)) &&
-          !(object.class.respond_to?(:virtual_attribute?) && object.class.virtual_attribute?(attr))
+        klass = object
+        klass = object.class if object.kind_of?(ActiveRecord::Base)
+        (klass.respond_to?(:has_attribute?) && klass.has_attribute?(attr)) &&
+          !(klass.respond_to?(:virtual_attribute?) && klass.virtual_attribute?(attr))
       end
 
       def attr_split(attr)
--- a/spec/requests/vms_spec.rb
+++ b/spec/requests/vms_spec.rb
@@ -83,8 +83,10 @@ describe "Vms API" do
                                     {"vendor" => "openstack"}])
     end
 
-    def add_hardware_to_vms
+    def add_hardware_and_os_to_vms
       [vm, vm1, vm2, vm_openstack, vm_openstack1, vm_openstack2].each do |vm_record|
+        cs = ComputerSystem.create
+        OperatingSystem.create(:name => "linux", :vm_or_template => vm_record, :computer_system => cs)
         FactoryBot.create(:hardware, :vm_or_template => vm_record, :host => host)
       end
     end
@@ -93,8 +95,24 @@ describe "Vms API" do
       /SELECT.*FROM\s"(?:#{tables.flatten.join("|")})"/m
     end
 
+    context "with nested indirect virtual attribute ('operating_system.computer_system.created_at')" do
+      before { add_hardware_and_os_to_vms }
+
+      it "removes N+1's from the index query for subcollections/virtual_attributes" do
+        api_basic_authorize action_identifier(:vms, :read, :resource_actions, :get)
+        query_match = query_match_regexp("vms", "operating_systems", "computer_systems")
+
+        expect {
+          get api_vms_url, :params => {
+            :expand     => "resources",
+            :attributes => "operating_system.computer_system.created_at,name"
+          }
+        }.to make_database_queries(:count => 3, :matching => query_match)
+      end
+    end
+
     context "with indirect virtual attribute ('hardware.cpu_sockets')" do
-      before { add_hardware_to_vms }
+      before { add_hardware_and_os_to_vms }
 
       it "removes N+1's from the index query for subcollections/virtual_attributes" do
         api_basic_authorize action_identifier(:vms, :read, :resource_actions, :get)
@@ -110,7 +128,7 @@ describe "Vms API" do
     end
 
     context "with direct virtual attribute ('num_cpus')" do
-      before { add_hardware_to_vms }
+      before { add_hardware_and_os_to_vms }
 
       it "removes N+1's from the index query for subcollections/virtual_attributes" do
         api_basic_authorize action_identifier(:vms, :read, :resource_actions, :get)
--- a/spec/requests/vms_spec.rb
+++ b/spec/requests/vms_spec.rb
@@ -82,6 +82,48 @@ describe "Vms API" do
                                    [{"vendor" => "openstack"},
                                     {"vendor" => "openstack"}])
     end
+
+    def add_hardware_to_vms
+      [vm, vm1, vm2, vm_openstack, vm_openstack1, vm_openstack2].each do |vm_record|
+        FactoryBot.create(:hardware, :vm_or_template => vm_record, :host => host)
+      end
+    end
+
+    def query_match_regexp(*tables)
+      /SELECT.*FROM\s"(?:#{tables.flatten.join("|")})"/m
+    end
+
+    context "with indirect virtual attribute ('hardware.cpu_sockets')" do
+      before { add_hardware_to_vms }
+
+      it "removes N+1's from the index query for subcollections/virtual_attributes" do
+        api_basic_authorize action_identifier(:vms, :read, :resource_actions, :get)
+        query_match = query_match_regexp("vms", "hardwares")
+
+        expect {
+          get api_vms_url, :params => {
+            :expand     => "resources",
+            :attributes => "hardware.cpu_sockets,name"
+          }
+        }.to make_database_queries(:count => 3, :matching => query_match)
+      end
+    end
+
+    context "with direct virtual attribute ('num_cpus')" do
+      before { add_hardware_to_vms }
+
+      it "removes N+1's from the index query for subcollections/virtual_attributes" do
+        api_basic_authorize action_identifier(:vms, :read, :resource_actions, :get)
+        query_match = query_match_regexp("vms", "hardwares")
+
+        expect {
+          get api_vms_url, :params => {
+            :expand     => "resources",
+            :attributes => "num_cpu,name"
+          }
+        }.to make_database_queries(:count => 3, :matching => query_match)
+      end
+    end
   end
 
   context 'Vm edit' do
--- a/app/controllers/api/base_controller.rb
+++ b/app/controllers/api/base_controller.rb
@@ -48,13 +48,14 @@ module Api
     def index
       klass = collection_class(@req.subject)
       res, subquery_count = collection_search(@req.subcollection?, @req.subject, klass)
+      res_count = (res.kind_of?(ActiveRecord::Relation) ? res.except(:select) : res).count
       opts = {
         :name                  => @req.subject,
         :is_subcollection      => @req.subcollection?,
         :expand_actions        => true,
         :expand_custom_actions => false,
         :expand_resources      => @req.expand?(:resources),
-        :counts                => Api::QueryCounts.new(klass.count, res.count, subquery_count)
+        :counts                => Api::QueryCounts.new(klass.count, res_count, subquery_count)
       }
       render_collection(@req.subject, res, opts)
     end


NickLaMuro:
@miq-bot assign @jrafanie 
@miq-bot add_labels performance, jansa/yes?

cc @abellotti @Fryguy 

Not sure if this deserved to be backported, but figured I would mark the same as #877

app/controllers/api/base_controller/renderer.rb
```diff
@@ -540,6 +546,47 @@ def render_options(resource, data = {})
...
+        virtual_attributes_for(klass).each do |vattr|
+          attr_name, attr_base = split_virtual_attribute(vattr)
+          next if attr_base.present?
+          next if virtual_attribute_accessor(type, attr_name)
+          next unless klass.attribute_supported_by_sql?(attr_name)
+
+          extra_cols << attr_name.to_sym
+        end
+
+        extra_cols.empty? ? nil : extra_cols
```
jrafanie, line 588: Both of these methods look similar.  Is this common behavior something worthy of a named method?

NickLaMuro: I might be able to push some of this shared logic into `virtual_attributes_for`... let me test some things out with that.

jrafanie: Well, I'm really wondering what is this common behavior called? It might make sense to name it since even I can see similarities 🤣   So, yeah, if there's parts that can be named as actual things, it might make sense to organize them by the behavior they're doing.  If not, that's fine also.  I'm just commenting on what appeared to be a "thing" that could want a name.

NickLaMuro: > Well, I'm really wondering what is this common behavior is called?

Well, unfortunately a lot of this is dissecting a bunch of already existing methods to pull out the piece(s) that are relevant to create these new ones.  The problem and why it isn't easy to just share these method fragments else where is the area that was pulled from (the `*_to_jbuilder` methods) fetch/expand the attributes for the result as it is generated, and so breaking down the attributes is done slowly in phases for each subcollection.

Where in the case here, we are trying determine if there are any relations to these methods up front so we don't have the N+1s.  The `fetch_direct_virtual_attribute` and `fetch_indirect_virtual_attribute` are where a lot of these came from, but there are some other places that I can't think of off the top of my head as well.

app/controllers/api/base_controller/renderer.rb
```diff
@@ -540,6 +546,47 @@ def render_options(resource, data = {})
...
+
+      def determine_include_for_find(klass)
+        return nil unless klass.respond_to?(:reflect_on_association)
+
+        type             = @req.subject
+        include_for_find = []
+
+        virtual_attributes_for(klass).each do |vattr|
+          attr_name, attr_base = split_virtual_attribute(vattr)
+          next if attr_base.blank?
```
NickLaMuro, line 562: Going to go through each of these individually and explain why they exist:

This one basically is checking if there is a `attr_base`, and if there is, then this is something like `hardware.cpu_sockets`.  In that case, `"hardware"` would be the result here so we would want to keep going.

app/controllers/api/base_controller/renderer.rb
```diff
@@ -540,6 +546,47 @@ def render_options(resource, data = {})
...
+          next if attr_base.include?(".") # FIXME: Allow nested relations
```
NickLaMuro, line 563: If the `attr_base` includes a `"."`, then it probably is something like `"hardware.host.name"` for the full attribute (and `"hardware.host"` for the `attr_base`), and this first iteration doesn't support that at the moment.

app/controllers/api/base_controller/renderer.rb
```diff
@@ -540,6 +546,47 @@ def render_options(resource, data = {})
...
+          next if virtual_attribute_accessor(type, attr_name)
```
NickLaMuro, line 564: These are custom methods written into the controllers with a `fetch_` prefix.  In this case, we don't want to mess with the custom code, so skip over these and just eat the performance hit.

app/controllers/api/base_controller/renderer.rb
```diff
@@ -540,6 +546,47 @@ def render_options(resource, data = {})
...
+          next if Rbac::Filterer::CLASSES_THAT_PARTICIPATE_IN_RBAC.include?(attr_base)
```
NickLaMuro, line 565: This is just ensuring we aren't circumventing any Rbac checks.  If the `attr_base` participates in `Rbac`, then skip it and let it be fetched normally.

app/controllers/api/base_controller/renderer.rb
```diff
@@ -540,6 +546,47 @@ def render_options(resource, data = {})
...
+
+      def determine_extra_cols(klass)
+        return nil unless klass.respond_to?(:reflect_on_association)
+
+        type       = @req.subject
+        extra_cols = []
+
+        virtual_attributes_for(klass).each do |vattr|
+          attr_name, attr_base = split_virtual_attribute(vattr)
+          next if attr_base.present?
```
NickLaMuro, line 581: Compared to the similar line in `determine_include_for_find`, this does the opposite.  If `attr_base` exists, that means it is exists as part of a relation, and isn't a "direct virtual attribute" on the top level collection.  Since we currently don't support nested virtual attributes, we skip.

app/controllers/api/base_controller/renderer.rb
```diff
@@ -540,6 +546,47 @@ def render_options(resource, data = {})
...
+          next if virtual_attribute_accessor(type, attr_name)
```
NickLaMuro, line 582: Same as https://github.com/ManageIQ/manageiq-api/pull/887/files#r481454895

app/controllers/api/base_controller/renderer.rb
```diff
@@ -540,6 +546,47 @@ def render_options(resource, data = {})
...
+          next unless klass.attribute_supported_by_sql?(attr_name)
```
NickLaMuro, line 583: This actually checks if we can use this `virtual_attribute` via SQL.  If so, continue.

Fryguy:
Is this a copy/similar to what we use in reporting?  include_for_find is literally the same thing it's called there, hence what triggered my mind - https://github.com/ManageIQ/manageiq/blob/69f14c83c53418b9549862e176934fd5afecefe5/app/models/miq_report/generator.rb#L339

NickLaMuro:
I would say "no" simply because what is being worked on there is a `Hash`:

https://github.com/ManageIQ/manageiq/blob/b8e8f2e6250f4497ab44c695c1c5c4c1c53262e8/product/reports/400_Operations-%20Virtual%20Machines/010_Registered%20VMs%20by%20Free%20Space.yaml#L28-L35

And what we are working with here is a 1D `Array` with values that can be nested.  We would have to transform the data and then feed it into that.  Unsure if it is worth the effort, or a route we want to go down, but I don't really see it replacing much of what I have here.

Might be helpful with the `TODO` comment though.

app/controllers/api/base_controller/renderer.rb
NickLaMuro: @jrafanie I took a stab at refactoring this to DRY things up:

https://github.com/ManageIQ/manageiq-api/pull/887/commits/934ee3e13321f72e002d134e54c56dfbaad6a307

Let me know what you think.

app/controllers/api/base_controller/renderer.rb
```diff
@@ -350,15 +352,19 @@ def attr_accessible?(object, attr)
...
       end
 
       def attr_physical?(object, attr)
         return true if ID_ATTRS.include?(attr)
-        (object.class.respond_to?(:has_attribute?) && object.class.has_attribute?(attr)) &&
-          !(object.class.respond_to?(:virtual_attribute?) && object.class.virtual_attribute?(attr))
+        klass = object
+        klass = object.class if object.kind_of?(ActiveRecord::Base)
+        (klass.respond_to?(:has_attribute?) && klass.has_attribute?(attr)) &&
+          !(klass.respond_to?(:virtual_attribute?) && klass.virtual_attribute?(attr))
```
jrafanie, line 367: Are these cosmetic changes or is this changing behavior for performance reasons?  I'm ok with cosmetic changes but I don't see how this makes a meaningful performance gain, so I want to make sure I am understanding...

jrafanie: (this block the similar change above it)

jrafanie:
Looks good so far, just asking questions to see if I'm understanding the changes that affect performance

app/controllers/api/base_controller/renderer.rb
chessbyte: these are not cosmetic - these seem to be working on `klass` instead of `object.class`

NickLaMuro: > these are not cosmetic

These are cosmetic from a "`klass` and `object.class`" are the same, but the rational as to why I changed it because normally these methods took an instance (as it was traversing through the `jbuilder` methods... aka "N+1ing"), but this allows it to just pass the `klass` itself.

jrafanie: > but the rational as to why I changed it because normally these methods took an instance (as it was traversing through the `jbuilder` methods... aka "N+1ing"), but this allows it to just pass the `klass` itself.

Thanks... that was the detail I was trying to understand.  👍

app/controllers/api/base_controller/renderer.rb
```diff
@@ -350,15 +352,19 @@ def attr_accessible?(object, attr)
       def attr_virtual?(object, attr)
         return false if ID_ATTRS.include?(attr)
         primary = attr_split(attr).first
-        (object.class.respond_to?(:reflect_on_association) && object.class.reflect_on_association(primary)) ||
-          (object.class.respond_to?(:virtual_attribute?) && object.class.virtual_attribute?(primary)) ||
-          (object.class.respond_to?(:virtual_reflection?) && object.class.virtual_reflection?(primary))
+        klass   = object
+        klass   = object.class if object.kind_of?(ActiveRecord::Base)
+        (klass.respond_to?(:reflect_on_association) && klass.reflect_on_association(primary)) ||
```
Fryguy, line 357: `klass.try(:reflect_on_association)` ?  The .try does a respond_to check

Fryguy:
As discussed, while support for the `:arel` clause is good, it's not complete, and we need support for the `:uses` clause as well.

app/controllers/api/base_controller/renderer.rb
Fryguy: Though I do see the old code had it that way.

NickLaMuro:
Will un-`[WIP]` once we  determine what to do with that last commit.

app/controllers/api/base_controller/renderer.rb
```diff
@@ -540,6 +546,45 @@ def render_options(resource, data = {})
...
+        return nil unless klass.respond_to?(:reflect_on_association)
+
+        type    = @req.subject
+        results = []
+
+        validate_attr_selection(klass).last.each do |vattr|
+          next if vattr == "href_slug"
+
+          attr_name, attr_base = split_virtual_attribute(vattr)
+          filtered_attr        = yield type, attr_name, attr_base
```
Fryguy, line 560: The block returning filtered_attr is kind of confusing as that's not what I'd expect from a method named `virtual_attributes_for`, but I can't think of something better at the moment. 😕

spec/requests/vms_spec.rb
```diff
@@ -82,6 +82,24 @@ def update_raw_power_state(state, *vms)
                                    [{"vendor" => "openstack"},
                                     {"vendor" => "openstack"}])
     end
+
+    context "with indirect virtual attribute ('hardware.cpu_sockets')" do
```
Fryguy, line 86: Is there a spec for something like `hardware.host.name` as per your example above?

spec/support/shared_examples/expand_resources_includes.rb
```diff
@@ -0,0 +1,25 @@
...
+      expands    = ["resources"]
+      from_match = [resource]
+
+      if defined?(uses)
+        from_match << uses
+      end
+
+      attrs                = defined?(attributes) ? attributes : "resources"
+      query_match          = /SELECT.*FROM\s"(?:#{from_match.join("|")})"/m
+      expected_query_count = defined?(query_count) ? query_count : 3
```
Fryguy, line 15: Can you just add a comment on why the magic number 3 (or perhaps move into a named constant?)

Fryguy (approved):

Fryguy:
> Will un-`[WIP]` once we determine what to do with that last commit. (keep or drop)

I'm fine with keeping it.

app/controllers/api/base_controller/renderer.rb
NickLaMuro: @Fryguy Would maybe just adding a comment explaining how this method functions be a good middle ground for now?

spec/support/shared_examples/expand_resources_includes.rb
NickLaMuro: Sure, I will do both.

app/controllers/api/base_controller/renderer.rb
Fryguy: Sounds good

app/controllers/api/base_controller/renderer.rb
NickLaMuro: Yeah, I didn't change the code here on purpose.  I was just refactoring it to allow it to accept a class directly instead of an instance and converting it to a class.

Explained already here:  https://github.com/ManageIQ/manageiq-api/pull/887#discussion_r482307292

jrafanie:
> > Will un-`[WIP]` once we determine what to do with that last commit. (keep or drop)
> 
> I'm fine with keeping it.

@NickLaMuro I'm fine with it too.  👍  You don't seem happy with it so if you have good reasons to not include it, we can drop it, just let us know what you think.

NickLaMuro:
@jrafanie nah, I am fine with it, I just ran into a bug I noticed in my tests and still haven't got them working yet.  I have already rebased in the changes locally, just haven't pushed until I have a proper fix.

spec/support/shared_examples/expand_resources_includes.rb
NickLaMuro: Added the comment and made it a `let` attribute.

spec/requests/vms_spec.rb
NickLaMuro: So I did this, but I needed to add support for it so it would allow support for nested attributes.

Left it as a separate commit instead of rebasing, so see the last commit for info.

app/controllers/api/base_controller/renderer.rb
NickLaMuro: Done.

app/controllers/api/base_controller/renderer.rb
```diff
@@ -540,6 +546,64 @@ def render_options(resource, data = {})
...
+
+          attr_base
+        end
+
+        # Handle nested relationships and convert to a hash
+        if attrs
+          attrs.each_with_object({}) do |key, include_for_find|
+            if key.include?(".")
+              nested = include_for_find
+              key.split(".").each {|k| nested = nested[k] ||= {} }
```
NickLaMuro, line 590: So this line is a little funky, so let me explain.

`key` in this case would be something like `"hardware.host"` (for an attribute of `"hardware.host.name"`).  The code here is then splitting it into `["hardware", "host"]`, and "recursively but not recursively" setting the `nested` variable (and the key at the current level) so the output ends up being:

```ruby
{"hardware" => {"host" => {}}}
```

This should also de-dup at the same time, but haven't tested this to be honest.

spec/requests/vms_spec.rb
```diff
@@ -82,6 +82,43 @@ def update_raw_power_state(state, *vms)
                                    [{"vendor" => "openstack"},
                                     {"vendor" => "openstack"}])
     end
+
+    def add_hardware_and_os_to_vms
+      [vm, vm1, vm2, vm_openstack, vm_openstack1, vm_openstack2].each do |vm_record|
+        cs = ComputerSystem.create
+        os = OperatingSystem.create(:name => "linux", :vm_or_template => vm_record, :computer_system => cs)
+        FactoryBot.create(:hardware, :vm_or_template => vm_record, :host => host)
```
NickLaMuro, line 90: Added this because there is a quirk with the API where it doesn't render nested keys if the relationship is `nil`, and this was a way to ensure that the records were being loaded as expected.

NickLaMuro:
> and we need support for the `:uses` clause as well.

So, this is going to be a bit rougher than I thought, and will probably take some time to think over how to implement (I have some half baked ideas, but nothing perfect at the moment).

I had to do a bit of gymnastics to make the "dot syntax" for relations to work, and probably have to do something different for `:uses` since it can be a collection of relations.  The `virtual_columns` on `VmOrTemplate` provide a good example of what we would have to try and support:

https://github.com/ManageIQ/manageiq/blob/9a23708582de38c9ae0cc8332f7e46fc1a53ba08/app/models/vm_or_template.rb#L139-L140

Because they do an `Array` of relations (`[:operating_system, :hardware]`), instead of say `"hardware.host.name"`, which expands to `{:hardware => {:host => {}}`, I am still trying to figure out how to pass this back through the `virtual_attributes_for` method.

jrafanie:
@NickLaMuro I think I follow your last few commit changes. Thanks for explaining. Did you choose to not fix the remaining style issues?  The useless assignment seems legit.  The others are meh.

I think I'm good with this.  Are the performance numbers in the description still the same after the last commit?

NickLaMuro:
> Did you choose to not fix the remaining style issues? The useless assignment seems legit. The others are meh.

One or two I agree with, so I will fix them an re-push in a bit.

> I think I'm good with this. Are the performance numbers in the description still the same after the last commit?

I hadn't re-run with the new changes, though said changes are also only expanding the types of attributes we support preloading for, and shouldn't affect the existing metrics.  Regardless, I think you bring up a good point that I should re-test to confirm all is well, so I will do that as well.

NickLaMuro:
@jrafanie re-ran the **"After"** benchmarks I ran previously, and got _basically_ the same results:

```
$ bundle exec miqperf benchmark -ac 5 "/api/vms?expand=resources&attributes=name,hardware.cpu_sockets"
/api/vms
| 1757 |      12 |        140 | 2041 |
|  367 |      10 |       10.5 |  559 |
|  328 |      10 |        9.7 |  559 |
|  373 |      10 |       10.6 |  559 |
|  270 |      10 |        9.4 |  559 |
$ bundle exec miqperf benchmark -ac 5 "/api/vms?expand=resources,hardware&attributes=num_cpu,name"
/api/vms
| 1578 |      12 |         24 | 2041 |
|  330 |      10 |        7.3 |  559 |
|  351 |      10 |        8.3 |  559 |
|  205 |      10 |        8.1 |  559 |
|  198 |      10 |        7.3 |  559 |
```

Fixed the `rubocop` errors as well.

chessbyte:
@jrafanie @Fryguy anything holding up this PR from getting merged?

app/controllers/api/base_controller/renderer.rb
```diff
@@ -540,6 +546,61 @@ def render_options(resource, data = {})
...
+          next if Rbac::Filterer::CLASSES_THAT_PARTICIPATE_IN_RBAC.include?(attr_base)
+
+          attr_base
+        end
+
+        # Handle nested relationships and convert to a hash
+        if attrs
+          attrs.each_with_object({}) do |key, include_for_find|
+            nested = include_for_find
+            key.split(".").each { |k| nested = nested[k] ||= {} }
```
kbrock, line 590: commented on the other side that I think you want to return nested here
(but you probably already commented over there that this is right... )

NickLaMuro: Yes I did, it is here:

https://github.com/ManageIQ/manageiq-api/pull/887#discussion_r486694177

kbrock: sorry - this really doesn't make sense to me.

for `"host.hardware"` I read:

```ruby
nested = {host: {hardware: {}}}
return [:host, :hardware] # <===
```

NickLaMuro: @kbrock "`nested`" is not the return value, it is a temp value for a single iteration of the `.each_with_object`, and a "tracker variable" for the nested `.each`.  `include_for_find` is the return value of this method.

`nested` is keeping track of the location as we traverse through `include_for_find` as we split the `key` for each iteration.  Because Ruby is "pass by reference", we are manipulating the value of `include_for_find` by passing a reference to the "nested" object each time we traverse deeper into the hash.

This means that for something like `"hardware.host.name"`, the loop for the `.each` (inner loop) would be:

### 1.

```ruby
key = "hardware"
include_for_find_before = {}
include_for_find_after  = {"hardware" => {}}
nested_before           = {} # the value of `include_for_find_before`
nested_after            = {} # the value of `include_for_find_after["hardware"]
```

### 2.
```ruby
key = "host"
include_for_find_before = {"hardware" => {}}
include_for_find_after  = {"hardware" => {"host" => {}}}
nested_before           = {"hardware" => {}} # the value of `include_for_find_before`
nested_after            = {} # the value of `include_for_find_after["hardware"]["host"]`
```

This is also why I wrote a test for this...

> "with nested indirect virtual attribute ('operating_system.computer_system.created_at')"

simaishi:
Backported as a part of #890

PR #888, opened 2020-09-01
User/Role/Feature lookup improvements

NickLaMuro:
Requires https://github.com/ManageIQ/manageiq/pull/20506

Makes use of new scopes to improve the number of queries required for login and even regular requests (though, the latter is less significant).

Also avoids some duplicate `User` lookups on login.

Benchmarks
----------

Note:  Improvements are in number of **queries**.

### Before

```
$ bundle exec miqperf benchmark -ac2 "/api/users"
$ bundle exec miqperf report --last
/api/auth
| 5088 |      26 |        183 |   10 |
/api/users
|  267 |      12 |         13 | 1494 |
|   17 |      10 |        3.9 |   12 |
```

### After

```
$ bundle exec miqperf benchmark -ac2 "/api/users"
$ bundle exec miqperf report --last
/api/auth
| 4949 |      12 |        181 |    1 |
/api/users
|  258 |       7 |         12 | 1489 |
|   15 |       5 |        3.3 |    7 |
```

Links
-----

* https://github.com/ManageIQ/manageiq/pull/20506
* More improvements as part of this effort:  https://github.com/ManageIQ/manageiq-api/issues/880

### DIFF
--- a/app/controllers/api/auth_controller.rb
+++ b/app/controllers/api/auth_controller.rb
@@ -5,7 +5,7 @@ module Api
     def show
       requester_type = fetch_and_validate_requester_type
       token_service = Environment.user_token_service
-      auth_token = token_service.generate_token(User.current_user.userid, requester_type)
+      auth_token = token_service.generate_token(User.current_user, requester_type)
       token_info = token_service.token_mgr(requester_type).token_get_info(auth_token)
       res = {
         :auth_token => auth_token,
--- a/app/controllers/api/base_controller/authentication.rb
+++ b/app/controllers/api/base_controller/authentication.rb
@@ -87,7 +87,7 @@ module Api
       end
 
       def auth_user(userid)
-        auth_user_obj = User.lookup_by_identity(userid)
+        auth_user_obj = User.lookup_by_identity(userid, lookup_scope: :api_includes)
         authorize_user_group(auth_user_obj)
         validate_user_identity(auth_user_obj)
         User.current_user = auth_user_obj
@@ -154,7 +154,7 @@ module Api
 
       def basic_authentication(username, password)
         timeout = ::Settings.api.authentication_timeout.to_i_with_method
-        user = User.authenticate(username, password, request, :require_user => true, :timeout => timeout)
+        user = User.authenticate(username, password, request, :require_user => true, :timeout => timeout, :lookup_scope => :api_includes)
         auth_user(user.userid)
       rescue MiqException::MiqEVMLoginError => e
         raise AuthenticationError, e.message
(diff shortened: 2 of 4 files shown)


app/controllers/api/base_controller/authentication.rb
```diff
@@ -91,8 +91,9 @@ def api_token_mgr
         Environment.user_token_service.token_mgr('api')
       end
 
-      def auth_user(userid)
-        auth_user_obj = User.lookup_by_identity(userid)
+      def auth_user(user_or_id)
+        auth_user_obj   = user_or_id if user_or_id.kind_of?(User)
+        auth_user_obj ||= User.lookup_by_identity(user_or_id, lookup_scope: :api_includes)
```
kbrock, line 96: is it possible to write this as an `if / else`, or `?:`

lib/services/api/user_token_service.rb
```diff
@@ -23,9 +23,13 @@ def api_config
       @api_config ||= ::Settings[base_config[:module]].to_hash
     end
 
-    def generate_token(userid, requester_type, token_ttl: nil)
-      userid = userid.downcase
-      validate_userid(userid)
+    def generate_token(user_or_id, requester_type, token_ttl: nil)
```
kbrock, line 26: This is great

kbrock (approved):
this is simple and really nice

one question on the if else, but just a nit

We've made changes exactly like this in the past, so they follow a common pattern.
But I am surprised how much of an impact it made.
very nice find

app/controllers/api/base_controller/authentication.rb
NickLaMuro: Yeah, I am down with that.  Probably would be much easier to grok.  👍

app/controllers/api/auth_controller.rb
```diff
@@ -5,7 +5,7 @@ class AuthController < BaseController
     def show
       requester_type = fetch_and_validate_requester_type
       token_service = Environment.user_token_service
-      auth_token = token_service.generate_token(User.current_user.userid, requester_type)
```
kbrock, line 8: aside, there is `User.current_userid` for future ref
you are deleting this reference, so not relevant here

NickLaMuro:
Looks like I have a lot of broken tests to fix... I will look into that today.

lib/services/api/user_token_service.rb
```diff
@@ -23,9 +23,13 @@ def api_config
...
+      if user_or_id.kind_of?(User)
+        userid = user_or_id.userid.downcase
```
Fryguy, line 28: do we still need to do the validation? (at least the "in my region" aspect of what the validate method does)?

NickLaMuro: I think I may have screwed up the `if` here, and the `end` should be a line earlier... let me look into this...

(sorry I took so long to respond to this...)

Fryguy:
cc @kbrock

PR #895, opened 2020-09-23
[WIP] Expose the safe_delete operation on cloud volumes

skateman:
The new AutoSDE provider exposes a safe delete operation that takes care of cleaning up some stuff around cloud volumes (which I don't really understand). In order to make this work, the `safe_delete_volume_queue` should be implemented on the related model analogously to `delete_volume_queue`. 

@abellotti I'm not sure how the API should handle the fact that not all cloud volumes support this kind of delete. Would be much easier if the (safe) deletion could be decided on the model level, but I was told that both the safe and unsafe delete should be available for an AutoSDE cloud volume.

### DIFF
--- a/config/api.yml
+++ b/config/api.yml
@@ -750,6 +750,8 @@
         :identifier: cloud_volume_show_list
       - :name: delete
         :identifier: cloud_volume_delete
+      - :name: delete
+        :identifier: cloud_volume_safe_delete
     :resource_actions:
       :get:
       - :name: read
@@ -757,6 +759,8 @@
       :post:
       - :name: delete
         :identifier: cloud_volume_delete
+      - :name: delete
+        :identifier: cloud_volume_safe_delete
       :delete:
       - :name: delete
         :identifier: cloud_volume_delete
--- a/app/controllers/api/cloud_volumes_controller.rb
+++ b/app/controllers/api/cloud_volumes_controller.rb
@@ -2,6 +2,14 @@ module Api
   class CloudVolumesController < BaseController
     include Subcollections::Tags
 
+    def safe_delete_resource(type, id, _data = {})
+      delete_action_handler do
+        cloud_volume = resource_search(id, type, collection_class(:cloud_volumes))
+        task_id = cloud_volume.safe_delete_volume_queue(User.current_user)
+        action_result(true, "Safely Deleting Cloud Volume #{cloud_volume.name}", :task_id => task_id)
+      end
+    end
+
     def delete_resource(type, id, _data = {})
       delete_action_handler do
         cloud_volume = resource_search(id, type, collection_class(:cloud_volumes))


skateman:
> Checked commit [skateman@090b8fb](https://github.com/skateman/manageiq-api/commit/090b8fb67a32ad7dc3b1fb7e5482ac1f54ed9afc) with ruby 2.6.3, rubocop 0.69.0, haml-lint 0.28.0, and yamllint
> 1 file checked, 1 offense detected
> 
> **app/controllers/api/cloud_volumes_controller.rb**
> 
> * [ ]  exclamation - [Line 5](https://github.com/skateman/manageiq-api/blob/090b8fb67a32ad7dc3b1fb7e5482ac1f54ed9afc/app/controllers/api/cloud_volumes_controller.rb#L5), Col 36 - [Naming/UncommunicativeMethodParamName](http://rubydoc.info/gems/rubocop/0.69.0/RuboCop/Cop/Naming/UncommunicativeMethodParamName) - Method parameter must be at least 3 characters long.

Oh come on @miq-bot, using an `id` as an argument is totally fine. We're in the API, should I use `identifier` to make you happy? :laughing:

agrare:
@skateman the AutoSDE provider hasn't been accepted into the ManageIQ org yet and we're still working through the schema changes.  I think we should hold off on this API until the feature has been added to the model layer

config/api.yml
```diff
@@ -750,13 +750,17 @@
...
+      - :name: delete
+        :identifier: cloud_volume_safe_delete
     :resource_actions:
       :get:
       - :name: read
         :identifier: cloud_volume_show
       :post:
       - :name: delete
         :identifier: cloud_volume_delete
+      - :name: delete
```
abellotti, line 762: are we exposing both types of delete ?      if so, maybe renaming second one to safe_delete

abellotti:
actions can be exposed or not for a resource, by providing a model method like _validate_safe_delete_ https://github.com/ManageIQ/manageiq-api/blob/3a7a8c1fec795538ca9b21ad2a1e8b566f09f0af/app/controllers/api/base_controller/renderer.rb#L537

agrare:
> Would be much easier if the (safe) deletion could be decided on the model level

Safe delete sounds more like an option to the normal delete action (like `delete(force: true)` or `delete(force: false)`), would auto-sde also have an unsafe delete method as an option?

config/api.yml
```diff
@@ -750,13 +750,17 @@
         :identifier: cloud_volume_show_list
       - :name: delete
         :identifier: cloud_volume_delete
+      - :name: delete
```
llivne, line 753: name should be safe_delete

config/api.yml
llivne: I also think that the name here need to be safe_delete

skateman: Yeah, I was copy-pasting too fast (sorry)

skateman:
> > Would be much easier if the (safe) deletion could be decided on the model level
> 
> Safe delete sounds more like an option to the normal delete action (like `delete(force: true)` or `delete(force: false)`), would auto-sde also have an unsafe delete method as an option?

I agree, my original idea was to have this sorted out on the model level and keep just the delete, but I was told that both operations should be available.

Issue #885, opened 2020-08-19
Parameterize endpoint to revoke user sessions by type


jvlcek:
This is a sub-issue of issue **_Ability to track and revoke sessions_** [#20378](https://github.com/ManageIQ/manageiq/issues/20378)

Issue **Add endpoint to revoke user sessions** [#884](https://github.com/ManageIQ/manageiq-api/issues/884) should be parameterized by type (api, ui, ws, etc). The default is all. 

This issue will track adding the support for specifying the parameter for the specific type.

Issue #884, opened 2020-08-19
Add endpoint to revoke user sessions


jvlcek:
This is a sub-issue of issue **_Ability to track and revoke sessions_** [#20378](https://github.com/ManageIQ/manageiq/issues/20378)

**Brief:**

A new endpoint needs to be added to allow a user sessions to be revoked

**Details:**

Introduce revoke session backend support and an api revoke session endpoint.

A user should be able to revoke their own sessions. A super-admin should be able to revoke anybody's sessions.

Might consider supporting `GET /api/revoke/user_session/:id`, if it is thought other resources could be added to revoke in the future.
If not might consider using `GET /api/revoke_user_session/:id `

Need to confirm that MCM needs an API GET call to end user sessions on other clients. POST might be more appropriate to this functionality;

Should perhaps consider

```
POST /api/revoke_session
{
  "userid" : "testuser", # default to current user, but support any user for the admin to use
  "type" : { "all" } # default to all
}

```

Parameterized support by type ("api", "ui", "ws", etc) needs to be supported with the default to "all". This work will be done as a follow up and tracked in Issue: **Parameterize endpoint to revoke user sessions by type** [#885](https://github.com/ManageIQ/manageiq-api/issues/885)

abellotti:
Thanks for writing this @jvlcek, I kind of also prefer the POST since we're doing/changing something in the environment.

Here's what I'm thinking:

Revoking all session types for the authenticated user, no need to send _userid_ or _session_types_ since the userid is the authenticated user for the API call.

```
POST /api/revoke
```

Revoking one or more _session_types_ for the authenticated user:

```
POST /api/revoke
{
  "session_types" : ["ui", "api"]
}
```

Omitting _session_types_ is essentially an "all".

For an administrative account to deleting sessions for a user, with optional _session_types_:

```
POST /api/revoke
{
  "userid" : "any_user"
  "session_types" : ["ui"]
}
```

PR #900, opened 2020-09-30
Add action "revoke sessions" to users_controllers

lpichler:
Add endpoint to revoke user's session as described here in this issue: https://github.com/ManageIQ/manageiq-api/issues/884

Current Signature:
https://github.com/ManageIQ/manageiq-api/pull/900#issuecomment-707271568

## Links
Issue - https://github.com/ManageIQ/manageiq-api/issues/884 
Parent Issue - https://github.com/ManageIQ/manageiq/issues/20378
Next Issue (follow up) - https://github.com/ManageIQ/manageiq-api/issues/885

## Needs to be merged before (reason for WIP)
- [x] https://github.com/ManageIQ/manageiq/pull/20601
- [x] https://github.com/ManageIQ/manageiq/pull/20633
- [x] https://github.com/ManageIQ/manageiq/pull/20692

### DIFF
--- a/app/controllers/api/users_controller.rb
+++ b/app/controllers/api/users_controller.rb
@@ -62,7 +62,33 @@ module Api
       raise BadRequestError, "Cannot set current_group - #{err}"
     end
 
+    def revoke_sessions_collection(type, data)
+      revoke_sessions_resource(type, current_user.id, data)
+    end
+
+    def revoke_sessions_resource(type, id, _data)
+      api_action(type, id) do |klass|
+        user = target_user(id, type, klass)
+        api_log_info("Revoking all sessions of user #{user.userid}")
+
+        user.revoke_sessions
+        action_result(true, "All sessions revoked successfully for user #{user.userid}.")
+      rescue => err
+        action_result(false, err.to_s)
+      end
+    end
+
     private
+
+    def target_user(id, type, klass)
+      if id == current_user.id
+        current_user
+      elsif current_user.role_allows?(:identifier => 'revoke_user_sessions')
+        resource_search(id, type, klass)
+      else
+        raise ForbiddenError, "The user is not authorized for this task or item."
+      end
+    end
 
     def update_target_is_api_user?
       User.current_user.id == @req.collection_id.to_i
--- a/config/api.yml
+++ b/config/api.yml
@@ -4238,6 +4238,7 @@
         :identifier: rbac_user_tags_edit
       - :name: unassign_tags
         :identifier: rbac_user_tags_edit
+      - :name: revoke_sessions
     :resource_actions:
       :get:
       - :name: read
@@ -4247,6 +4248,7 @@
         :identifier: rbac_user_edit
       - :name: delete
         :identifier: rbac_user_delete
+      - :name: revoke_sessions
       - :name: set_current_group
       :delete:
       - :name: delete
--- a/spec/requests/users_spec.rb
+++ b/spec/requests/users_spec.rb
@@ -613,4 +613,161 @@ RSpec.describe "users API" do
       expect(response).to have_http_status(:ok)
     end
   end
+
+  context "Revoking Sessions" do
+    %w[cache sql memory].each do |session_store|
+      before do
+        ::Settings.server.session_store = session_store
+      end
+
+      it "revokes all own sessions of authenticated user with #{session_store}" do
+        api_basic_authorize
+
+        FactoryBot.create(:session, :user_id => @user.id)
+        expect(Session.where(:user_id => @user.id).count).to eq(1)
+
+        TokenStore.token_caches
+
+        ts = TokenStore.acquire("api", 100)
+
+        ts.create_user_token("my_token", {:userid => @user.userid, :expires_on => Time.zone.now + 100.days}, {:expires_in => 100})
+
+        expect(ts.read("my_token")).not_to be_nil
+
+        if ::Settings.server.session_store == "sql"
+          expect(ts.read("my_token")[:userid]).to eq(@user.userid)
+        else
+          expect(ts.read("tokens_for_#{@user.userid}")).not_to be_nil
+        end
+
+        ts.create_user_token("his_token", {:userid => user1.userid, :expires_on => Time.zone.now + 100.days}, {:expires_in => 100})
+        expect(ts.read("his_token")).not_to be_nil
+
+        if ::Settings.server.session_store == "sql"
+          expect(ts.read("my_token")[:userid]).to eq(@user.userid)
+        else
+          expect(ts.read("tokens_for_#{@user.userid}")).not_to be_nil
+        end
+
+        post(api_users_url, :params => {'action' => "revoke_sessions"})
+
+        expect(Session.where(:user_id => @user.id).count).to eq(0)
+
+        expect(response).to have_http_status(:ok)
+        expect(ts.read("my_token")).to be_nil
+
+        unless ::Settings.server.session_store == "sql"
+          expect(ts.read("tokens_for_#{@user.userid}")).to be_nil
+        end
+
+        expect(ts.read("his_token")).not_to be_nil
+
+        unless ::Settings.server.session_store == "sql"
+          expect(ts.read("tokens_for_#{user1.userid}")).not_to be_nil
+        end
+      end
+
+      context "target user is other than authenticated user with #{session_store}" do
+        it "revokes sessions of resource user" do
+          api_basic_authorize :revoke_user_sessions
+          user1.miq_groups << @user.miq_groups.first
+          user1.save
+
+          post(api_user_url(nil, user1), :params => {'action' => "revoke_sessions"})
+
+          expect(response).to have_http_status(:ok)
+
+          expect(response.parsed_body["success"]).to be_truthy
+          expect(response.parsed_body["message"]).to eq("All sessions revoked successfully for user #{user1.userid}.")
+        end
+
+        it "doesn't allow to revoke resource user sessions when authenticated user doesn't have access to resource user due to RBAC" do
+          api_basic_authorize :revoke_user_sessions
+
+          post(api_user_url(nil, user_admin), :params => {'action' => "revoke_sessions"})
+
+          expect(response.parsed_body["success"]).to be_falsey
+          expect(response.parsed_body["message"]).to eq("Access to the resource users/#{user_admin.id} is forbidden")
+        end
+
+        it "doesn't allow to revoke resource user sessions when authenticated user doesn't have access to resource user due to missing entitlement" do
+          api_basic_authorize
+          user1.miq_groups << @user.miq_groups.first
+          user1.save
+
+          post(api_user_url(nil, user1), :params => {'action' => "revoke_sessions"})
+
+          expect(response.parsed_body["success"]).to be_falsey
+          expect(response.parsed_body["message"]).to eq("The user is not authorized for this task or item.")
+        end
+
+        let(:user_admin) do
+          User.all.detect(&:super_admin_user?) || FactoryBot.create(:user_admin, :userid => "admin")
+        end
+
+        it "does allow to revoke resource user sessions when authenticated user is super_admin" do
+          api_basic_authorize
+          @user = user_admin
+          User.current_user = user_admin
+          allow(User).to receive(:current_user).and_return(user_admin)
+
+          post(api_user_url(nil, user1), :params => {'action' => "revoke_sessions"})
+
+          expect(response.parsed_body["success"]).to be_truthy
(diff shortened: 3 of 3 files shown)


lpichler:
cc @gtanzillo @NickLaMuro

app/controllers/api/revoke_controller.rb
```diff
@@ -0,0 +1,30 @@
...
+    end
+
+    def target_user
+      userid = params[:userid]
+      if userid
+        user_object = User.find_by(:userid => userid)
+        raise BadRequestError, "Unable to find user #{userid}" unless user_object
+
+        filtered_user = Rbac::Filterer.filtered_object(user_object)
+        raise ForbiddenError, "The user is not authorized for this task or item." unless filtered_user
```
abellotti, line 14: we might also need to check if target user is not the current user, then current_user.super_admin_user? should be true unless we introduce a new entitlement for this, in which case something liek current_user.role_allows?(:identifier => 'revoke_user_sessions') could be introduced.

app/controllers/api/revoke_controller.rb
```diff
@@ -0,0 +1,30 @@
...
+
+        filtered_user
+      else
+        User.current_user
```
gtanzillo, line 18: I'm thinking that the current user should be able to revoke his own sessions regardless of RBAC, just like he'd be able to log himself out. Do you think we should skip RBAR for this case?

I'm also thinking this would be the most common way this API will be accessed

app/controllers/api/revoke_controller.rb
lpichler: @abellotti I updated PR as you suggested:
- I added check if target user is not the current user 
- I added feature `revoke_user_sessions` in core for this operation
- I refactored it a bit.

can you take a look please ? thanks

app/controllers/api/revoke_controller.rb
lpichler: yes I think also that we don't need to do RBAC check on current user. 

Newly I added here `role_allows?` check for product feature `revoke_current_user_sessions` and
thanks to that we can distinguish privileges
for revoking current user's sessions and 
for revoking sessions of requester users from params. 

thanks

app/controllers/api/revoke_controller.rb
```diff
@@ -0,0 +1,50 @@
...
+    end
+
+    def current_user_allowed?
+      current_user.role_allows?(:identifier => 'revoke_current_user_sessions')
+    end
+
+    def target_user
+      if user_from_params_allowed?(userid = params[:userid])
+        filtered_user_from_params(userid)
+      elsif current_user_allowed?
```
abellotti, line 37: I just want to make sure of the following:
- A user can revoke his own sessions with or without the userid param and without needing the revoke_user_sessions entitlement.
- A user can revoke someone else's session if and only if they have the revoke_user_sessions entitlement.

lpichler: I think it is not possible so simply as you suggesting and we need to use more entitlements.

- we need one entitlement(product feature added [here](https://github.com/ManageIQ/manageiq/pull/20633) 1- `revoke_session`) just to allow access method `update` -  this is done generally and this check is done in https://github.com/ManageIQ/manageiq-api/blob/master/app/controllers/api/base_controller/parser.rb#L178.
It finds entitlement in api.yml(added in this PR ) for edit action and then it is checked whether current user has this entitlement(product feature added)
- > - A user can revoke his own sessions with or without the userid param and without needing the revoke_user_sessions entitlement.

We are not able to revoke without `revoke_user_sessions ` but we need also  entitlement `revoke_session` to have on current user from first point and thanks to that we need to add new entitlement just for this case so I added `revoke_current_user_sessions`.

- > - A user can revoke someone else's session if and only if they have the revoke_user_sessions entitlement.

Yes this is correct, this operation can be performed only if current (authenticated) user has `revoke_user_sessions` entitlement(product feature) and if current (authenticated) user has RBAC access to this user.


Hopefully this is more clear and it makes sense now.
If you have any other concerns/suggestions please let me know.

thanks

gtanzillo (approved):

app/controllers/api/revoke_controller.rb
abellotti: I wonder if you need to update https://github.com/ManageIQ/manageiq/blob/master/db/fixtures/miq_user_roles.yml  in the https://github.com/ManageIQ/manageiq/pull/20633 PR so you add the `revoke_user_sessions` to let's say admins and super-admins, and add `revoke_sessions` to every other role.

lpichler: yes that's a good note, I did update -
`super_administrator` and `administrator` 
have privileges(`revoke_sessions` as parent entitlement) to perform both revoke operations and
other roles have privileges(`revoke_current_user_sessions`) to perform **only** revoke own session of authenticated (current)user. 

thanks

gtanzillo:
Regarding 
>A user can revoke his own sessions with or without the userid param and without needing the revoke_user_sessions entitlement.

Couldn't we just change the method to do `||` for the three conditions? Like -
```
    def current_user_allowed?(userid)
      (userid.nil? || userid == current_user.userid) || current_user.role_allows?(:identifier => 'revoke_current_user_sessions')
    end
```

lpichler:
> Regarding
> 
> > A user can revoke his own sessions with or without the userid param and without needing the revoke_user_sessions entitlement.
> 

@gtanzillo yes we could but this statement is not reachable because we need one product feature to access  `update` which is done by common code. I described it more [here](https://github.com/ManageIQ/manageiq-api/pull/900#discussion_r498455085)

I used that one feature as parent feature for other features which are tied with operations:
1. revoke own sessions of authenticated (current)user
2. revoke sessions of requested user from parameters

then the mentioned method have different semantic - method checks requirement on `revoke_current_user_sessions` for current user but only when **case 1. is valid** (no `userid` is present in request parameters - and yes there is one exception when passed `userid` is same asi current user)


let me know if have more questions, thanks

lpichler:
@gtanzillo @abellotti thanks for today's discussion and  I updated PR as suggested:

- the only new entitlement `revoke_user_sessions` has added [here](https://github.com/ManageIQ/manageiq/pull/20633) 
- super admin a users with role `EvmRole-administrator` can revoke own's sessions and other's user sessions
- other users can revoke only own's sessions by default  (if they want to revoke other user's session they have to have enabled  permission and access to user regard to our rbac filterer)

PR ready for review again 👍

config/api.yml
```diff
@@ -3046,6 +3046,15 @@
       :post:
       - :name: request_download
         :identifier: miq_report_view
+  :revokes:
```
Fryguy, line 3049: I find this weird...I would expect a revoke action on the /users/ endpoint.

Fryguy: Either that or a revoke action on the /auth endpoint.

Fryguy:
Code wise, everything looks good to me.  Ergonomics-wise I would never expect to hit a revokes endpoint when this is a function/action of either users or auth.

app/controllers/api/revokes_controller.rb
```diff
@@ -0,0 +1,41 @@
+module Api
+  class RevokesController < BaseController
+    skip_before_action :validate_api_action, :only => :create
+
+    def filtered_user_from_params(userid)
```
Fryguy, line 5: Most of these methods should be made private except the ones exposed to routes.

config/api.yml
lpichler: yes `/user` sounds better to me and based on some api suprises with current way `/user` fits better to our api.

it is based on @abellotti [suggestion](https://github.com/ManageIQ/manageiq-api/issues/884#issuecomment-67668246), 

@abellotti it is ok If I will change it to `/user` endpoint ?

thanks

abellotti:
Since we can revoke ui session too for the user, I tend to lean toward the users endpoint since /api/auth is intended for the API session token.  Since we're not required to specify a user id, but target the authenticated user, we can leverage collection level action, i.e. https://github.com/ManageIQ/manageiq-api/pull/629


Maybe introducing a "revoke_sessions" action as follows:

```
POST /api/users
{
   "action" : "revoke_sessions",
   "session_types" : [ "ui", "api" ]
}
```

This will target the authenticated user,

But also support the resource level action:

```
POST /api/users/:id
{
   "action" : "revoke_sessions",
   "session_types" : [ "ui", "api", "ws" ]
}
```

And nothing stops us from supporting the collection level (bulk) action for revoking sessions of multiple users:

```
POST /api/users
{
   "action" : "revoke_sessions",
   "resources" : [
       { "id" : ":id1", "session_types" : ["ui"] },
       { "id" : ":id2", "session_types" : ["api"] },
       { "id" : ":id3" },
       { "id" : ":id4", "session_types" : ["ui", "api"]}
   ]
}
```

Fryguy:
@abellotti Yes that is exactly what I was thinking... I just realized I put /api/revoke in the original requirements and that was my mistake...I actually meant to write something like `/api/user (action=revoke)`

Fryguy:
I could also see an revoke_session (singular) action on the `api/user/<id>` endpoint to revoke the session for a specific user.

lpichler:
@abellotti @gtanzillo @Fryguy I did changes according to signature in https://github.com/ManageIQ/manageiq-api/pull/900#issuecomment-707271568.

app/controllers/api/users_controller.rb
```diff
@@ -62,8 +62,34 @@ def set_current_group_resource(_type, id, data)
       raise BadRequestError, "Cannot set current_group - #{err}"
     end
 
+    def revoke_sessions_collection(type, data)
+      revoke_sessions_resource(type, current_user.id, data)
+    end
+
+    def revoke_sessions_resource(type, id, _data)
```
Fryguy, line 69: Should these 2 methods be private since they aren't routable?

Fryguy (approved):
LGTM - will defer to @abellotti for merge.

spec/requests/users_spec.rb
```diff
@@ -613,4 +613,136 @@
...
+          "message" => "Couldn't find User with 'id'=999999",
+          "href"    => "http://www.example.com/api/users/999999"
+        }
+      end
+
+      let(:expected_results) do
+        [expected_result_user3, expected_result_user1, expected_result_invalid_id, expected_result_not_found]
+      end
+
+      it "tries to revoke resource users sessions when authenticated user doens't have access to all users" do
```
abellotti, line 736: minor -> doesn't

spec/requests/users_spec.rb
```diff
@@ -613,4 +613,136 @@
...
+        api_basic_authorize :revoke_user_sessions
+
+        user1.miq_groups << @user.miq_groups.first
+        user1.save
+
+        post(api_users_url, :params => {"action" => "revoke_sessions", "resources" => resource_parameters})
```
abellotti, line 742: Nice to see the bulk action working 👍

spec/requests/users_spec.rb
```diff
@@ -613,4 +613,136 @@
...
+
+      expect(response).to have_http_status(:ok)
+      expect(ts.read("my_token")).to be_nil
+      expect(ts.read("tokens_for_#{@user.userid}")).to be_nil
+
+      expect(ts.read("his_token")).not_to be_nil
+      expect(ts.read("tokens_for_#{user1.userid}")).not_to be_nil
+    end
+
+    context "target user is other then authenticated user" do
```
abellotti, line 648: minor -> other than ...

app/controllers/api/users_controller.rb
```diff
@@ -62,8 +62,34 @@ def set_current_group_resource(_type, id, data)
...
     private
 
+    def target_user(id, type, klass)
+      if id == current_user.id
+        current_user
+      elsif current_user.role_allows?(:identifier => 'revoke_user_sessions')
+        resource_search(id, type, klass)
+      else
+        raise ForbiddenError, "The user is not authorized for this task or item."
+      end
```
abellotti, line 90: I like how you're checking entitlement without needing the role entitlement in the api.yaml.  👍

abellotti:
@lpichler this looks great and straightforward implementation.  Just minor typos, but other than that LGTM!!

app/controllers/api/users_controller.rb
lpichler: those method could be probably private but we don't have such methods set as private in other controllers, so let's leave as it is for now, maybe it is intentionally for some reason.

lpichler:
@abellotti fix typos,
@Fryguy @abellotti for review 👍

abellotti (approved):

abellotti:
LGTM @lpichler !!  Thanks for the API Enhancement. will merge when green.

PR #905, opened 2020-10-06
Added api for adding new physical storage

IlyaKessel:
As part of the storage modifications (see https://github.com/ManageIQ/manageiq-ui-classic/pull/7282 for the entire modifications details.) , we added the ability of directly adding physical storage for supporting storage managers.

This PR handles the required API mechanism.

### DIFF
--- a/app/controllers/api/physical_storages_controller.rb
+++ b/app/controllers/api/physical_storages_controller.rb
@@ -12,6 +12,14 @@ module Api
       end
     end
 
+    def create_resource(_type, _id = nil, data = {})
+      ext_management_system = ExtManagementSystem.find(data['ems_id'])
+      task_id = PhysicalStorage.create_physical_storage_queue(session[:userid], ext_management_system, data)
+      action_result(true, "Creating Physical Storage #{data['name']} for Provider: #{ext_management_system.name}", :task_id => task_id)
+    rescue => err
+      action_result(false, err.to_s)
+    end
+
     private
 
     def ensure_resource_exists(type, id)
--- a/config/api.yml
+++ b/config/api.yml
@@ -2280,6 +2280,8 @@
       - :name: read
         :identifier: physical_storage_show_list
       :post:
+      - :name: create
+        :identifier: physical_storage_new
       - :name: refresh
         :identifier: physical_storage_refresh
     :resource_actions:
--- a/spec/requests/physical_storages_spec.rb
+++ b/spec/requests/physical_storages_spec.rb
@@ -1,4 +1,26 @@
 describe "Physical Storages API" do
+  context "POST /api/physical_storages" do
+    it "creates new storage" do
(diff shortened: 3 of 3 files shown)


config/api.yml
```diff
@@ -2280,6 +2280,8 @@
       - :name: read
         :identifier: physical_storage_show_list
       :post:
+      - :name: create_new_physical_storage
```
abellotti, line 2283: This post action name should just be "create"

app/controllers/api/physical_storages_controller.rb
```diff
@@ -12,6 +12,14 @@ def refresh_resource(type, id, _data = nil)
       end
     end
 
+    def create_new_physical_storage_resource(_type, _id = nil, data = {})
```
abellotti, line 15: just: `def create_resource(type, _id, _data = {})`  should do the trick

config/api.yml
```diff
@@ -2280,6 +2280,8 @@
...
+        :identifier: create_new_physical_storage
```
abellotti, line 2284: probably `physical_storage_new` as declared in ManageIQ's db/fixtures/miq_product_features.yml

abellotti:
Specs needed for the new create action.  Thanks!!

app/controllers/api/physical_storages_controller.rb
```diff
@@ -12,6 +12,14 @@ def refresh_resource(type, id, _data = nil)
       end
     end
 
+    def create_resource(_type, _id = nil, data = {})
+      ext_management_system = ExtManagementSystem.find(data['ems_id'])
+      task_id = PhysicalStorage.create_physical_storage_queue(session[:userid], ext_management_system, data)
+      action_result(true, 'Creating Physical Storage', :task_id => task_id)
```
abellotti, line 18: might be nice to return in the action result the target in addition to the  generic message, maybe:

```
Creating Physical Storage #{data["name"]} for Provider id: #{data["ems_id"]}
```

spec/requests/physical_storages_spec.rb
```diff
@@ -1,4 +1,31 @@
...
+      }
+
+      post(api_physical_storages_url, :params => request)
+
+      expected = {
+        "success" => true,
+        "message" => "Creating Physical Storage",
+      }
+      expect(response).to have_http_status(:ok)
+      expect(response.parsed_body['results'].first).to include(expected)
```
abellotti, line 25: You can replace above expected and the 2 expects with the following helper:

```
expect_single_action_result(:result => true, :message => /Creating Physical Storage test_storage/, :task => true)
```

IlyaKessel: changed to expect_multiple_action_result, because create action is done on the collection_actions context

abellotti:
Thank you @IlyaKessel for the API Enhancement. LGTM!! 👍

abellotti (approved):

PR #914, opened 2020-10-08
Expose the DDF schema for CloudVolume subclasses via OPTIONS

skateman:
Similarly to the provider forms, the CloudVolume forms also have provider-specific fields that we're moving to `params_for_create` methods under the provider-specific subclasses. This change exposes these schemas through the OPTIONS endpoint. The only necessary param is the `id` of a StorageManager that will be selected from a dropdown defined in the common schema. 

The current name of the param is `ems_id` but I'm open for better ideas. Also I'm not sure if all the storage manager subclasses follow the `"#{provider.class}::CloudVolume"` convention on which this PR heavily relies.

```js
// OPTIONS /api/cloud_volumes?ems_id=XY
{
  // ...
  "form_schema": {
    fields: [
      // ...
    ]
  }  
}
```

Parent issue: https://github.com/ManageIQ/manageiq-ui-classic/issues/7334

@miq-bot add_reviewer @agrare 

### DIFF
--- a/app/controllers/api/cloud_volumes_controller.rb
+++ b/app/controllers/api/cloud_volumes_controller.rb
@@ -2,6 +2,20 @@ module Api
   class CloudVolumesController < BaseController
     include Subcollections::Tags
 
+    def create_resource(_type, _id = nil, data = {})
+      ext_management_system = ExtManagementSystem.find(data['ems_id'])
+
+      klass = CloudVolume.class_by_ems(ext_management_system)
+
+      validate = klass.validate_create_volume(ext_management_system)
+      raise validate[:message] unless validate[:available]
+
+      task_id = klass.create_volume_queue(session[:userid], ext_management_system, data)
+      action_result(true, "Creating Cloud Volume #{data['name']} for Provider: #{ext_management_system.name}", :task_id => task_id)
+    rescue => err
+      action_result(false, err.to_s)
+    end
+
     def delete_resource(type, id, _data = {})
       delete_action_handler do
         cloud_volume = resource_search(id, type, collection_class(:cloud_volumes))
--- a/app/controllers/api/cloud_volumes_controller.rb
+++ b/app/controllers/api/cloud_volumes_controller.rb
@@ -9,5 +9,19 @@ module Api
         action_result(true, "Deleting Cloud Volume #{cloud_volume.name}", :task_id => task_id)
       end
     end
+
+    def options
+      return super unless params[:ems_id]
+
+      ems = ExtManagementSystem.find(params[:ems_id])
+
+      raise BadRequestError, "No CloudVolume support for - #{klass}" unless defined?(ems.class::CloudVolume)
+
+      klass = ems.class::CloudVolume
+
+      raise BadRequestError, "No DDF specified for - #{klass}" unless klass.respond_to?(:params_for_create)
+
+      render_options(:cloud_volumes, :form_schema => klass.params_for_create(ems))
+    end
   end
 end
--- a/config/api.yml
+++ b/config/api.yml
@@ -801,6 +801,8 @@
       - :name: read
         :identifier: cloud_volume_show_list
       :post:
+      - :name: create
+        :identifier: cloud_volume_new
       - :name: query
         :identifier: cloud_volume_show_list
       - :name: delete
--- a/spec/requests/cloud_volumes_spec.rb
+++ b/spec/requests/cloud_volumes_spec.rb
@@ -128,4 +128,38 @@ describe "Cloud Volumes API" do
     expect(response.parsed_body).to include(expected)
     expect(response).to have_http_status(:ok)
   end
+
+  it 'it can create cloud volumes through POST' do
+    zone = FactoryBot.create(:zone, :name => "api_zone")
+    provider = FactoryBot.create(:ems_autosde, :zone => zone)
+
+    api_basic_authorize collection_action_identifier(:cloud_volumes, :create, :post)
+
+    post(api_cloud_volumes_url, :params => {:ems_id => provider.id, :name => 'foo', :size => 1234})
+
+    expected = {
+      'results' => a_collection_containing_exactly(
+        a_hash_including(
+          'success' => true,
+          'message' => a_string_including('Creating Cloud Volume')
+        )
+      )
+    }
+
+    expect(response.parsed_body).to include(expected)
+    expect(response).to have_http_status(:ok)
+  end
+
+  it 'returns a DDF schema when available via OPTIONS' do
+    zone = FactoryBot.create(:zone, :name => "api_zone")
+    provider = FactoryBot.create(:ems_autosde, :zone => zone)
+
+    allow(provider.class::CloudVolume).to receive(:params_for_create).and_return('foo')
+
+    options(api_cloud_volumes_url, :params => {:ems_id => provider.id})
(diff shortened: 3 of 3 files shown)


skateman:
@agrare I have a feeling that this will only work for Amazon, because the OpenStack cloud volumes are subclassed differently. The goal here is to have an endpoint that is able to tell me the `CloudVolume` subclass for the given storage manager and call its `params_for_create`.

@miq-bot add_label question

agrare:
@skateman we could do `ems.cloud_volumes_klass`

skateman:
> @skateman we could do `ems.cloud_volumes_klass`

Perfect!

app/controllers/api/cloud_volumes_controller.rb
```diff
@@ -9,5 +9,15 @@ def delete_resource(type, id, _data = {})
         action_result(true, "Deleting Cloud Volume #{cloud_volume.name}", :task_id => task_id)
       end
     end
+
+    def options
+      return super unless params[:ems_id]
+
+      klass = ManageIQ::Providers::StorageManager.find(params[:ems_id]).class.cloud_volumes_klass
```
agrare, line 16: For OpenStack will the ems_id be the StorageManager here or will it be the CloudManager?

skateman: It gets me `ManageIQ::Providers::Openstack::StorageManager::CinderManager`

agrare: Hmm that's going to be an issue for Google then which like OpenStack has `CloudManager::CloudVolume` but it doesn't have a `StorageManager`

I can add this to the `BaseManager` and return `self::CloudVolume` and override it in amazon, but that depends on us getting the "right" ems_id back here.

skateman: Yup, it's not returning anything for google:
```
  "resources": [
    {
      "href": "http://localhost:3000/api/providers/40",
      "id": "40",
      "name": "Amazon EBS Storage Manager",
      "type": "ManageIQ::Providers::Amazon::StorageManager::Ebs",
      "supports_block_storage": true
    },
    {
      "href": "http://localhost:3000/api/providers/5",
      "id": "5",
      "name": "OpenStack Cinder Manager",
      "type": "ManageIQ::Providers::Openstack::StorageManager::CinderManager",
      "supports_block_storage": true
    }
  ],

```

skateman: This however means that there's no way to create a google cloud volume, only editing is allowed...

agrare: Yeah exactly, I don't think checking supports_block_storage isn't exactly correct.  We should probably change the query to check for managers that support cloud_volumes that way we can assume that `ems.class::CloudVolume` will exist.

gtanzillo, line 16: Whare does `cloud_volumes_klass` get defined?

skateman: Umm @agrare https://github.com/ManageIQ/manageiq-api/pull/914#issuecomment-708486870

skateman: :+1:

skateman:
@agrare are you okat with these changes? Can we get this merged?

agrare (approved):

app/controllers/api/cloud_volumes_controller.rb
```diff
@@ -2,12 +2,37 @@ module Api
...
     def delete_resource(type, id, _data = {})
       delete_action_handler do
         cloud_volume = resource_search(id, type, collection_class(:cloud_volumes))
         task_id = cloud_volume.delete_volume_queue(User.current_user)
         action_result(true, "Deleting Cloud Volume #{cloud_volume.name}", :task_id => task_id)
       end
     end
+
+    def options
+      return super unless params[:ems_id]
```
Fryguy, line 28: Does this imply the user has to pass in the ems_id?  (Or do we know the ems_id based on the incoming endpoint?

That is, if a user does `OPTIONS /api/cloud_volumes/123` I would expect under the covers we would look up the ems_id and not force the user to enter it into the query.

app/controllers/api/cloud_volumes_controller.rb
```diff
@@ -2,12 +2,37 @@ module Api
   class CloudVolumesController < BaseController
     include Subcollections::Tags
 
+    def create_resource(_type, _id = nil, data = {})
+      ext_management_system = ExtManagementSystem.find(data['ems_id'])
+
+      klass = CloudVolume.class_by_ems(ext_management_system)
+
+      validate = klass.validate_create_volume(ext_management_system)
+      raise validate[:message] unless validate[:available]
```
Fryguy, line 11: @agrare Is this specific to CloudVolume?  Asking because if we have this pattern for checking things, then should be apply that to the other PRs, like [this one](https://github.com/ManageIQ/manageiq-api/pull/962/files#diff-aa014dae2b42e9d76dcfe82a22847466b59bdb401f5c066446059d2bee5410f2R6)?

agrare: @Fryguy this is the older "AvailabilityMixin" style versus SupportsFeature

app/controllers/api/cloud_volumes_controller.rb
skateman: You can't really do `OPTIONS /api/cloud_volumes/123` if you want to create the very first cloud volume :confused:

abellotti:
If possible, my preference here is to specify the type of the CloudVolume instead of the id of the hosting provider. This is for consistency with what we do with the OPTIONS on providers, i.e. do something like this:

```
OPTIONS /api/cloud_volumes?type=ManageIQ::Providers::Amazon::StorageManager::Ebs::CloudVolume
```

Also, with this PR, I'm getting a 500 when the provider specified does not have a ::CloudVolume.   Getting the following:

```
{
  "error":  {
     "kind":"internal_server_error",
     "message":"uninitialized constant ManageIQ::Providers::Vmware::InfraManager::CloudVolume","klass":"NameError"
    }
}
```

Probably need to check the klass exists before using it.

```
      ems = ExtManagementSystem.find(params[:ems_id])
      klass = ems.class::CloudVolume

      raise BadRequestError, "No DDF specified for - #{klass}" unless klass.respond_to?(:params_for_create)
```

chessbyte:
@abellotti do we really want to expose `type=ManageIQ::Providers::Amazon::StorageManager::Ebs::CloudVolume` in our API?  /cc @Fryguy @gtanzillo @agrare

Fryguy:
Unfortunately, we've already exposed those types.  It's not uncommon to query an endpoint for a particular type or sub-type.

skateman:
@abellotti we're asking with an `ems_id` and not a class because we don't really have access to it. The dropdown is being feeded with `name` and `emd_id` using `loadOptions` and an `onChange` picks up any change and fires the OPTIONS request. Theoretically we can cache the result of the `loadOptions` but it would be super hacky and definitely not *The React Way&trade;* of doing things. Also we're submitting the same `ems_id` with the form, so the dropdown can't just be altered to have both the class and the `ems_id` as its `value`...

abellotti:
Thanks @skateman for looking into it and adding the check for the class existence. LGTM!! 👍

abellotti (approved):

Issue #898, opened 2020-09-25
Trying to edit a Zone's authentication(s) crashes with AssociationTypeMismatch


skateman:
When I try this request:
```js
# POST /api/zones/X
{
  "action": "edit",
  "resource": {
    "authentications": [{ 
      "userid": "foo",
      "password": "bar", 
    }]
  }
}
```

It fails with:
```
[ActiveRecord::AssociationTypeMismatch] Authentication(#70163535308460) expected, got {"userid"=>"foo", "password"=>"bar"} which is an instance of Hash(#13895400).
```

I guess *The Right Way&trade;* would be to send three requests instead. One to check if there are any authentications, one to update the authentication and one to update the zone. Which might raise transactional concerns, because AFAIK there is no rollback support for chained API requests and I'm not sure if we want to compensate this feature on the frontend.

Anyway, the authentications is a special case in the `Zone` model as there's either zero or only one assinged. I am pretty sure I can fix this, but I have way too many ideas of how and I'm not entirely sure which one to choose.

Related issue: https://github.com/ManageIQ/manageiq-ui-classic/issues/6917

@gtanzillo @abellotti @Fryguy 
@miq-bot add_label question

skateman:
@miq-bot assign @gtanzillo

NickLaMuro:
@skateman So I looked into this by literally writing a test to replicate the failure:

```diff
diff --git a/app/controllers/api/base_controller/generic.rb b/app/controllers/api/base_controller/generic.rb
index 6b943e8f..228c4f07 100644
--- a/app/controllers/api/base_controller/generic.rb
+++ b/app/controllers/api/base_controller/generic.rb
@@ -50,6 +50,9 @@ module Api
       def edit_resource(type, id, data)
         klass = collection_class(type)
         resource = resource_search(id, type, klass)
+        puts
+        puts data.except(*ID_ATTRS)
+        puts
         resource.update!(data.except(*ID_ATTRS))
         resource
       end
diff --git a/spec/requests/zones_spec.rb b/spec/requests/zones_spec.rb
index 3e955c81..40db8098 100644
--- a/spec/requests/zones_spec.rb
+++ b/spec/requests/zones_spec.rb
@@ -25,6 +25,22 @@ RSpec.describe "Zones" do
   end
 
   context "edit", :edit do
+    it "can update zone authentications" do
+      api_basic_authorize action_identifier(:zones, :edit)
+
[...]

PR #916, opened 2020-10-09
[WIP][ZonesController] Allow authentications to be created/edited via a single zone update

NickLaMuro:
This hopefully will address https://github.com/ManageIQ/manageiq-api/issues/898 but arguably should be split up into multiple PRs.

This does the following:

- Updates `Api::BaseController::Generic#add_resource` and `Api::BaseController::Generic#edit_resource` to work within a transaction
- Updates `Api::BaseController::Generic#edit_resource` to `add_subcollection_data_to_resource` (`.add_resource` already had this functionality)
- Refactored some code around `add_subcollection_data_to_resource`
  - just allowed for it to be shared between `.edit_resource` and `.add_resource`, no functional changes made
- Added `Authentications` as a subcollection for `Zone`
- Added `authentications_assign_resource` to the `Subcollections::Authentications`, allowing for authentications to be created/edit via a zone action (and within a transaction).

The last item there is what was required for https://github.com/ManageIQ/manageiq-api/issues/898

Example Request
---------------

```
# POST /api/zones/X
{
  "action": "edit",
  "resource": {
    "description": "Updated Zone Description",
    "authentications": [{
      "href": "/",
      "windows_domain": {
        "userid": "foo",
        "password": "bar"
      }
    }]
  }
}
```

### DIFF
--- a/app/controllers/api/base_controller/generic.rb
+++ b/app/controllers/api/base_controller/generic.rb
@@ -13,17 +13,12 @@ module Api
       def add_resource(type, _id, data)
         assert_id_not_specified(data, "#{type} resource")
         klass = collection_class(type)
-        subcollection_data = collection_config.subcollections(type).each_with_object({}) do |sc, hash|
-          if data.key?(sc.to_s)
-            hash[sc] = data[sc.to_s]
-            data.delete(sc.to_s)
-          end
-        end
+        sc_data = subcollection_data(type, data)
         validate_type(klass, data['type']) if data['type']
         resource = klass.new(data)
         klass.transaction do
           if resource.save
-            add_subcollection_data_to_resource(resource, type, subcollection_data)
+            add_subcollection_data_to_resource(resource, type, sc_data)
             resource
           else
             raise BadRequestError, "Failed to add a new #{type} resource - #{resource.errors.full_messages.join(', ')}"
@@ -152,6 +147,15 @@ module Api
         klass.descendant_get(type)
       rescue ArgumentError => err
         raise BadRequestError, "Invalid type #{type} specified - #{err}"
+      end
+
+      def subcollection_data(type, data)
+        collection_config.subcollections(type).each_with_object({}) do |sc, hash|
+          if data.key?(sc.to_s)
+            hash[sc] = data[sc.to_s]
+            data.delete(sc.to_s)
+          end
+        end
       end
 
       def add_subcollection_data_to_resource(resource, type, subcollection_data)
--- a/app/controllers/api/base_controller/generic.rb
+++ b/app/controllers/api/base_controller/generic.rb
@@ -46,9 +46,13 @@ module Api
 
       def edit_resource(type, id, data)
         klass = collection_class(type)
+        sc_data = subcollection_data(type, data)
         resource = resource_search(id, type, klass)
-        resource.update!(data.except(*ID_ATTRS))
-        resource
+        klass.transaction do
+          resource.update!(data.except(*ID_ATTRS))
+          add_subcollection_data_to_resource(resource, type, sc_data)
+          resource
+        end
       end
 
       def delete_resource(type, id = nil, _data = nil)
--- a/app/controllers/api/base_controller/generic.rb
+++ b/app/controllers/api/base_controller/generic.rb
@@ -21,11 +21,13 @@ module Api
         end
         validate_type(klass, data['type']) if data['type']
         resource = klass.new(data)
-        if resource.save
-          add_subcollection_data_to_resource(resource, type, subcollection_data)
-          resource
-        else
-          raise BadRequestError, "Failed to add a new #{type} resource - #{resource.errors.full_messages.join(', ')}"
+        klass.transaction do
+          if resource.save
+            add_subcollection_data_to_resource(resource, type, subcollection_data)
+            resource
+          else
+            raise BadRequestError, "Failed to add a new #{type} resource - #{resource.errors.full_messages.join(', ')}"
+          end
         end
       end
 
--- a/app/controllers/api/subcollections/authentications.rb
+++ b/app/controllers/api/subcollections/authentications.rb
@@ -11,6 +11,22 @@ module Api
       rescue => err
         action_result(false, err.to_s)
       end
+
+      def authentications_assign_resource(parent, _type, _id, data)
+        authentication_data = data.except("href").with_indifferent_access
+        validate_authentication_type(authentication_data)
+
+        parent.update_authentication(authentication_data, {:save => true})
+      end
+
+      private
+
+      # Requires implementing valid_authentication_types in included subclass
+      def validate_authentication_type(data)
+        if data.keys != 1 && !valid_authentication_types.include?(data.keys.first)
+          raise BadRequestError, "Invalid type specified for authentication"
+        end
+      end
     end
   end
 end
--- a/app/controllers/api/zones_controller.rb
+++ b/app/controllers/api/zones_controller.rb
@@ -3,6 +3,7 @@ module Api
     include Subcollections::Authentications
 
     INVALID_ZONES_ATTRS = ID_ATTRS + %w[created_on updated_on].freeze
+    VALID_AUTH_TYPES    = %w[windows_domain].freeze
 
     # Edit an existing zone. Certain fields are meant for internal use only
     # and may not be edited. Attempting to edit one of the forbidden fields
@@ -29,6 +30,10 @@ module Api
       return nil unless data
 
       data.keys.select { |key| INVALID_ZONES_ATTRS.include?(key) }.compact.join(", ")
+    end
+
+    def valid_authentication_types
+      VALID_AUTH_TYPES
     end
   end
 end
--- a/app/controllers/api/zones_controller.rb
+++ b/app/controllers/api/zones_controller.rb
@@ -1,5 +1,7 @@
 module Api
   class ZonesController < BaseController
+    include Subcollections::Authentications
+
     INVALID_ZONES_ATTRS = ID_ATTRS + %w[created_on updated_on].freeze
 
     # Edit an existing zone. Certain fields are meant for internal use only
--- a/spec/requests/zones_spec.rb
+++ b/spec/requests/zones_spec.rb
@@ -25,6 +25,42 @@ RSpec.describe "Zones" do
   end
 
   context "edit", :edit do
+    it "can create zone authentications" do
+      api_basic_authorize action_identifier(:zones, :edit)
+
+      zone   = FactoryBot.create(:zone, :description => "Current Zone description")
+      params = gen_request(:edit,
+                           "description"     => "Updated Zone Description",
+                           "authentications" => [{"href" => "/", "windows_domain" => {"userid" => "foo", "password" => "bar"}}])
+
+      expect(zone.authentications).to eq([]) # sanity check
+
+      post api_zone_url(nil, zone), :params => params
+      zone.reload
+
+      expect(zone.description).to                    eq("Updated Zone Description")
+      expect(zone.authentications.first.userid).to   eq("foo")
+      expect(zone.authentications.first.password).to eq("bar")
+    end
+
+    it "rejects authentication if not of a valid authentication type" do
+      api_basic_authorize action_identifier(:zones, :edit)
+
+      zone   = FactoryBot.create(:zone, :description => "Current Zone description")
+      params = gen_request(:edit,
+                           "description"     => "Updated Zone Description",
+                           "authentications" => [{"href" => "/", "unix_domain" => {"userid" => "foo", "password" => "bar"}}])
+
+      expect(zone.authentications).to eq([]) # sanity check
+
+      post api_zone_url(nil, zone), :params => params
+      zone.reload
+
+      expect(response).to                   have_http_status(:bad_request)
+      expect(zone.description).to           eq("Current Zone description")
+      expect(zone.authentications.count).to eq(0)
+    end
+
     it "will fail if you try to edit invalid fields" do
       api_basic_authorize action_identifier(:zones, :edit)
 
--- a/config/api.yml
+++ b/config/api.yml
@@ -275,6 +275,8 @@
       :post:
       - :name: create
         :identifier: embedded_automation_manager_credentials_add
+      - :name: edit
+        :identifier: embedded_automation_manager_credentials_edit
   :automate:
     :description: Automate
     :options:
@@ -4800,6 +4802,7 @@
     - :collection
     :subcollections:
     - :settings
+    - :authentications
     :verbs: *gpppd
     :klass: Zone
     :collection_actions:


NickLaMuro:
@skateman this probably needs a lot of scrutiny from @abellotti @Fryguy and others, but hopefully this should give you what you were looking for out of the API from https://github.com/ManageIQ/manageiq-api/issues/898

cc @gtanzillo

spec/requests/zones_spec.rb
```diff
@@ -25,6 +25,25 @@
   end
 
   context "edit", :edit do
+    it "can create zone authentications" do
+      api_basic_authorize action_identifier(:zones, :edit)
+
+      zone   = FactoryBot.create(:zone, :description => "Current Zone description")
+      params = gen_request(:edit,
+                           "description"     => "Updated Zone Description",
+                           "authentications" => [{"href" => "/", "userid" => "foo", "password" => "bar"}])
```
NickLaMuro, line 34: Just a heads up that `"href"` is wrong here, but I didn't want to spend too much more time getting a POC for this out, and since it isn't used for our purposes, this just allows the `Href` code in `add_subcollection_data_to_resource` to function without error for this spec.

Moving forward, I should figure out the #RightWay™ to do this, but for now, I just wanted a POC working.

NickLaMuro: A note to @skateman that the `"href"` key will be required, which is different from your https://github.com/ManageIQ/manageiq-api/issues/898 format, but I don't see that being much of an issue.

This is existing functionality I am working off of here.

app/controllers/api/subcollections/authentications.rb
```diff
@@ -11,6 +11,18 @@ def authentications_create_resource(parent, _type, _id, data)
       rescue => err
         action_result(false, err.to_s)
       end
+
+      def authentications_assign_resource(parent, _type, _id, data)
+        parent.update_authentication(
+          {
+            :windows_domain => {
```
Fryguy, line 18: I wasn't expecting this to be hardcoded... Is this intentional?

Fryguy: Or rather hardcoded in the base method.  I could kind of see it hardcoded at the zones layer, since that is zone specific, but even that hardcoding I'm not sure about

Fryguy (commented):
One thing I noticed

app/controllers/api/subcollections/authentications.rb
NickLaMuro: I referenced this in my ":book:ish" comment:

https://github.com/ManageIQ/manageiq-api/issues/898#issuecomment-702913963

But I was pulling from the `OpsController::Settings::Zones#zone_set_record_vars` method for reference:

https://github.com/ManageIQ/manageiq-ui-classic/blob/92cc572b/app/controllers/ops_controller/settings/zones.rb#L116

Where it was hard coded.

* * *

That said, I don't disagree with your point, so I wouldn't be opposed to forcing this from a client to fill in via the `{:type => {:userid => "userid", :password => "..."}}` format, just started with how it was in the UI.

NickLaMuro: Of note:  This would be different from the expected query format that was described in #898 (in the issue description).

Fryguy: Oh I see now where you said "...taking all of the deleted code from the controller in the UI and slapping it back into the API".  I guess that feels wrong to me because we feel the controller code is _also_ not great in that it hardcodes.

I _think_ I'm just not ok with hardcoding in `subcollections/authentications.rb` which is a common base, but I am _kind of_ ok with hardcoding in `api/zones_controller.rb` as that's where the hardcoding waas previously in the UI controller.

My biggest concern is that once we hardcode it and the incoming API takes on a specific shape, then later if we want to change it to add back in the authetication_type, for whatever reason, then we have to introduce a breaking change.

NickLaMuro: > I think I'm just not ok with hardcoding in `subcollections/authentications.rb`

That works.  This was mostly a POC to prove that doing the `authentications_assign_resource` format works, and then we can work out the proper implementation of it.  We still need to add the subcollection, but I think the method can live in `app/controllers/api/zones_controller.rb`.

I am also not opposed to telling the clients they have to pass the `:type` along with their request, since if that specific logic is required, the hard coding of it should live where it needs to be hard coded, and not in the UI.  It also allows authentications for Zones in the API to be more open ended in the future, and only requires the clients to adapt as necessary.

skateman:
Any update on this?

NickLaMuro:
@skateman sorry for the delay on this, but I addressed the feedback from Jason, and had to do a bit of refactoring as a result, but I think this is in a good spot now.

The spec has changed for the API payload where you now have to pass the `:windows_domain => { ... }` as a key wrapping the authentication record data, but that allows this subcollection to be used more universally.  Some of this logic might make sense being pushed back into the model in core, but for simplicity of making this 1 PR instead of two, I didn't do that just yet.  There is some architectural implementations with doing that as well which i think needs to be discussed, so punting on that for now.

skateman:
Can you please post an example request?

NickLaMuro:
@skateman see commit message and tests.

skateman:
I think I can work with this one, thanks!

spec/requests/zones_spec.rb
abellotti: for the usage, href is not required, so let's not require it and check for it internally, from what I see in add_subcollection_data_to_resource we use it to zap identifying attrs, maybe we can just check for it first.

abellotti:
I think for formally exposing the authentication as a subcollection we need to make sure the related calls work as expected. 

since we can add/edit via a zone update, the following also works:

```
GET /api/zones/2/authentications
```

that means we `should` be able to do:

```
POST /api/zones/2/authentications
PATCH /api/zones/2/authentications/3
DELETE /api/zones/2/authentications/3
```

Some things need visiting:

```
DELETE /api/zones/2/authentications/3

{
  "error": {
    "message": "undefined method `[]' for nil:NilClass",
    "klass": "NoMethodError"
  }
}
```

```
POST /api/zones/2/authentications

{
  "windows_domain" : {
    "userid" : "foo",
    "password" : "bar"
  }
}

{
  "results": [
    {    
      "success": false,
      "message": "undefined method `manager' for #<Zone:0x00007f8365f63980>"
    }    
  ]
}
```

I think to keep authentications enabled as a subcollection, we can control what subcollection action we want enabled
for zones in the config/api.yml zone definition itself via an authentications_subcollection_actions: section.

Another option, is not to declare as a subcollection, but have the create and update methods of the zones controller deal with it separately. Similar to what we do with provider endpoints during create/updates. This will also remove the changes done in  generic.rb and keep that in the zones controller.

app/controllers/api/zones_controller.rb
```diff
@@ -1,6 +1,9 @@
 module Api
   class ZonesController < BaseController
+    include Subcollections::Authentications
+
     INVALID_ZONES_ATTRS = ID_ATTRS + %w[created_on updated_on].freeze
+    VALID_AUTH_TYPES    = %w[windows_domain].freeze
```
abellotti, line 6: should be made available to the API via app/models/zone.rb

Fryguy: You mean, like the "possible" values?  I don't think so, because we don't do that for other things.  This specfic value, currently, would not be used in anything like a drop down, so there would be no caller for that kind of method.

abellotti: Interesting, even though "windows_domain" is a value we look for in host and vms ?  (just doing a git grep windows_domain). You'd think it's a model declared list (valid auth types).

Fryguy: Oh sorry...I misunderstood your request.  I thought you were stating you wanted it exposed over the API as opposed to what you really meant which was declaring it in the model.  On that I agree...this list should be in the model  See also my comment: https://github.com/ManageIQ/manageiq-api/pull/916#discussion_r502706666

Fryguy: As per 

> I am _kind of_ ok with hardcoding in api/zones_controller.rb as that's where the hardcoding waas previously in the UI controller

I'd rather it ultimately made it into the Zone model, but I'd be ok with a follow-up PR.

abellotti: 👍

spec/requests/zones_spec.rb
NickLaMuro: @abellotti I will double check this tomorrow (I am only reading this now), but I am almost positive this was necessary for some reason, otherwise the tests wouldn't pass...

Again, though, will provide a better answer tomorrow.

app/controllers/api/zones_controller.rb
NickLaMuro: > I'd rather it ultimately made it into the Zone model, but I'd be ok with a follow-up PR.

I agree with this as well.

I think a follow up makes the most sense at this point just so we can get something out to the UI team, and then fix in a follow up (which will require a little more coordination than a single PR and merge)

NickLaMuro:
> > I think for formally exposing the authentication as a subcollection we need to make sure the related calls work as expected...
> 
> Some things need visiting...
> 
> etc. ...

I think I will need to take some time to digest this statement, based on my limited understanding (comparatively) of the API to fully respond, but sounds reasonable from what I do understand.  Will look into this more tomorrow and respond if I have questions (which is very likely that I will).

Fryguy (commented):
This PR LGTM, and I can approve once the request in https://github.com/ManageIQ/manageiq-api/pull/916#issuecomment-759702973 is done.  Assigning to @gtanzillo.

NickLaMuro:
@Fryguy that is what I have been looking into, and I don't think either is an easy solution... it will either be a re-write ("Another option") or having to add more to the API to support having the `*_subcollection_actions` actually behave in a exclusive manner, not inclusive (how it is currently).

NickLaMuro:
@Fryguy Okay, I should have just tested this again with `/api/cloud_tenant/:id/tags` instead of spending half a day (the other day) trying to understand how this "should" work.  And I assume by the " 😕 " emoji, you didn't quite understand what I was talking about previously.

So for reference, I was referencing this portion of the `config/api.yml`:

https://github.com/ManageIQ/manageiq-api/blob/e174d80c96ca3f9f796ad5aa70c680752a99f9c3/config/api.yml#L768-L773

Which, by the explanation given to me by @abellotti made it seem like adding `:authentications_subcollection_actions:` section to the `:zones:` section in he `config/api.yml`, this would be an "allow list" of sorts for which actions could be used.

However, when I tested this with `/api/cloud_tenant/:id/tags`, I got the following response:

```console
$ curl --silent -u admin:smartvm -X GET http://localhost:3000/api/cloud_tenants/1/tags | json
{
    "actions": [
        {
            "href": "http://localhost:3000/api/cloud_tenants/1/tags",
            "method": "post",
            "name": "assign"
        },
        {
            "href": "http://localhost:3000/api/cloud_tenants/1/tags",
            "method": "post",
            "name": "unassign"
        }
    ],
    "count": 231,
    "links": {
        "first": "http://localhost:3000/api/cloud_tenants/1/tags?offset=0",
        "last": "http://localhost:3000/api/cloud_tenants/1/tags?offset=0",
        "self": "http://localhost:3000/api/cloud_tenants/1/tags?offset=0"
    },
    "name": "tags",
    "pages": 1,
    "resources": [],
    "subcount": 0
}
```

```console
$ curl --silent -u admin:smartvm -X GET http://localhost:3000/api/tags/1 | json
{
    "actions": [
        {
            "href": "http://localhost:3000/api/tags/1",
            "method": "post",
            "name": "edit"
        },
        {
            "href": "http://localhost:3000/api/tags/1",
            "method": "post",
            "name": "delete"
        },
        {
            "href": "http://localhost:3000/api/tags/1",
            "method": "delete",
            "name": "delete"
        }
    ],
    "href": "http://localhost:3000/api/tags/1",
    "id": "1",
    "name": "/managed/location"
}
```

And it seems like the `routes` are also not defined in a way to exclude the routes that aren't allowed by the `*_subcollection_actions` as well:

```
$ bundle exec rake routes | grep api | grep tags | grep cloud_tenant
                      OPTIONS          /api(/:version)/cloud_tenants/:c_id/tags(.:format)         api/cloud_tenants#options {:format=>"json", :subcollection=>true, :version=>/v\d+(\.[\da-zA-Z]+)*(\-[\da-zA-Z\-\.]+)?/}
api_cloud_tenant_tags GET              /api(/:version)/cloud_tenants/:c_id/tags(.:format)         api/cloud_tenants#index {:format=>"json", :version=>/v\d+(\.[\da-zA-Z]+)*(\-[\da-zA-Z\-\.]+)?/}
 api_cloud_tenant_tag GET              /api(/:version)/cloud_tenants/:c_id/tags/:s_id(.:format)   api/cloud_tenants#show {:format=>"json", :version=>/v\d+(\.[\da-zA-Z]+)*(\-[\da-zA-Z\-\.]+)?/}
                      POST             /api(/:version)/cloud_tenants/:c_id/tags(.:format)         api/cloud_tenants#create {:format=>"json", :version=>/v\d+(\.[\da-zA-Z]+)*(\-[\da-zA-Z\-\.]+)?/}
                      POST             /api(/:version)/cloud_tenants/:c_id/tags(/:s_id)(.:format) api/cloud_tenants#update {:format=>"json", :version=>/v\d+(\.[\da-zA-Z]+)*(\-[\da-zA-Z\-\.]+)?/}
                      DELETE           /api(/:version)/cloud_tenants/:c_id/tags/:s_id(.:format)   api/cloud_tenants#destroy {:format=>"json", :version=>/v\d+(\.[\da-zA-Z]+)*(\-[\da-zA-Z\-\.]+)?/}
```

That said, it seems like this was intended to not be handled in the routes, but as `before_action` filters:

https://github.com/ManageIQ/manageiq-api/blob/e174d80c96ca3f9f796ad5aa70c680752a99f9c3/app/controllers/api/base_controller.rb#L27-L29
https://github.com/ManageIQ/manageiq-api/blob/e174d80c96ca3f9f796ad5aa70c680752a99f9c3/app/controllers/api/base_controller/parser.rb#L171-L172
https://github.com/ManageIQ/manageiq-api/blob/e174d80c96ca3f9f796ad5aa70c680752a99f9c3/lib/api/collection_config.rb#L57-L59

But I didn't fully grasp how those were supposed to work, and clearly by the above, it isn't working.

* * *

So, what I was saying is that the first suggestion by Alberto just didn't do what he said it would do.  The other option suggested, as I mentioned previously, is re-write of what I have here and I have also pushed back on the idea in the past:

https://github.com/ManageIQ/manageiq-api/issues/898#issuecomment-702913963

> ...
> 
> ### Possible solutions
> 
> * Make a custom update action for the ZoneController, and allow that to do specific items for itself
>   * This is the quick solution, but basically assumes custom documentation and coding for this particular endpoint, and I would argue this opens the gates for more of these types of "quick and dirty" solutions in the future
>   * The result is we are basically taking all of the deleted code from the controller in the UI and slapping it back into the API
>

I personally think that decision will lead to a maintenance nightmare in the API, but at this point, it is not a hill I am willing to die on to defend.


cc @gtanzillo

app/controllers/api/base_controller/generic.rb
```diff
@@ -152,6 +153,15 @@ def validate_type(klass, type)
...
       end
 
+      def subcollection_data(type, data)
+        collection_config.subcollections(type).each_with_object({}) do |sc, hash|
+          if data.key?(sc.to_s)
+            hash[sc] = data[sc.to_s]
+            data.delete(sc.to_s)
+          end
+        end
+      end
```
Fryguy, line 163: Is this all just basically `Hash#extract!`?
```suggestion
      def subcollection_data(type, data)
        data.extract!(collection_config.subcollections(type).map(&:to_s))
      end
```

app/controllers/api/subcollections/authentications.rb
```diff
@@ -11,6 +11,22 @@ def authentications_create_resource(parent, _type, _id, data)
...
+        authentication_data = data.except("href").with_indifferent_access
+        validate_authenication_type(authentication_data)
+
+        parent.update_authentication(authentication_data, {:save => true})
+      end
+
+      private
+
+      # Requires implementing validate_authenication_type in included subclass
+      def validate_authenication_type(data)
```
Fryguy, line 25: Typo

```suggestion
      def validate_authentication_type(data)
```

Fryguy:
As discussed, the `authentications_subcollection_actions` will only filter *actions*, but does nothing for *verbs*, which is really what is needed in this case.  So, the choices are 

1. add support for filtering verbs somehow (maybe a `authentications_subcollection_verbs`?), then filter on verbs
2. fully add the code to support all of the verbs in the subcollection
3. create a custom subcollection that only implements the methods in question.

To me it feels like option 2 is the best option, but I don't know how much more work that is.

app/controllers/api/base_controller/generic.rb
NickLaMuro: Not going to exit this because this was me just extracting this out of the previous method to share it.

https://github.com/ManageIQ/manageiq-api/pull/916/files#diff-6388f564e8b184b4d6d0380cfd06f502fff6fe45fbd1972cf5c66a47857912d4L16-L21

Refactoring like this can come at a later date.

NickLaMuro:
> 3. create a custom subcollection that only implements the methods in question.
>
> Option 3 feels like technical debt waiting to happen (as @NickLaMuro said, "I personally think that decision will lead to a maintenance nightmare in the API")

For the record, this isn't what I was referring to, as I was saying implementing a **custom edit action** in the `ZoneController` as being the cause to future technical debt.  This was in reference to both what was suggested by Alberto and David:

> Another option, is not to declare as a subcollection, but have the create and update methods of the zones controller deal with it separately. Similar to what we do with provider endpoints during create/updates. [[ref](https://github.com/ManageIQ/manageiq-api/pull/916#issuecomment-759702973)]

> @NickLaMuro we have this sorted out for [providers](https://github.com/ManageIQ/manageiq/blob/master/app/models/ext_management_system.rb#L62) [[ref](https://github.com/ManageIQ/manageiq-api/issues/898#issuecomment-702594772)]

NickLaMuro:
Closing as this was solved by just removing this feature from Zones in the UI:

https://github.com/ManageIQ/manageiq-ui-classic/pull/7650

Issue #918, opened 2020-10-13
Entity creation is not asynchronous where it should be


skateman:
I found a bunch of API endpoints that don't create entities *The Right Way&trade;*, i.e.using an asynchronous task initiated by `create_xy_queue` but just by creating the DB record:
* [ ] auth_key_pairs - https://github.com/ManageIQ/manageiq-api/pull/921

@miq-bot add_label bug
FYI @agrare

skateman:
@miq-bot assign @gtanzillo 

This is kinda big, the API is basically broken for these endpoints and doesn't actually create the entities that should be created, only their DB counterparts.

skateman:
@miq-bot add_label help wanted

skateman:
Okay, I figured out that the ones I thought are done wrong actually are nonexistent endpoints, so the issue is just with `auth_key_pairs`.

PR #921, opened 2020-10-14
Make the auth key pair creating asynchronous

skateman:
The keypair creation was only defined implicitly in the API, which means that the request payload was mapped to the database. However, this never actually creates a keypair in the provider which is a bug. As a solution I'm using the same method I found in the UI controller that returns with a task...

Fixes https://github.com/ManageIQ/manageiq-api/issues/918

@miq-bot add_label bug
fyi @agrare

### DIFF
--- a/app/controllers/api/auth_key_pairs_controller.rb
+++ b/app/controllers/api/auth_key_pairs_controller.rb
@@ -1,5 +1,19 @@
 module Api
   class AuthKeyPairsController < BaseController
+    def create_resource(_type, _id = nil, data = {})
+      ext_management_system = ExtManagementSystem.find(data['ems_id'])
+
+      klass = ManageIQ::Providers::CloudManager::AuthKeyPair.class_by_ems(ext_management_system)
+
+      validate = klass.validate_create_key_pair(ext_management_system)
+      raise validate[:message] unless validate[:available]
+
+      task_id = klass.create_key_pair_queue(session[:userid], ext_management_system, data)
+      action_result(true, "Creating Cloud Key Pair #{data['name']} for Provider: #{ext_management_system.name}", :task_id => task_id)
+    rescue => err
+      action_result(false, err.to_s)
+    end
+
     def delete_resource(type, id, _data = {})
       delete_action_handler do
         key_pair = resource_search(id, type, collection_class(type))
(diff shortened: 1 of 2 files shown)


app/controllers/api/auth_key_pairs_controller.rb
```diff
@@ -1,5 +1,13 @@
 module Api
   class AuthKeyPairsController < BaseController
+    def create_resource(_type, _id = nil, data = {})
+      ext_management_system = ExtManagementSystem.find(data['ems_id'])
+      task_id = ManageIQ::Providers::CloudManager::AuthKeyPair.create_key_pair_queue(session[:userid], ext_management_system, data)
```
agrare, line 5: We should probably check `.validate_create_key_pair` before calling `create_`

skateman: Well, in that case I will need a way to determine the namespaced subclass of the `AuthKeyPair` under a given ems ...

agrare: Isn't that what `class_by_ems` does?  e.g. https://github.com/ManageIQ/manageiq/blob/master/app/models/manageiq/providers/cloud_manager/auth_key_pair.rb#L41

app/controllers/api/auth_key_pairs_controller.rb
```diff
@@ -1,5 +1,13 @@
...
+      action_result(true, "Creating Cloud Key Pair #{data['name']} for Provider: #{ext_management_system.name}", :task_id => task_id)
```
gtanzillo, line 6: So the response is now a task_id. Assuming the original code returned the instance created in the DB, this could be a breaking change for existing consumers.

skateman: Well, if an existing customer ever used this feature, it never actually worker for them. This is one of those issues where we mapped the HTTP payload into our DB but we never relayed the information towards the provider.

skateman:
Can we get this in?

app/controllers/api/auth_key_pairs_controller.rb
agrare: My opinion, if it was broken we need to fix it, and if we break the API in doing so it was already not working

@Fryguy @abellotti what do you think?

gtanzillo: I agree with both @agrare @skateman comments. Perhaps we can work this out in the API meeting in a little while

PR #927, opened 2020-10-19
Allow patch/put/delete methods on the CloudNetwork resource

skateman:
REST should be REST :wink:

FYI @gubbe505 
@miq-bot assign @lpichler

### DIFF
--- a/config/api.yml
+++ b/config/api.yml
@@ -531,7 +531,7 @@
     - :collection
     - :subcollection
     - :custom_actions
-    :verbs: *gp
+    :verbs: *gpppd
     :klass: CloudNetwork
     :subcollections:
     - :tags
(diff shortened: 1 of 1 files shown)


Fryguy:
Why is this WIP?

skateman:
> Why is this WIP?

because I forgot :see_no_evil:

lpichler (approved):

abellotti:
Missing specs

abellotti:
Nothing automatic spec wise for routes. Put/Patch logic was tested via Post edit action, true, but not routes. However delete logic would be 100% new.  

Also, this is leveraging he generic delete, so sync not async delete.  Is this ok for cloud network resources ?

skateman:
How can I force this to leverage the non-generic delete? There's a `delete_resource` method defined in the controller...

abellotti:
ahh, nm!!  If delete_resource code is already there (didn't see it in this PR), then that's what gets called from the generic handler.

PR #932, opened 2020-10-21
Added api for remove physical storage

llivne:
Adding functionality for deleting supported physical storages via rest-API

links
-----------
the following are related PRs:
https://github.com/ManageIQ/manageiq/pull/20713
https://github.com/ManageIQ/manageiq-providers-autosde/pull/38
https://github.com/ManageIQ/manageiq-ui-classic/pull/7433

### DIFF
--- a/app/controllers/api/physical_storages_controller.rb
+++ b/app/controllers/api/physical_storages_controller.rb
@@ -29,7 +29,7 @@ module Api
         begin
           physical_storage = resource_search(id, type, klass)
           unless physical_storage.supports?(:delete)
-            error_msg = "#{physical_storage.name} raised the following error: " + physical_storage.unsupported_reason(:delete)
+            error_msg = "Failed to delete #{physical_storage.name}: #{physical_storage.unsupported_reason(:delete)}"
             raise error_msg
           end
           msg = "Detaching #{physical_storage_ident(physical_storage)}"
--- a/app/controllers/api/physical_storages_controller.rb
+++ b/app/controllers/api/physical_storages_controller.rb
@@ -20,6 +20,26 @@ module Api
       action_result(false, err.to_s)
     end
 
+    def delete_resource(type, id, _data = nil)
+      raise BadRequestError, "Must specify an id for deleting a #{type} resource" if id.blank?
+
+      api_action(type, id) do |klass|
+        begin
+          physical_storage = resource_search(id, type, klass)
+          unless physical_storage.supports?(:delete)
+            error_msg = "#{physical_storage.type} provider type does not support deleting. #{physical_storage.name} could not be deleted."
+            raise error_msg
+          end
+          msg = "Detaching #{physical_storage_ident(physical_storage)}"
+          api_log_info(msg)
+          task_id = physical_storage.delete_physical_storage_queue(User.current_user)
+          action_result(true, msg, :task_id => task_id)
+        rescue => err
+          action_result(false, err.to_s)
+        end
+      end
+    end
+
     private
 
     def ensure_resource_exists(type, id)
--- a/app/controllers/api/physical_storages_controller.rb
+++ b/app/controllers/api/physical_storages_controller.rb
@@ -23,6 +23,8 @@ module Api
     def delete_resource(type, id, _data = nil)
       raise BadRequestError, "Must specify an id for deleting a #{type} resource" if id.blank?
 
+      ensure_resource_exists(type, id) if single_resource?
+
       api_action(type, id) do |klass|
         begin
           physical_storage = resource_search(id, type, klass)
--- a/app/controllers/api/physical_storages_controller.rb
+++ b/app/controllers/api/physical_storages_controller.rb
@@ -27,7 +27,7 @@ module Api
         begin
           physical_storage = resource_search(id, type, klass)
           unless physical_storage.supports?(:delete)
-            error_msg = "#{physical_storage.type} provider type does not support deleting. #{physical_storage.name} could not be deleted."
+            error_msg = "#{physical_storage.name} raised the following error: " + physical_storage.unsupported_reason(:delete)
             raise error_msg
           end
           msg = "Detaching #{physical_storage_ident(physical_storage)}"
--- a/config/api.yml
+++ b/config/api.yml
@@ -2317,6 +2317,8 @@
         :identifier: physical_storage_new
       - :name: refresh
         :identifier: physical_storage_refresh
+      - :name: delete
+        :identifier: physical_storage_delete
     :resource_actions:
       :get:
       - :name: read
@@ -2324,6 +2326,8 @@
       :post:
       - :name: refresh
         :identifier: physical_storage_refresh
+      - :name: delete
+        :identifier: physical_storage_delete
   :physical_switches:
     :description: Physical Switches
     :identifier: physical_switch
--- a/config/api.yml
+++ b/config/api.yml
@@ -2305,7 +2305,7 @@
     :identifier: physical_storage
     :options:
     - :collection
-    :verbs: *gp
+    :verbs: *gpd
     :klass: PhysicalStorage
     :subcollections:
     :collection_actions:
--- a/spec/requests/physical_storages_spec.rb
+++ b/spec/requests/physical_storages_spec.rb
@@ -21,7 +21,24 @@ describe "Physical Storages API" do
     end
   end
 
-  context "Delete physical storage" do
+  context "Physical Storages delete action" do
+    it "with an invalid id" do
+      api_basic_authorize(action_identifier(:physical_storages, :delete, :resource_actions, :post))
+
+      post(api_physical_storage_url(nil, 999_999), :params => gen_request(:delete))
+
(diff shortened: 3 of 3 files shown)


llivne:
Travis fails because he need the changes done in https://github.com/ManageIQ/manageiq/pull/20713. (you can see that all is green in https://github.com/ManageIQ/manageiq-cross_repo-tests/pull/196)

app/controllers/api/physical_storages_controller.rb
```diff
@@ -20,6 +20,22 @@ def create_resource(_type, _id = nil, data = {})
       action_result(false, err.to_s)
     end
 
+    def delete_resource(type, id, _data = nil)
+      raise BadRequestError, "Must specify an id for deleting a #{type} resource" if id.blank?
+
+      ensure_resource_exists(type, id) if single_resource?
+
+      api_action(type, id) do |klass|
+        physical_storage = resource_search(id, type, klass)
```
agrare, line 29: @llivne Re: my comment here https://github.com/ManageIQ/manageiq/pull/20713/files#r510124331

You can use `supports?(:delete)` here e.g.`raise physical_storage.unsupported_reason(:delete) unless physical_storage.supports?(:delete)`

llivne: done

llivne:
@agrare I cant figure out why the codecilmate fails. Can you assist me ?

app/controllers/api/physical_storages_controller.rb
```diff
@@ -20,6 +20,21 @@ def create_resource(_type, _id = nil, data = {})
...
+      raise BadRequestError, "Must specify an id for deleting a #{type} resource" if id.blank?
+
+      api_action(type, id) do |klass|
+        physical_storage = resource_search(id, type, klass)
+        raise physical_storage.unsupported_reason(:delete) unless physical_storage.supports?(:delete)
+        msg = "Detaching #{physical_storage_ident(physical_storage)}"
+        api_log_info(msg)
+        task_id = physical_storage.delete_physical_storage_queue(User.current_user)
+        action_result(true, msg, :task_id => task_id)
+      rescue => err
```
agrare, line 33: @llivne I believe this is causing your codeclimate issues, try adding a `begin` around this like
```
api_action(type, id) do |klass|
  begin
    physical_storage = resource_search(id, type, klass)
    etc...
  rescue => err
  end
end
```

llivne: You were right. fixed

app/controllers/api/physical_storages_controller.rb
```diff
@@ -20,6 +20,26 @@ def create_resource(_type, _id = nil, data = {})
...
     end
 
+    def delete_resource(type, id, _data = nil)
+      raise BadRequestError, "Must specify an id for deleting a #{type} resource" if id.blank?
+
+      api_action(type, id) do |klass|
+        begin
+          physical_storage = resource_search(id, type, klass)
+          unless physical_storage.supports?(:delete)
+            error_msg = "#{physical_storage.type} provider type does not support deleting. #{physical_storage.name} could not be deleted."
```
agrare, line 30: I'm not so sure that we want to override the unsupported error message, this would prevent possibly helpful messages for providers which do support physical_storage deletion but not on a specific instance.

We do `raise record.unsupported_reason(:feature) unless record.supports?(:feature)` in a lot of places and we should continue to for consistency with the rest of the API

config/api.yml
```diff
@@ -2311,13 +2311,17 @@
         :identifier: physical_storage_new
       - :name: refresh
         :identifier: physical_storage_refresh
+      - :name: delete
+        :identifier: physical_storage_delete
```
agrare, line 2315: Believe you need to update the `:verbs` to include `delete` so `:verbs: *gpd`

app/controllers/api/physical_storages_controller.rb
llivne: Ok, I thought it would be nice to have a more informative msg (like I said, if for example we mark 10 storages for deletion but only 1 does not support delete we will get an error msg with the info about the storage that was not deleted and not a general error msg), but I can understand why you want to keep the same logic. I will change it back to raise record.unsupported_reason(:feature)

spec/requests/physical_storages_spec.rb
```diff
@@ -21,6 +21,44 @@
...
+    it "Deletion of a single Physical Storage" do
+      provider = FactoryBot.create(:ems_autosde, :name => 'Autosde')
+      physical_storage = FactoryBot.create("ManageIQ::Providers::Autosde::StorageManager::PhysicalStorage", :name => 'test_storage', :ext_management_system => provider)
+      api_basic_authorize('physical_storage_delete')
+
+      post(api_physical_storage_url(nil, physical_storage), :params => gen_request(:delete))
+
+      expect_single_action_result(:success => true, :message => /Detaching Physical Storage id:#{physical_storage.id}/i, :href => api_physical_storage_url(nil, physical_storage))
+    end
+
```
agrare, line 34: Can you add specs for some of the error handling?  E.g. PhysicalStorage that doesn't support `:delete`, an ID that doesn't exist returns a 404, etc...

agrare:
@bdunne @gtanzillo @abellotti could you guys take a look when you get a chance?

app/controllers/api/physical_storages_controller.rb
llivne: @agrare what about this:
```
          unless physical_storage.supports?(:delete)
            error_msg = "#{physical_storage.name} raised the following error: " + physical_storage.unsupported_reason(:delete)
            raise error_msg
          end
```
so we will get error msg similar to that in the gui:
image

if you still dont like it let me know and I will revert it to just raise unsupported_reason(:delete)

spec/requests/physical_storages_spec.rb
llivne: done

config/api.yml
llivne: done

app/controllers/api/physical_storages_controller.rb
```diff
@@ -20,6 +20,28 @@ def create_resource(_type, _id = nil, data = {})
...
+    def delete_resource(type, id, _data = nil)
+      raise BadRequestError, "Must specify an id for deleting a #{type} resource" if id.blank?
+
+      ensure_resource_exists(type, id) if single_resource?
+
+      api_action(type, id) do |klass|
+        begin
+          physical_storage = resource_search(id, type, klass)
+          unless physical_storage.supports?(:delete)
+            error_msg = "#{physical_storage.name} raised the following error: " + physical_storage.unsupported_reason(:delete)
```
llivne, line 32: @agrare I've done all the requested modifications. The only issue left is my suggestion for error_msg. if you prefer to keep it simple I will change it to
`raise physical_storage.unsupported_reason(:delete) unless physical_storage.supports?(:delete)`

agrare:
@bdunne @gtanzillo @abellotti can you review?

app/controllers/api/physical_storages_controller.rb
agrare: You should use string interpolation instead of `+` so `"#{physical_storage.name} raised the following error: #{physical_storage.unsupported_reason(:delete)}"`

agrare: I'm actually surprised rubocop didn't complain about this cc @bdunne

llivne: > You should use string interpolation instead of `+` so `"#{physical_storage.name} raised the following error: #{physical_storage.unsupported_reason(:delete)}"`

fixed. I changed the error msg a bit to: `error_msg = "Failed to delete #{physical_storage.name}: #{physical_storage.unsupported_reason(:delete)}"`

abellotti:
LGTM!!  Maybe just the rubocop issues in app/controllers/api/physical_storages_controller.rb.

Fryguy (approved):
LGTM...will give @agrare a pass as well.

llivne:
> LGTM!! Maybe just the rubocop issues in app/controllers/api/physical_storages_controller.rb.

@abellotti, about the rubocop issues, without the begin the codeclimate fails (see https://github.com/ManageIQ/manageiq-api/pull/932#discussion_r512037141)

app/controllers/api/physical_storages_controller.rb
Fryguy: That's weird... do/rescue was introduced in Ruby 2.5, so it _should_ work

Fryguy:
@llivne That's fine...we'll have to review why that issue is happening on the bot (cc @bdunne)

agrare (approved):

app/controllers/api/physical_storages_controller.rb
agrare: Yeah codeclimate seems to still be checking ruby 2.4 syntax even though rubocop isn't.

abellotti (approved):

gtanzillo (approved):

PR #938, opened 2020-10-22
don't consider schema querys in query count

kbrock:
these query counts make tests too brittle
we did this over in core repo and it made things more stable over there


kbrock:
ugh - issues, opened another

PR #939, opened 2020-10-22
don't consider schema querys in query count

kbrock:
these query counts make tests too brittle
we did this over in core repo and it made things more stable over there

### DIFF
--- a/spec/support/db_query_matchers.rb
+++ b/spec/support/db_query_matchers.rb
@@ -1,0 +1,3 @@
+DBQueryMatchers.configure do |config|
+  config.schemaless = true
+end


kbrock:
/cc @NickLaMuro this will hopefully make it not necessary for some of the filtering on the 

@agrare added the filtering queries. I could have tried to do a single query for that or put into the ems select clauses but seems not worth the time for us right now

NickLaMuro:
> /cc @NickLaMuro this will hopefully make it not necessary for some of the filtering on the

@kbrock I think this is a partial thought.  I have an idea what you might mean, but could you finish this thought before I go into guessing?

Fryguy (commented):
LGTM ... My only other thought is maybe we should put this in the shared part of core so we don't have to re-implemented it in every repo/plugin

i.e. The stuff in [here](https://github.com/ManageIQ/manageiq-api/blob/master/spec/spec_helper.rb#L6) is loaded from core, so if we change the location in core, then every plugin benefits from this.

kbrock:
@NickLaMuro I notice that you often use filter clauses to test query counts.
I wasn't sure if you did this to avoid the schema queries botching up your specs (happened quite often for me) or if you're trying to avoid counting login and authentication type queries.

So I thought this may help.

NickLaMuro:
It will remove some of the junk a bit, but a lot of times, I am using it with request specs (or, at least I have been lately with performance PRs).  In those cases, auth and user role/feature lookup is what I am filtering out, so this probably won't help.

Probably doesn't hurt, but would save me from having to do what I am doing anyway.

Fryguy:
Should be handled now via https://github.com/ManageIQ/manageiq/pull/20723

PR #933, opened 2020-10-21
[Renderer] Fix duplicate includes with extra_cols

NickLaMuro:
When using just `klass.virtual_includes`, it only checks to see if the given `virtual_attribute`/`virtual_column` also includes associated with it.  On the surface, that seems like a sufficient check, but since a `virtual_attribute` can have a `:uses` associated with it, as well as `:arel`, that means the column could be included as both an entry to `:extra_cols`, and for `:include_for_find`.

The problem with certain columns for `:include_for_find` is that they can very easily lead to "LEFT JOIN bombs", as in the example shown in the specs.

Background:
-----------

`virtual_total :total_vms` and `virtual_aggregate :aggregate_vm_cpus` both use the `:vms` relationship on `ExtManagementSystem`, but the latter also has a secondary relationship to hardware, and is aliased as `:vm_hardwares` as a `.includes` target.  The resulting options that would be passed to Rbac would be:

```ruby
{
  :user             => #<User id: 1, name: "Administrator", ... >,
  :offset           => 0,
  :limit            => 1000,
  :extra_cols       => [:aggregate_vm_cpus, :total_vms],
  :include_for_find => {
    :vm_hardwares => {},
    :vms          => {}
  }
}
```

As a result, both of those two columns combined in this fashion previously caused the `vms` table to be `LEFT OUTER JOIN`'d twice, meaning it included all of the columns of that table (60+) twice for each row.  On a large dataset, with LEFT OUTER JOINS also causing a huge matrix of multiple number of duplicate rows meaning the SQL query for even a system with 1000+ Vms can be crippled with this one query.

This change just ensures that we are avoiding the `:include_for_find` when that will be present and gathered via a `:extra_cols` elsewhere, hopefully avoiding some of these "LEFT JOIN bombs" as a result.

Links
-----

* Fixes https://github.com/ManageIQ/manageiq-api/issues/923

PR #940, opened 2020-10-23
[Renderer] Use next, not return

NickLaMuro:
Fixes Jansa Failures: https://github.com/ManageIQ/manageiq-api/pull/933#issuecomment-715313395

In the context of this block, using `return` will eject from the iteration of the block, AND also the method, so in all of usages of `return` previously, it will `nil`, which is a valid return value for the method.

With `next`, it will just skip anything after the `yield` and move on to the next item in the iterator (which isn't obvious we are in one with this code change).  This is the behavior that was intended, but because the specs around this allow skipped at the proper times with the previous implementation, it wasn't know that this was a problem.

`next` is also used in the `determine_include_for_find` method, so this should be a relatively safe and tested change.

Links
-----

* Noticed in https://github.com/ManageIQ/manageiq-api/pull/933

### DIFF
--- a/app/controllers/api/base_controller/renderer.rb
+++ b/app/controllers/api/base_controller/renderer.rb
@@ -609,9 +609,9 @@ module Api
 
       def determine_extra_cols(klass)
         virtual_attributes_for(klass) do |type, attr_name, attr_base|
-          return if attr_base.present?
-          return if virtual_attribute_accessor(type, attr_name)
-          return unless klass.attribute_supported_by_sql?(attr_name)
+          next if attr_base.present?
+          next if virtual_attribute_accessor(type, attr_name)
+          next unless klass.attribute_supported_by_sql?(attr_name)
 
           attr_name.to_sym
         end
--- a/spec/requests/providers_spec.rb
+++ b/spec/requests/providers_spec.rb
@@ -1588,26 +1588,42 @@ describe "Providers API" do
   end
 
   context 'GET /api/providers' do
-    # Make sure we don't add :include_for_find for columns that end up in
-    # :extra_cols, since that is wasted includes.
-    it "loads multiple direct virtual attributes with includes properly" do
-      ems_1 = FactoryBot.create(:ext_management_system)
-      ems_2 = FactoryBot.create(:ext_management_system)
-      vm_1  = FactoryBot.create(:vm_vmware, :name => "myvmware_1", :ems_id => ems_1.id)
-      vm_2  = FactoryBot.create(:vm_vmware, :name => "myvmware_2", :ems_id => ems_2.id)
-      vm_3  = FactoryBot.create(:vm_vmware, :name => "myvmware_3", :ems_id => ems_2.id)
+    context "with multiple direct virtual attributes" do
+      before do
+        api_basic_authorize action_identifier(:providers, :read, :collection_actions, :get)
 
-      query_match = /SELECT.*FROM\s"(?:ext_management_systems|vms)"(?!(?:.+LEFT OUTER JOIN))/m
+        ems_1 = FactoryBot.create(:ext_management_system)
+        ems_2 = FactoryBot.create(:ext_management_system)
+        FactoryBot.create(:vm_vmware, :name => "myvmware_1", :ems_id => ems_1.id)
+        FactoryBot.create(:vm_vmware, :name => "myvmware_2", :ems_id => ems_1.id)
+        FactoryBot.create(:vm_vmware, :name => "myvmware_3", :ems_id => ems_2.id)
+        FactoryBot.create(:vm_vmware, :name => "myvmware_4", :ems_id => ems_2.id)
+        FactoryBot.create(:vm_vmware, :name => "myvmware_5", :ems_id => ems_2.id)
+      end
 
-      api_basic_authorize action_identifier(:providers, :read, :collection_actions, :get)
+      # Make sure we don't add :include_for_find for columns that end up in
+      # :extra_cols, since that is wasted includes.
+      it "avoids doing a LEFT OUTER JOIN" do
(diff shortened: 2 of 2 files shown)


NickLaMuro:
@miq-bot add_label bug, jansa/yes?

This should be `jansa/yes`, as it is not only fixing a hidden bug on `master`, but also the failures currently on `jansa` described here:

https://github.com/ManageIQ/manageiq-api/pull/933#issuecomment-715313395

cc @jrafanie @Fryguy @kbrock

Fryguy:
I guess this is sort of a side-effect of what I had concerns about here - https://github.com/ManageIQ/manageiq-api/pull/887#discussion_r485845138.  That is, the name of the method in no way implies that it's part of a looping construct, and so the caller doesn't know this effect will occur.  I wonder if it's better to make the block return nil instead of using next/return, thus making it feel more like a filter block a la `select`.

NickLaMuro:
> I wonder if it's better to make the block return nil instead of using next/return, thus making it feel more like a filter block a la select.

I kind of agree, the only reason I am hesitant is that `determine_include_for_find` also uses `next`, so I would want to be consistent and change that as well, and I tend to prefer backporting only what is necessary.  I also like how he `next` style reads since you can more easily see what each of the conditions are at a glance, but agree clearly the legibility is causing bugs and obviously that is more important to avoid those.

Finishing a commit up now that fixes the tests as well (so they are a little less brittle), but let me know if you want the change from a `next` to a wrapping `if` statement.

kbrock (approved):
Great find nick.

Great tests that detect n+1 and  edge cases too.

NickLaMuro:
Regarding the test failure:

I checked back on `master` and I got the same failure that I am getting here.  Even undid the changes from https://github.com/ManageIQ/manageiq-api/pull/933 and it was still present.

I think what changed is `ManageIQ/manageiq`, so I will have to do some digging to determine what failed there and is causing the issues here.

Sorry this is taking so long to fix... 😞

NickLaMuro:
So after looking into master... I think the only PR(s) that could affect the output `spec/request/providers_spec.rb` was from here:

- https://github.com/ManageIQ/manageiq/pull/20725
- https://github.com/ManageIQ/manageiq-providers-openstack/pull/649

But I am not sure... I have commented everything from this and myself and can still not determine what is causing the issue...

NickLaMuro:
Okay, after looking at this with a fresh set up eyes, turns out the culprit for the failures is not David, but @kbrock 

Test that is failing:

https://github.com/ManageIQ/manageiq-api/blob/b39c665878f6ae04b195d30cf5caeab05bb1cdcb/spec/requests/providers_spec.rb#L199-L232

PR responsible:

https://github.com/ManageIQ/manageiq/pull/20701

I unfortunately don't understand the code intent around this change, but this is affecting my ability to merge this PR, and thus fix the `jansa` which broke because of https://github.com/ManageIQ/manageiq-api/pull/933

Thanks.

cc @Fryguy @agrare

app/controllers/api/base_controller/renderer.rb
```diff
@@ -609,9 +609,9 @@ def determine_include_for_find(klass)
 
       def determine_extra_cols(klass)
         virtual_attributes_for(klass) do |type, attr_name, attr_base|
-          return if attr_base.present?
-          return if virtual_attribute_accessor(type, attr_name)
-          return unless klass.attribute_supported_by_sql?(attr_name)
+          next if attr_base.present?
+          next if virtual_attribute_accessor(type, attr_name)
+          next unless klass.attribute_supported_by_sql?(attr_name)
```
jrafanie, line 614: wow, weird.  I don't follow what's happening here.  Is this method returning a collection of `attr_name.to_sym` that aren't skipped in the prior 3 conditionals?  Is it returning only the last `attr_name.to_sym`?  

I think this mistake happened because it's unclear it's looping through a list of attributes without an `each`.

agrare:
@NickLaMuro I think this is a bug in how Openstack is building the child managers, it looks like it is calling `ensure_managers_zone_and_provider_region` before calling `ensure_cinder_manager` so the zone/region/tenant aren't being set properly.

I'll push a PR to openstack, thanks for tracking this down!

NickLaMuro:
@agrare cool, thanks!

cc @kbrock (since I threw you under the bus for breaking this via https://github.com/ManageIQ/manageiq/pull/20701 )

NickLaMuro:
@kbrock @jrafanie we good with this now?  Currently need this PR to "un-break" `jansa`

app/controllers/api/base_controller/renderer.rb
jrafanie: This can be looked at later, it just looked odd to me so I thought it was worth mentioning

jrafanie (approved):
LGTM, my comment above can be looked at later.  For solving this bug, this PR is fine.

simaishi:
Jansa backport details:

```
$ git log -1
commit 6dd642c080b1af8bc71f792af7420af6e54d3ef3
Date:   Mon Oct 26 14:57:10 2020 -0400

    Merge pull request #940 from NickLaMuro/renderer-use-next-for-method-nested-blocks

    [Renderer] Use next, not return

    (cherry picked from commit d2990e2f94c78a7dff9d0696d66149da2f355287)
```

app/controllers/api/base_controller/renderer.rb
NickLaMuro: @jrafanie while I am not going to "blame you" for why this looks like it does, but you definitely inspired the change:

https://github.com/ManageIQ/manageiq-api/pull/887#discussion_r481410812

Half :trollface:ing

jrafanie: Yeah, we can come back to it later.  It could be the wrong abstraction.  :trollface:

PR #941, opened 2020-10-23
Fix .rubocop_local.yml

Fryguy:
@bdunne Please review


Fryguy:
Actually this seems to be a problem in all repos...going to PR blaster outer this.

PR #942, opened 2020-10-26
Expose POST/PATCH/PUT/DELETE actions on Cloud Tenants

skateman:
There was a read-only access to cloud tenants, but the UI requires the ability add/edit/delete them.

### DIFF
--- a/spec/requests/cloud_tenants_spec.rb
+++ b/spec/requests/cloud_tenants_spec.rb
@@ -47,6 +47,58 @@ RSpec.describe 'CloudTenants API' do
     end
   end
 
+  describe 'POST /api/cloud_tenants' do
+    it 'creates a cloud tenant' do
+      ems = FactoryBot.create(:ems_openstack)
+      api_basic_authorize collection_action_identifier(:cloud_tenants, :create, :post)
+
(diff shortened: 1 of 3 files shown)


spec/requests/cloud_tenants_spec.rb
```diff
@@ -47,6 +47,59 @@
     end
   end
 
+  describe 'POST /api/cloud_tenants' do
+    it 'creates a cloud tenant' do
+      ems = FactoryBot.create(:ems_openstack)
+      cloud_tenant = FactoryBot.create(:cloud_tenant, :ext_management_system => ems)
```
gtanzillo, line 53: It doesn't seem like you need to create a cloud_tenant here

gtanzillo (commented):
Looks good, just one minor comment

agrare:
@skateman should we have e.g.: `supports?(:create_cloud_tenant)` checks here?  Also are provider operations usually exposed via `PATCH/DELETE` ?  I would expect for example `DELETE /vm/1` to delete it from VMDB not from the provider.

skateman:
@agrare I don't understand :confused:

agrare:
When I think of `PATCH` or `DELETE` I think of modifying the API resource not the underlying provider resource.  Modifying the provider resource is more of an operation that I would have expected a `POST` to do

PR #943, opened 2020-10-26
Expose DDF for creating snapshots when calling OPTIONS

skateman:
Exposing the DDF for creating a snapshot when sending an `OPTIONS` request to a `/api/vms/:id/snapshots` subcollection URL.

@miq-bot assign @lpichler 

### DIFF
--- a/app/controllers/api/subcollections/snapshots.rb
+++ b/app/controllers/api/subcollections/snapshots.rb
@@ -47,6 +47,12 @@ module Api
         action_result(false, e.to_s)
       end
 
+      def snapshots_subcollection_options(parent)
+        raise BadRequestError, "No DDF specified for snapshots in #{parent}" unless parent.respond_to?(:params_for_create_snapshot)
+
+        {:snapshot_form_schema => parent.params_for_create_snapshot}
+      end
+
       private
 
       def snapshot_ident(parent)
--- a/app/controllers/api/vms_controller.rb
+++ b/app/controllers/api/vms_controller.rb
@@ -301,6 +301,17 @@ module Api
       end
     end
 
+    def options
+      return super unless @req.subcollection?
+
+      # Try to look for subcollection specific options
+      subcollection_options_method = "#{@req.subject}_subcollection_options"
+      return super unless respond_to?(subcollection_options_method)
+
+      vm = resource_search(params[:c_id], @req.collection, collection_class(@req.collection))
+      render_options(@req.collection.to_sym, send(subcollection_options_method, vm))
+    end
+
     private
 
     def miq_server_message(miq_server)
--- a/spec/requests/vms_spec.rb
+++ b/spec/requests/vms_spec.rb
@@ -2153,6 +2153,19 @@ describe "Vms API" do
     end
   end
 
+  describe "/api/vms/:id/options" do
+    it 'returns the snapshot DDF schema for the given VM' do
+      api_basic_authorize
+
+      vm = FactoryBot.create(:vm_vmware)
+
+      options(api_vm_snapshots_url(nil, vm))
+
+      expect(response.parsed_body['data']['snapshot_form_schema']).to be_an_instance_of(Hash)
+      expect(response).to have_http_status(:ok)
+    end
+  end
+
   describe "/api/vms central admin" do
     let(:resource_type) { "vm" }
 


skateman:
The CI breaks because the `options` method gets redefined in the controllers where the subcollection is loaded. Unfortunately the subcollections are calling the parents' `options` method, which is probably not the best way to implement this.

My idea would be to forward these routes to methods defined as `collection_name_options`:
```diff
diff --git a/config/routes.rb b/config/routes.rb
index ff955ac1..53eec15f 100644
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -61,7 +61,7 @@
 
         Array(collection.subcollections).each do |subcollection_name|
           # OPTIONS action for each subcollection
-          match "/:c_id/#{subcollection_name}", :controller => collection_name, :action => :options, :via => :options, :as => nil, :subcollection => true
+          match "/:c_id/#{subcollection_name}", :controller => collection_name, :action => "#{subcollection_name}_options".to_sym, :via => :options, :as => nil, :subcollection => true
 
           if subcollection_name == :settings
             match(
```

This however would require us to explicitly define these methods for subcollections without a defined method, which is for regular collections could never happen because it always falls back to [`BaseController#options`](https://github.com/ManageIQ/manageiq-api/blob/master/app/controllers/api/base_controller.rb#L109-L115). A possible solution is to implement all of these missing methods and just default them to `BaseController#options` or maybe have an implicit `alias` set up for each of these or we could juggle with `method_missing`.

Anyway, I need help @abellotti @gtanzillo @NickLaMuro @lpichler

lpichler:
yes, @skateman I agree with your suggestion
and it would be nice to define just `subcollection_name_options` method.

For example for snapshots subcollection:

```ruby
def snapshots_options
...
end
```

It it looks like that we can also update [BaseController#options](https://github.com/ManageIQ/manageiq-api/blob/master/app/controllers/api/base_controller.rb#L109-L115) to 
call `subcollection_name_options`  if exists
in subcollections mixins.

Do you have any other insights @abellotti ? Thanks!

abellotti:
OPTIONS in the API is not intended for subcollections, the route was updated to honor options /api/collection/:id/subcollection for supporting CORS preflight requests.  If you look at the response signature for either, it's the same, showing attributes, relationships as well as subcollections of the primary collection as the base information.  Unless snapshots is a primary collection that you can query OPTIONS /api/snapshots, you could expose subcollection specific data for the primary via the data portion.  In this case:

```
OPTIONS /api/vms

{
  "attributes" : ...
  "virtual_attributes" : ...
  "relationships": ...
  "subcollections": ...
  "data": {
     "subcollections_options" : {
          "snapshots" : {
          }
      },
      ...
    }
```

You can drive the data->subcollection_options entries via meta calls as you laid out above. 

If this is is just a one-off for vms/snapshots, this is fine.  However, if we see this being expanded for many more subcollections for different collections, we may want to entertain having this driven by the lib/services/api's OptionsSerializer and implemented generically for all.    For each collection's subcollection, honor the methods subcollection_\<subcollection_name\>_options via the serializer and populate a new subcollection_data section:

```
OPTIONS /api/vms

{
   "attributes": ...
   "virtual_attributes" : ...
   "relationships": ...
   "subcollections": ...
   "subcollection_data": {
       "snapshots" : {
       },
       ...
   }
   "data": ...
}
```

skateman:
@abellotti the only problem with your approach is that the `OPTIONS` request in my case would be sent against a specific VM:
```
OPTIONS /api/vms/:vm_id/snapshots
```
Which would provide me the DDF schema specific for the given VM. Unfortunately, these schemas are provider-specific...

Fryguy:
> Unfortunately, these schemas are provider-specific...

Then technically, you need OPTIONS for a provider _type_ right?  If so, then what if it was an OPTIONS request to /api/vms, with a specific type in the OPTIONS payload (or if no type given it returns all types)?

skateman:
> > Unfortunately, these schemas are provider-specific...
> 
> Then technically, you need OPTIONS for a provider _type_ right? If so, then what if it was an OPTIONS request to /api/vms, with a specific type in the OPTIONS payload (or if no type given it returns all types)?

No type set -> default options handler is called as it should be
Type is set -> default options response is enhanced with the DDF schema for the given type

Fryguy:
What I'm asking is you said you need to hit ` /api/vms/:vm_id`, but that's really hitting an _instance_ of a VM as opposed to the _type_ of the VM... In my mind asking questions of the type could be done at `/api/vms`.

skateman:
I guess I can work with the `OPTIONS /api/vms?type=xy` here and pass the schema to that response if @abellotti agrees...

skateman:
> I guess I can work with the `OPTIONS /api/vms?type=xy` here and pass the schema to that response if @abellotti agrees...

No, I can't do it like this, I need `OPTIONS /api/vms/:id` because I need the powerstate for the memory snapshot...

gtanzillo:
@skateman @abellotti @Fryguy @lpichler Reading back, it seems like we haven't gotten a consensus on how we should proceed with this. 

Are you guys ok with just merging this as is?

skateman:
Can we get this in? The CI is green now!

app/controllers/api/subcollections/snapshots.rb
```diff
@@ -47,6 +47,12 @@ def snapshots_revert_resource(parent, type, id, _data)
         action_result(false, e.to_s)
       end
 
+      def snapshots_options(parent)
```
abellotti, line 50: would prefer the following naming:

```
snapshot_subcollection_options(parent)
```

abellotti:
Minor method name change above, and we'd also need specs for OPTIONS /api/vms/:id/snapshots. 
Thanks @skateman for the API Enhancement.

skateman:
@abellotti added the spec, but it will fail until the dependent provider PRs are merged. I could possibly create a mock using `allow_any_instance_of` but it doesn't seem clean to me, so I'd rather wait for the dependencies to be merged...

skateman:
Dependencies are in, CI is green so it's ready from my side.

abellotti (approved):

abellotti:
LGTM!! Thanks @skateman for the API Enhancement 👍

agrare (approved):

Issue #717, opened 2019-12-04
[API] Convert to OpenAPI


Fryguy:

abellotti:
With the OpenAPI spec versioning to match ManageIQ's, bringing in the API & MiQ versioning that were on separate version paths to now being one in the same.

PR #945, opened 2020-10-26
Automatically generate an OpenApi specification

agrare:
Passes validation https://editor-next.swagger.io/

```
$ bundle exec rake app:manageiq:api:openapi_generate
```

TODO
- [x] Load models to build schema
- [x] Add descriptions if a column has a comment
- [ ] Detect and build parameters
- [ ] Handle multiple post actions

https://github.com/ManageIQ/manageiq-api/issues/717

### DIFF
--- a/lib/manageiq/api/open_api/generator.rb
+++ b/lib/manageiq/api/open_api/generator.rb
@@ -34,7 +34,16 @@ module ManageIQ
         end
 
         def build_schemas
-          ::Api::ApiConfig.collections.each_with_object({}) do |(_collection_name, collection), schemas|
+          schemas = {
+            "ID" => {
+              "type"        => "string",
+              "description" => "ID of the resource",
+              "pattern"     => "^\\d+$",
+              "readOnly"    => true,
+            }
+          }
+
+          ::Api::ApiConfig.collections.each do |_collection_name, collection|
             next unless collection.klass
 
             model = collection.klass.constantize
@@ -45,6 +54,8 @@ module ManageIQ
               "additionalProperties" => false
             }
           end
+
+          schemas
         end
 
         def build_schema_properties(model)
@@ -78,17 +89,6 @@ module ManageIQ
 
             properties_value
           end
-        end
-
-        def schemas
-          @schemas ||= {
-            "ID" => {
-              "type"        => "string",
-              "description" => "ID of the resource",
-              "pattern"     => "^\\d+$",
-              "readOnly"    => true,
-            }
-          }
         end
 
         def skeletal_openapi_spec
--- a/lib/manageiq/api/open_api/generator.rb
+++ b/lib/manageiq/api/open_api/generator.rb
@@ -46,9 +46,10 @@ module ManageIQ
           ::Api::ApiConfig.collections.each do |_collection_name, collection|
             next unless collection.klass
 
-            model = collection.klass.constantize
+            model       = collection.klass.constantize
+            schema_name = model.name.gsub("::", "_")
 
-            schemas[model.name] = {
+            schemas[schema_name] = {
               "type"                 => "object",
               "properties"           => build_schema_properties(model),
               "additionalProperties" => false
--- a/lib/manageiq/api/open_api/generator.rb
+++ b/lib/manageiq/api/open_api/generator.rb
@@ -94,10 +94,12 @@ module ManageIQ
         def skeletal_openapi_spec
           {
             "openapi"    => OPENAPI_VERSION,
-            "info"       => {},
-            "security"   => [],
+            "info"       => {
+              "version"     => api_version,
+              "title"       => ::Api::ApiConfig.base.name,
+              "description" => ::Api::ApiConfig.base.description
+            },
             "paths"      => {},
-            "servers"    => [],
             "components" => {
               "parameters" => {},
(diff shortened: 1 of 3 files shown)


agrare:
This mostly just populates the schema from the models, not building out the paths yet

lib/manageiq/api/open_api/generator.rb
```diff
@@ -0,0 +1,139 @@
...
+
+        def manageiq_api_engine
+          @manageiq_api_engine ||= Vmdb::Plugins.all.detect { |e| e.name == "ManageIQ::Api::Engine" }
+        end
+
+        def skeletal_openapi_spec
+          {
+            "openapi"    => openapi_version,
+            "info"       => {},
+            "secuirty"   => [],
```
Fryguy, line 127:
```suggestion
            "security"   => [],
```

lib/manageiq/api/open_api/generator.rb
```diff
@@ -0,0 +1,112 @@
...
+          }
+        end
+
+        def skeletal_openapi_spec
+          {
+            "openapi"    => OPENAPI_VERSION,
+            "info"       => {
+              "version"     => api_version,
+              "title"       => "API",
+              "description" => "REST API"
```
agrare, line 100: This matches what `GET /api` returns:
```
{
  "name": "API",
  "description": "REST API",
  "version": "4.4.0-pre",
```

lib/manageiq/api/open_api/generator.rb
```diff
@@ -0,0 +1,113 @@
...
+              "pattern"     => "^\\d+$",
+              "readOnly"    => true,
+            }
+          }
+
+          ::Api::ApiConfig.collections.each do |_collection_name, collection|
+            next unless collection.klass
+
+            model       = collection.klass.constantize
+            schema_name = model.name.gsub("::", "_")
```
agrare, line 50: image

lib/manageiq/api/open_api/generator.rb
```diff
@@ -0,0 +1,113 @@
...
+
+            schemas[schema_name] = {
+              "type"                 => "object",
+              "properties"           => build_schema_properties(model),
+              "additionalProperties" => false
+            }
+          end
```
Fryguy, line 57: After we build the schemas, can we sort by schema_name?  Right now they seem to come in a strange order.

Fryguy: Or perhaps "ID" followed by our models sorted by name.

agrare: :+1: yeah definitely, right now they are coming in in the order of the API yaml definition but alphabetically sorted is better

lib/manageiq/api/open_api/generator.rb
```diff
@@ -0,0 +1,113 @@
...
+        end
+
+        def build_schema_properties(model)
+          model.columns_hash.each_with_object({}) do |(key, value), properties|
+            properties[key] = build_schema_properties_value(model, key, value)
+          end
+        end
+
+        def build_schema_properties_value(model, key, value)
+          if key == model.primary_key || key.ends_with?("_id")
```
Fryguy, line 69: We know there are a few fields where they are named `_id` but aren't actually ids.  Perhaps instead of this check here we check `value.sql_type_metadata.type` down below, and in the integer case we do this `key == model.primary_key || key.ends_with?("_id")` over there.

This way we are effective checking `("id" || "_id") && is_integer?`

agrare: Good catch
```
           "provider_segmentation_id": {
-            "$ref": "#/components/schemas/ID"
+            "type": "string"
           },
```

lib/manageiq/api/open_api/generator.rb
```diff
@@ -0,0 +1,123 @@
...
+          else
+            properties_value["type"] = "string"
+          end
+
+          # If there is a comment present on the column and the property is
+          # a $ref we have to use allOf since all other properties of a $ref
+          # are ignored
+          if value.comment.present?
+            if properties_value.key?("$ref")
+              properties_value = {"allOf" => [properties_value, {"description" => value.comment}]}
```
agrare, line 96: If we have a description on a foreign_key we have to use `allOf: []` because any other properties with a `$ref` are ignored

lib/manageiq/api/open_api/generator.rb
```diff
@@ -0,0 +1,123 @@
...
+        def build_schema_properties_value(model, key, value)
+          properties_value = {}
+
+          case value.sql_type_metadata.type
+          when :datetime
+            properties_value["type"]   = "string"
+            properties_value["format"] = "date-time"
+          when :integer
+            if key == model.primary_key || key.ends_with?("_id")
+              properties_value = {"$ref" => "##{SCHEMAS_PATH}/ID"}
```
Fryguy, line 77: Super minor but to be consistent with the other branches, this could be
```suggestion
              properties_value["$ref"] = "##{SCHEMAS_PATH}/ID"
```

agrare: :+1: done

agrare:
@Fryguy believe I addressed all existing comments, PTAL

PR #947, opened 2020-10-27
Handle DDF type API_ATTRIBUTES in AuthenticationsSpec

agrare:
If an Ansible credential has DDF type attributes then we don't get a val in the loop.  Still trying to work out if this is the right fix or a band aid, this doesn't seem to be an issue for all credentials just some

https://github.com/ManageIQ/manageiq/pull/20568

### DIFF
--- a/spec/requests/authentications_spec.rb
+++ b/spec/requests/authentications_spec.rb
@@ -507,7 +507,7 @@ RSpec.describe 'Authentications API' do
         next unless defined? subklass::API_OPTIONS
         subklass::API_OPTIONS.tap do |options|
           options[:attributes].each do |_k, val|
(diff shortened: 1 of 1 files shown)


agrare:
Okay so all of the `AnsibleTower` credentials still use the old style: https://github.com/ManageIQ/manageiq-providers-ansible_tower/blob/master/app/models/manageiq/providers/ansible_tower/automation_manager/amazon_credential.rb#L3-L13
The EmbeddedAnsible credentials use the DDF style: https://github.com/ManageIQ/manageiq/blob/master/app/models/manageiq/providers/embedded_ansible/automation_manager/amazon_credential.rb#L3-L21

chessbyte (approved):

PR #949, opened 2020-10-27
[KASPAROV] Test with ruby 2.6.6 on Kasparov

jrafanie:
Test with ruby 2.6.6 on Kasparov


jrafanie:
Failure with blaster outer... starting over.

Issue #912, opened 2020-10-06
Missing endpoint for editing Network Routers


skateman:
The current verb is `*gpd` and we would probably need `*gpppd`

@miq-bot add_label help wanted
@miq-bot assign @gtanzillo

PR #960, opened 2020-11-04
Add endpoint for editing Network Routers

lfu:
Add endpoint for editing Network Routers

https://github.com/ManageIQ/manageiq-api/issues/912

### DIFF
--- a/app/controllers/api/network_routers_controller.rb
+++ b/app/controllers/api/network_routers_controller.rb
@@ -2,6 +2,16 @@ module Api
   class NetworkRoutersController < BaseController
     include Subcollections::Tags
 
+    def edit_resource(type, id, data)
+      network_router = resource_search(id, type, collection_class(:network_routers))
+      raise BadRequestError, "Update for #{network_router_ident(network_router)}: #{network_router.unsupported_reason(:update)}" unless network_router.supports_update?
+
+      task_id = network_router.update_network_router_queue(User.current_user.id, data.deep_symbolize_keys)
+      action_result(true, "Updating #{network_router_ident(network_router)}", :task_id => task_id)
+    rescue => err
+      action_result(false, err.to_s)
+    end
+
     def delete_resource(type, id, _data = {})
       delete_action_handler do
         network_router = resource_search(id, type, collection_class(:network_routers))
--- a/spec/requests/network_routers_spec.rb
+++ b/spec/requests/network_routers_spec.rb
@@ -58,6 +58,34 @@ RSpec.describe 'NetworkRouters API' do
     end
   end
 
+  describe "POST /api/network_routers/:id" do
+    let(:ems) { FactoryBot.create(:ems_openstack) }
+    let(:tenant) { FactoryBot.create(:cloud_tenant_openstack, :ext_management_system => ems) }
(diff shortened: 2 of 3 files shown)


gtanzillo:
@lfu Please add a comment here with an example of the request and response for editing a network router, thanks!

lfu:
POST /api/network_routers/10000000000009
```
{
    "action": "edit",
    "status": "test"
}
```
RESPONSE
```
{
    "href": "http://localhost:3000/api/network_routers/10000000000009",
    "id": "10000000000009",
    "status": "test",
    "type": "ManageIQ::Providers::Openstack::NetworkManager::NetworkRouter",
    "name": "redhat-ext-router",
    "ems_ref": "211740f8-df06-48ce-b320-726c980d94db",
    "ems_id": "10000000000013",
    "cloud_network_id": "10000000000017",
    "admin_state_up": "t",
    "cloud_tenant_id": "10000000000004",
    "extra_attributes": {
        "distributed": null,
        "routes": [],
        "high_availability": false,
        "external_gateway_info": {
            "network_id": "67763b08-aec6-4ebe-8dee-872afbd86d7f",
            "enable_snat": true,
            "external_fixed_ips": [
                {
                    "subnet_id": "02176573-719a-4d65-8710-d6a3e0d1c875",
                    "ip_address": "10.9.60.148"
                }
            ]
        }
    },
    "network_group_id": null
}
```

skateman (approved):
The Seal of Approval

spec/requests/network_routers_spec.rb
```diff
@@ -58,6 +58,32 @@
...
+    it "can queue the updating of a network router" do
+      api_basic_authorize(action_identifier(:network_routers, :edit))
+
+      post(api_network_router_url(nil, network_router), :params => {:action => 'edit', :status => "inactive"})
+
+      expected = {
+        'success' => true,
+        'message' => a_string_including('Updating Network Router')
+      }
+      expect(response.parsed_body).to include(expected)
```
abellotti, line 75: should check that we got a task id and href returned.

app/controllers/api/network_routers_controller.rb
```diff
@@ -2,6 +2,16 @@ module Api
   class NetworkRoutersController < BaseController
     include Subcollections::Tags
 
+    def edit_resource(type, id, data)
+      network_router = resource_search(id, type, collection_class(:network_routers))
+      raise "Update not supported for #{network_router_ident(network_router)}" unless network_router.respond_to?(:update_network_router_queue)
```
Fryguy, line 7: Do we need a supports_feature / validate_ check ?  @agrare ?

agrare: :+1: yes `ManageIQ::Providers::Openstack::NetworkManager::NetworkRouter` has `supports :update` so we can replace this `respond_to?` with `supports_update?` and use the standard `unsupported_reason(:update)`

agrare: This one is a little strange in that the `update_network_router_queue` doesn't live in the base `NetworkRouter` model which is why the `respond_to?` would have "worked" but supports feature is still the right way to do this

agrare (approved):

gtanzillo (approved):

PR #961, opened 2020-11-05
Introduce error code for errors

### LINKED ISSUE #919: Raising an exception in a POST request returns with a 200 response
The following request:

```
curl -XPOST -d '{"name": "e", "type": "ManageIQ::Providers::EmbeddedAnsible::AutomationManager::VaultCredential", "vault_password": "x"}' "http://admin:smartvm@localhost:3000/api/authentications" | jq
```

Returns a `200` response:

```
{
  "results": [
    {
      "success": false,
      "message": "must supply a manager resource"
    }
  ]
}
```

Is this correct? Because the UI codebase expects HTTP errors if the request failed :confused:  

@miq-bot add_label bug

kbrock:
Fixes #919 

We are currently returning an ok (200) http status from all api requests, even the ones that fail.
This makes routing difficult in the ui.

It is not so cut and dry because some requests take multiple records and end up with multiple statuses, some failing and some succeeding.

### Before 

We returned an ok for all create and update requests, even when they failed.

### After

Now if there is one resource, including a collection of one resource, and it fails creating or updating, it will return a failed status code (4xx)

The status of `406` (not processable) is the best status available to us.

### DIFF
--- a/app/controllers/api/base_controller/renderer.rb
+++ b/app/controllers/api/base_controller/renderer.rb
@@ -14,7 +14,7 @@ module Api
       # Helper proc to render a single resource
       #
       def render_resource(type, resource, opts = {})
-        render :json => resource_to_jbuilder(type, gen_reftype(type, opts), resource, opts).target!
+        render :json => resource_to_jbuilder(type, gen_reftype(type, opts), resource, opts).target!, :status => status_from_resource(resource)
       end
 
       #
@@ -614,6 +614,32 @@ module Api
           next unless klass.attribute_supported_by_sql?(attr_name)
 
           attr_name.to_sym
+        end
+      end
+
+      # given a response to render, determine the proper return code
+      # index pages have a response array of objects or hash that represents an object
+      # show pages have resource as an object, or a hash representing an object
+      # all of these pages just want to return ok
+      #
+      # the create and update pages are the ones we want to give a status
+      # if there are multiple response entries, we just return ok
+      # but if there is a single response, we'll use status to determine the resposne code
+      #
+      # So there are many ways we want to render and only a fraction of them will
+      # use success to determine the return status - that is why so much short circuit logic
+      #
+      # @param [Object,Array[Object],Hash,Hash{String=>Array[Hash]}] resource
+      # @return [Symbol] http status code
+      def status_from_resource(resource)
+        return :ok if @req.bulk? || !resource.kind_of?(Hash)
+        return resource[:success] ? :ok : :bad_request if resource.key?(:success)
+
+        results = resource["results"]
+        if !results.kind_of?(Array) || !results.first.kind_of?(Hash) || !results.first.key?(:success) || results.first[:success]
+          :ok
+        else
+          :bad_request
         end
       end
     end
--- a/lib/api/request_adapter.rb
+++ b/lib/api/request_adapter.rb
@@ -99,6 +99,13 @@ module Api
       json_body.fetch("resource", json_body.except("action"))
     end
 
+    # A bulk request is of the form:
+    # {:action => :create, resources => [{},{}]}
+    # @returns true if this is a bulk request
+    def bulk?
+      json_body.kind_of?(Hash) && json_body.key?("resources")
+    end
+
     private
 
     def expand_requested
--- a/spec/requests/authentications_spec.rb
+++ b/spec/requests/authentications_spec.rb
@@ -213,7 +213,7 @@ RSpec.describe 'Authentications API' do
           { 'success' => false, 'message' => 'must supply a manager resource' }
         ]
       }
-      expect(response).to have_http_status(:ok)
+      expect(response).to have_http_status(:bad_request)
       expect(response.parsed_body).to include(expected)
     end
 
@@ -227,7 +227,7 @@ RSpec.describe 'Authentications API' do
           { 'success' => false, 'message' => 'type not currently supported' }
         ]
       }
-      expect(response).to have_http_status(:ok)
+      expect(response).to have_http_status(:bad_request)
       expect(response.parsed_body).to include(expected)
     end
 
@@ -247,7 +247,7 @@ RSpec.describe 'Authentications API' do
       expect(response.parsed_body).to include(expected)
     end
 
-    it 'can create authentications in bulk' do
+    it 'creates and fails in a single request' do
       api_basic_authorize collection_action_identifier(:authentications, :create, :post)
 
       expected = {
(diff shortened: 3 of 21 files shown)


app/controllers/api/base_controller/renderer.rb
```diff
@@ -616,6 +616,18 @@ def determine_extra_cols(klass)
...
+      def status_from_resource(resource)
+        return :ok unless resource.kind_of?(Hash)
+
+        if resource.key?(:success)
+          return resource[:success] ? :ok : :not_acceptable
+        end
+        results = resource["results"]
+        return :ok unless results.kind_of?(Array) && results.size == 1
+        return :ok unless results.first.kind_of?(Hash)
+        !results.first.key?(:success) || results.first[:success] ? :ok : :not_acceptables
```
chessbyte, line 629: `: not_acceptables` is the plural a typo?

kbrock: totally - thanks

Fryguy:
I don't understand using 406 for singles that fail...feels weird.   406 is like when the client does something like `Accept: text/xml` content and the server doesn't provide xml content, so it replies not acceptable.

Fryguy:
The only 400-level codes that makes sense to me, in this case, are 

- 422 - Unprocessable Entity https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/422
- 400 - Bad Request

422 was designed for WebDAV, which is why I'm leaning towards just a normal 400 error.

Fryguy (changes requested):
Code wise this looks good to me, though I'd like another set of eyes.  My request changes is mainly around the choice of error code, as I think 406 is incorrect.

app/controllers/api/base_controller/renderer.rb
```diff
@@ -616,6 +616,18 @@ def determine_extra_cols(klass)
           attr_name.to_sym
         end
       end
+
+      def status_from_resource(resource)
+        return :ok unless resource.kind_of?(Hash)
```
Fryguy, line 621: Can you comment on this line...I understand it only because I understand the issue, but a casual reader won't understand why a non-hash is atuomatically :ok

kbrock: we sure have a lot of ways we can render objects. I hope the documentation will serve well for people (including me) in the future

abellotti:
Maybe a quick sanity check with https://github.com/ManageIQ/manageiq-api-client's usage just to make sure we're not breaking central admin.

kbrock:
> Maybe a quick sanity check with https://github.com/ManageIQ/manageiq-api-client's usage just to make sure we're not breaking central admin.

I do see a response code of 400 triggering a special response in [`connection.rb`](https://github.com/ManageIQ/manageiq-api-client/blob/master/lib/manageiq/api/client/connection.rb#L109) 

I'm having trouble knowing if we do or do not want that exception thrown. Since we weren't before, then I'm assuming we would not like that now. Then again, before it is possible that we had to add special logic to handle his and that will be fixed.

do you know of any other cases in our code that may also react negatively to this code?

TL;DR - nice find. I'm stumped where to go from here

Fryguy (approved):
Code-wise seems good to me.  @abellotti Please also review.  As @abellotti we should just make sure that the manageiq-api-client responds accordingly.  If it's a bad request, I think I'd expect the client to raise an Exception.  Since we use the client in core in order to do the central admin stuff (and also manageiq-automation_engine I believe), I'm assuming there will be tests for it, so perhaps a cross-repo test is enough?

lib/api/request_adapter.rb
```diff
@@ -99,6 +99,15 @@ def resource
       json_body.fetch("resource", json_body.except("action"))
     end
 
+    # A bulk request is of the form:
+    # {:action => :create, resources => [{},{}]}
+    # OR
+    # [{:action => :create, ...}, {:action => :delete, ...}]
+    # @returns true if this is a bulk request
+    def bulk?
+      !json_body.kind_of?(Hash) || json_body.key?("resources")
```
abellotti, line 108: may be easier to understand `json_body.kind_of?(Hash) && json_body.key?("resources")`

spec/requests/authentications_spec.rb
```diff
@@ -258,13 +258,12 @@
...
-            'success' => true,
-            'message' => 'Creating Authentication',
-            'task_id' => a_kind_of(String)
+            'success' => false,
+            'message' => 'must supply a manager resource'
           )
         ]
       }
-      post(api_authentications_url, :params => { :resources => [create_params, create_params] })
+      post(api_authentications_url, :params => {:resources => [create_params, create_params.except(:manager_resource)]})
```
abellotti, line 266: why is this needed in this PR ?

lib/api/request_adapter.rb
Fryguy: @abellotti That's not the same thing...

`!A || B` -> `A && !B`

abellotti: But it is what we want though, a Hash that has a key resources.

Fryguy: I was assuming the `!json_body.kind_of?(Hash)` was referring to the Array notation as listed in the comment.  That is I interpreted it as:

```
!json_body.kind_of?(Hash) || json_body.key?("resources")

Array || Hash-with-resources
```

kbrock: |json_body|bulk?|
|---|---|
|hash with :resources|true|
|hash without :resources|false|
|array of hashes|true|
|nil| don't care|
|some other class|don't care


I will make the successes more explicit rather than the vague `not`

abellotti: I may be missing something but I don't recall the second signature as bulk. @kbrock which API endpoint implements this ?

I know of the PATCH /api/:collection/:id with the similar signature as per https://manageiq.org/docs/reference/latest/api.overviiew/http.html example of PATCH /api/services/42, Array of attribute update specs, so really a single resource update, not bulk.

spec/requests/authentications_spec.rb
kbrock: deleted it... remembered why it is there.

I wanted to ensure that if we have a successful and a failed creation of an authentication (it requires a manager_resource to create the record) that it would still return a 200.

kbrock:
@Fryguy 

The api client simply raises for any response code >= 400 (with a special case for 404). So this will be handled.

From what I can tell, the api client tests do not use the api to test so a cross repo would probably not buy us anything.

kbrock:
Kicking

kbrock:
pushed a rebase (no changes)

I was reproducing locally and after this rebase, it fixed it locally.

gtanzillo (approved):
This looks good to me. I'd like to get a final 👍  from @abellotti before merging if it can wait another day

Fryguy:
@gtanzillo I think this thread is still unresolved - https://github.com/ManageIQ/manageiq-api/pull/961#discussion_r529727436

kbrock:
update:

i had thought that passing an array of a hash in the specs meant that it supported an array of 2 hashes. this was wrong. The `Hash(param)` where `param=[{}]` voided it out.
I had thought that meant that just passing a hash always worked, but this turned out to be wrong as well. While all but the chargeback example worked, chargeback was throwing odd errors - and not changing this behavior seemed best. especially since it made no sense to me.

abellotti (approved):

Fryguy:
@miq-bot cross-repo-test manageiq-ui-classic,manageiq

Fryguy (approved):
LGTM - just waiting on cross repo test.

Issue #901, opened 2020-10-06
Missing endpoint for creating/editing/deleting floating IPs


skateman:
The current verb is `*gp` and we would need `*gpppd` in the UI to move further with https://github.com/ManageIQ/manageiq-ui-classic/issues/6875

skateman:
@miq-bot assign @gtanzillo 
@miq-bot add_label help wanted

PR #962, opened 2020-11-09
Add endpoint for creating/editing/deleting floating IPs

lfu:
```
POST /api/floating_ips/10000000000054
{
    "action": "edit",
    "status": "test"
}
```
RESPONSE
```
{
    "success": true,
    "message": "Updating Floating Ip id:10000000000054 name: '10.9.60.154'",
    "task_id": "10000000053300",
    "task_href": "http://localhost:3000/api/tasks/10000000053300"
}
```

https://github.com/ManageIQ/manageiq-api/issues/901

### DIFF
--- a/app/controllers/api/floating_ips_controller.rb
+++ b/app/controllers/api/floating_ips_controller.rb
@@ -1,4 +1,40 @@
 module Api
   class FloatingIpsController < BaseController
+    def create_resource(_type, _id = nil, data = {})
+      ems = ExtManagementSystem.find(data['ems_id'])
+      klass = FloatingIp.class_by_ems(ems)
+      raise BadRequestError, "Create floating ip for Provider #{ems.name}: #{klass.unsupported_reason(:create)}" unless klass.supports?(:create)
+
+      task_id = ems.create_floating_ip_queue(session[:userid], data.deep_symbolize_keys)
+      action_result(true, "Creating Floating Ip #{data['name']} for Provider: #{ems.name}", :task_id => task_id)
+    rescue => err
+      action_result(false, err.to_s)
+    end
+
+    def edit_resource(type, id, data)
+      floating_ip = resource_search(id, type, collection_class(:floating_ips))
+      raise BadRequestError, "Update for #{floating_ip_ident(floating_ip)}: #{floating_ip.unsupported_reason(:update)}" unless floating_ip.supports?(:update)
+
+      task_id = floating_ip.update_floating_ip_queue(session[:userid], data.deep_symbolize_keys)
+      action_result(true, "Updating #{floating_ip_ident(floating_ip)}", :task_id => task_id)
+    rescue => err
+      action_result(false, err.to_s)
+    end
+
+    def delete_resource(type, id, _data = {})
+      delete_action_handler do
+        floating_ip = resource_search(id, type, collection_class(:floating_ips))
+        raise BadRequestError, "Delete for #{floating_ip_ident(floating_ip)}: #{floating_ip.unsupported_reason(:delete)}" unless floating_ip.supports?(:delete)
+
+        task_id = floating_ip.delete_floating_ip_queue(session[:userid])
+        action_result(true, "Deleting #{floating_ip_ident(floating_ip)}", :task_id => task_id)
+      end
+    end
+
+    private
+
+    def floating_ip_ident(floating_ip)
+      "Floating Ip id:#{floating_ip.id} name: '#{floating_ip.name}'"
+    end
   end
 end
--- a/spec/requests/floating_ips_spec.rb
+++ b/spec/requests/floating_ips_spec.rb
@@ -53,5 +53,148 @@ RSpec.describe 'FloatingIp API' do
       post(api_floating_ips_url, :params => gen_request(:query, ""))
       expect(response).to have_http_status(:forbidden)
     end
+
+    it "queues the creating of floating ip" do
+      api_basic_authorize collection_action_identifier(:floating_ips, :create)
+      provider = FactoryBot.create(:ems_openstack, :name => 'test_provider')
+      request = {
+        "action"   => "create",
+        "resource" => {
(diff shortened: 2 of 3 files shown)


app/controllers/api/floating_ips_controller.rb
```diff
@@ -1,4 +1,39 @@
 module Api
   class FloatingIpsController < BaseController
+    def create_resource(_type, _id = nil, data = {})
+      ems = ExtManagementSystem.find(data['ems_id'])
+      raise BadRequestError, "Create floating ip not supported for Provider: #{ems.name}" unless ems.respond_to?(:create_floating_ip_queue)
```
agrare, line 5: We should use `SupportsFeatureMixin` for this, e.g. `ems.supports?(:create_floating_ip)

app/controllers/api/floating_ips_controller.rb
```diff
@@ -1,4 +1,39 @@
...
+
+      task_id = ems.create_floating_ip_queue(session[:userid], data.deep_symbolize_keys)
+      action_result(true, "Creating Floating Ip #{data['name']} for Provider: #{ems.name}", :task_id => task_id)
+    rescue => err
+      action_result(false, err.to_s)
+    end
+
+    def edit_resource(type, id, data)
+      floating_ip = resource_search(id, type, collection_class(:floating_ips))
+      raise BadRequestError, "Update not supported for #{floating_ip_ident(floating_ip)}" unless floating_ip.respond_to?(:update_floating_ip_queue)
```
agrare, line 15: `FloatingIp` already has `floating_ip.supports?(:update)`

app/controllers/api/floating_ips_controller.rb
```diff
@@ -1,4 +1,39 @@
...
+      task_id = floating_ip.update_floating_ip_queue(session[:userid], data.deep_symbolize_keys)
+      action_result(true, "Updating #{floating_ip_ident(floating_ip)}", :task_id => task_id)
+    rescue => err
+      action_result(false, err.to_s)
+    end
+
+    def delete_resource(type, id, _data = {})
+      delete_action_handler do
+        floating_ip = resource_search(id, type, collection_class(:floating_ips))
+        raise BadRequestError, "Delete not supported for #{floating_ip_ident(floating_ip)}" unless floating_ip.respond_to?(:delete_floating_ip_queue)
```
agrare, line 26: Same, `floating_ip.supports?(:delete)` should work

app/controllers/api/floating_ips_controller.rb
agrare: You can also raise `.unsupported_reason(:feature)` for each of these

app/controllers/api/floating_ips_controller.rb
```diff
@@ -1,4 +1,39 @@
...
+      ems = ExtManagementSystem.find(data['ems_id'])
+      raise BadRequestError, "Create floating ip not supported for Provider: #{ems.name}" unless ems.respond_to?(:create_floating_ip_queue)
+
+      task_id = ems.create_floating_ip_queue(session[:userid], data.deep_symbolize_keys)
+      action_result(true, "Creating Floating Ip #{data['name']} for Provider: #{ems.name}", :task_id => task_id)
+    rescue => err
+      action_result(false, err.to_s)
+    end
+
+    def edit_resource(type, id, data)
```
agrare, line 13: Re: rubocop warnings...I feel like `id` should be an allowed variable name

app/controllers/api/floating_ips_controller.rb
```diff
@@ -1,4 +1,40 @@
 module Api
   class FloatingIpsController < BaseController
+    def create_resource(_type, _id = nil, data = {})
```
Fryguy, line 3: Does this come from a PUT or from a POST with "create" action (or both)?  What does the incoming payload look like and does it match the shape of the GET request?

Fryguy (approved):
LGTM - @agrare Can you review the changes (specifically the supports feature stuff?)

app/controllers/api/floating_ips_controller.rb
```diff
@@ -1,4 +1,40 @@
...
+
+      task_id = ems.create_floating_ip_queue(session[:userid], data.deep_symbolize_keys)
+      action_result(true, "Creating Floating Ip #{data['name']} for Provider: #{ems.name}", :task_id => task_id)
+    rescue => err
+      action_result(false, err.to_s)
+    end
+
+    def edit_resource(type, id, data)
+      floating_ip = resource_search(id, type, collection_class(:floating_ips))
+      raise BadRequestError, "Update for #{floating_ip_ident(floating_ip)}: #{floating_ip.unsupported_reason(:update)}" unless floating_ip.supports?(:update)
```
agrare, line 16: I just noticed that the supports feature is `:update` but the unsupported reason is `:update_floating_ip` [[ref]](https://github.com/ManageIQ/manageiq-providers-openstack/blob/master/app/models/manageiq/providers/openstack/network_manager/floating_ip.rb#L18) so you likely won't see the proper error here but that looks like a bug on the openstack side not here

agrare (approved):
:+1: LGTM

spec/requests/floating_ips_spec.rb
```diff
@@ -53,5 +53,148 @@
...
+    it "forbids multiple floating ip deletion without an appropriate role" do
+      floating_ip1, floating_ip2 = FactoryBot.create_list(:floating_ip, 2)
+      api_basic_authorize
+
+      post(api_floating_ips_url, :params => {:action => "delete", :resources => [{:id => floating_ip1.id}, {:id => floating_ip2.id}]})
+
+      expect(response).to have_http_status(:forbidden)
+    end
+
+    it 'raises an error when delete not supported for floating ip' do
```
abellotti, line 186: minor, quoting not consistent with other tests, ' vs "

gtanzillo (approved):

PR #965, opened 2020-11-17
[WIP] [api_config.yml] Fixes :indentifier: updates

NickLaMuro:
Should hopefully fix failing tests on `master`:  https://travis-ci.com/github/ManageIQ/manageiq-api/jobs/439594527#L1991-L1994

Background
----------

Changes were made in Core:

https://github.com/ManageIQ/manageiq/pull/20438

That updated the product feature `:identifier:` key/values to support the Control Explorer split.

This caused the `spec/lib/api/api_config_spec.rb` to fail when it checked the difference between the `api_config.yml` and the configured `MiqProductFeature` records to confirm they are in sync.

This trues up the config with the new changes from the aforementioned PR above to allow the specs to pass.

TODO
----

Determine the answer to this: https://github.com/ManageIQ/manageiq/pull/20438/files#r525547913

Links
-----

- https://travis-ci.com/github/ManageIQ/manageiq-api/jobs/439594527#L1991-L1994

PR #966, opened 2020-11-18
Fixed endpoint feature mapping for policy-related endpoints

skateman:
The policy de-explorerization had some [RBAC feature changes](https://github.com/ManageIQ/manageiq/pull/20438) and renaming which causes the CI fail on this repo as the feature mapping has not been updated. I went through these missing features and replaced them with the existing ones.

I'm not entirely sure about the `policy_profile_assign` as it seems like it maps to two new features `miq_policy_conditions_assignment` and `miq_policy_events_assignment`.

Another issue I found was that the `/alerts` endpoint is actually querying `MiqAlertStatus` entities. Here only the top-level feature identifier was wrongly set, so I adjusted it.

@lpichler @NickLaMuro @gtanzillo @abellotti please verify if I'm doing stuff the right way
@h-kataria can you please check if these features map properly and if the migrations were ran against the deleted ones to rename them to the new names

@miq-bot add_label bug, rbac, kasparov/no

### DIFF
--- a/config/api.yml
+++ b/config/api.yml
@@ -38,7 +38,7 @@
     :klass: Account
   :actions:
     :description: Actions
-    :identifier: action
+    :identifier: miq_action
     :options:
     - :collection
     :verbs: *gpppd
@@ -46,37 +46,37 @@
     :collection_actions:
       :get:
       - :name: read
-        :identifier: action_show_list
+        :identifier: miq_action_show_list
       :post:
       - :name: query
-        :identifier: action_show_list
+        :identifier: miq_action_show_list
       - :name: create
-        :identifier: action_new
+        :identifier: miq_action_new
       - :name: edit
-        :identifier: action_edit
+        :identifier: miq_action_edit
       - :name: delete
-        :identifier: action_delete
+        :identifier: miq_action_delete
       :delete:
       - :name: delete
-        :identifier: action_delete
+        :identifier: miq_action_delete
     :resource_actions:
       :get:
       - :name: read
-        :identifier: action_show
+        :identifier: miq_action_show
       :post:
       - :name: edit
-        :identifier: action_edit
+        :identifier: miq_action_edit
       - :name: delete
-        :identifier: action_delete
+        :identifier: miq_action_delete
       :patch:
       - :name: edit
-        :identifier: action_edit
+        :identifier: miq_action_edit
       :put:
       - :name: edit
-        :identifier: action_edit
+        :identifier: miq_action_edit
       :delete:
       - :name: delete
-        :identifier: action_delete
+        :identifier: miq_action_delete
   :alert_actions:
     :description: Alert Actions
     :identifier: alert_action
@@ -176,8 +176,8 @@
       - :name: delete
         :identifier: alert_definition_delete
   :alerts:
-    :description: Alerts
-    :identifier: alert
+    :description: Alert Statuses
+    :identifier: alert_status
     :options:
     - :collection
     :verbs: *g
@@ -1441,7 +1441,7 @@
         :identifier: event_streams_show
   :events:
     :description: Events
-    :identifier: event
+    :identifier: miq_event
     :options:
     - :collection
     - :subcollection
@@ -1450,14 +1450,14 @@
     :collection_actions:
       :get:
       - :name: read
-        :identifier: event
+        :identifier: miq_event
       :post:
       - :name: query
-        :identifier: event
+        :identifier: miq_event
     :resource_actions:
       :get:
       - :name: read
-        :identifier: event
+        :identifier: miq_event
   :features:
     :description: Product Features
     :options:
@@ -2387,7 +2387,7 @@
         :identifier: picture_new
   :policies:
     :description: Policies
-    :identifier: policy
+    :identifier: miq_policy
     :options:
     - :collection
     - :subcollection
@@ -2400,28 +2400,28 @@
     :collection_actions:
       :get:
       - :name: read
-        :identifier: policy_view
+        :identifier: miq_policy_view
       :post:
       - :name: query
-        :identifier: policy_view
+        :identifier: miq_policy_view
       - :name: create
-        :identifier: policy_new
+        :identifier: miq_policy_new
       - :name: edit
-        :identifier: policy_edit
+        :identifier: miq_policy_edit
       - :name: delete
-        :identifier: policy_delete
+        :identifier: miq_policy_delete
     :resource_actions:
       :get:
       - :name: read
-        :identifier: policy_view
+        :identifier: miq_policy_view
       :post:
       - :name: edit
-        :identifier: policy_edit
+        :identifier: miq_policy_edit
       - :name: delete
-        :identifier: policy_delete
+        :identifier: miq_policy_delete
       :delete:
       - :name: delete
-        :identifier: policy_delete
+        :identifier: miq_policy_delete
     :subcollection_actions:
       :post:
       - :name: assign
@@ -2434,7 +2434,7 @@
       - :name: resolve
   :policy_actions:
     :description: Actions
-    :identifier: control_explorer
+    :identifier: control
     :options:
     - :collection
     - :subcollection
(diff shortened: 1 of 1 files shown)


config/api.yml
```diff
@@ -2464,47 +2464,55 @@
...
       - :name: edit
-        :identifier: profile_edit
+        :identifier: miq_policy_set_edit
     :policies_subcollection_actions:
       :post:
       - :name: assign
-        :identifier: policy_profile_assign
+        :identifier:
+        - miq_policy_conditions_assignment
+        - miq_policy_events_assignment
```
NickLaMuro, line 2500: This was the reason why I asked this question to @h-kataria 

https://github.com/ManageIQ/manageiq/pull/20438/files#r525530422

In my PR, I think I was a bit confused, and am not sure I did the right thing:

https://github.com/ManageIQ/manageiq-api/pull/965/files#diff-0f4eeb68b6115c7dcee6074cbf57d5e60b4f0abd7abe5b2ec5b1f8cab58d8d35R2498

So that probably should not match mine, but we probably should determine what is the correct permission identifier here.  I would assume it would be under the `miq_policy_set*` based on the other permissions, but perhaps I am missing something.  That said, the PR that changed all these in core didn't really provide much context for me to go off of.

cc @gtanzillo @h-kataria

config/api.yml
```diff
@@ -176,8 +176,8 @@
       - :name: delete
         :identifier: alert_definition_delete
   :alerts:
-    :description: Alerts
-    :identifier: alert
+    :description: Alert Statuses
```
NickLaMuro, line 179: There a reason you decided to change the name here?

This wasn't even a change in `ManageIQ/manageiq`:

https://github.com/ManageIQ/manageiq/pull/20438/files#diff-cdcd147fb85ef58d008a4ad0cfa3ddfea649382cb062407e2342aff07362b6cfL2049

so not sure what is prompting the change here.

config/api.yml
```diff
@@ -176,8 +176,8 @@
...
+    :identifier: alert_status
```
NickLaMuro, line 180: Seems like this should just be changed to `miq_alert`, based on what the previous key was:

https://github.com/ManageIQ/manageiq/pull/20438/files#diff-cdcd147fb85ef58d008a4ad0cfa3ddfea649382cb062407e2342aff07362b6cfL2054

Also what I did in my PR:

https://github.com/ManageIQ/manageiq-api/pull/965/files#diff-0f4eeb68b6115c7dcee6074cbf57d5e60b4f0abd7abe5b2ec5b1f8cab58d8d35L180

config/api.yml
```diff
@@ -2443,17 +2443,17 @@
     :collection_actions:
       :get:
       - :name: read
-        :identifier: control_explorer_view
+        :identifier: miq_action_show_list
```
NickLaMuro, line 2446: For these next three, what was the rational for using this permission?

The reason I ask is that this was the other permission that I had a question on, since in https://github.com/ManageIQ/manageiq/pull/20438 it was removed:

https://github.com/ManageIQ/manageiq/pull/20438/files#r525547913

Your response really didn't provide much context.  Am I to assume that `miq_action_show_list` is the same permission then?

NickLaMuro (changes requested):
Well... would have been nice if you had just responded to the PR I already had opened doing mostly the same thing...

https://github.com/ManageIQ/manageiq-api/pull/965

But I guess whats done is done...

* * *

I have a few comments regarding some of the decisions you made that differed from what was done in my PR, and would at least like for those to be answered prior to me approving this PR.

config/api.yml
skateman: Because this endpoint is not for alerts but alert statuses, wrongly named (see the class and the child endpoints below)

config/api.yml
skateman: Anything containing `explorer` in the API feature identifiers is probably wrong. The explorer is a (bad) UI concept and these features should not be tied to the API at all. The endpoint here is retrieving you a list of policy actions, for this we have this dedicated RBAC feature identifier...

config/api.yml
skateman: If you look at the stuff below, it says `alert_status_*` for all child features + the `klass` is also `MiqAlertStatus` so here the top route and its naming was wrong (IMO).

config/api.yml
NickLaMuro: Fair enough.

config/api.yml
NickLaMuro: However, unlike the name change above, I could see this having issues when a user upgrades because what was previously associated with one `MiqProductFeature`(?) is now associated with another.

Does this change also require a migration?

config/api.yml
NickLaMuro: "Probably wrong" or not, there are probably deployments of MIQ where this is currently the assumed as the identifier gating these permissions.

If this changes, this then either enables or disables configured permissions for some users based on what is currently configured, making it a breaking change.

This is similar to what I am getting at here:  https://github.com/ManageIQ/manageiq-api/pull/966#discussion_r526986231

config/api.yml
skateman: Well, I can go with `miq_alert` which would not break compatibility but that's not the correct feature for this class name. Also there's no [migration](https://github.com/ManageIQ/manageiq-schema/pull/509) for [this](https://github.com/ManageIQ/manageiq/pull/20438/files#diff-cdcd147fb85ef58d008a4ad0cfa3ddfea649382cb062407e2342aff07362b6cfR2002) rename. cc @h-kataria

config/api.yml
skateman: Yeah, this will also need a migration as it was missed in the original PRs :confused:

h-kataria: @NickLaMuro @skateman prior to Control explorer split and renaming of the product features we only had a parent feature exposed in product features tree and that was `control_explorer` any children features under that used to be hidden features and were granted access if user role had access to `control_explorer`. User roles only had `control_explorer` feature that is renamed to `control`

config/api.yml
h-kataria: @NickLaMuro https://github.com/ManageIQ/manageiq/blob/master/db/fixtures/miq_product_features.yml#L2025-L2139 features do not have corresponding screens in UI these features were added for API usage only, and remain unchanged with my Control explorer work.
See screenshots from before the Control explorer split
screen_shot_2020-11-20_at_11 40 11_am

screen_shot_2020-11-20_at_11 39 27_am

All you need to do is fix the identifier/description for `alert`

config/api.yml
skateman: The `miq_policy_set_view` is definitely not enough, as you're modifying the policy profile by assigning conditions or events to it. Also these are both exposed in the UI...

Fryguy:
Renaming RBAC generally requires a database migration in order to ensure existing groups have the roles properly attributed - is one of those in place for these changes?

h-kataria:
> Renaming RBAC generally requires a database migration in order to ensure existing groups have the roles properly attributed - is one of those in place for these changes? (is it this one? [ManageIQ/manageiq-schema#509](https://github.com/ManageIQ/manageiq-schema/pull/509) )

@Fryguy that's the correct PR to address migration for the control explorer related features.

Fryguy:
I only brought it up, because I don't see the same RBAC features in this PR (unless these features are existing features, and no transfer is needed?)

NickLaMuro (approved):
We discussed this off line, but my two remaining comments will be addressed in a data migration:

- For https://github.com/ManageIQ/manageiq-api/pull/966/files#r526141669 we just have to make sure if existing permissions for the old permission happen to exist, they get transferred to the new appropriate permission this is now assigned to
- For https://github.com/ManageIQ/manageiq-api/pull/966/files#r526147366 the expanded permission this was previously, `controller_explorer_view`, now needs to migrate to include all of the permissions previously

But these are just a safety net for those who might have included these permissions through a console, and are probably edge cases at best, if not non-existent in the real world.  Shouldn't stop this PR from going out, so good to go. 👍

PR #970, opened 2020-11-30
Introducing a simple GET-only endpoint for shortcuts

skateman:
The list of shortcuts for startpages is being stored in the database, so we need to expose them through the API to access them from the UI. It is probably necessary to expose it to all users, but the feature still needs to be created. Also the collection entities should be displayed only if the user has access to them based on their `rbac_feature_name` attribute.

So I think I need some help here to filter the results based on this field @gtanzillo 

Depends on https://github.com/ManageIQ/manageiq/pull/20844

### DIFF
--- a/app/controllers/api/shortcuts_controller.rb
+++ b/app/controllers/api/shortcuts_controller.rb
@@ -1,0 +1,4 @@
+module Api
+  class ShortcutsController < BaseController
+  end
+end
--- a/spec/requests/collections_spec.rb
+++ b/spec/requests/collections_spec.rb
@@ -402,6 +402,11 @@ describe "Rest API Collections" do
       FactoryBot.create(:miq_search)
       test_collection_query(:search_filters, api_search_filters_url, MiqSearch)
     end
+
+    it 'query shortcuts' do
+      FactoryBot.create(:miq_shortcut, :rbac_feature_name => 'shortcuts')
+      test_collection_query(:shortcuts, api_shortcuts_url, MiqShortcut)
+    end
   end
 
   context "Collections Bulk Queries" do
--- a/config/api.yml
+++ b/config/api.yml
@@ -4003,6 +4003,21 @@
       :get:
       - :name: read
         :identifier: ops_settings
+  :shortcuts:
+    :description: Shortcuts
+    :options:
+    - :collection
+    :identifier: shortcuts
+    :klass: MiqShortcut
+    :verbs: *g
+    :collection_actions:
+      :get:
+      - :name: read
+        :identifier: shortcuts
+    :resource_actions:
+      :get:
+      - :name: read
+        :identifier: shortcuts
   :snapshots:
     :description: Snapshots
     :options:


app/controllers/api/shortcuts_controller.rb
```diff
@@ -0,0 +1,4 @@
+module Api
+  class ProvidersController < BaseController
```
Fryguy, line 2:
```suggestion
  class ShortcutsController < BaseController
```

skateman: Lol, I forgot to push :rofl:

lpichler:
> The list of shortcuts for startpages is being stored in the database, so we need to expose them through the API to access them from the UI. It is probably necessary to expose it to all users, but the feature still needs to be created. Also the collection entities should be displayed only if the user has access to them based on their `rbac_feature_name` attribute.
> 
> So I think I need some help here to filter the results based on this field @gtanzillo 

Possible solution https://github.com/ManageIQ/manageiq/pull/20848.

Fryguy:
LGTM...is this dependent on ManageIQ/manageiq#20848 ?

abellotti:
specs needed, probably just spec/requests/collections_spec.rb update.

lpichler:
> LGTM...is this dependent on [ManageIQ/manageiq#20848](https://github.com/ManageIQ/manageiq/pull/20848) ?

yes, otherwise RBAC will not work for shortcuts as we need.

skateman:
> specs needed, probably just spec/requests/collections_spec.rb update.

@abellotti can you please be more specific? I can't see any case in that file that explicitly tests the root route :confused:

abellotti:
@skateman the spec/requests/collections_spec.rb has helper methods to test collection queries (and bulk query if that action if defined). In cases like a GET only endpoint like shortcuts where you don't need to test all methods/actions/etc in a separate spec file, you can simply add a short test in that file for that new collection.  i.e. adding something like:

```
it "query Shortcuts" do
     FactoryBot.create(:shortcut)
     test_collection_query(:shortcuts, api_shortcuts_url, MiqShortcut)
end
```

Will create a shortcut and do a GET on /api/shortcuts testing the new collection route and response.

skateman:
Oh sorry, I thought this is the timezones endpoint :see_no_evil:

spec/requests/collections_spec.rb
```diff
@@ -402,6 +402,11 @@ def test_collection_bulk_query(collection, collection_url, klass, id = nil)
       FactoryBot.create(:miq_search)
       test_collection_query(:search_filters, api_search_filters_url, MiqSearch)
     end
+
+    it 'query shortcuts' do
+      FactoryBot.create(:shortcut)
+      test_collection_query(:shortcuts, api_shortcuts_url, MiqShortcut)
+    end
```
Fryguy, line 409: Not for this PR, I feel like we should move this into dedicated files.  Yeah, the files will be small, but then the pattern is clearer to people implementing a new endpoint - they see a spec file and they copy the pattern (as opposed to completely missing it in this generic file).

Fryguy:
@skatem do you need also need a test for bulk query?

skateman:
> @skateman do you need also need a test for bulk query?

nope as it is not supported, if you look into the `api.yml` there's no action for querying and I don't think we need it...

Fryguy (approved):

abellotti (approved):

Fryguy:
This is good to go, just needs specs to pass.

skateman:
> This is good to go, just needs specs to pass.

done!

PR #971, opened 2020-11-30
Make api debuggable from browser again, accept cookie without csrf for GET

himdel:
HTTP Basic is no longer really supported by the API, without the auth request, the browser ignores URLs like http://admin:smartvm@localhost:3000 , and never opens the login popup.
So, there is currently no way to open an api request directly from the browser window, for debugging.

Since there was [presumably](https://github.com/ManageIQ/manageiq-api/commit/800747976be6bbe5c5033f769e18da7a612e4be0) a reason to break http basic that way (it still works if the client ignores it and sends the right headers, so it does work for curl),
but there's also reason to make this accessible from the browser.

So, adjusting the ui_session code to:

* treat any request with the right *cookie* as an UI request, not only when CSRF present (retaining csrf logic in case it's needed when using openid)
* don't require CSRF token for GET requests

That makes it possible to open an API tab as long as the user is logged in into manageiq UI in the browser.

Cc @skateman , @jvlcek

### DIFF
--- a/app/controllers/api/base_controller/authentication.rb
+++ b/app/controllers/api/base_controller/authentication.rb
@@ -15,10 +15,9 @@ module Api
         elsif request.headers["HTTP_AUTHORIZATION"]
           # For AJAX requests the basic auth type should be distinguished
           request.headers['X-REQUESTED-WITH'] == 'XMLHttpRequest' ? :basic_async : :basic
-        elsif request.x_csrf_token
+        elsif request.cookies['_vmdb_session'] || request.x_csrf_token
           # Even if the session cookie is not set, we want to consider a request
-          # as a UI authentication request. Otherwise the response would force
-          # the browser to throw an undesired HTTP basic authentication dialog.
+          # as a UI authentication request. (Previously that would force a login popup.)
           :ui_session
         else
           # no attempt at authentication, usually falls back to :basic
@@ -139,8 +138,8 @@ module Api
 
       def valid_ui_session?
         [
-          valid_authenticity_token?(session, request.x_csrf_token), # CSRF token be set and valid
-          session[:userid].present?,                                # session has a userid stored
+          valid_authenticity_token?(session, request.x_csrf_token) || request.method == 'GET', # CSRF token be set and valid (or GET)
+          session[:userid].present?, # session has a userid stored
           request.origin.nil? || request.origin == request.base_url # origin header if set matches base_url
         ].all?
       end


himdel:
Cc @Fryguy , what broke this is 800747976 (https://github.com/ManageIQ/manageiq-api/pull/886)

Fryguy:
We intentionally removed basic auth for security reasons (specifically a possible CSRF).

Fryguy:
@kbrock Please review.

himdel:
(test failure unrelated, that's on master as well)

skateman:
CI should be fixed by https://github.com/ManageIQ/manageiq-api/pull/966

Fryguy:
I publicized the security issue (it was fixed a while ago, actually) - https://github.com/ManageIQ/manageiq/security/advisories/GHSA-fwcf-7r2j-3vq7

Fryguy (approved):
I think this is good, but want @kbrock to verify.

Fryguy:
>  So, there is currently no way to open an api request directly from the browser window, for debugging.

FWIW (and for others reading along) basic auth works fine if you are able to pass the header on the request directly without waiting for the challenge.  This is why curl works (because it sends the header immediately unless you pass `--anyauth`).  If you want to use the browser still you can use a browser extension like [ModHeader](https://chrome.google.com/webstore/detail/modheader/idgpnmonknjnojddfkpgkljpfnnfcklj?hl=en), and craft the Basic Auth header (which is simply `"Authorization: Basic #{Base64.encode("user:password")}"`)

This PR is relying on the local cookie you'd get from the UI, so you'd log into the UI, have the cookie locally, and then in a separate tab, the browser will send that along with the API request.  I'm not sure if this opens us up to the similar problem we solved with the basic auth challenge, but I think the CSRF protections make this different and also secure.

Fryguy:
@abellotti Can you also review?

kbrock:
I really don't like this. :( speaking with @himdel and will post more on this thread

kbrock:
Sorry 📖 

## Overall Problem

When you login into an app, the browser remembers the user's authentication credentials and sends the credentials on all subsequent requests.

Cookies can be deleted, session ids can be made invalid.
But basic-auth headers can not be deleted without shutting down the browser. Credentials can only be made invalid by changing the username and password.

## Securing the sandbox

The default javascript sandboxing blocks foreign javascript from accessing our cookies and trusted code. The malicious script gets around the sandbox constraint with two common attack vectors:

1. pasting a form or script into our page. This code will look like it is trusted and run in our sandbox.
2. asking our browser to make a request via something like a server redirect, Flash redirect, javascript redirect, or popup.

### Script protection

We protect the sandbox with the following methods:

- `CORS` allows us to punch a hole in that sandbox and trusting javascript from other sites to call our code.  E.g.: CDN
- `IFRAME` allows us to paste content into our page but mark it as untrusted. E.g.: advertising
- the `h` in  `<%= h user.field %>` escapes content, making any script or form variable in user content not run able to run in our page (or the database).

### Request protection

It is the protection of the redirect that is tricky and our current concern.

We introduce an XSRF token, which is unique per page request. So while the credentials are automatically included in every request, the server rejects requests that do not also have the token. Our code can either puts the token into all forms or add the tokens to XHR requests.

The problem with XSRF tokens is they are dependent upon the previous request, which is not very REST. So for the REST API we do not want to require XSRF tokens.

|method|contents|browser support|XSRF tokens|
|---|---|---|---|
|Basic-Auth|user/password|No|No|
|jwt tokens|session_id|No|No|
|Cookie|session_id|Yes|Yes|

### Locking down Basic-Auth

The way to prevent the browser from using Basic-Auth is by not sending a challenge and advertising Basic-Auth support. The browser doesn't even know that Basic-auth is a viable option browser.

JWT tokens are not part of standard browser functionality. So if they stay out of cookies or headers, we are all set.

## Proposal

The proposal is to add a trigger that will tell the server to send the challenge, and basically enable Basic-Auth in the browser. Make the browser vulnerable to XSRF - but only when the trigger is present. In this case, the trigger is a particular cookie.

We need to make sure that our customers do not have this flag set in normal usage.
It is also important for customers to know that by using the flag they will be in a vulnerable state and how to get out of it.

Outside example: allow the user to login as root, but know they are vulnerable and logout as soon as they can.

## Concerns

Every time we provide a backdoor, it tends to be used for standard business practices. So it transforms from a back door to  a front door. And a front door without some form of protection is dangerous.

app/controllers/api/base_controller/authentication.rb
```diff
@@ -139,8 +138,8 @@ def validate_system_token_timestamp(timestamp)
 
       def valid_ui_session?
         [
-          valid_authenticity_token?(session, request.x_csrf_token), # CSRF token be set and valid
-          session[:userid].present?,                                # session has a userid stored
+          valid_authenticity_token?(session, request.x_csrf_token) || request.method == 'GET', # CSRF token be set and valid (or GET)
```
kbrock, line 141: Is there a way to make a GET request but pass an action so it performs a mutable operation?

app/controllers/api/base_controller/authentication.rb
```diff
@@ -15,10 +15,9 @@ def auth_mechanism
         elsif request.headers["HTTP_AUTHORIZATION"]
           # For AJAX requests the basic auth type should be distinguished
           request.headers['X-REQUESTED-WITH'] == 'XMLHttpRequest' ? :basic_async : :basic
-        elsif request.x_csrf_token
+        elsif request.cookies['_vmdb_session'] || request.x_csrf_token
```
kbrock, line 18: I think our apps all include vmdb session, so won't this essentially turn of csrf for all browser requests?

kbrock (commented):
already commented in a book. I think this review better highlights my specific implementation concerns.

Maybe a custom, non webapp url parameter would work?
Or a non webapp custom cookie would work? (custom cookie seems more tricky)

Or a specific endpoint that would return a challenge?

abellotti:
@himdel is the debugging needed on production systems ? if not, could this be solved if we enable the Basic-Auth challenge if  RAILS_ENV=development ?

himdel:
@kbrock I'm sorry too, but I think you've completely missed the point here:

1. this does not attempt to re-introduce HTTP Basic in any way. I disagree with us having broken HTTP Basic the way we did (IMO no http basic request = no http basic), but I'm not fighting that, I undestand the reasons there. (Even though it means most non-debugging tools won't be able to log in, without crafting a header manually.)
This only makes the cookie authentication behave similar to the UI.

2. you're seeing this as a back door. It's not. It does the exact same thing our UI does. Cookie is used for reading data, cookie + csrf is used for writing data. That's the normal behaviour, there is absolutely no reason to require CSRF for GET requests, and we don't anywhere else.

So I'm ignoring your comment as I think you were really thinking of http basic when you were writing that :), can you re-review please? :).

---

> is the debugging needed on production systems ? if not, could this be solved if we enable the Basic-Auth challenge if RAILS_ENV=development ?

I mean, this is not *strictly* necessary even in development. There is just no reason not to do that (again, GET is never used to write data), and it's equally useful when debugging production and development.

I would just really appreciate having a way of opening the API response in a new browser tab. I get why you dislike http basic, this is me trying to come with a more secure alternative.

As far as I'm concerned this is it, does anyone have any actual argument why this breaks?

himdel:
> Proposal .. The proposal is to add a trigger that will tell the server to send the challenge, and basically enable Basic-Auth in the browser.

I would be OK with dropping my PR, and re-enabling HTTP Basic challenge when an attribute is present.
But, attribute, not a cookie, that's nonsense.

himdel:
Oh, in case anyone is wondering about the use case...

I'm logged in in manageiq as $user, want to see an API response *for that particular user*.

Now I *could* open postman, configure the username and password in a second place, and keep 2 different places logged in, but that's too cumbersome, and inevitably, you will forget to change users at some point.

I could also just use `API.get...` in the miq browser console, but then you're limited to the developer tools tab.

Crucially,  our API uses HATEOAS links .. and they're entirely useless now. 
It *would* be a great way of exploring the API, seeing the related endpoints, etc.

But .. how do you open a HATEOAS link? You're right, in a new tab, and it's not authorized :).

app/controllers/api/base_controller/authentication.rb
himdel: No. Action is only allowed for post.

app/controllers/api/base_controller/authentication.rb
himdel: ..how? 

Before: a cookie was ignored if csrf was not found
Now: a cookie is enough to get to the place that checks for csrf or get

kbrock (approved):
Thanks so much @himdel LGTM

kbrock:
@Fryguy Merge away

app/controllers/api/base_controller/authentication.rb
Fryguy: There _might_ be GETs with side effects, but in general, I think we've tried to avoid that.  If we were to find one, we could always mark that one as needing CSRF protections.

@himdel Would it make sense to move your `request.method == 'GET'` into a method named `request_allowed?` or something?  This way, if we need to deny a particular get request we have a method where we can encapsulate that?

Fryguy:
> does anyone have any actual argument why this breaks?

Nope, we all just wanted to be absolutely sure we understood what exactly this change entails, and the possible side effects (considering it took us a while to come to the conclusion to remove the basic auth challenge, which we did begrudgingly).  Thanks for entertaining the discussion.

himdel:
Thanks! :)

Re the possibilty of using GET to do a POST-like request... GET has no body, so, I *think* the only rails-magic way of doing that would be `GET /api/vms/123?method=delete` or  `GET /api/vms/123?_method=delete`. (As suggested by https://medium.com/@moulayjam/how-put-patch-and-delete-methods-work-with-ruby-on-rails-9cfc1cff2dde)

I've tried both, neither seems to have attempted a delete. I've also tried `?action=delete`, all of those behave like a GET without params.

Fryguy:
Well, I was more talking about cases where there are *side effects* of a GET...like let's say someone created a hypthetical `GET /api/vms/123/delete_me` endpoint for whatever reason (obviously someone wouldn't create that endpoint, but it demonstrates the idea of a GET having a side-effect that could be detrimental.

PR #973, opened 2020-12-01
Add the list of available rails timezones to the base endpoint

skateman:
There's no way to retrieve the list of available timezones from the frontend in a Rails-compatible way, so I'm adjusting the `/api` endpoint to provide a list of these. 

```js
# GET /api

{
  "name": "API",
  "description": "REST API",
  ...
  "timezones": [
    {
      "name": "International Date Line West",
      "description": "(GMT-12:00) International Date Line West"
    },
    {
      "name": "American Samoa",
      "description": "(GMT-11:00) American Samoa"
    },
    ...
  ],
  ...
}
```

### DIFF
--- a/app/controllers/api/api_controller.rb
+++ b/app/controllers/api/api_controller.rb
@@ -6,14 +6,15 @@ module Api
 
     def index
       res = {
-        :name          => ApiConfig.base.name,
-        :description   => ApiConfig.base.description,
-        :version       => ManageIQ::Api::VERSION,
-        :versions      => entrypoint_versions,
-        :settings      => user_settings,
-        :identity      => auth_identity,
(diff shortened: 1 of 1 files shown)


Fryguy:
I don't think we should expose a proper endpoint for something that isn't the domain of our application (imagine, for example, github exposed a /api/timezones endpoint, but it was because of some quirk of their UI...it just doesn't make sense).

Instead, can we just expose this from /api or OPTIONS /api ?  I imagine this is a one-time thing that needs to be queried and cached by the UI.

skateman:
@Fryguy updated

Fryguy (approved):
@gtanzillo Please also review.

gtanzillo (approved):

PR #985, opened 2021-01-17
Implement define host initiator

llivne:
Part of https://github.com/ManageIQ/manageiq/issues/21003


llivne:
travis will pass once core changes will be merged. see https://github.com/ManageIQ/manageiq-cross_repo-tests/pull/265

agrare:
@llivne can you add spec tests covering these new APIs

llivne:
> @llivne can you add spec tests covering these new APIs

Done. I saw a bug that cause the tests to fail, I opened PR on it:
https://github.com/ManageIQ/manageiq/pull/20991

llivne:
@agrare I dont think travis fails because of my changes. can you please take a look on it?

agrare:
Hey @llivne yes manageiq-api is currently red on master, looks like your tests passed

app/controllers/api/host_initiators_controller.rb
```diff
@@ -0,0 +1,48 @@
...
+      act_refresh(host_initiator, method_name, role)
+    rescue => err
+      action_result(false, err.to_s)
+    end
+
+    def host_initiator_ident(host_initiator)
+      "Host Initiator id:#{host_initiator.id} name:'#{host_initiator.name}'"
+    end
+
+    def act_refresh(host_initiator, method_name, role)
```
agrare, line 42: The content of this method looks like it should just go in `refresh_host_initiator` then we wouldn't need method_name and role as arguments.

It'd look much simpler as:
```
def refresh_host_initiator(host_initiator)
  task_id = queue_object_action(host_initiator, desc, :method_name => "refresh_ems", :role => "ems_operations")
  action_result(true, "#{host_initiator_ident(host_initiator)} refreshing", :task_id => task_id)
rescue => err
end
end
```

config/api.yml
```diff
@@ -1880,6 +1880,30 @@
...
+    :subcollections:
+    :collection_actions:
+      :get:
+      - :name: read
+        :identifier: host_initiator_show_list
+      :post:
+      - :name: create
+        :identifier: host_initiator_new
+      - :name: refresh
+        :identifier: host_initiator_refresh
```
agrare, line 1899: What does a refresh action on the collection mean?  I would expect refresh only on the resource

agrare: Ahh okay I see you're doing a bulk refresh action https://github.com/ManageIQ/manageiq-api/pull/985/files#diff-3dbf099b30332a2d5c69d3f876bb4510c784579cd46a1c9f5ae6bddb399be467R102
Carry on

agrare:
Just the one comment about the `act_refresh` method, @abellotti @Fryguy please review

app/controllers/api/host_initiators_controller.rb
llivne: done

llivne:
> Just the one comment about the `act_refresh` method, @abellotti @Fryguy please review

done

app/controllers/api/host_initiators_controller.rb
```diff
@@ -0,0 +1,41 @@
...
+
+      api_action(type, id) do |klass|
+        host_initiator = resource_search(id, type, klass)
+        api_log_info("Refreshing #{host_initiator_ident(host_initiator)}")
+        refresh_host_initiator(host_initiator)
+      end
+    end
+
+    def create_resource(_type, _id = nil, data = {})
+      ext_management_system = ExtManagementSystem.find(data['ems_id'])
```
abellotti, line 16: Since we're using ems_id, probably check for its existence before using it and throw a BadRequestError accordingly.   Maybe add a test for that case.

abellotti:
@llivne PR looks great, just adding the one check/spec. Thanks for the API Enhancement 👍

app/controllers/api/host_initiators_controller.rb
llivne: @abellotti I added the check for ems_id existence and test for it.

llivne:
> @llivne PR looks great, just adding the one check/spec. Thanks for the API Enhancement 👍

done 👍

app/controllers/api/host_initiators_controller.rb
```diff
@@ -0,0 +1,43 @@
...
+
+      api_action(type, id) do |klass|
+        host_initiator = resource_search(id, type, klass)
+        api_log_info("Refreshing #{host_initiator_ident(host_initiator)}")
+        refresh_host_initiator(host_initiator)
+      end
+    end
+
+    def create_resource(_type, _id = nil, data = {})
+      raise BadRequestError, "ems_id not defined for #{_type} resource" if data['ems_id'].blank?
```
agrare, line 16: Interesting, we don't do this for any other EMS lookups by `date["ems_id"]`, I would have expected the `.find()` to raise an exception and the API to return an appropriate error but this works too.
@abellotti is this a pattern you'd like to follow more broadly?

We're doing similar lookups here without this check:
```
app/controllers/api/auth_key_pairs_controller.rb:      ext_management_system = ExtManagementSystem.find(data['ems_id'])
app/controllers/api/cloud_tenants_controller.rb:      ext_management_system = resource_search(data['ems_id'], :providers, collection_class(:providers))
app/controllers/api/physical_storages_controller.rb:      ext_management_system = ExtManagementSystem.find(data['ems_id'])
app/controllers/api/host_aggregates_controller.rb:      ext_management_system = resource_search(data['ems_id'], :providers, collection_class(:providers))
app/controllers/api/floating_ips_controller.rb:      ems = ExtManagementSystem.find(data['ems_id'])
app/controllers/api/cloud_volumes_controller.rb:      ext_management_system = ExtManagementSystem.find(data['ems_id'])
```

PR #988, opened 2021-01-18
Fix klass.superclass not matching with acts_as_leaf_class

agrare:
When a manager class subclasses from another non-core manager-type class.  For example if `ManageIQ::Providers::Openshift::ContainerManager` inherits from `ManageIQ::Providers::Kubernetes::ContainerManager` instead of `ManageIQ::Providers::ContainerManager` then using superclass doesn't work.

Required for: https://github.com/ManageIQ/manageiq/pull/20756

### DIFF
--- a/app/controllers/api/providers_controller.rb
+++ b/app/controllers/api/providers_controller.rb
@@ -139,7 +139,7 @@ module Api
 
     def authorize_provider(typed_provider_klass)
       create_action = collection_config["providers"].collection_actions.post.detect { |a| a.name == "create" }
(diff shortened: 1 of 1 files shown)


agrare:
Cross-repo tests https://github.com/ManageIQ/manageiq-cross_repo-tests/pull/239

Fryguy (approved):

Fryguy:
Technically this is a bug fix...we just happen to not have any callers that can cause this to happen at the moment.  However, with ManageIQ/manageiq#20756 coming, it will be an issue.

Issue #936, opened 2020-10-22
Endpoint for Policy simulation


skateman:
The policy simulation itself is an asynchronous task initiated by [`.rsop_async`](https://github.com/ManageIQ/manageiq/blob/master/app/models/mixins/miq_policy_mixin.rb#L141) from the UI controller. 

![Screenshot from 2020-10-22 13-14-30](https://user-images.githubusercontent.com/649130/96864351-8a0a0980-1468-11eb-9586-73a1b9ca4e7c.png)

The VM selection filelds in the form can be probably feeded using existing endpoints. The event selection is however, just partially available, there's no endpoint to retrieve the `MiqEventDefinitionSet`, only the underlying `MiqEventDefinition` entities.

@miq-bot add_label help wanted
@miq-bot assign @gtanzillo

skateman:
The `/api/event_definition_sets` endpoint seems okay to me. Now we would need an endpoint that implements `MiqPolicyMixin#rsop_async`, I think the best location would be `POST /api/event_definition/:id` with a `simulate` action, but @abellotti should confirm. cc @gtanzillo

abellotti:
You meant event_definition_sets ? If so, that makes sense,  resource endpoint **POST /api/event_definition_sets/:id** with action "**simulate**"

skateman:
The PR @lfu created is for listing `event_definition_sets` but the next step in the policy simulation is to select an actual `event_definition` and run the simulation against it.

abellotti:
Ahh ok, so for MiqEventDefinition that is currently exposed via /api/events, so sounds like that would be a **POST /api/events/:id** action simulate instead

skateman:
> Ahh ok, so for MiqEventDefinition that is currently exposed via /api/events, so sounds like that would be a **POST /api/events/:id** action simulate instead

I'm good with that, so @lfu we need a `simulate` action under `/api/events/:id` with basically the same behavior as `rsop_async` but returning the task ID as any other async API request would do.

lfu:
`MiqPolicyMixin` is included by many objects that support policy. `MiqPolicyMixin#rsop_async` is a class method of the event targets. So it seems not right to have a `simulate` action for the events.

In my opinion, `simulate` action should be added for those objects that support policy, like `Vm`, `ExtManagementSystem`, `Host` etc. There are about 15 classes that could call `MiqPolicyMixin#rsop_async`.

Is there a way to define a shared action once for all?

[...]

PR #993, opened 2021-01-27
Add simulate action for policy simulation.

lfu:
https://github.com/ManageIQ/manageiq-api/issues/936

@miq-bot add_label enhancement, kasparov/no

```
POST /api/vms
{
    "action": "simulate_policy",
    "event": "request_vm_poweroff",
    "targets": [
        {"id" : "10000000000001"},
        {"href": "http://localhost:3000/api/vms/10000000000003"},
        {"guid": "dcb564ce-801f-429a-8865-01d252d5ece7"}
    ]
}

{
    "success": true,
    "message": "Simulating policy on event request_vm_poweroff for targets [vms] ID: [10000000000001, 10000000000003, 10000000000870]",
    "task_id": "10000000053328",
    "task_href": "http://localhost:3000/api/tasks/10000000053328"
}
```

```
POST /api/vms
{
    "action": "simulate_policy",
    "resources": [
        {"id" : "10000000000001", "event": "request_vm_poweroff"},
        {"id" : "10000000000002", "event": "request_vm_poweroff"}
    ]
}

{
    "results": [
        {
            "success": true,
            "message": "Vm id:10000000000001 name: 'jrnettest' simulating policy on event request_vm_poweroff",
            "task_id": "10000000053329",
            "task_href": "http://localhost:3000/api/tasks/10000000053329",
            "href": "http://localhost:3000/api/vms/10000000000001"
        },
        {
            "success": true,
            "message": "Vm id:10000000000002 name: 'RHEL72_1' simulating policy on event request_vm_poweroff",
            "task_id": "10000000053330",
            "task_href": "http://localhost:3000/api/tasks/10000000053330",
            "href": "http://localhost:3000/api/vms/10000000000002"
        }
    ]
}
```

```
POST /api/vms/10000000000001
{
    "action": "simulate_policy",
    "event": "vm_poweroff"
}

{
    "success": true,
    "message": "Vm id:10000000000001 name: 'jrnettest' simulating policy on event vm_poweroff",
    "task_id": "10000000053331",
    "task_href": "http://localhost:3000/api/tasks/10000000053331",
    "href": "http://localhost:3000/api/vms/10000000000001"
}
```

### DIFF
--- a/app/controllers/api/mixins/policy_simulation.rb
+++ b/app/controllers/api/mixins/policy_simulation.rb
@@ -1,0 +1,50 @@
+module Api
+  module Mixins
+    module PolicySimulation
+      def simulate_policy_collection(type, data = {})
+        raise BadRequestError, "Must specify an event for policy simulation" if data["event"].blank?
+        raise BadRequestError, "Must specify targets for policy simulation" if data["targets"].blank?
+
+        target_ids = parse_collection_targets(type, data["targets"])
+        klass = collection_class(type)
+        invalid_ids = target_ids - klass.where(:id => target_ids).pluck(:id)
+        raise "Invalid target ids #{invalid_ids.sort.join(', ')} specified for #{klass.name}" if invalid_ids.present?
+
+        desc = "Simulating policy on event #{data["event"]} for targets [#{type}] ID: [#{target_ids.join(', ')}]"
+        task_id = klass.rsop_async(data["event"], target_ids, User.current_user)
+        action_result(true, desc, :task_id => task_id)
+      rescue => e
+        action_result(false, e.to_s)
+      end
+
+      def simulate_policy_resource(type, id, data = {})
+        raise BadRequestError, "Must specify an event for policy simulation" if data["event"].blank?
+
+        resource = resource_search(id, type, collection_class(type))
+
+        api_action(type, id) do
+          api_log_info("Simulating policy for #{resource_ident(resource)}")
+          request_policy_simulation(resource, data["event"])
+        end
+      end
+
+      def parse_collection_targets(type, targets)
+        targets.collect do |target|
+          parse_id(target, type) || parse_by_attr(target, type)
+        end.compact.uniq
+      end
+
+      def request_policy_simulation(resource, event)
+        desc = "#{resource_ident(resource)} simulating policy on event #{event}"
+        task_id = resource.class.base_class.rsop_async(event, [resource], User.current_user)
+        action_result(true, desc, :task_id => task_id)
+      rescue => err
+        action_result(false, err.to_s)
+      end
+
+      def resource_ident(resource)
+        "#{resource.class.name.demodulize.underscore.humanize} id:#{resource.id} name: '#{resource.name}'"
+      end
+    end
+  end
+end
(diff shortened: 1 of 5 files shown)


lfu:
Is this looking good as a request body for bulk action?
```
{
    "action": "simulate",
    "policy": {
      "event": "reqquest_vm_poweroff",
      "resources": [
        {"id" : "10000000000001"},
        {"id" : "10000000000002"}
      ]
    }  
}
```

abellotti:
bulk action in the API is intended for triggering an action while targeting multiple resources in a single request, i.e. in https://www.manageiq.org/docs/reference/latest/api/reference/vms.html look at starting multiple vms.

bulk action on vms in this PR would mean triggering the simulate action on multiple VMS, i.e.:

```
POST /api/vms

{
  "action" : "simulate",
  "resources" : [
      { "href" : "http://localhost:3000/api/vms/111", "policy" : { "event" : "request_vm_poweroff" } },
      { "href" : "http://localhost:3000/api/vms/112", "policy" : { "event" : "request_vm_poweroff" } },
      ...
   ]
}
```

Just using href in this example to show we're picking Vms, id, guid, etc. can also be used.

gtanzillo:
After discussion, we agreed on this format for the request -
```
{
    "action": "simulate",
    "event": "request_vm_poweroff",
    "resources": [
      {"id" : "10000000000001"},
      {"id" : "10000000000002"}
    ]
}
```

chessbyte:
@gtanzillo since there are only ids, without hrefs, will the event itself force the determination of which type of resource each is?

Fryguy:
@chessbyte The action applies to a collection (e.g. `/api/vms`) or a resource (e.g. `/api/vms/1`), so the object type is implied.  Note that while the example there uses `id`s, one can also pass hrefs, and I'd expect we use the common helper method that parses resource objects.

We also discussed naming the method `simulate_vm_policy` for clarity.  In fact, now that I read that, `simulate` isn't a good enough action name by itself, and at a minimum should have the word `policy` in there somewhere to qualify it.

Fryguy:
To add a little more clarity based on a question that was over in the issue.  If you were targeting non-VMs (e.g. the clusters dropdown), the payload would be similar:

`POST /api/cluster`

```json
{
    "action": "simulate_vm_policy",
    "event": "request_vm_poweroff",
    "resources": [
      {"id" : "10000000000001"},
      {"id" : "10000000000002"}
    ]
}
```

where those are cluster ids.  You're effectively asking "run the policies on all of the vms in these clusters, kicking it with the `request_vm_poweroff` event"

lfu:
If we want to use a method like `simulate_collection` to handle the bulk action, `resources` key can't exist at the root level of the request body hash due to [this line](https://github.com/ManageIQ/manageiq-api/blob/e174d80c96ca3f9f796ad5aa70c680752a99f9c3/app/controllers/api/base_controller/manager.rb#L9). Once there is `resources` key at the root level of the hash, `_collection` method would be skipped and `_resource` method would be called.

Could I suggest the following format:
```
POST /api/vms

{
    "action": "simulate_policy",
    "event": "request_vm_poweroff",
    "targets": [
      {"id" : "10000000000001"},
      {"id" : "10000000000002"}
    ]
}
```
Or something like:
```
{
    "action": "simulate_policy",
    "event": "request_vm_poweroff",
    "target_ids": [
      "10000000000001",
      "10000000000002"
    ]
}
```

abellotti (approved):

gtanzillo (approved):

PR #994, opened 2021-01-28
save after setting parent on vm update

d-m-u:
when we call replace_children on vm edit, it should be the responsibility of the caller to save the parent, so the update should be _after_
see the related https://github.com/ManageIQ/manageiq/pull/20990

@miq-bot add_label bug 

broken in https://github.com/ManageIQ/manageiq/pull/20274

### DIFF
--- a/app/controllers/api/vms_controller.rb
+++ b/app/controllers/api/vms_controller.rb
@@ -103,9 +103,9 @@ module Api
       attrs = validate_edit_data(data)
       parent, children = build_parent_children(data)
       resource_search(id, type, collection_class(type)).tap do |vm|
-        vm.update!(attrs)
         vm.replace_children(children)
         vm.set_parent(parent)
+        vm.update!(attrs)
       end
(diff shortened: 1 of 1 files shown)


app/controllers/api/vms_controller.rb
```diff
@@ -106,6 +106,7 @@ def edit_resource(type, id, data)
         vm.update!(attrs)
         vm.replace_children(children)
         vm.set_parent(parent)
+        vm.save!
```
agrare, line 109: We're doing one query to `update!` [above](https://github.com/ManageIQ/manageiq-api/pull/994/files#diff-7f92a67a59f1578ea3c05b75fc01361b0d9a576c9bc4473bb6b5e81b2aa66cceR106) then another `save!` here

Could we `assign_attributes(attrs)` then `.save!` with one query?

Fryguy:
All of the failures are the known widget spec issues, so merging on red.

Issue #1000, opened 2021-02-03
settings subcollection is not traversable like the main collection


Fryguy:
This works: `localhost:3000/api/settings/workers`
This does not: `localhost:3000/api/servers/1/settings/workers`

IMO they should be consistent.

PR #1002, opened 2021-02-05
Add Subcollections::Settings

NickLaMuro:
This code has been living in the base controller, when it really should just exist as a subcollection, so moving it there so it is matching conventions (easier to find), to better represent what it does, and only apply the method to controllers that actually use it.

Links
-----

Original PRs around this topic:

- https://github.com/ManageIQ/manageiq-api/pull/275
- https://github.com/ManageIQ/manageiq-api/pull/304

### DIFF
--- a/app/controllers/api/base_controller.rb
+++ b/app/controllers/api/base_controller.rb
@@ -114,43 +114,7 @@ module Api
       end
     end
 
-    def settings
-      id       = @req.collection_id
-      type     = @req.collection
-      klass    = collection_class(@req.collection)
-      resource = resource_search(id, type, klass)
-
-      case @req.method
-      when :patch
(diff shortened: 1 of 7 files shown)


Fryguy:
@NickLaMuro does this fix #1000 ?

NickLaMuro:
@Fryguy no, just some refactoring in settings to address some duplicate code and weirdness.  Felt like it should be it's own thing prior to addressing #1000

Fryguy:
Only thing I can't see is whether or not this code is used by the base collection as well?  (I'd expect `/api/settings/...` to be equivalent to `/api/server/<my_server_id>/settings/...`)

NickLaMuro:
@Fryguy `/api/settings` comes from `Api::SettingsController` (`app/controllers/api/settings_controller.rb`), and doesn't use this code, hence why it is confusing.

PR #1005, opened 2021-02-08
[config/api.yml] Fix for new miq_event identifiers

NickLaMuro:
This was changed in https://github.com/ManageIQ/manageiq/pull/21016 and due to https://github.com/ManageIQ/manageiq/pull/21016 it won't let the specs pass if there is a name mismatch.

Considerations
--------------

Currently in the UI, we have more granular filters for show/index, but in the API, we just use the single permission.

Perhaps it makes sense to make use of `miq_event_definition_show_list` and `miq_event_definition_show` instead of the broader permission?

### DIFF
--- a/config/api.yml
+++ b/config/api.yml
@@ -1490,7 +1490,7 @@
         :identifier: event_streams_show
   :events:
     :description: Events
-    :identifier: miq_event
+    :identifier: miq_event_definition
     :options:
     - :collection
     - :subcollection
@@ -1499,14 +1499,14 @@
     :collection_actions:
       :get:
       - :name: read
-        :identifier: miq_event
+        :identifier: miq_event_definition_show_list
       :post:
       - :name: query
-        :identifier: miq_event
+        :identifier: miq_event_definition
     :resource_actions:
       :get:
       - :name: read
-        :identifier: miq_event
+        :identifier: miq_event_definition_show
   :features:
     :description: Product Features
     :options:


config/api.yml
```diff
@@ -1499,14 +1499,14 @@
     :collection_actions:
       :get:
       - :name: read
-        :identifier: miq_event
+        :identifier: miq_event_definition
```
NickLaMuro, line 1502: Here, it might make sense to use `miq_show_list` instead of the current one.

config/api.yml
```diff
@@ -1499,14 +1499,14 @@
...
       :post:
       - :name: query
-        :identifier: miq_event
+        :identifier: miq_event_definition
     :resource_actions:
       :get:
       - :name: read
-        :identifier: miq_event
+        :identifier: miq_event_definition
```
NickLaMuro, line 1509: Here, it might make sense to use `miq_event_show_list` instead of the current one.

NickLaMuro: I believe since both `miq_event_show` and `miq_event_show_list` are children of `miq_event_definition`, changing to either of the children wouldn't require a migration to update as those with the current permission wouldn't loose anything, and it would only allow for more granular permissions later.

However, correct me if I am wrong on this.

h-kataria: @NickLaMuro it makes sense to go with granular permissions.

NickLaMuro: Roger.  Done.

h-kataria (approved):

agrare (approved):

PR #1014, opened 2021-02-17
[zones_spec.rb] Add tests for authentication

NickLaMuro:
**REQUIRES https://github.com/ManageIQ/manageiq/pull/21047 to be merged**

Adds some specs to cover creating and deleting authentications attached to zones using `accepts_nested_attributes_for :authentication`

Links
-----

* Core PR:  https://github.com/ManageIQ/manageiq/pull/21047
* Cross Repo tests:  https://github.com/ManageIQ/manageiq-cross_repo-tests/pull/319
* Related to https://github.com/ManageIQ/manageiq-api/issues/898

### DIFF
--- a/spec/requests/zones_spec.rb
+++ b/spec/requests/zones_spec.rb
@@ -25,6 +25,42 @@ RSpec.describe "Zones" do
   end
 
   context "edit", :edit do
+    it "can create zone authentication" do
+      api_basic_authorize action_identifier(:zones, :edit)
+
+      zone   = FactoryBot.create(:zone, :description => "Current Zone description")
+      params = gen_request(:edit,
+                           "description"               => "Updated Zone Description",
+                           "authentication_attributes" => {"userid" => "foo", "password" => "bar"})
+
+      expect(zone.authentication).to eq(nil) # sanity check
+
+      post api_zone_url(nil, zone), :params => params
+      zone.reload
+
+      expect(zone.description).to             eq("Updated Zone Description")
+      expect(zone.authentication.userid).to   eq("foo")
+      expect(zone.authentication.password).to eq("bar")
+    end
+
+    it "can delete an existing authentication" do
+      api_basic_authorize action_identifier(:zones, :edit)
+
+      zone   = FactoryBot.create(:zone, :description               => "Current Zone description",
+                                        :authentication_attributes => {:userid => "foo", :password => "bar"})
+      params = gen_request(:edit,
+                           "description"               => "Updated Zone Description",
+                           "authentication_attributes" => {"_destroy" => "true"})
+
+      expect(zone.authentication.userid).to eq("foo") # sanity check
+
+      post api_zone_url(nil, zone), :params => params
+      zone.reload
+
+      expect(zone.description).to    eq("Updated Zone Description")
+      expect(zone.authentication).to eq(nil)
+    end
+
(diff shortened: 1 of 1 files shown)


spec/requests/zones_spec.rb
```diff
@@ -25,6 +25,42 @@
   end
 
   context "edit", :edit do
+    it "can create zone authentication" do
+      api_basic_authorize action_identifier(:zones, :edit)
+
+      zone   = FactoryBot.create(:zone, :description => "Current Zone description")
+      params = gen_request(:edit,
+                           "description"               => "Updated Zone Description",
+                           "authentication_attributes" => {"userid" => "foo", "password" => "bar"})
```
bdunne, line 34: Should this be `default_authentication` rather than `authentication_attributes`?

spec/requests/zones_spec.rb
```diff
@@ -25,6 +25,42 @@
...
+
+      expect(zone.authentication).to eq(nil) # sanity check
```
bdunne, line 36: Shouldn't this be `zone.authentications` since it's a `has_many`?

NickLaMuro: @bdunne no, since this is meant to pull in the changes from here:

https://github.com/ManageIQ/manageiq/pull/21047

bdunne (commented):
Sorry, but I can't remember the use case for having authentications on a Zone.  Do we still use this?

spec/requests/zones_spec.rb
NickLaMuro: Again, this is a change that was introduced as a part of https://github.com/ManageIQ/manageiq/pull/21047 so `authentication_attributes` is from the `allow_nested_attributes_for` change made there.

NickLaMuro:
@bdunne it was a muddled response over here:

https://github.com/ManageIQ/manageiq-api/issues/898

But I think the resulting answer (not really said in the issue) was "let's not get rid of it, since we aren't sure".  Not a great answer, but to allow the UI conversions to continue moving forward, this was the answer I came up with.

bdunne:
> @bdunne it was a muddled response over here:
> 
> #898
> 
> But I think the resulting answer (not really said in the issue) was "let's not get rid of it, since we aren't sure". Not a great answer, but to allow the UI conversions to continue moving forward, this was the ([second](https://github.com/ManageIQ/manageiq-api/pull/916)) answer I came up with.

Reading through that more, it sounds like we may need to support multiple authentications hanging off a zone (windows_domain and ssh)

NickLaMuro:
@bdunne I feel like we can then cross that bridge when (and if) that becomes relevant, since this code hasn't been updated much at all lately, and these changes currently don't require a migration to change again in the future.  I didn't add support for the `has_many` relationship this time, because there is many more touch points for that style of relationship, but `allowed_nested_attributes_for` supports `has_many` as well:

https://api.rubyonrails.org/v6.1.0/classes/ActiveRecord/NestedAttributes/ClassMethods.html#module-ActiveRecord::NestedAttributes::ClassMethods-label-One-to-many

Also, currently the UI hard codes support for only for a single `:windows_domain`:

https://github.com/ManageIQ/manageiq-ui-classic/pull/7629/files

So the change in core is just lifting the loosely encoded relationship in the UI and enforcing it from a model perspective instead.

Issue #926, opened 2020-10-18
PXE server credential validation endpoint


skateman:
There's no endpoint for validating PXE server credentials. This is how it's done currently in the UI:
```ruby
  def pxe_server_async_cred_validation
    begin
      PxeServer.verify_depot_settings(params[:pxe])
    rescue StandardError => bang
      render :json => {:status => 'error', :message => _("Error during 'Validate': %{error_message}") % {:error_message => bang.message}}
    end

    render :json => {:status => 'success', :message => _('PXE Credentials successfuly validated')}
  end
```

@miq-bot add_label help wanted

PR #1017, opened 2021-02-23
Add PXE server credential validation endpoint

lfu:
Depends on https://github.com/ManageIQ/manageiq/pull/21013.

https://github.com/ManageIQ/manageiq-api/issues/926

@miq-bot add_label enhancement, kasparov/no

```
POST /api/pxe_servers
{"action" => "verify_credentials",
 "resource" =>
  {:name => "test", :uri => "smb://tmp/foo", :username => "test", :password => "foo"}
}
```

```
RESPONSE
{"results" =>
  [{"success" => true,
    "message" => "Credentials sent for verification",
    "task_id" => "13000000000006",
    "task_href" => "http://localhost:3000/api/tasks/13000000000006"}]
}
```

### DIFF
--- a/app/controllers/api/pxe_servers_controller.rb
+++ b/app/controllers/api/pxe_servers_controller.rb
@@ -55,6 +55,15 @@ module Api
       end
     end
 
+    def verify_credentials_resource(_type, id = nil, data = {})
+      zone_name = data.delete('zone_name')
+      data['id'] = id if id
+      task_id = PxeServer.verify_depot_settings_queue(User.current_user.userid, zone_name, data)
+      action_result(true, "Credentials sent for verification", :task_id => task_id)
+    rescue => err
+      action_result(false, err.to_s)
+    end
+
     private
 
     def validate_data_for(klass, data)
--- a/config/api.yml
+++ b/config/api.yml
@@ -3030,6 +3030,8 @@
         :identifier: pxe_server_edit
       - :name: delete
         :identifier: pxe_server_delete
+      - :name: verify_credentials
+        :identifier: pxe_server_new
     :resource_actions:
       :get:
       - :name: read
--- a/spec/requests/pxe_servers_spec.rb
+++ b/spec/requests/pxe_servers_spec.rb
@@ -267,4 +267,29 @@ RSpec.describe 'PxeServers API' do
       expect(response).to have_http_status(:forbidden)
     end
   end
+
+  describe 'post /api/pxe_servers with verify_credentials action' do
+    it 'queues the request to verify the credentials' do
+      api_basic_authorize collection_action_identifier(:pxe_servers, :verify_credentials)
+      request = {
+        "action"    => "verify_credentials",
+        "zone_name" => "zone",
+        "resource"  => {:name => "test", :uri => "smb://tmp/foo", :username => "test", :password => "foo"}
+      }
+      post(api_pxe_servers_url, :params => request)
+      expected = {
+        "results" => a_collection_containing_exactly(
+          a_hash_including(
+            "message"   => a_string_matching(/Credentials sent for verification/),
+            "success"   => true,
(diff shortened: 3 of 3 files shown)


lfu:
@miq-bot cross-repo-tests ManageIQ/manageiq#21013

app/controllers/api/pxe_servers_controller.rb
```diff
@@ -55,6 +55,13 @@ def edit_resource(type, id, data)
       end
     end
 
+    def verify_credentials_resource(_type, _id = nil, data = {})
+      task_id = PxeServer.verify_depot_settings_queue(User.current_user.userid, data)
```
agrare, line 59: @lfu if you're re-verifying credentials (e.g. if you edit an existing resource) we don't require the user to enter a password again but instead load it from vmdb if it doesn't exist (e.g. they only change the username or IP).

We do this by passing in `data["id"] = id if id` here and handling it on the `verify_depot_settings` side if an `"id"` is passed.

I don't know if that's something that we want to do on the UI for PxeServers but I expect it probably is and would keep parity with provider authentication verification.

agrare: @lfu (or @Fryguy @bdunne) do you need to pass PxeServers a zone?  I don't see one in the table right now but with any "local" resource it might not be reachable from every zone.

bdunne: That's a good idea

lfu: > @lfu if you're re-verifying credentials (e.g. if you edit an existing resource) we don't require the user to enter a password again but instead load it from vmdb if it doesn't exist (e.g. they only change the username or IP).

@agrare  `verify_depot_settings` always creates a [new instance](https://github.com/ManageIQ/manageiq/blob/de4c23643816b171243b26c5d48dc03add943873/app/models/mixins/file_depot_mixin.rb#L30) to verify the settings.

agrare: @lfu yes the problem with that is since we don't pull passwords back over the API if the user is editing an existing instance they have to re-type their password even if they don't intend on changing it.

We can handle that in a follow-up if you want but I don't think that's the best user experience.

Fryguy: @agrare Do we have a similar pattern for this with provider validation you can point to?  I think this probably is big enough that we should fix for this PR.

Fryguy: Oh sorry ... didn't realize this was an outdated comment.  Is this done then?

Fryguy: Mostly asking because I don't see anything that looks up the existing instance on the core side's method - https://github.com/ManageIQ/manageiq/pull/21013

agrare: @Fryguy the API now matches the pattern for provider validation, core still doesn't lookup the instance yet not sure if you want that for this PR set or as a follow-up

Fryguy: I think we need the core change then.  Otherwise, we have a "broken" API in the interim.

chessbyte:
@miq-bot cross-repo-tests

kbrock:
is this good to go?

Fryguy (changes requested):
Looks like we are still waiting on https://github.com/ManageIQ/manageiq-api/pull/1017#discussion_r582320690.  This was mainly being introduced for UI changes, and I'm not sure if there's priority on it.  @kavyanekkalapu would know better.  Otherwise, we can just leave this open until we get back to that screen.

kbrock:
Looks like https://github.com/ManageIQ/manageiq/pull/21013 has been merged.
Does that mean we can merge this one?

PR #1021, opened 2021-03-17
new POST API method allowing import of images from one provider into another

iv1111:
- this action is implemented in the context of a cloud template collection

- added import API endpoint for the image import operation in 'api.conf'

- added 'import_resource' implementation to the 'cloud_templates_controller.rb'

- calling static queueing method of the corresponding cloud template

- validation of input data is present and throws exceptions in case of malformed input

- minimal payload for a corresponding sample POST request
  '{"action": "import", "dst_provider_id": 10, "src_provider_id": 14, "src_image_id": 23 }'
  API path: /api/cloud_templates

- sibling PR in the base ManageIQ repository introducing static queuing method
  into base Template class is: https://github.com/ManageIQ/manageiq/pull/21106

- sibling PR in the base ManageIQ introducing a new feature is:
  https://github.com/ManageIQ/manageiq/pull/21112

### DIFF
--- a/app/controllers/api/cloud_templates_controller.rb
+++ b/app/controllers/api/cloud_templates_controller.rb
@@ -8,12 +8,31 @@ module Api
       end
     end
 
+    def import_resource(_type, _id, data = {})
+      params = %w[dst_provider_id src_provider_id src_image_id]
+      raise BadRequestError, "Parameter 'data' has to contain non-empty values for the keys '#{params.join(", ")}', received: '#{data.to_json}'" if data.values_at(*params).any?(&:blank?)
+      raise BadRequestError, "Source and destination provider identifiers must differ" if data['dst_provider_id'] == data['src_provider_id']
+
+      ems_dst   = resource_search(data['dst_provider_id'], :providers, collection_class(:providers))
+      ems_src   = resource_search(data['src_provider_id'], :providers, collection_class(:providers))
+      src_image = resource_search(data['src_image_id'],    :providers, collection_class(:templates))
+
+      raise BadRequestError, "Source image specified by the id '#{data['src_image_id']}' does not belong to the source provider with id '#{ems_src.id}'" if src_image.ems_id != ems_src.id
+
+      task_id = ManageIQ::Providers::CloudManager::Template.import_image_queue(session[:userid], ems_dst, data)
+      msg = "Importing image '#{src_image.name}' from '#{ems_src.name}' into '#{ems_dst.name}'."
+      api_log_info(msg)
+
+      action_result(true, msg, :task_id => task_id)
+    rescue => err
+      action_result(false, err.to_s)
+    end
+
     private
 
     def cloud_template_ident(cloud_template)
       "CloudTemplate id:#{cloud_template.id} name:'#{cloud_template.name}'"
     end
-
 
     def request_compliance_check(cloud_template)
       desc = "#{cloud_template_ident(cloud_template)} check compliance requested"
--- a/config/api.yml
+++ b/config/api.yml
@@ -699,6 +699,8 @@
         :identifier: image_show_list
       - :name: set_ownership
         :identifier: image_ownership
+      - :name: import
+        :identifier: image_import
     :resource_actions:
       :get:
       - :name: read


iv1111:
@agrare here is the corresponding PR for the image-import support

iv1111:
Is it alright to ignore the only codeclimate rubocop suggestion in this particular case?

iv1111:
What is the reason for the build failure? Looking at a log seems that some spec test fails ...

iv1111:
Looking at the spec code, seems that MiqProductFeature doesn't contain the _import_image_. Where else except for _api.yml_ do I have to add the new API declaration?

agrare:
@iv1111 that's a product feature that you have to add to https://github.com/ManageIQ/manageiq/blob/master/db/fixtures/miq_product_features.yml

agrare:
@miq-bot cross-repo-tests manageiq#21112 manageiq-ui

agrare:
@miq-bot cross-repo-tests manageiq#21112, manageiq-ui

agrare:
@iv1111 https://travis-ci.com/github/ManageIQ/manageiq-cross_repo-tests/jobs/492762624#L2497-L2500 looks like a valid spec failure

iv1111:
@agrare Not sure what the reason is, as I have added new feature definition to the specified file.
Adding @h-kataria to the discussion. Do you know what is missing here?

h-kataria:
> @agrare Not sure what the reason is as I have added new feature definition to the specified file.
> Adding @h-kataria to the discussion. Do you know what is missing here?

@iv1111 it is expecting this feature to be present in miq_product_features.yml in core repo

iv1111:
@h-kataria Hi. This PR adds it: https://github.com/ManageIQ/manageiq/pull/21112
@agrare has run a built using that PR and it still fails although I've added it there.
So, what else is missing?

I've added it like this there:

    ```
:name: Import Cloud Template
    :description: Import Cloud Template
    :feature_type: control
    :identifier: import_image
```

**Comment by agrare** [comment]
That is very strange, the spec passes locally with your PR merged but not on cross repo.  Let me try running cross-repo locally and see if that also fails

File: config/api.yml [hunk 1]
```diff
@@ -679,6 +679,8 @@
         :identifier: image_show_list
       - :name: set_ownership
         :identifier: image_ownership
+      - :name: import
+        :identifier: image_import
```
h-kataria, line 683: https://github.com/ManageIQ/manageiq/pull/21112/files#diff-cdcd147fb85ef58d008a4ad0cfa3ddfea649382cb062407e2342aff07362b6cfR6734 has `import_image` not sure which one of these is correct

agrare:
Still failing on cross-repo locally, good catch @h-kataria that looks like it is it

iv1111:
@miq-bot cross-repo-tests manageiq#21112, manageiq-ui

iv1111:
@miq-bot cross-repo-tests manageiq#21112, manageiq-api#1021, manageiq-ui-classic

config/api.yml
iv1111: Thanks you @h-kataria. My mistake, I've swapped the words for some reason. I started now a new cross-repo test to be sure.

agrare:
@abellotti could you take a look?

agrare:
@iv1111 this is going to need a spec test

app/controllers/api/cloud_templates_controller.rb
```diff
@@ -8,6 +8,29 @@ def check_compliance_resource(type, id, _data = nil)
       end
     end
 
+    def import_resource(_type, _id, data = {})
+      params = %w(dst_provider_id src_provider_id src_image_id)
+      raise "Parameter 'data' has to contain non-empty values for the keys '#{params}', received: '#{data}'" if data.values_at(*params).any?(&:blank?)
```
agrare, line 13: I'm guessing these should all be `raise BadRequestError, ""`

agrare: `the keys '#{params}', received: '#{data}'"`

You should `#{params.join(", ")}` so this prints a little nicer

iv1111:
@agrare I've incorporated your suggestions. I also added new keys: cloud object storage id (cos_provider_id) and the bucket id (cos_bucket_id). Their validation is added too. I've tested by adding a sample COS provider and creating a dummy bucket within.

app/controllers/api/cloud_templates_controller.rb
iv1111: done.

app/controllers/api/cloud_templates_controller.rb
```diff
@@ -8,6 +8,35 @@ def check_compliance_resource(type, id, _data = nil)
...
+      raise BadRequestError, "Parameter 'data' has to contain non-empty values for the keys '#{params.join(", ")}', received: '#{data}'" if data.values_at(*params).any?(&:blank?)
+      raise BadRequestError, "Source and destination provider identifiers must differ"  if data['dst_provider_id'] == data['src_provider_id']
+
+      ems_dst = resource_search(data['dst_provider_id'], :providers, collection_class(:providers))
+      raise BadRequestError, "Specified destination provider cannot be found by this id: '#{data['dst_provider_id']}'" if ems_dst.nil?
+
+      ems_src = resource_search(data['src_provider_id'], :providers, collection_class(:providers))
+      raise BadRequestError, "Specified source provider cannot be found by this id: '#{data['src_provider_id']}'" if ems_src.nil?
+
+      src_image = resource_search(data['src_image_id'], :providers, collection_class(:templates))
```
agrare, line 22: @iv1111 We'll need to check that `src_image.supports?(:import_image)` here

app/controllers/api/cloud_templates_controller.rb
```diff
@@ -8,6 +8,35 @@ def check_compliance_resource(type, id, _data = nil)
...
+      raise BadRequestError, "Source image specified by the id '#{data['src_image_id']}' cannot be found in the source provider with id '#{ems_src.id}'" if src_image.nil? || src_image.ems_id != ems_src.id
+
+      cos = resource_search(data['cos_provider_id'], :providers, collection_class(:providers))
```
agrare, line 25: Is the this different from the destination provider?  I thought the ibm cloud object storage manager was the target provider

iv1111: no, it is a transient storage, the final destination is the _ems_dst_.

app/controllers/api/cloud_templates_controller.rb
iv1111: You probably mean _ems_src_. It's the properties of source and destination providers to support import.

Right now I'm not sure how to implement this properly. It has to meet at both ends. Namely, instance denoted by _ems_dst_ has to support import from the instance denoted as _ems_src_. E.g. hypothetical situation - PVS supports import from PVC but not OpenStack, while OpenStack supports import from PVS (i.e. a directed graph).

Also something to note. Support of import can't be defined fixed for all instances of a given class. It depends on whether, say, user has done some preparations on the PowerVC server, prior to caling this API method.

iv1111: If possible, It'd be great if we could postpone the validation of this particular case and create a work item for it not to forget, mentioning it in this PR.

app/controllers/api/cloud_templates_controller.rb
iv1111: Hmm. But I see the problem now. If this API were to be generic it should probably not contain anything about the transient storage as it this step is specific to the PVS <----> PVC scenario. I guess I'll have to validate these parameters in PVS/PVC and not in the API controller.

iv1111:
@agrare I'm not sure now if checking _ems_src_ and _ems_dst_ and throwing an exception is needed. I noticed that _resource_search_ throws exceptions itself. Maybe I should remove my own checks of this type.

app/controllers/api/cloud_templates_controller.rb
iv1111: I think it's more like we have to do something like this:

`ok = dst_provider.supports_import_from?(src_provider_id) && src_provider.supports_export_to?(dst_provider_id)`

This way one can be sure that the _dst_provider_ has done manual or any other preparations (if any) to support the importing from this particular _src_provider_ type and that _src_provider_ is capable of providing the image to _dst_provider_. I guess that doing it this way would also defy the purpose for defining a graph using "allowed_*" .

However, in such case _supports_import_from_ and _supports_export_to_ would have to check for class names and set _supports_ value accordingly, which isn't that nice.

iv1111: Imagine PVS with id = 1 generally supports import from PVC but this particular PVC with id=6  isn't prepared for that (and not even menat to), while PVC instance with id = 8 is prepared. So having a statically defined graph somewhere doesn't help you to determine if the execution can continue.

agrare: If you have to check that a provider is configured to import from/export to a specific other provider then `supports feature` is not the right way to do this.

agrare: Okay well the API isn't the right place to be doing this, we are going to need to look at filtering down what are "eligible" selections when building the dialog dropdowns the API should just take what it is given and pass it along.

agrare:
@iv1111 can you open an issue in https://github.com/ManageIQ/manageiq-providers-ibm_cloud/issues to track all of these related pull requests, and we can discuss the design for the dialogs/validations there?

agrare (approved):

agrare:
Take out of WIP when you're ready

iv1111:
@agrare I've removed WIP and squashed my commits into one.
Referencing the parent issue here: [180](https://github.com/ManageIQ/manageiq-providers-ibm_cloud/issues/180)

agrare (approved):

PR #1028, opened 2021-04-16
Add edit cloud volume function and route

kavyanekkalapu:
Issue: https://github.com/ManageIQ/manageiq-ui-classic/issues/7712

**After:**

Screen Shot 2021-04-16 at 12 53 13 PM
Screen Shot 2021-04-16 at 12 53 41 PM

UI PR:https://github.com/ManageIQ/manageiq-ui-classic/pull/7714

@miq-bot assign @Fryguy 
@miq-bot add-reviewer @Fryguy 
@miq-bot add-label  lasker/yes?

### DIFF
--- a/app/controllers/api/cloud_volumes_controller.rb
+++ b/app/controllers/api/cloud_volumes_controller.rb
@@ -18,15 +18,17 @@ module Api
 
     def edit_resource(type, id, data = {})
       raise BadRequestError, "Must specify an id for editing a #{type} resource" unless id
+
       cloud_volume = resource_search(id, type, collection_class(:cloud_volumes))
+
       raise BadRequestError, cloud_volume.unsupported_reason(:update) unless cloud_volume.supports?(:update)
-  
+
       task_id = cloud_volume.update_volume_queue(User.current_user, data)
       action_result(true, "Updating #{cloud_volume.name}", :task_id => task_id)
     rescue => err
       action_result(false, err.to_s)
     end
-  
+
     def safe_delete_resource(type, id, _data = {})
       delete_action_handler do
         cloud_volume = resource_search(id, type, collection_class(:cloud_volumes))
--- a/app/controllers/api/cloud_volumes_controller.rb
+++ b/app/controllers/api/cloud_volumes_controller.rb
@@ -19,6 +19,7 @@ module Api
     def edit_resource(type, id, data = {})
       raise BadRequestError, "Must specify an id for editing a #{type} resource" unless id
       cloud_volume = resource_search(id, type, collection_class(:cloud_volumes))
+      raise BadRequestError, cloud_volume.unsupported_reason(:update) unless cloud_volume.supports?(:update)
   
       task_id = cloud_volume.update_volume_queue(User.current_user, data)
       action_result(true, "Updating #{cloud_volume.name}", :task_id => task_id)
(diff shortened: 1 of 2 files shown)


Fryguy:
LGTM - @agrare Please also review.

app/controllers/api/cloud_volumes_controller.rb
```diff
@@ -16,6 +16,16 @@ def create_resource(_type, _id = nil, data = {})
       action_result(false, err.to_s)
     end
 
+    def edit_resource(type, id, data = {})
+      raise BadRequestError, "Must specify an id for editing a #{type} resource" unless id
+      cloud_volume = resource_search(id, type, collection_class(:cloud_volumes))
+
```
agrare, line 22: We should have a supports_feature check that this cloud_volume supports being updated since out of the 7 providers which collect cloud_volumes only 3 support updating them.

It looks like the old UI probably used `CloudVolume#validate_update_volume` (and maybe [[still]](https://github.com/ManageIQ/manageiq-ui-classic/blob/master/app/controllers/cloud_volume_controller.rb#L293) [[does]](https://github.com/ManageIQ/manageiq-ui-classic/blob/master/app/controllers/cloud_volume_controller.rb#L310))

kavyanekkalapu: @agrare Could you share a sample from api code where we are using similar check in any of the api calls, im not really sure how to do it, thanks.

agrare: Of course, just look for anything that does `.supports?(` for example https://github.com/ManageIQ/manageiq-api/blob/master/app/controllers/api/network_routers_controller.rb#L7

agrare: @kavyanekkalapu https://github.com/ManageIQ/manageiq/pull/21181 adds the supports feature for you to use here

kavyanekkalapu: @agrare thanks i updated it by looking at safe_delete func, please check and let me know if that is not sufficient.

chessbyte:
I don't think this can be backported to Lasker as is since https://github.com/ManageIQ/manageiq/pull/21046 is marked as `lasker/no`.

kavyanekkalapu:
> I don't think this can be backported to Lasker as is since [ManageIQ/manageiq#21046](https://github.com/ManageIQ/manageiq/pull/21046) is marked as `lasker/no`.

@chessbyte Interesting, is it ok to move that to lasker? otherwise edit cloud volumes page would be broken.

agrare:
> I don't think this can be backported to Lasker as is since ManageIQ/manageiq#21046 is marked as lasker/no.

Would like @Fryguy's opinion to make sure, but I think it is actually okay because `supports` and `supports_not` already existed in the original `SupportsFeatureMixin`, and the `:update` feature existed in the `QUERYABLE_FEATURES` hash.

agrare:
@kavyanekkalapu can you fix the rubocop issues (except for `Method parameter must be at least 3 characters long.` since `id` is a valid variable name in this instance)

agrare (approved):
:+1: LGTM, I'll hold off on merging until we have a consensus on backport to lasker

Fryguy:
Backported to `lasker` in commit ee805486c657867a5387902bdb96fc02df5cde48.

```text
commit ee805486c657867a5387902bdb96fc02df5cde48
Date:   Mon Apr 19 16:40:16 2021 -0400

    Merge pull request #1028 from kavyanekkalapu/fix_edit_cloudvolume
    
    Add edit cloud volume function and route
    
    (cherry picked from commit abf5c3b13f3037d630df86c89df8d12e09292884)
```

PR #1029, opened 2021-04-19
Allow `id` as a method parameter name

agrare:
Most API method signatures look like `def api_action(type, id, data = {})` and rubocop warns that `id` is too short to be a helpful variable name.

### DIFF
--- a/.rubocop_local.yml
+++ b/.rubocop_local.yml
@@ -1,3 +1,6 @@
+Naming/MethodParameterName:
+  AllowedNames:
+  - id
 Style/GlobalVars:
   AllowedVariables:
   - $api_log


agrare:
Example: https://github.com/ManageIQ/manageiq-api/pull/1028#issuecomment-822710425
```
Checked commits kavyanekkalapu/manageiq-api@7529cd3~...ca9decc with ruby 2.6.3, rubocop 0.82.0, haml-lint 0.35.0, and yamllint
1 file checked, 4 offenses detected

app/controllers/api/cloud_volumes_controller.rb

exclamation - Line 19, Col 29 - Naming/MethodParameterName - Method parameter must be at least 3 characters long.
```

abellotti (approved):

PR #1035, opened 2021-05-11
[ZonesController] Allow "setting" settings

NickLaMuro:
The Zone model/api-controller has two overriding features:

- A subcollection for "settings"
- A hash column called "settings"

The code in `Api::BaseController::Generic#add_resource` has a block in it that takes any key from the params hash that is part of a subcollection, and passes it to create those sub-resources.  This overrides the behavior of being able to write the `settings` hash directly on create (this is not a behavior that is replicated generically for `edit_resource`)

To allow for editing of `Zone#settings` on create, we override `add_resource` so that it will not include this sections of code for this particular controller.

Note:  The `alias` also needed to be re-applied, as `create_resource` is what is actually called by `Api::BaseController::Generic#create`

### DIFF
--- a/app/controllers/api/base_controller/generic.rb
+++ b/app/controllers/api/base_controller/generic.rb
@@ -13,16 +13,11 @@ module Api
       def add_resource(type, _id, data)
         assert_id_not_specified(data, "#{type} resource")
         klass = collection_class(type)
-        subcollection_data = collection_config.subcollections(type).each_with_object({}) do |sc, hash|
-          if data.key?(sc.to_s)
-            hash[sc] = data[sc.to_s]
-            data.delete(sc.to_s)
-          end
-        end
+        extract_subcollection_data!(collection_config.subcollections(type), data)
         validate_type(klass, data['type']) if data['type']
         resource = klass.new(data)
         if resource.save
-          add_subcollection_data_to_resource(resource, type, subcollection_data)
+          add_subcollection_data_to_resource(resource, type)
           resource
         else
           raise BadRequestError, "Failed to add a new #{type} resource - #{resource.errors.full_messages.join(', ')}"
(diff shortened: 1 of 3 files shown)


kavyanekkalapu (approved):

kavyanekkalapu:
@bdunne We need this for zones from react conversion, Could you take a look at this pr for review?

NickLaMuro:
@Fryguy I like that idea of minimizing the impact, though, we would then have to override two methods instead of just one.  Both that new extracted method that you identified, as well as `add_subcollection_data_to_resource`.  I think that is fine, it just isn't as clean as your comment might suggest.

Do we want to do this as a HACK for now, and do a follow up to clean things up and attempt the suggested?

Fryguy:
I'm not sure we need to override `add_subcollection_data_to_resource`, because that method only works on the "extracted" subcollections...if we override the new method to no longer have settings in the "extracted" subcollections, then it won't be considered in `add_subcollection_data_to_resource`

NickLaMuro:
@Fryguy oh, good catch.  Yeah, in that respect, then I think it makes sense to move forward.

I will look at doing that then.

app/controllers/api/zones_controller.rb
```diff
@@ -30,5 +30,16 @@ def data_includes_invalid_attrs(data)
 
       data.keys.select { |key| INVALID_ZONES_ATTRS.include?(key) }.compact.join(", ")
     end
+
+    # Since there is a `settings` attribute on Zone, and a subcollection for
+    # zones called `settings`, allow "setting" of the `:settings` column, and
+    # don't allow setting the "settings" subcollection data on this particular
+    # resource.
```
NickLaMuro, line 37: > 🤣 I challenge you to get more "settings" in there 😂

Challenge accepted!

accepted!

app/controllers/api/base_controller/generic.rb
```diff
@@ -146,14 +141,23 @@ def refresh_dialog_fields_action(dialog, refresh_fields, resource_ident)
 
       private
 
+      def collect_subcollection_data(subcollections, data)
```
Fryguy, line 144: Since this modifies the data, I think we should name this something like `extract_subcollect_data!`

NickLaMuro: I'm on board with that.  Will change.

PR #1036, opened 2021-05-11
Specify a hardcoded feature_type when dynamically creating features

jrafanie:
We create product features on demand for various tests instead of seeding the correct features.
We shouldn't hardcode a feature_type here but since it's a UI concern and it's now required, it shouldn't affect the API.
See also: https://github.com/ManageIQ/manageiq/pull/21207

### DIFF
--- a/spec/support/api/helpers.rb
+++ b/spec/support/api/helpers.rb
@@ -22,7 +22,14 @@ module Spec
         def api_basic_authorize(*identifiers, user: @user.userid, password: @user.password)
           if identifiers.present?
             identifiers.flatten.collect do |identifier|
-              @role.miq_product_features << MiqProductFeature.find_or_create_by(:identifier => identifier) if identifier
+              if identifier
+                # TODO:  We create product features on demand for various tests instead of seeding the correct features.
+                # We shouldn't hardcode a feature_type here but since it's a UI concern and it's now required, it shouldn't affect the API.
+                # See also: https://github.com/ManageIQ/manageiq/pull/21207
+                feature = MiqProductFeature.find_by(:identifier => identifier)
+                feature ||= MiqProductFeature.create(:identifier => identifier, :feature_type => "view")
+                @role.miq_product_features << feature
+              end
             end
             @role.save
 


spec/support/api/helpers.rb
```diff
@@ -22,7 +22,14 @@ def init_api_spec_env
...
             identifiers.flatten.collect do |identifier|
-              @role.miq_product_features << MiqProductFeature.find_or_create_by(:identifier => identifier) if identifier
+              if identifier
+                # TODO:  We create product features on demand for various tests instead of seeding the correct features.
+                # We shouldn't hardcode a feature_type here but since it's a UI concern and it's now required, it shouldn't affect the API.
+                # See also: https://github.com/ManageIQ/manageiq/pull/21207
+                feature = MiqProductFeature.find_by(:identifier => identifier)
+                feature ||= MiqProductFeature.create(:identifier => identifier, :feature_type => "view")
+                @role.miq_product_features << feature
+              end
```
Fryguy, line 32: Minor, but we can do this with Rails directly

```suggestion
              if identifier
                # TODO:  We create product features on demand for various tests instead of seeding the correct features.
                # We shouldn't hardcode a feature_type here but since it's a UI concern and it's now required, it shouldn't affect the API.
                # See also: https://github.com/ManageIQ/manageiq/pull/21207
                @role.miq_product_features = MiqProductFeature.create_with(:feature_type => "view").find_or_create_by(:identifier => identifier)
              end
```

Fryguy: Even so, I'm going to merge as is

Fryguy:
Backported to `lasker` in commit 2a6712da6118482492df33ff71d9f3cad1a6e4c4.

```text
commit 2a6712da6118482492df33ff71d9f3cad1a6e4c4
Date:   Wed May 12 11:46:02 2021 -0400

    Merge pull request #1036 from jrafanie/specify_feature_type_when_dynmamically_creating_features
    
    Specify a hardcoded feature_type when dynamically creating features
    
    (cherry picked from commit b00a7572ce08caa0661ad43ace70058c8e776ffa)
```

PR #1037, opened 2021-05-12
Use the seed_specific_features method to accurately create features

Fryguy:
Follow up to #1036 and #1038

- [x] Depends on https://github.com/ManageIQ/manageiq/pull/21217

@jrafanie Please review.

### DIFF
--- a/config/api.yml
+++ b/config/api.yml
@@ -3292,7 +3292,7 @@
         - :klass: VmReconfigureRequest
           :identifier: vm_reconfigure
         - :klass: VmCloudReconfigureRequest
-          :identifier: vm_cloud_reconfigure
+          :identifier: instance_resize
         - :klass: VmMigrateRequest
           :identifier: vm_migrate
         - :klass: ServiceTemplateProvisionRequest
--- a/spec/requests/authentication_spec.rb
+++ b/spec/requests/authentication_spec.rb
@@ -444,7 +444,7 @@ describe "Authentication API" do
 
     context "actions with single role identifier" do
       it "are rejected when user is not authorized with the single role identifier" do
-        stub_api_action_role(:vms, :collection_actions, :get, :read, "vm_view_role1")
+        stub_api_action_role(:vms, :collection_actions, :get, :read, "vm_show_list")
         api_basic_authorize
 
         get api_vms_url
@@ -453,8 +453,8 @@ describe "Authentication API" do
       end
 
       it "are accepted when user is authorized with the single role identifier" do
-        stub_api_action_role(:vms, :collection_actions, :get, :read, "vm_view_role1")
-        api_basic_authorize "vm_view_role1"
+        stub_api_action_role(:vms, :collection_actions, :get, :read, "vm_show_list")
+        api_basic_authorize "vm_show_list"
 
         get api_vms_url
 
@@ -464,7 +464,7 @@ describe "Authentication API" do
 
     context "actions with multiple role identifiers" do
       it "are rejected when user is not authorized with any of the role identifiers" do
-        stub_api_action_role(:vms, :collection_actions, :get, :read, %w(vm_view_role1 vm_view_role2))
+        stub_api_action_role(:vms, :collection_actions, :get, :read, %w[vm_show_list sui_vm_details_view])
         api_basic_authorize
 
         get api_vms_url
@@ -473,8 +473,8 @@ describe "Authentication API" do
       end
 
       it "are accepted when user is authorized with at least one of the role identifiers" do
-        stub_api_action_role(:vms, :collection_actions, :get, :read, %w(vm_view_role1 vm_view_role2))
-        api_basic_authorize "vm_view_role2"
+        stub_api_action_role(:vms, :collection_actions, :get, :read, %w[vm_show_list sui_vm_details_view])
+        api_basic_authorize "sui_vm_details_view"
 
         get api_vms_url
 
--- a/spec/support/api/helpers.rb
+++ b/spec/support/api/helpers.rb
@@ -20,20 +20,11 @@ module Spec
         end
 
         def api_basic_authorize(*identifiers, user: @user.userid, password: @user.password)
+          identifiers = identifiers.flatten.compact.map(&:to_s)
           if identifiers.present?
-            identifiers.flatten.collect do |identifier|
-              if identifier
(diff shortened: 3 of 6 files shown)


spec/support/api/helpers.rb
```diff
@@ -21,19 +21,11 @@ def init_api_spec_env
...
-                # See also: https://github.com/ManageIQ/manageiq/pull/21207
-                feature = MiqProductFeature.find_by(:identifier => identifier)
-                feature ||= MiqProductFeature.create(:identifier => identifier, :feature_type => "view")
-                @role.miq_product_features << feature
-              end
-            end
-            @role.save
-
-            MiqProductFeature.seed_tenant_miq_product_features if identifiers & MiqProductFeature::TENANT_FEATURE_ROOT_IDENTIFIERS == identifiers
+            identifiers = identifiers.flatten
```
Fryguy, line 24: Not sure this is needed, but it was in the original code, so I kept it.

spec/support/api/helpers.rb
```diff
@@ -21,19 +21,11 @@ def init_api_spec_env
...
+
+            EvmSpecHelper.seed_specific_product_features(*identifiers)
```
jrafanie, line 26: It's too bad that this doesn't return the features it seeded.

I like this solution much better.  👍

spec/support/api/helpers.rb
Fryguy: Ok, it's needed - seems the helper methods that return a set of features from the api.yml pass them as an array, and thus `*identifiers` becomes an array of arrays.

spec/support/api/helpers.rb
Fryguy: Yeah I agree...maybe we could move towards that.

config/api.yml
```diff
@@ -3292,7 +3292,7 @@
         - :klass: VmReconfigureRequest
           :identifier: vm_reconfigure
         - :klass: VmCloudReconfigureRequest
-          :identifier: vm_cloud_reconfigure
+          :identifier: instance_resize
```
Fryguy, line 3295: @jrafanie @agrare I have no idea if this is the right value here, but I'm not sure what to put - do you know?  (`vm_cloud_reconfigure` doesn't exist as a feature)

Fryguy:
Leaving the rest of the rubocops for in-file consistency

config/api.yml
jrafanie: in DM, was suggesting the same one:

```
      - :name: Reconfigure Instance
        :description: Reconfigure Instance
        :feature_type: control
        :identifier: instance_resize
```

jrafanie: for whatever reason, we're not consistent... we call it Reconfigure Instance but the identifier is instance_resize... even though it's service_reconfigure, vm_reconfigure, etc. for everything else.

agrare: It is called `instance_resize` because the underlying method that is being called is `ManageIQ::Providers::CloudManager::Vm#resize`

I'm guessing they wanted something different from reconfigure because what you are able to change on a cloud instance is far less than you can change with a vm_reconfigure call on the infra side.

As for why the display name doesn't match, got me

spec/requests/authentication_spec.rb
```diff
@@ -453,8 +453,8 @@ def systoken(server_guid, userid, timestamp)
       end
 
       it "are accepted when user is authorized with the single role identifier" do
-        stub_api_action_role(:vms, :collection_actions, :get, :read, "vm_view_role1")
-        api_basic_authorize "vm_view_role1"
```
jrafanie, line 457: 🤦

spec/support/api/helpers.rb
```diff
@@ -19,21 +19,15 @@ def init_api_spec_env
                                       :miq_groups => [@group])
         end
 
-        def api_basic_authorize(*identifiers, user: @user.userid, password: @user.password)
+        def api_basic_authorize(*identifiers, user: @user.userid, password: @user.password, extra_identifiers: nil)
+          identifiers = identifiers.flatten.compact.map(&:to_s)
+          features    = identifiers + Array(extra_identifiers)
```
jrafanie, line 24: Why can't you pass the `extra_identifiers` in the array of identifiers?  I'm not sure why we need to pass them separately.

Fryguy: yeah, I couldn't find a way around this...it's basically "I need these identifiers to exist in the database, but I don't want to assign them to the user's role".

This is really only needed for the 2 tests that test rbac tenant features which are dynamic features.  For example,

```ruby
        api_basic_authorize "rbac_tenant_manage_quotas_tenant_#{tenant_omega.id}", :extra_identifiers => "rbac_tenant_manage_quotas"
```

So, `rbac_tenant_manage_quotas_tenant_1` is dynamically created based on the tenants (in this case, tenant `1`), _BUT_ it's only dynamically created if `rbac_tenant_manage_quotas` feature exists in the database.  For the purpose of the test, I need the `rbac_tenant_manage_quotas` to be in the database, but I don't want the user to have that specific role otherwise they can see all quotas from all tenants.

---

Now that I wrote this 📖 , I realize this is only in 2 specific tests, so I'm considering just seeding those specific features right in that test.

Fryguy: Ok, I dropped this in favor of putting it right in the tenant_quotas_spec, and in doing so, that spec got a _lot_ simpler, because it was introducing unnecessary redundancies

Fryguy:
@jrafanie I reworked to remove the "extra_identifiers", and while it's much cleaner and I prefer it, I now depend on https://github.com/ManageIQ/manageiq/pull/21217 in core.

jrafanie (approved):

jrafanie:
This depends on code added in https://github.com/ManageIQ/manageiq-api/pull/1036 so that will have to be backported to lasker first.

Fryguy:
Backported to `lasker` in commit 94cac65f15ef1126f629604d704f3a62ed7e1608.

```text
commit 94cac65f15ef1126f629604d704f3a62ed7e1608
Date:   Fri May 14 11:58:53 2021 -0400

    Merge pull request #1037 from Fryguy/use_feature_seeder
    
    Use the seed_specific_features method to accurately create features
    
    (cherry picked from commit 9d82210b3bd6d4f1453de4910bd8dc807f502817)
```

PR #1039, opened 2021-05-12
[BaseController::Render] Fix subquery_res.count

NickLaMuro:
Very simply, the API can be provided with a filter + attribute params that can be reduced down to something like this:

```
https://localhost:3000/api/vms?expand=resources&filter[]=ems_id!=null&attributes=name,provisioned_storage
```

In the `bin/rails c`, it can be simplified to:

```ruby
irb> Vm.select(Vm.arel_table[Arel.star], :provisioned_storage).count
```

## Nitty Gritty Details

(that LJ is too lazy to read...)

This, in Rails 6, does not seem to work, as a `.join` is done to the `Vm.arel_table[Arel.star]` and the `:provisioned_storage` columns deep within the `ActiveRecord` library, and ultimately the `Vm.arel_table...` portion will be converted into a `struct` when passed as a SQL string, and not a column name or function, looking something like this:

```
SELECT COUNT(#<struct Arel::Attributes::Attribute relation=#...
```

This obviously isn't valid SQL, and while it might make sense to fix this in Rails (as it seems to be a regression), the solution isn't terribly obvious, and this is a simple fix on our end.

* * *

Side note:  The following also doesn't seem to be valid:

```
irb> Vm.select(:id, :provisioned_storage).count
(0.5ms)  SELECT COUNT(vendor, id) FROM "vms" WHERE ...
PG::UndefinedFunction: ERROR:  function count(character varying, bigint) does not exist
LINE 1: SELECT COUNT(vendor, id) FROM "vms" WHERE "vms"."type" IN ($...
               ^
HINT:  No function matches the given name and argument types. You might need to add explicit type casts.
```

So it seems like the logic here is possibly flawed is Rails 6+, but what lead to this isn't clear.

### DIFF
--- a/app/controllers/api/base_controller/renderer.rb
+++ b/app/controllers/api/base_controller/renderer.rb
@@ -193,7 +193,7 @@ module Api
 
       def filter_results(miq_expression, res, options)
         if miq_expression.present? && options.key?(:limit) && options.key?(:offset)
-          subquery_res = Rbac.filtered(res, options.except(:offset, :limit))
+          subquery_res = Rbac.filtered(res, options.except(:offset, :limit, :extra_cols))
           [Rbac.filtered(res, options), subquery_res.count]
         else
           [Rbac.filtered(res, options)]
--- a/spec/requests/vms_spec.rb
+++ b/spec/requests/vms_spec.rb
@@ -1856,6 +1856,17 @@ describe "Vms API" do
       expect_result_resources_to_include_hrefs("resources", [api_vm_url(nil, vm2)])
     end
 
+    it "handles counts properly with virtual_attributes" do
+      api_basic_authorize action_identifier(:vms, :read, :resource_actions, :get)
+      get api_vms_url, :params => {
+        :expand     => "resources",
+        :filter     => ["ems_id!=null"],
+        :attributes => "name,provisioned_storage"
+      }
+
+      expect(response.parsed_body["subcount"]).to eq 1
+    end
+
     it "assigns a tag to a Vm without appropriate role" do
       api_basic_authorize
 


NickLaMuro:
@miq-bot assign @jrafanie 

Would you be willing to look at this since you correctly identified this probably is a Rails 6.0 issue?

cc @Fryguy @kbrock

NickLaMuro:
@miq-bot add_labels bug, jansa/no, kasparov/yes?

NickLaMuro:
Threw together a reproducer, and it seems like it might have been a change to the VM model, as I can reproduce in both Rails 5.2 AND Rails 6.0:

```ruby
require "bundler/inline"

rails_version = ENV["RAILS_VERSION"] || '6.0'
rails_db      = ENV["RAILS_DB"]      || 'sqlite3'

gemfile(true) do
  source "https://rubygems.org"

  git_source(:github) { |repo| "https://github.com/#{repo}.git" }

  gem "activerecord",  "~> #{rails_version}"
  gem "activemodel",   "~> #{rails_version}"
  gem "activesupport", "~> #{rails_version}"

  if rails_db == "sqlite3"
    gem "sqlite3"
  else
    gem "pg"
  end
end

require "active_record"

if rails_db == "sqlite3"
  ActiveRecord::Base.establish_connection(adapter: "sqlite3", database: ":memory:")
else
  db_config = {:adapter  => "postgresql", :username => "nicklamuro", :database => "rails_count_testing"}
  pg_config = db_config.merge(:database => "postgres", :schema_search_path => "public")

  ActiveRecord::Base.establish_connection pg_config
  ActiveRecord::Base.connection.recreate_database(db_config[:database])
  ActiveRecord::Base.establish_connection db_config
end


ActiveRecord::Schema.define do
  create_table :users, :force => true do |t|
    t.string :username
  end
end

class User < ActiveRecord::Base
end
                                                                  
puts User.select(User.arel_table[Arel.star], :username).count

# Usage:
#
# $ ruby script.rb
# $ RAILS_VERSION=5.2 ruby script.rb
# $ RAILS_VERSION=5.2 RAILS_DB=pg ruby script.rb
```

So, while I think this is a bug in Rails, it doesn't seem to be a regression, and so something might have changed on the `Vm` model side or `Rbac` side to have caused this.

I don't think the defensive fix here is a bad solution, but it might not hurt to also determine the root cause at some point.

* * *

Regardless, looks like I have some tests to fix...

NickLaMuro:
So, probably similar failures in travis, but this is what I wrote for a test to cover this case:

```diff
diff --git a/spec/requests/vms_spec.rb b/spec/requests/vms_spec.rb
index e2cef0bc..c55bdf16 100644
--- a/spec/requests/vms_spec.rb
+++ b/spec/requests/vms_spec.rb
@@ -1856,6 +1856,18 @@ describe "Vms API" do
       expect_result_resources_to_include_hrefs("resources", [api_vm_url(nil, vm2)])
     end
 
+    it "handles counts properly with virtual_attributes" do
+      api_basic_authorize action_identifier(:vms, :read, :resource_actions, :get)
+      get api_vms_url, :params => {
+        :expand => "resources",
+        :filter => ["ems_id!=null"],
+        :attributes => "name,provisioned_storage"
+      }
+
+      puts response.parsed_body
+      expect(response.parsed_body["subcount"]).to eq 1
+    end
+
     it "assigns a tag to a Vm without appropriate role" do
       api_basic_authorize
```

While this covers the current issue, when applying the current fix it will break with a different error.  I don't think that `Arel.star` is going to work here, but I think just doing `except(:select)` will work.  Will try that an see if it works.

NickLaMuro:
I think I have settled on a solution, though it isn't the one originally proposed above.

Since `Rbac.filtered` can sometime return an `Array` instead of a relation, it is safer to avoid a `.select` from being created instead of having to adjust `subquery_res` _**only if**_ it is an `ActiveRecord::Relation`.

jrafanie:
@NickLaMuro can you include the original "outside in" API request that worked in prior versions but doesn't work now?  I believe it was `ems_id!=null ` but it's unclear what minimal example API requests fail and would be fixed by this change.

NickLaMuro:
@jrafanie The test that I added is just that, the minimal I was able to replicate the issue with.

```ruby
get api_vms_url, :params => {
  :expand => "resources",
  :filter => ["ems_id!=null"],
  :attributes => "name,provisioned_storage"
}
```

Which translates to:

```
https://localhost:3000/api/vms?expand=resources&filter[]=ems_id!=null&attributes=name,provisioned_storage
```

jrafanie:
> @jrafanie The test that I added is just that, the minimal I was able to replicate the issue with.

Thanks!  Can you put that in the description?  It's currently very heavy on the Arel details and I won't know what it is in 2 hours when I forget this discussion.  😉

app/controllers/api/base_controller/renderer.rb
```diff
@@ -193,7 +193,7 @@ def collection_filterer(res, type, klass, is_subcollection = false)
 
       def filter_results(miq_expression, res, options)
         if miq_expression.present? && options.key?(:limit) && options.key?(:offset)
-          subquery_res = Rbac.filtered(res, options.except(:offset, :limit))
+          subquery_res = Rbac.filtered(res, options.except(:offset, :limit, :extra_cols))
           [Rbac.filtered(res, options), subquery_res.count]
```
kbrock, line 197: would this work?

```suggestion
          [Rbac.filtered(res, options), subquery_res.count(:all)]
```

kbrock (commented):
This has crept up over time. I seem to remember fully qualifying the count (i.e.: `count(:all)`) as a good solution.

For some reason I would prefer not ignoring the `extra_cols` but if that doesn't solve it, maybe this is just as good

app/controllers/api/base_controller/renderer.rb
NickLaMuro: @kbrock Not unless this would work:

```ruby
irb> [1,2,3].count(:all)
#=> 0
```

NickLaMuro: Resolving since this will not be something we are going to use.

Fryguy:
@NickLaMuro You sent this back to kasparov, but should it also then go to lasker?  (seems it also affects master, so I think so)

Fryguy (approved):

kbrock:
bringing in a side conversation: would be nice to patch rails, either at `count` or `activerecord/relation/calculations.rb` and ignore `includes` (without references) and ignore `:select_values` for counting.

It has bitten us in a number of places. the solution has been sql only (i.e. use `count(:all)` ) but since that is not ruby friendly, change to be ruby friendly would be best.

jrafanie:
> (that LJ is too lazy to read...)

🤣 

It's easy to write it from the inside out but most of us will forget the original reported problem if it's not stated up front.  I hope you remind me to do the same.

jrafanie (approved):
LGTM

jrafanie:
> @NickLaMuro You sent this back to kasparov, but should it also then go to lasker? (seems it also affects master, so I think so)

Agreed.  Lasker and master are rails.  Kasparov and older are rails 5.2.  Shouldn't this only be for lasker/master?

jrafanie:
> Agreed. Lasker and master are rails 6. Kasparov and older are rails 5.2. Shouldn't this only be for lasker/master?

I'll drop kasparov and add lasker for now so I can merge this confidently... feel free to comment on the choice of backports @NickLaMuro if I am mistaken.  Thanks!

NickLaMuro:
`@miq-bot add_label laskar-yes?`
`@miq-bot remove_label kasparov-yes?`

> You sent this back to kasparov, but should it also then go to lasker? (seems it also affects master, so I think so)

Woops, looks I can't _**alphabet**_, and jumped back one too many versions.

Rails 6 was merged in `laskar`, so I think that is all we need to bring this back to.  Also, this just got merged into `master`, so I think we are covered on that front... and looks like @jrafanie didn't give me a chance to get back to him before updating the labels, so I guess I won't update them again...

NickLaMuro:
> I'll drop kasparov and add lasker for now so I can merge this confidently... feel free to comment on the choice of backports @NickLaMuro if I am mistaken. Thanks!

@jrafanie Yes, this was correct, and you correctly guess my intention.  Thanks!

Fryguy:
Backported to `lasker` in commit 24fdffe0aa3bbce34f59247b0e2f186a8a8a9044.

```text
commit 24fdffe0aa3bbce34f59247b0e2f186a8a8a9044
Date:   Fri May 14 12:37:30 2021 -0400

    Merge pull request #1039 from NickLaMuro/fix_subquery_counts
    
    [BaseController::Render] Fix subquery_res.count
    
    (cherry picked from commit 10779fadb0f296a2d88249db0fae15f4f78b3c01)
```

PR #1038, opened 2021-05-12
Fix Provider create product feature

agrare:
`provider_foreman_add_provider` was renamed to `ems_configuration_add_provider` in https://github.com/ManageIQ/manageiq/pull/19949

PR #1040, opened 2021-05-13
Fix spec that verifies features to actually see all features

Fryguy:
@jrafanie Please review

Part of https://github.com/ManageIQ/manageiq/issues/21216

~~Built on #1037 so it will go green.~~ When this spec is run on master (prior to #1038), I get

```
  1) API configuration (config/api.yml) collections identifiers contains only valid miq_feature identifiers
     Failure/Error: expect(api_feature_identifiers - MiqProductFeature.pluck(:identifier)).to be_empty
       expected `#<Set: {"provider_foreman_add_provider", "vm_cloud_reconfigure"}>.empty?` to be truthy, got false
     # ./spec/lib/api/api_config_spec.rb:46:in `block (4 levels) in <top (required)>'
     # /Users/jfrey/.gem/ruby/2.7.2/gems/webmock-3.12.2/lib/webmock/rspec.rb:37:in `block (2 levels) in <top (required)>'
```

which is what I'd expect.

~~WIP because~~
- [x] ~~I want to add some more checks in this same spec file that ensures that collection actions are built properly~~
- [x] I'm not going to do this one...the test makes no sense. ~~I don't understand the SUI test and removed it for now, but I _think_ it really just wants to check that anytime an sui feature is added there is also a non-sui feature added.  However the spec as is written now doesn't do that and is just checking that Ruby .select is a subset, which is always is.~~

### DIFF
--- a/spec/lib/api/api_config_spec.rb
+++ b/spec/lib/api/api_config_spec.rb
@@ -22,60 +22,31 @@ describe 'API configuration (config/api.yml)' do
 
     describe 'identifiers' do
       let(:api_feature_identifiers) do
-        feature_identifiers { |set, id| set.add(id) }
-      end
+        collection_settings.each_with_object(Set.new) do |(_key, config), ids|
+          action_configs  = config.to_h.select { |k, v| k.to_s.end_with?("_actions") }.values
+          action_configs += Array(config.resource_entities).map { |e| Array(e.entity_actions) }
 
-      let(:sui_product_features) do
-        feature_identifiers { |set, id| set.add(id) if /^sui_/ =~ id }
+          action_configs.each do |action_config|
+            action_config.each do |_verb, actions|
+              actions.each do |action|
+                ids.merge(Array(action[:identifier]))
+                ids.merge(Array(action[:identifiers]).flat_map { |i| Array(i[:identifier]) })
+              end
+            end
+          end
+
+          ids.merge(Array(config[:identifier]))
+        end
       end
 
       it 'is not empty' do
         expect(api_feature_identifiers).not_to be_empty
       end
 
-      it 'contains only valid miq_feature identifiers' do
+      it 'contains only valid MiqProductFeature identifiers' do
         MiqProductFeature.seed_features
-        dangling = Array(api_feature_identifiers).reject { |feature| MiqProductFeature.feature_exists?(feature) }
-        expect(dangling).to be_empty
-      end
-
-      it 'contains valid sui specific miq_feature identifiers' do
-        expect(sui_product_features.subset?(api_feature_identifiers))
-      end
-
-      def feature_identifiers
-        collection_settings.each_with_object(Set.new) do |(_, cfg), set|
-          Array(cfg[:identifier]).each { |id| set.add(id) }
-          keys = %i(collection_actions resource_actions subcollection_actions subresource_actions)
-          Array(cfg[:subcollections]).each do |s|
-            keys << "#{s}_subcollection_actions" << "#{s}_subresource_actions"
-          end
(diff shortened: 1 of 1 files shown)


spec/lib/api/api_config_spec.rb
```diff
@@ -22,11 +22,19 @@
...
+            config[action_type].each do |_verb, actions|
+              actions.map do |action|
+                ids.merge(Array(action[:identifier]))
+                ids.merge(Array(action[:identifiers]).map { |i| i[:identifier] }.compact)
+              end
+            end
+          end
 
-      let(:sui_product_features) do
-        feature_identifiers { |set, id| set.add(id) if /^sui_/ =~ id }
```
Fryguy, line 29: This is basically just taking a subset of the features from the full features list, but then later it checks:

```ruby
expect(sui_product_features.subset?(api_feature_identifiers))
```

which is _always_ going to be true.  This test makes no sense.

spec/lib/api/api_config_spec.rb
```diff
@@ -22,11 +22,19 @@
...
       let(:api_feature_identifiers) do
-        feature_identifiers { |set, id| set.add(id) }
-      end
+        collection_settings.each_with_object(Set.new) do |(_key, config), ids|
+          action_types = config.keys.select { |k| k.to_s.end_with?("_actions") }
+          action_identifiers = action_types.each do |action_type|
+            config[action_type].each do |_verb, actions|
+              actions.map do |action|
+                ids.merge(Array(action[:identifier]))
+                ids.merge(Array(action[:identifiers]).map { |i| i[:identifier] }.compact)
```
Fryguy, line 31: This effectively fixes the bug in the test...it never checked `:identifiers` (plural) before.

spec/lib/api/api_config_spec.rb
```diff
@@ -22,11 +22,21 @@
...
-      end
+        collection_settings.each_with_object(Set.new) do |(_key, config), ids|
+          action_configs  = config.to_h.select { |k, v| k.to_s.end_with?("_actions") }.values
+          action_configs += Array(config.resource_entities).map { |e| Array(e.entity_actions) }
+
+          action_configs.each do |action_config|
+            action_config.each do |_verb, actions|
+              actions.each do |action|
+                ids.merge(Array(action[:identifier]))
+                ids.merge(Array(action[:identifiers]).flat_map { |i| Array(i[:identifier]) })
```
Fryguy, line 33: This effectively fixes the bug in the test...it never checked :identifiers (plural) before.

Fryguy:
@jrafaniw - unWIPed - this should be good to go.

spec/lib/api/api_config_spec.rb
jrafanie: WAT.  I don't understand either.

Fryguy:
Now I _actually_ unWIPed :|

spec/lib/api/api_config_spec.rb
jrafanie: It looks like you more than cover the rest of the changes since this one wasn't even really testing anything that either of us can tell...

Fryguy:
Backported to `lasker` in commit 58b4210ab2b9c3c4ebd8555864991e57c9757d43.

```text
commit 58b4210ab2b9c3c4ebd8555864991e57c9757d43
Date:   Fri May 14 16:19:33 2021 -0400

    Merge pull request #1040 from Fryguy/fix_spec_to_verify_features
    
    Fix spec that verifies features to actually see all features
    
    (cherry picked from commit ee146ebf5466a03acc2ea3743515aeb3f2fd1073)
```

Issue #1042, opened 2021-06-03
DDF schema parameters are not getting translated in api response.


kavyanekkalapu:
**Issue:** When user switches to any other language other than English, wherever UI calling backend from ddf schema, those strings/labels are not coming back as  translated strings.

Sample api call: http://localhost:3000/api/providers?type=ManageIQ::Providers::Openstack::CloudManager, you can see UI for this call in Compute--> Clouds--> Add New Provider
 
<img width="910" alt="Screen Shot 2021-06-03 at 2 33 48 PM" src="https://user-images.githubusercontent.com/37085529/120694935-c11db000-c478-11eb-8008-eb32eddcdda0.png">

@miq-bot add-label bug
@miq-bot assign @NickLaMuro 
@miq-bot assign @Fryguy

kavyanekkalapu:
I am able to assign it to only one person CC @NickLaMuro

NickLaMuro:
@kavyanekkalapu Thanks, I cheated and used "super powers" to add myself.  Probably something we can fix with the bot, but this approach works for now.

I can take a look at this later today then, since it is blocking your team.

kavyanekkalapu:
> @kavyanekkalapu Thanks, I cheated and used "super powers" to add myself. Probably something we can fix with the bot, but this approach works for now.
> 
> I can take a look at this later today then, since it is blocking your team.

Thank you very much, this affects every form which is loading ddf schema from backend/api, if we can fix this, this will close many issues.

NickLaMuro:
Cool.  I don't see this as being too difficult (compared to other things), so I can't see why it won't take long to provide a fix.

NickLaMuro:
@kavyanekkalapu I think this PR should mostly address things for you:

https://github.com/ManageIQ/manageiq-api/pull/1043

There are still cases that need to be addressed as well, but this should be at least make sure the places that are properly configured to be translated are.

NickLaMuro:
Going to close this as I think this was addressed in #1043 .  Please re-open if there are further issues that need to be addressed.

PR #1043, opened 2021-06-03
[BaseController] Localize OPTIONS (allow optional user auth)

NickLaMuro:
This allows translating of OPTION data if a user is logged in.

To achieve this, the commit does a few things:

- Split up `.require_api_user_or_token` to pull out the "authentication" step into it's own method:  `.authenticate_user`
- Create a new method called `.optional_api_user_or_token` for only `OPTIONS` requests
  - Logs in a user in on `OPTIONS` calls, but no-op if it fails
  - in otherwords:  if you don't pass auth, still work as we did, but allow a user to log in and retrieve settings

By doing this, we can make sure that `:set_gettext_locale` can have a user if one is provided, and allow translations to be processed further down the line.

FYI
---

There are places where we are using `_N()` instead of `_()`, which will cause some of the translations to still result in English:

https://github.com/ManageIQ/manageiq-providers-kubernetes/blob/9b5b107dfcf702901683f02ac19263e2730d217a/app/models/manageiq/providers/kubernetes/container_manager/options.rb#L5-L13

This probably should be addressed, but can be handled on a case by case basis.

Links
-----

- Partially addresses: https://github.com/ManageIQ/manageiq-api/issues/1042

Testing/QA
----------

The spec in `spec/requests/providers_spec.rb` is a pretty good framework for seeing if this works:

- Update your user record to use a different locale ("es" for Spanish)
- Get an API auth token
- `curl` the OPTIONS endpoint:
  
  ```
  $ curl -X OPTIONS "localhost:3000?type=ManageIQ::Providers::Openstack::CloudManager"
  ```
  
(the `curl` above is missing some options for auth.  I use `miqperf` from `manageiq-performance` myself so I don't have to worry about these options)

### DIFF
--- a/app/controllers/api/base_controller/authentication.rb
+++ b/app/controllers/api/base_controller/authentication.rb
@@ -25,10 +25,7 @@ module Api
         end
       end
 
-      #
-      # REST APIs Authenticator and Redirector
-      #
-      def require_api_user_or_token
+      def authenticate_user
         case auth_mechanism
         when :system
           authenticate_with_system_token(request.headers[HttpHeaders::MIQ_TOKEN])
@@ -44,12 +41,28 @@ module Api
           raise AuthenticationError unless success
         end
         log_api_auth
+      end
+
+      #
+      # REST APIs Authenticator and Redirector
+      #
+      def require_api_user_or_token
+        authenticate_user
       rescue AuthenticationError => e
         api_log_error("AuthenticationError: #{e.message}")
         response.headers["Content-Type"] = "application/json"
         error_message = ErrorSerializer.new(:unauthorized, e).serialize(true).to_json
         render :status => 401, :json => error_message
         log_api_response
+      end
+
+      #
+      # Attempt auth if desired, but flow through on failure
+      #
+      def optional_api_user_or_token
+        authenticate_user
+      rescue AuthenticationError => e
+        api_log_warn("AuthenticationError: #{e.message} (on #{request.method} that doesn't require it... ignoring)")
       end
 
       def user_settings
--- a/app/controllers/api/base_controller.rb
+++ b/app/controllers/api/base_controller.rb
@@ -23,6 +23,7 @@ module Api
     before_action :log_request_initiated
     before_action :clear_cached_current_user
     before_action :require_api_user_or_token, :except => [:options, :product_info]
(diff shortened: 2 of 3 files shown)


NickLaMuro:
@miq-bot add_label bug
@miq-bot assign @Fryguy

@Fryguy assigning you as you seem familiar with this code, but feel free to delegate to someone else.

NickLaMuro:
cc @kavyanekkalapu

Fryguy:
wow this is cool...great work @NickLaMuro

Fryguy:
Wondering if we should cross repo test this

NickLaMuro:
@Fryguy not opposed to it.  Do you have some repos in mind where this makes sense?  Not sure what might be using the API in their specs myself.  Maybe the UI I guess?

kavyanekkalapu:
@NickLaMuro I verified the fix, it is working , tomorrow i will try to more testing on other pages, i am not sure if we are calling options in all pages or not, Thank you.
Screen Shot 2021-06-03 at 9 42 40 PM

NickLaMuro:
@miq-bot cross-repo-test ManageIQ/manageiq-ui-clasic

app/controllers/api/base_controller/authentication.rb
```diff
@@ -52,6 +56,15 @@ def require_api_user_or_token
         log_api_response
       end
 
+      #
+      # Attempt auth if desired, but flow through on failure
+      #
+      def optional_api_user_or_token
+        authenticate_user
+      rescue AuthenticationError => e
+        api_log_error("AuthenticationError: #{e.message} (on #{request.method}... ignoring)")
```
Fryguy, line 65: wondering if we should log this more as a warning, since it's not really an error.

Fryguy (approved):
One minor comment

app/controllers/api/base_controller/authentication.rb
NickLaMuro: Yeah, I just threw something in there for getting something pushed up, but definitely wasn't satisfied when I wrote it.

We can absolutely change it to a warning.  I will see if there is a method for this somewhere.  Happy to change anything else with this if you can think of anything better as well.

NickLaMuro:
@Fryguy let me know if that phrasing change is also good.

PR #930, opened 2020-10-20
Expose the MiqAction types on the endpoint using OPTIONS

skateman:
```js
// curl -XOPTIONS "http://admin:smartvm@localhost:3000/api/actions" | jq
{
  "attributes": [
    "action_type",
    "created_on",
    "description",
    "guid",
    "id",
    "name",
    "options",
    "updated_on"
  ],
  "virtual_attributes": [
    "action_type_description",
    "href_slug",
    "region_description",
    "region_number",
    "v_synchronicity"
  ],
  "relationships": [
    "all_relationships",
    "miq_policy_contents",
    "taggings",
    "tags"
  ],
  "subcollections": [],
  "data": {
    "action_types": {
      "create_snapshot": "Create a Snapshot",
      "email": "Send an E-mail",
      "snmp_trap": "Send an SNMP Trap",
      "tag": "Tag",
      "reconfigure_memory": "Reconfigure Memory",
      "reconfigure_cpus": "Reconfigure CPUs",
      "custom_automation": "Invoke a Custom Automation",
      "evaluate_alerts": "Evaluate Alerts",
      "assign_scan_profile": "Assign Profile to Analysis Task",
      "set_custom_attribute": "Set a Custom Attribute in vCenter",
      "inherit_parent_tags": "Inherit Parent Tags",
      "remove_tags": "Remove Tags",
      "delete_snapshots_by_age": "Delete Snapshots by Age",
      "run_ansible_playbook": "Run Ansible Playbook"
    }
  }
}
```

@miq-bot assign @gtanzillo 
@miq-bot add_label enhancement

PR #1045, opened 2021-06-08
[WIP] Expose MiqSnmp Types

mengxiwuu:


Fryguy:
Hi @mengxiwuu  - we probably won't expose a dedicated endpoint for something that is basically static options.  This is probably more appropriate as part of the `/api` endpoint via the OPTIONS call.

Fryguy:
@NickLaMuro @agrare @jrafanie Opinions on where something like this should live?  I was originally thinking in the /api endpoint

https://github.com/ManageIQ/manageiq-api/blob/ee146ebf5466a03acc2ea3743515aeb3f2fd1073/app/controllers/api/api_controller.rb#L9-L17

but now I'm not sure.  It should definitely be options, but it feels weird to create a new endpoint for this.

agrare:
Is there anything else that we could configure via an `/api/snmp` API?  We do have e.g. provider options exposed via `OPTIONS /api/providers` so I could see this as `OPTIONS /api/snmp`

NickLaMuro:
So, not terribly familiar with the snmp functionality of MIQ, but my thought is if we plan on having a bunch of these, it might make things more "discoverable" by wrapping it in a `/api/options/*` umbrella.  

So something like...

- `/api/options`  (everything)
- `/api/options/snmp`
- `/api/options/providers`
- etc.

mengxiwuu:
@Fryguy it is in http://localhost:3000/miq_action/new#/    
Action Type -> Send an SNMP Trap -> Variables -> Type

Fryguy:
@mengxiwuu Ah that's interesting - maybe then it makes sense to expose it as an extension to https://github.com/ManageIQ/manageiq-api/pull/930, by having another section in the data somehow.

Fryguy:
So right now it's 

```json
  "data": {
    "action_types": {
      "create_snapshot": "Create a Snapshot",
      "email": "Send an E-mail",
      "snmp_trap": "Send an SNMP Trap",
      "tag": "Tag",
      "reconfigure_memory": "Reconfigure Memory",
      "reconfigure_cpus": "Reconfigure CPUs",
      "custom_automation": "Invoke a Custom Automation",
      "evaluate_alerts": "Evaluate Alerts",
      "assign_scan_profile": "Assign Profile to Analysis Task",
      "set_custom_attribute": "Set a Custom Attribute in vCenter",
      "inherit_parent_tags": "Inherit Parent Tags",
      "remove_tags": "Remove Tags",
      "delete_snapshots_by_age": "Delete Snapshots by Age",
      "run_ansible_playbook": "Run Ansible Playbook"
    }
  }
```

And I'm wondering if we addin something like:
```json
  "data": {
    "action_types": {
      "create_snapshot": "Create a Snapshot",
      "email": "Send an E-mail",
      "snmp_trap": "Send an SNMP Trap",
      "tag": "Tag",
      "reconfigure_memory": "Reconfigure Memory",
      "reconfigure_cpus": "Reconfigure CPUs",
      "custom_automation": "Invoke a Custom Automation",
      "evaluate_alerts": "Evaluate Alerts",
      "assign_scan_profile": "Assign Profile to Analysis Task",
      "set_custom_attribute": "Set a Custom Attribute in vCenter",
      "inherit_parent_tags": "Inherit Parent Tags",
      "remove_tags": "Remove Tags",
      "delete_snapshots_by_age": "Delete Snapshots by Age",
      "run_ansible_playbook": "Run Ansible Playbook"
    },
    "snmp_trap" : {  # This is named like this to match the action type
       # all of the options here ...
    }
  }
```

kavyanekkalapu:
@Fryguy This snmp_options is used in one more place too, Alerts page--> http://localhost:3000/miq_alert/new#/.

Fryguy:
Well, that changes things 😅

Fryguy:
alerts and actions are related in that alerts can trigger actions - I think actions still makes a good place.  @kavyanekkalapu or @mengxiwuu  Do you have a screenshot of where the usage is in alerts?

kavyanekkalapu:
@Fryguy 

Screen Shot 2021-06-10 at 5 06 56 PM

app/controllers/api/actions_controller.rb
```diff
@@ -11,7 +11,14 @@ def edit_resource(type, id = nil, data = {})
     end
 
     def options
-      render_options(:actions, :action_types => MiqAction::TYPES)
+      render_options(:actions, build_action_options)
+    end
+
+    def build_action_options
```
Fryguy, line 17: Can you make this a private method?  Public methods are ones expected to be routed to via the API, so this should be private.

Fryguy:
Also can you add tests for this in, I think, https://github.com/ManageIQ/manageiq-api/blob/master/spec/requests/actions_spec.rb

I don't think we have any `options` specs already on there, so you'll have to copy the pattern from another spec.  I think it's sufficient to just check for the existence of the keys in the response payload and maybe a specific key/value

app/controllers/api/actions_controller.rb
NickLaMuro: The other option is just to multiline the hash in the existing method:

```ruby
def options
  render_options(:actions, :action_types => MiqAction::TYPES,
                           :snmp_trap    => MiqSnmp.available_types)
```

NickLaMuro (approved):
Looks good to me.  @Fryguy any other opinions, or is this good to merge?

PR #1050, opened 2021-06-21
Simplify validate_*_for_action methods

chessbyte:
Now that VMs have been converted to leverage SupportsFeatureMixin for the actions that call these methods, remove the check for SupportsFeatureMixin vs AvailabilityMixin

### DIFF
--- a/app/controllers/api/instances_controller.rb
+++ b/app/controllers/api/instances_controller.rb
@@ -172,12 +172,7 @@ module Api
     end
 
     def validate_instance_for_action(instance, action)
-      if instance.respond_to?("supports_#{action}?")
-        action_result(instance.public_send("supports_#{action}?"), instance.unsupported_reason(action.to_sym))
-      else
-        validation = instance.send("validate_#{action}")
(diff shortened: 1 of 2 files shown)


agrare:
@chessbyte looks like one more `validate_*` method, https://github.com/ManageIQ/manageiq/blob/master/app/models/vm_or_template/operations/smart_state.rb#L4-L6

There is already a `smartstate_analysis` supports_feature, looks like API is just checking for the wrong thing.

agrare (changes requested):
Need to change the `Vm#scan_resource` method to check for `supports?("smartstate_analysis")` instead of `supports?("scan")`:
```
@@ -155,7 +155,7 @@ module Api
         vm = resource_search(id, type, klass)
         api_log_info("Scanning #{vm_ident(vm)}")
 
-        result = validate_vm_for_action(vm, "scan")
+        result = validate_vm_for_action(vm, "smartstate_analysis")
```

agrare (approved):

PR #1051, opened 2021-06-24
[/api/configuration_script_sources] Test for verify_ssl

NickLaMuro:
Requires:  https://github.com/ManageIQ/manageiq/pull/21298

Create a spec that ensures `verify_ssl` is a valid attribute for `EmbeddedAnsible::AutomationManager::ConfigurationScriptSource` records

TODO
----

Merge the following:

* [x] https://github.com/ManageIQ/manageiq/pull/21298

Links
-----

* https://github.com/ManageIQ/manageiq/pull/21298
  - Required PR before merging
* https://github.com/ManageIQ/manageiq-ui-classic/pull/7768
  - API field will be used in this PR

### DIFF
--- a/spec/requests/configuration_script_sources_spec.rb
+++ b/spec/requests/configuration_script_sources_spec.rb
@@ -32,23 +32,36 @@ RSpec.describe 'Configuration Script Sources API' do
   end
 
   describe 'GET /api/configuration_script_sources/:id' do
-    it 'will show a configuration script source with an appropriate role' do
-      repository = FactoryBot.create(:configuration_script_source)
-      api_basic_authorize collection_action_identifier(:configuration_script_sources, :read, :get)
+    let(:repository) { FactoryBot.create(:configuration_script_source) }
 
+    it 'will show a configuration script source with an appropriate role' do
+      api_basic_authorize collection_action_identifier(:configuration_script_sources, :read, :get)
       get(api_configuration_script_source_url(nil, repository))
 
+      expected = {'href' => api_configuration_script_source_url(nil, repository)}
+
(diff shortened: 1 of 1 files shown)


NickLaMuro:
@miq-bot assign @Fryguy 

Another quick one.  Just making sure the API is ensuring this attribute is available in the future.

NickLaMuro:
Bah... ignore the above commit.  Accidental push while working on something else.  Need to move it to another branch...

spec/requests/configuration_script_sources_spec.rb
```diff
@@ -32,23 +32,36 @@
...
       get(api_configuration_script_source_url(nil, repository))
 
+      expected = {'href' => api_configuration_script_source_url(nil, repository)}
+
+      expect(response.parsed_body).to include(expected)
+      expect(response).to have_http_status(:ok)
+    end
+
+    it 'allows accessing verify_ssl for EmbeddedAnsible::AutomationManager::ConfigurationScriptSource' do
+      embedded_ansible_repository = FactoryBot.create(:embedded_ansible_configuration_script_source)
```
bdunne, line 48: Is this specific to EmbeddedAnsible or does it also work on the base class?

NickLaMuro: No, it does not, since nothing else has `:git_repository` as a relation like the specific version does:

https://github.com/ManageIQ/manageiq/blob/5aa0822a056a28e1cd7790768a05343cda835684/app/models/manageiq/providers/embedded_ansible/automation_manager/configuration_script_source.rb#L13

Which was added here:

https://github.com/ManageIQ/manageiq/pull/21298

Issue #1053, opened 2021-06-27
Group filter_expression not setting thru API


svinopterix:
Filter expression isn't working when set thru API. No errors returned, group is successfully created, but no filters set on it. I’ve found that a record is created in “entitlements” table in both cases. But if through WebUI, filter_expression is ruby/object:MiqExpression. Аnd if through API, it’s ruby/hash:ActiveSupport::HashWithIndifferentAccess.

Through API:

```
vmdb_production=# select miq_group_id, filter_expression from entitlements;
 miq_group_id |                        filter_expression
--------------+-----------------------------------------------------------------
.......
           80 | --- !ruby/hash:ActiveSupport::HashWithIndifferentAccess        +
              | exp: !ruby/hash:ActiveSupport::HashWithIndifferentAccess       +
              |   CONTAINS: !ruby/hash:ActiveSupport::HashWithIndifferentAccess+
              |     tag: managed-prov_scope                                    +
              |     value: all                                                 +
```

Through WebUI:

```
vmdb_production=# select miq_group_id, filter_expression from entitlements;
 miq_group_id |       filter_expression
--------------+--------------------------------
......
           80 | --- !ruby/object:MiqExpression+
              | exp:                          +
              |   CONTAINS:                   +
              |     tag: managed-prov_scope   +
              |     value: all                +
              | context_type:                 +
```

I'm using API request as follows. MIQ is jansa-2.

```
{
  "action": "create",
  "resource": {
    "description": "test_group_1",
    "tenant": {
     "href": "/api/tenants/1"
    },
    "role": {
      "id": "2"
    },
    "filter_expression": {
      "exp": {
        "CONTAINS": {
          "tag": "managed-prov_scope",
          "value": "all"
        }    
} } } }
```

PR #1054, opened 2021-07-06
[GroupsController] Fix filter_expression save format

NickLaMuro:
Fixes https://github.com/ManageIQ/manageiq-api/issues/1053

When saving through the UI, the data passed from the browser are converted into a proper `MiqExpression` object:

https://github.com/ManageIQ/manageiq-ui-classic/blob/d0e9a1b319718f6a91d1a0c5ca3edb792bf1a8ee/app/controllers/ops_controller/ops_rbac.rb#L1267

However, in the API, this was just passed and saved as an `Hash` (`HashWithIndifferentAccess`), and this seems cause issues with being editable and usable when the unserialized hash data is used.

This fixes that by match the API with what we save it as in the UI.

### DIFF
--- a/app/controllers/api/groups_controller.rb
+++ b/app/controllers/api/groups_controller.rb
@@ -18,7 +18,7 @@ module Api
       parse_set_role(data)
       parse_set_tenant(data)
       parse_set_filters(data)
-      group = collection_class(:groups).create(data)
+      group = collection_class(:groups).create(data.to_h)
       if group.invalid?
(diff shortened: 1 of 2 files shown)


kbrock (approved):
surprised that the `data.to_h` was necessary, and even a little surprised that the stringify keys was needed for the filters.

But the `filter_expression` with the `MiqExpression.new` looks right and the test shows that the expression makes its way across.

LGTM

PR #1033, opened 2021-05-05
Allow creation of AutomationManagers via API

agrare:
https://talk.manageiq.org/t/bad-request-when-trying-to-add-create-ansible-tower-providers-via-api/5394/3

PR #1055, opened 2021-07-12
[LASKER] ems_automation_add_provider was from a refactoring on master, not lasker

jrafanie:
https://github.com/ManageIQ/manageiq-api/pull/1033 was fine on master but relied
on a refactoring/rewording of product features and models in https://github.com/ManageIQ/manageiq/pull/21108
and this refactoring did not get backported to lasker.

In other words, ems_automation_add_provider doesn't exist in lasker, but
automation_manager_add_provider does, so we'll use that.

cc @agrare

### DIFF
--- a/config/api.yml
+++ b/config/api.yml
@@ -2752,7 +2752,7 @@
       - :name: create
         :identifiers:
         - :klass: ManageIQ::Providers::AutomationManager
-          :identifier: ems_automation_add_provider
+          :identifier: automation_manager_add_provider
         - :klass: ManageIQ::Providers::InfraManager
           :identifier: ems_infra_new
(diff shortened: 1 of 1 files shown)


Fryguy (approved):

agrare (approved):
:+1: good catch, didn't realize `ems_automation_add_provider` was changed recently

jrafanie:
> 👍 good catch, didn't realize `ems_automation_add_provider` was changed recently

Yeah, I only knew because I was looking at all of these bogus product features that the UI was doing and silently wrongly failing feature checks for non-admin users.  Normally, it would make sense to backport simple product feature renames but since these reflected code changes in model names, I think we correctly chose not to backport it.

PR #1056, opened 2021-07-15
Add support for editing autosde cloud volume

llivne:
This PR is part of PRs that solve Autosde cloud volume editing

image

If specific info about a cloud volume is passed then we are under update mode (not creation) and the update field should be used.

Links
----------------
https://github.com/ManageIQ/manageiq-ui-classic/pull/7788
https://github.com/ManageIQ/manageiq-providers-autosde/pull/72

### DIFF
--- a/app/controllers/api/cloud_volumes_controller.rb
+++ b/app/controllers/api/cloud_volumes_controller.rb
@@ -59,7 +59,11 @@ module Api
 
       raise BadRequestError, "No DDF specified for - #{klass}" unless klass.respond_to?(:params_for_create)
 
-      render_options(:cloud_volumes, :form_schema => klass.params_for_create(ems))
+      if params["record_id"]
+        render_options(:cloud_volumes, :form_schema => klass.params_for_update(ems, params))
+      else
+        render_options(:cloud_volumes, :form_schema => klass.params_for_create(ems))
+      end
     end
   end
 end
--- a/app/controllers/api/cloud_volumes_controller.rb
+++ b/app/controllers/api/cloud_volumes_controller.rb
@@ -49,20 +49,19 @@ module Api
     end
 
     def options
-      return super unless params[:ems_id]
+      if params[:id]
+        cloud_volume = resource_search(params[:id], :cloud_volumes, CloudVolume)
+        render_options(:cloud_volumes, :form_schema => cloud_volume.params_for_update)
+      elsif params[:ems_id]
+        ems = resource_search(params[:ems_id], :ext_management_systems, ExtManagementSystem)
+        raise BadRequestError, "No CloudVolume support for - #{ems.class}" unless defined?(ems.class::CloudVolume)
 
-      ems = ExtManagementSystem.find(params[:ems_id])
+        klass = ems.class::CloudVolume
+        raise BadRequestError, "No DDF specified for - #{klass}" unless klass.respond_to?(:params_for_create)
 
-      raise BadRequestError, "No CloudVolume support for - #{ems.class}" unless defined?(ems.class::CloudVolume)
-
-      klass = ems.class::CloudVolume
-
-      raise BadRequestError, "No DDF specified for - #{klass}" unless klass.respond_to?(:params_for_create)
-
-      if params["record_id"]
-        render_options(:cloud_volumes, :form_schema => klass.params_for_update(ems, params))
-      else
         render_options(:cloud_volumes, :form_schema => klass.params_for_create(ems))
+      else
+        super
       end
     end
   end
--- a/app/controllers/api/cloud_volumes_controller.rb
+++ b/app/controllers/api/cloud_volumes_controller.rb
@@ -57,7 +57,7 @@ module Api
         raise BadRequestError, "No CloudVolume support for - #{ems.class}" unless defined?(ems.class::CloudVolume)
 
         klass = ems.class::CloudVolume
-        raise BadRequestError, "No DDF specified for - #{klass}" unless klass.respond_to?(:params_for_create)
+        raise BadRequestError, klass.unsupported_reason(:create) unless klass.supports?(:create)
 
         render_options(:cloud_volumes, :form_schema => klass.params_for_create(ems))
       else
--- a/spec/requests/cloud_volumes_spec.rb
+++ b/spec/requests/cloud_volumes_spec.rb
@@ -222,7 +222,6 @@ describe "Cloud Volumes API" do
       provider = FactoryBot.create(:ems_autosde, :zone => zone)
       cloud_volume = FactoryBot.create(:cloud_volume_autosde, :ext_management_system => provider)
 
-      allow(CloudVolume).to receive(:find).with(cloud_volume.id.to_s).and_return(cloud_volume)
       allow(Rbac).to receive(:filtered_object).and_return(cloud_volume)
       expect(cloud_volume).to receive(:params_for_update).and_return('foo')
       options("#{api_cloud_volumes_url}/#{cloud_volume.id}")
--- a/spec/requests/cloud_volumes_spec.rb
+++ b/spec/requests/cloud_volumes_spec.rb
@@ -8,7 +8,7 @@
 # GET /api/cloud_volumes/:id
 #
 
-fdescribe "Cloud Volumes API" do
+describe "Cloud Volumes API" do
   it "forbids access to cloud volumes without an appropriate role" do
     api_basic_authorize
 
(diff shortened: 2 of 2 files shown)


agrare:
Hey @llivne sorry for the delay on reviewing this I've been working on how we want to do the ddf params for editing records, so far we've only done `OPTIONS /api/collection`

I've added the ability to call `OPTIONS /api/collection/:id` which I think is a more natural way to do this.  So your API call will look like `API.options('/api/cloud_volumes/${recordId}')`

Then in this method you can handle this by checking for `params["id"]` to see if it was called on an instance rather than the collection.  You can then lookup the record by `resource_search()` and call an instance method `record.params_for_update` and not have to pass any params in.

app/controllers/api/cloud_volumes_controller.rb
```diff
@@ -59,7 +59,11 @@ def options
 
       raise BadRequestError, "No DDF specified for - #{klass}" unless klass.respond_to?(:params_for_create)
 
-      render_options(:cloud_volumes, :form_schema => klass.params_for_create(ems))
+      if params["record_id"]
```
agrare, line 62: I'm thinking this method can look like this now:
```
def options
  if params[:ems_id]
    ems = resource_search(params[:ems_id], :ext_management_systems, ExtManagementSystem)
    raise BadRequestError, "No CloudVolume support for - #{ems.class}" unless defined?(ems.class::CloudVolume)

    klass = ems.class::CloudVolume
    raise BadRequestError, "No DDF specified for - #{klass}" unless klass.respond_to?(:params_for_create)

    render_options(:cloud_volumes, :form_schema => klass.params_for_create(ems))
  elsif params[:id]
    cloud_volume = resource_search(params[:id], :cloud_volumes, CloudVolume)
    render_options(:cloud_volumes, :form_schema => cloud_volume.params_for_update)
  else
    super
  end
end
```

llivne: @agrare thank you for your comments. I updated all of the PRs to work like you suggested. please review the new changes

app/controllers/api/cloud_volumes_controller.rb
```diff
@@ -49,17 +49,20 @@ def delete_resource(type, id, _data = {})
...
-      render_options(:cloud_volumes, :form_schema => klass.params_for_create(ems))
+      if params[:id]
+        cloud_volume = resource_search(params[:id], :cloud_volumes, CloudVolume)
+        render_options(:cloud_volumes, :form_schema => cloud_volume.params_for_update)
+      elsif params[:ems_id]
+        ems = resource_search(params[:ems_id], :ext_management_systems, ExtManagementSystem)
+        raise BadRequestError, "No CloudVolume support for - #{ems.class}" unless defined?(ems.class::CloudVolume)
+
+        klass = ems.class::CloudVolume
+        raise BadRequestError, "No DDF specified for - #{klass}" unless klass.respond_to?(:params_for_create)
```
agrare, line 60: Prefer to check for `klass.supports?(:create)` rather than checking for methods existing.

```suggestion
        raise BadRequestError, klass.unsupported_reason(:create) unless klass.supports?(:create)
```

agrare:
Looks good, can you add a spec test for these two new code paths?  You can use https://github.com/ManageIQ/manageiq-api/pull/1058/files#diff-a0d92b93db6899bc0858bc8ccfbcec74b815ab90846b58784bc0de9acdefdee2R101-R129 as a reference

app/controllers/api/cloud_volumes_controller.rb
llivne: done

llivne:
> Looks good, can you add a spec test for these two new code paths? You can use https://github.com/ManageIQ/manageiq-api/pull/1058/files#diff-a0d92b93db6899bc0858bc8ccfbcec74b815ab90846b58784bc0de9acdefdee2R101-R129 as a reference

done :)

spec/requests/cloud_volumes_spec.rb
```diff
@@ -8,7 +8,7 @@
 # GET /api/cloud_volumes/:id
 #
 
-describe "Cloud Volumes API" do
+fdescribe "Cloud Volumes API" do
```
agrare, line 11: This looks like a typo

agrare:
@bdunne can you also review?

spec/requests/cloud_volumes_spec.rb
```diff
@@ -201,16 +201,36 @@
...
-    expect(response.parsed_body['data']['form_schema']).to eq('foo')
-    expect(response).to have_http_status(:ok)
+      expect(response.parsed_body['data']['form_schema']).to eq('foo')
+      expect(response).to have_http_status(:ok)
+    end
+  end
+
+  describe 'OPTIONS /api/cloud_volumes/:id' do
+    it 'returns a DDF schema for edit when available via OPTIONS' do
+      zone = FactoryBot.create(:zone, :name => "api_zone")
```
bdunne, line 221: Please don't specify `:name`, it isn't important to your test and the factory will automatically create one for you

spec/requests/cloud_volumes_spec.rb
```diff
@@ -201,16 +201,36 @@
...
+      provider = FactoryBot.create(:ems_autosde, :zone => zone)
+      volume = FactoryBot.create(:cloud_volume_autosde, :ext_management_system => provider, :name => "my_volume")
+
+      cloud_volume = provider.cloud_volumes.first
```
bdunne, line 225: Do you need this extra lookup?  It's the same as `volume` above, correct?

bdunne (dismissed):

spec/requests/cloud_volumes_spec.rb
bdunne: Hmm... somehow Travis passed 😕

bdunne: Well, it's [a real thing](https://relishapp.com/rspec/rspec-core/v/3-10/docs/filtering/filter-run-when-matching#groups-can-be-focused-with-%60fdescribe%60-or-%60fcontext%60), but I don't think it was intentional.

spec/requests/cloud_volumes_spec.rb
llivne: fixed

spec/requests/cloud_volumes_spec.rb
llivne: you are correct. fixed

spec/requests/cloud_volumes_spec.rb
llivne: @bdunne @agrare sorry about that. This actually was intentional, but it wasnt supposed to be part of the PR. in order to run specific specs we sometime use[ Inclusion filters](https://relishapp.com/rspec/rspec-core/v/2-6/docs/filtering/inclusion-filters). this is because running a specific file not always works. in short, I wrote a script that add  config.filter_run focus: true to spec_helper.rb (simply sed -i '/RSpec.configure.*/a \ \ config.filter_run focus: true' ./spec/spec_helper.rb) and then only tagged tests will be run. Tagging is done by adding ':focus => true' (or prepend an f) to it, specify, describe or context.
my script also remove the modified spec_helper.rb when it done but not the tagging and I forgot to remove it before the push. this shouldn't append, sorry about that.

llivne: I removed it now :)

llivne:
@bdunne @agrare I did all the required changes, waiting for your review/approval

ereztheodorou:
@miq-bot add_label morphy/yes?

spec/requests/cloud_volumes_spec.rb
```diff
@@ -201,16 +201,34 @@
...
+  end
+
+  describe 'OPTIONS /api/cloud_volumes/:id' do
+    it 'returns a DDF schema for edit when available via OPTIONS' do
+      zone = FactoryBot.create(:zone)
+      provider = FactoryBot.create(:ems_autosde, :zone => zone)
+      cloud_volume = FactoryBot.create(:cloud_volume_autosde, :ext_management_system => provider)
+
+      allow(CloudVolume).to receive(:find).with(cloud_volume.id.to_s).and_return(cloud_volume)
+      allow(Rbac).to receive(:filtered_object).and_return(cloud_volume)
```
bdunne, line 226: Do we need to stub these?

llivne: @bdunne `allow(CloudVolume).to receive(:find).with(cloud_volume.id.to_s).and_return(cloud_volume)` is not needed so I removed it. the Rbac is required for the OPTIONS to work

agrare (approved):

Fryguy:
Backported to `morphy` in commit d762bdf1ac80bdc6b5cf54d9d3509ea51e13024b.

```text
commit d762bdf1ac80bdc6b5cf54d9d3509ea51e13024b
Author: Adam Grare <adam@grare.com>
Date:   Tue Sep 7 09:12:18 2021 -0400

    Merge pull request #1056 from Autosde/fix_edit_cloud_volume
    
    Add support for editing autosde cloud volume
    
    (cherry picked from commit e6fa13e6e942051d8943e013538fdfb3c6a417d2)
```

PR #1057, opened 2021-07-15
Fix user_spec.rb test order failures

NickLaMuro:
Fixes a few issues with the `spec/requests/user_spec.rb` tests:

- Fixes a settings stub issue where only `memory` store was tested
- Fixes issue where running after `rspec/requests/authentication_spec.rb` causes the `user_spec.rb` tests to fail due to cache of `TokenStore.token_caches`
- Removes pointless call to `TokenStore.token_caches` in the middle of the specs

Links
-----

- Example failed build:  https://travis-ci.com/github/ManageIQ/manageiq-api/builds/231533940

### DIFF
--- a/spec/requests/users_spec.rb
+++ b/spec/requests/users_spec.rb
@@ -615,10 +615,14 @@ RSpec.describe "users API" do
   end
 
   context "Revoking Sessions" do
+    before do
+      # Keep this outside of the loop below with the `let` providing the value
+      # for the stub, otherwise it will not work as expected
+      stub_settings_merge(:server => {:session_store => session_store_value})
+    end
+
     %w[cache sql memory].each do |session_store|
-      before do
-        ::Settings.server.session_store = session_store
-      end
+      let(:session_store_value) { session_store }
 
       it "revokes all own sessions of authenticated user with #{session_store}" do
         api_basic_authorize
--- a/spec/requests/users_spec.rb
+++ b/spec/requests/users_spec.rb
@@ -633,8 +633,6 @@ RSpec.describe "users API" do
         FactoryBot.create(:session, :user_id => @user.id)
         expect(Session.where(:user_id => @user.id).count).to eq(1)
 
-        TokenStore.token_caches
-
         ts = TokenStore.acquire("api", 100)
 
         ts.create_user_token("my_token", {:userid => @user.userid, :expires_on => Time.zone.now + 100.days}, {:expires_in => 100})
--- a/spec/requests/users_spec.rb
+++ b/spec/requests/users_spec.rb
@@ -619,7 +619,10 @@ RSpec.describe "users API" do
       # Keep this outside of the loop below with the `let` providing the value
       # for the stub, otherwise it will not work as expected
       stub_settings_merge(:server => {:session_store => session_store_value})
+      TokenStore.token_caches.clear
     end
+
+    after { TokenStore.token_caches.clear }
 
     %w[cache sql memory].each do |session_store|
       let(:session_store_value) { session_store }
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,5 +6,7 @@ rvm:
 - 2.7.2
 addons:
   postgresql: '10'
+services:
+- memcached
 install: bin/setup
 after_script: bin/ci/after_script


spec/requests/users_spec.rb
```diff
@@ -616,18 +616,16 @@
...
   context "Revoking Sessions" do
     %w[cache sql memory].each do |session_store|
-      before do
-        ::Settings.server.session_store = session_store
-      end
+      before { TokenStore.token_caches.clear }
+      after  { TokenStore.token_caches.clear }
 
       it "revokes all own sessions of authenticated user with #{session_store}" do
+        stub_settings_merge(:server => {:session_store => session_store}) # Don't stub in before
```
Fryguy, line 623: The comment here doesn't make sense out of this PR's context, so I don't think it's necessary.

Fryguy: Also you should be able to stub in the before as long as you use `stub_settings_merge`

Fryguy:
Seems tests are still failing though :(

spec/requests/users_spec.rb
NickLaMuro: > Also you should be able to stub in the before as long as you use stub_settings_merge

No, I couldn't, because it just runs the `memory` test 3 times when I did, instead of running each one.  See the commit message for more details

> [user_spec.rb] Stub settings on per test basis
> 
> RSpec does something weird with the `before` block here where it will run the block times per test invocation, once for each value in the array that defines `session_store`.  Each time, it would set it as `memory` instead of running each set of tests for each store respectively (determined via some "`puts` debugging")
> 
> Stubbing in the tests blocks themselves (`it` blocks), so for the sake of the tests doing what they should instead of repeating the same set of tests 3 times, the "less DRY" approach was used.

If I did the following:

```ruby
before do
  puts "before #{::Settings.server.session_store.inspect}..."
  stub_settings_merge(:server => {:session_store => session_store})
  puts "before #{::Settings.server.session_store.inspect}..."
end

it "revokes all own sessions of authenticated user with #{session_store}" do
  puts "in test #{::Settings.server.session_store.inspect}..."
```

The output for a **single test** would look like:

```
before "memory"
after "cache"
before "cache"
after "sql"
before "sql"
after "memory"
in test "memory"
```

So every test would run with `"memory"`.

NickLaMuro:
> Seems tests are still failing though :(

Oh, I bet I know what it is.

Because these specs were always running with the `memory` version, we probably never had to worry about using `memcache` before, and I bet that `memcache` isn't even installed on CI...

@Fryguy Should I install memcache here and see if it fixes the test, or skip this test in CI?

spec/requests/users_spec.rb
NickLaMuro: Also, more proof that this was actually an issue is that this deprecation warning:

```
.......DEPRECATION: :dalli_store will be removed in Dalli 3.0.
Please use Rails' official :mem_cache_store instead.
https://guides.rubyonrails.org/caching_with_rails.html
F.......
```

found in the build:

https://app.travis-ci.com/github/ManageIQ/manageiq-api/jobs/524811951#L2085-L2086

Wasn't there before because we always used `memory` for these specs.

NickLaMuro: FYI, I addressed that here:

https://github.com/ManageIQ/manageiq/pull/21335

Fryguy: OH i see what the problem is...because the before block is inside of the loop, it effectively creates 3 before blocks, and the most recent one is going to win.  There shouldn't be a before/after block inside of the loop even after your changes here.

Even so, I feel there has go to be a way to dry this up, but I'm not going to hold up a merge for that.

Fryguy:
> @Fryguy Should I install memcache here and see if it fixes the test, or skip this test in CI?

If installing memcache is relatively simple, I think we should do that.

NickLaMuro:
@Fryguy installed memcache, but I think that I also solved the "`stub_settings_merge` in a `before`" issue, so going to push a change for that quickly as well.

Putting it as `[WIP]` briefly as I get that sorted.

spec/requests/users_spec.rb
NickLaMuro: Alright, fixed this so it is DRY.  Just had to use a `let` for the value that was set so we could pass it to the outer `before`.

Put a comment in there so people don't change it without thinking.

NickLaMuro:
@Fryguy This should be good to go now, pending Travis run.

spec/requests/users_spec.rb
```diff
@@ -615,19 +615,24 @@
...
+      TokenStore.token_caches.clear
+    end
+
+    after { TokenStore.token_caches.clear }
+
     %w[cache sql memory].each do |session_store|
-      before do
-        ::Settings.server.session_store = session_store
-      end
+      let(:session_store_value) { session_store }
```
Fryguy, line 628: I guess this doesn't clobber the next one in the loop?  Still seems weird.  I wonder if it's just safer to wrap the entire thing in a context.

Fryguy (approved):

spec/requests/users_spec.rb
kbrock: pretty sure the contexts are not kept across tests.
There were gems that reused contexts across tests to speed things up, but I'm pretty sure that stopped over 5 years go.

kbrock (approved):
I'm more comfortable having this type of infrastructure in before/after.

Great Job Nick

spec/requests/users_spec.rb
Fryguy: I'm not sure what you are saying - I'm suggesting:

%w[cache sql memory].each do |session_store|
  context "with #{session_store}" do
    let(:session_store_value) { session_store }

    it "revokes all own sessions of authenticated user with #{session_store}" do
...
```

Comment (NickLaMuro) [reply spec/requests/users_spec.rb:628]: @Fryguy Hmm... I didn't catch what you were saying before, but possibly using the `context` instead of the `let` would solve the issue with the `before` loop.  The internals of `RSpec` are a bit complex, to say the least, but I think if we created a new context and put the `before` inside that, it would only create one `before` for that single context`.

Unsure, but I can try that out later today.  Not sure it is worth it now that we have a working solution, but I will see what comes of that.

**Comment by Fryguy** [comment]
Backported to `morphy` in commit a8df72b1ecec51c5a3668dd3fa963ad444ac815e.

```text
commit a8df72b1ecec51c5a3668dd3fa963ad444ac815e
Date:   Thu Aug 19 00:59:36 2021 -0400

    Merge pull request #1057 from NickLaMuro/fix-token-store-test-failures
    
    Fix user_spec.rb test order failures
    
    (cherry picked from commit 4375b8c1159f43db62ff5189810fcdc72d11e59e)
```

Fryguy:
Backported to `lasker` in commit ecb9058857c7b80cbfc1feb01dcb7e15fcbb75b9.

```text
commit ecb9058857c7b80cbfc1feb01dcb7e15fcbb75b9
Date:   Thu Aug 19 00:59:36 2021 -0400

    Merge pull request #1057 from NickLaMuro/fix-token-store-test-failures
    
    Fix user_spec.rb test order failures
    
    (cherry picked from commit 4375b8c1159f43db62ff5189810fcdc72d11e59e)
```

PR #1058, opened 2021-07-22
Added options call for cloud networks

GilbertCherrie:
Depends on: #1060

Added an options call for cloud networks:

Calling example with no cloud_network_id:
OPTIONS "http://admin:smartvm@localhost:3000/api/cloud_networks?ems_id=8"  
`{
  "attributes": [
    "cidr",
    "cloud_tenant_id",
    "description",
    "ems_id",
    "ems_ref",
    "enabled",
    "external_facing",
    "extra_attributes",
    "id",
    "name",
    "orchestration_stack_id",
    "provider_network_type",
    "provider_physical_network",
    "provider_segmentation_id",
    "resource_group_id",
    "shared",
    "status",
    "type",
    "vlan_transparent"
  ],
  "virtual_attributes": [
    "href_slug",
    "maximum_transmission_unit",
    "port_security_enabled",
    "qos_policy_id",
    "region_description",
    "region_number",
    "total_vms"
  ],
  "relationships": [
    "cloud_subnets",
    "cloud_tenant",
    "custom_action_buttons",
    "custom_actions",
    "custom_button_events",
    "custom_button_sets",
    "custom_buttons",
    "ext_management_system",
    "floating_ips",
    "network_ports",
    "network_routers",
    "orchestration_stack",
    "private_networks",
    "public_network_routers",
    "public_network_vms",
    "public_networks",
    "resource_group",
    "security_groups",
    "taggings",
    "tags",
    "vms"
  ],
  "subcollections": [
    "tags"
  ],
  "data": {
    "form_schema": {
      "fields": [
        {
          "component": "sub-form",
          "title": "Placement",
          "id": "placement",
          "name": "placement",
          "fields": [
            {
              "component": "select",
              "id": "cloud_tenant",
              "name": "cloud_tenant",
              "key": "id-8",
              "label": "Cloud Tenant",
              "placeholder": "<Choose>",
              "validateOnMount": true,
              "validate": [
                {
                  "type": "required",
                  "message": "Required"
                }
              ],
              "isDisabled": false,
              "options": [
                {
                  "label": "cloud-user-demo",
                  "value": "1"
                },
                {
                  "label": "admin",
                  "value": "2"
                },
                {
                  "label": "openshift_demo",
                  "value": "3"
                },
                {
                  "label": "testtel",
                  "value": "4"
                },
                {
                  "label": "cloud-southeast",
                  "value": "5"
                },
                {
                  "label": "Loic Tenant",
                  "value": "6"
                },
                {
                  "label": "cloudwest",
                  "value": "7"
                },
                {
                  "label": "Massachusetts",
                  "value": "8"
                },
                {
                  "label": "testetot",
                  "value": "10"
                },
                {
                  "label": "test-ivy",
                  "value": "13"
                },
                {
                  "label": "test-ivr",
                  "value": "12"
                }
              ],
              "includeEmpty": true,
              "clearOnUnmount": true
            }
          ]
        },
        {
          "component": "sub-form",
          "title": "Network Provider Information",
          "id": "provider-information",
          "name": "provider-information",
          "condition": {
            "when": "cloud_tenant",
            "isNotEmpty": true
          },
          "fields": [
            {
              "component": "select",
              "id": "provider_network_type",
              "name": "provider_network_type",
              "label": "Provider Network Type",
              "placeholder": "Nothing selected",
              "options": [
                {
                  "label": "None"
                },
                {
                  "label": "Local",
                  "value": "local"
                },
                {
                  "label": "Flat",
                  "value": "flat"
                },
                {
                  "label": "GRE",
                  "value": "gre"
                },
                {
                  "label": "GENEVE",
                  "value": "geneve"
                },
                {
                  "label": "VLAN",
                  "value": "vlan"
                },
                {
                  "label": "VXLAN",
                  "value": "vxlan"
                }
              ],
              "isDisabled": false
            },
            {
              "component": "sub-form",
              "id": "subform-1",
              "name": "subform-1",
              "condition": {
                "when": "provider_network_type",
                "is": "flat"
              },
              "fields": [
                {
                  "component": "text-field",
                  "label": "Physical Network",
                  "maxLength": 128,
                  "id": "provider_physical_network",
                  "name": "provider_physical_network",
                  "isDisabled": false,
                  "validateOnMount": true,
                  "clearOnUnmount": true,
                  "validate": [
                    {
                      "type": "required",
                      "message": "Required"
                    }
                  ]
                }
              ]
            },
            {
              "component": "sub-form",
              "id": "subform-2",
              "name": "subform-2",
              "condition": {
                "when": "provider_network_type",
                "is": "gre"
              },
              "fields": [
                {
                  "component": "text-field",
                  "label": "Segmentation ID",
                  "maxLength": 128,
                  "id": "provider_segmentation_id",
                  "name": "provider_segmentation_id",
                  "isDisabled": false,
                  "validateOnMount": true,
                  "clearOnUnmount": true,
                  "validate": [
                    {
                      "type": "required",
                      "message": "Required"
                    }
                  ]
                }
              ]
            },
            {
              "component": "sub-form",
              "id": "subform-3",
              "name": "subform-3",
              "condition": {
                "when": "provider_network_type",
                "is": "vlan"
              },
              "fields": [
                {
                  "component": "text-field",
                  "label": "Physical Network",
                  "maxLength": 128,
                  "id": "provider_physical_network",
                  "name": "provider_physical_network",
                  "isDisabled": false,
                  "validateOnMount": true,
                  "clearOnUnmount": true,
                  "validate": [
                    {
                      "type": "required",
                      "message": "Required"
                    }
                  ]
                },
                {
                  "component": "text-field",
                  "label": "Segmentation ID",
                  "maxLength": 128,
                  "id": "provider_segmentation_id",
                  "name": "provider_segmentation_id",
                  "isDisabled": false,
                  "validateOnMount": true,
                  "clearOnUnmount": true,
                  "validate": [
                    {
                      "type": "required",
                      "message": "Required"
                    }
                  ]
                }
              ]
            },
            {
              "component": "sub-form",
              "id": "subform-4",
              "name": "subform-4",
              "condition": {
                "when": "provider_network_type",
                "is": "vxlan"
              },
              "fields": [
                {
                  "component": "text-field",
                  "label": "Segmentation ID",
                  "maxLength": 128,
                  "id": "provider_segmentation_id",
                  "name": "provider_segmentation_id",
                  "isDisabled": false,
                  "clearOnUnmount": true
                }
              ]
            }
          ]
        },
        {
          "component": "sub-form",
          "title": "Network Information",
          "id": "network-information",
          "name": "network-information",
          "fields": [
            {
              "component": "text-field",
              "id": "name",
              "name": "name",
              "validateOnMount": true,
              "label": "Network Name",
              "validate": [
                {
                  "type": "required",
                  "message": "Required"
                }
              ]
            }
          ]
        },
        {
          "component": "switch",
          "id": "cloud_network_external_facing",
          "name": "external_facing",
          "label": "External Router",
          "onText": "Yes",
          "offText": "No"
        },
        {
          "component": "switch",
          "id": "cloud_network_enabled",
          "name": "enabled",
          "label": "Adminstrative State",
          "onText": "Up",
          "offText": "Down"
        },
        {
          "component": "switch",
          "id": "cloud_network_shared",
          "name": "shared",
          "label": "Shared",
          "onText": "Yes",
          "offText": "No"
        }
      ]
    }
  }
}`

Can also call with a cloud_network_id in order to return only fields that are allowed to be edited while the rest are disabled.
OPTIONS "http://admin:smartvm@localhost:3000/api/cloud_networks/${cloud_network_id}

@miq-bot add_reviewer @Fryguy 
@miq-bot add-label dependencies

### DIFF
--- a/app/controllers/api/cloud_networks_controller.rb
+++ b/app/controllers/api/cloud_networks_controller.rb
@@ -15,12 +15,13 @@ module Api
 
       if params.key?("ems_id")
         ems = resource_search(params["ems_id"], :ext_management_systems, ExtManagementSystem)
+        klass = CloudNetwork.class_by_ems(ems)
+
         raise BadRequestError, "No Cloud Network support for - #{ems.class}" unless defined?(ems.class::CloudNetwork)
 
-        klass = ems.class::CloudNetwork
-        raise BadRequestError, "No DDF specified for - #{klass}" unless klass.respond_to?(:params_for_create)
+        raise BadRequestError, "No DDF specified for - #{klass}" unless klass.supports?(:create)
 
-        render_options(:cloud_networks, :form_schema => CloudNetwork.class_by_ems(ems).params_for_create(ems))
+        render_options(:cloud_networks, :form_schema => klass.params_for_create(ems))
       else
         cloud_network = resource_search(params["id"], :cloud_networks, CloudNetwork)
         render_options(:cloud_networks, :form_schema => cloud_network.params_for_edit)
--- a/app/controllers/api/cloud_networks_controller.rb
+++ b/app/controllers/api/cloud_networks_controller.rb
@@ -10,21 +10,29 @@ module Api
       action_result(true, "Deleting #{cloud_network.name}", :task_id => task_id)
     end
 
+    private def options_by_ems_id
+      ems = resource_search(params["ems_id"], :ext_management_systems, ExtManagementSystem)
+      klass = CloudNetwork.class_by_ems(ems)
+
+      raise BadRequestError, "No Cloud Network support for - #{ems.class}" unless defined?(ems.class::CloudNetwork)
+
+      raise BadRequestError, "No DDF specified for - #{klass}" unless klass.supports?(:create)
+
+      render_options(:cloud_networks, :form_schema => klass.params_for_create(ems))
+    end
+
+    private def options_by_id
+      cloud_network = resource_search(params["id"], :cloud_networks, CloudNetwork)
+      render_options(:cloud_networks, :form_schema => cloud_network.params_for_edit)
+    end
+
     def options
-      return super unless params.keys.include_any?(%w[id ems_id])
-
       if params.key?("ems_id")
-        ems = resource_search(params["ems_id"], :ext_management_systems, ExtManagementSystem)
-        klass = CloudNetwork.class_by_ems(ems)
-
-        raise BadRequestError, "No Cloud Network support for - #{ems.class}" unless defined?(ems.class::CloudNetwork)
-
-        raise BadRequestError, "No DDF specified for - #{klass}" unless klass.supports?(:create)
-
-        render_options(:cloud_networks, :form_schema => klass.params_for_create(ems))
+        options_by_ems_id
+      elsif params.key?("id")
+        options_by_id
       else
-        cloud_network = resource_search(params["id"], :cloud_networks, CloudNetwork)
-        render_options(:cloud_networks, :form_schema => cloud_network.params_for_edit)
+        super
       end
     end
   end
--- a/app/controllers/api/cloud_networks_controller.rb
+++ b/app/controllers/api/cloud_networks_controller.rb
@@ -9,5 +9,22 @@ module Api
       task_id = cloud_network.delete_cloud_network_queue(User.current_user.userid)
       action_result(true, "Deleting #{cloud_network.name}", :task_id => task_id)
     end
+
+    def options
+      return super unless params.keys.include_any?(%w[id ems_id])
+
+      if params.key?("ems_id")
+        ems = resource_search(params["ems_id"], :ext_management_systems, ExtManagementSystem)
+        raise BadRequestError, "No Cloud Network support for - #{ems.class}" unless defined?(ems.class::CloudNetwork)
+
+        klass = ems.class::CloudNetwork
+        raise BadRequestError, "No DDF specified for - #{klass}" unless klass.respond_to?(:params_for_create)
+
+        render_options(:cloud_networks, :form_schema => CloudNetwork.class_by_ems(ems).params_for_create(ems))
+      else
+        cloud_network = resource_search(params["id"], :cloud_networks, CloudNetwork)
+        render_options(:cloud_networks, :form_schema => cloud_network.params_for_edit)
+      end
+    end
   end
 end
--- a/spec/requests/cloud_networks_spec.rb
+++ b/spec/requests/cloud_networks_spec.rb
@@ -97,4 +97,35 @@ RSpec.describe 'Cloud Networks API' do
       expect(response).to have_http_status(:ok)
     end
   end
+
+  describe 'OPTIONS /api/cloud_networks' do
+    it 'returns a DDF schema for add when available via OPTIONS' do
+      zone = FactoryBot.create(:zone, :name => "api_zone")
+      provider = FactoryBot.create(:ems_network, :zone => zone)
+
+      allow(provider.class::CloudNetwork).to receive(:params_for_create).and_return('foo')
+
+      options(api_cloud_networks_url, :params => {:ems_id => provider.id})
+      options("#{api_cloud_networks_url}?ems_id=#{provider.id}")
+
+      expect(response.parsed_body['data']['form_schema']).to eq('foo')
+      expect(response).to have_http_status(:ok)
+    end
+  end
+
+  describe 'OPTIONS /api/cloud_networks/:id' do
+    it 'returns a DDF schema for edit when available via OPTIONS' do
+      provider = FactoryBot.create(:ems_amazon_with_cloud_networks)
+
+      cloud_network = provider.cloud_networks.first
+
+      allow(CloudNetwork).to receive(:find).with(cloud_network.id.to_s).and_return(cloud_network)
+      allow(Rbac).to receive(:filtered_object).and_return(cloud_network)
+      expect(cloud_network).to receive(:params_for_edit).and_return('foo')
+      options("#{api_cloud_networks_url}/#{cloud_network.id}")
+
+      expect(response.parsed_body['data']['form_schema']).to eq('foo')
+      expect(response).to have_http_status(:ok)
+    end
+  end
 end
--- a/spec/requests/cloud_networks_spec.rb
+++ b/spec/requests/cloud_networks_spec.rb
@@ -105,7 +105,6 @@ RSpec.describe 'Cloud Networks API' do
 
       allow(provider.class::CloudNetwork).to receive(:params_for_create).and_return('foo')
 
-      options(api_cloud_networks_url, :params => {:ems_id => provider.id})
       options("#{api_cloud_networks_url}?ems_id=#{provider.id}")
 
       expect(response.parsed_body['data']['form_schema']).to eq('foo')


agrare:
We need to standardize how we want to handle provider defined DDF params (`params_for_create` with an instance_id vs. standalone `params_for_update`) and how we access those from the API (`OPTIONS /collection?instance_id=:id` VS `OPTIONS /collection/:id`)

We happen to have two pending PRs doing this two different ways: https://github.com/ManageIQ/manageiq-api/pull/1056

cc @Fryguy

agrare:
If we're getting `OPTIONS` for a record, prefer to do `OPTIONS /api/cloud_networks/:id` rather than `OPTIONS /api/cloud_networks?cloud_network_id=:id`

Splitting up `CloudNetwork#params_for_edit` from `CloudNetwork.params_for_create` is also more obvious, calling params_for_create with an optional instance ID is confusing.  The EMS params_for_create gets away with this because nothing is not able to be edited but that doesn't hold true for all of these types.

NickLaMuro (changes requested):
@GilbertCherrie Can you please create a global `.gitignore` file:

https://docs.github.com/en/get-started/getting-started-with-git/ignoring-files#configuring-ignored-files-for-all-repositories-on-your-computer

and add the following files to it:

- `.yarn-integrity`
- `.DS_Store`
- `yarn.lock`

Also, please remove said files from your commits.

GilbertCherrie:
@agrare The params for edit would need the ems_id and cloud_network_id passed to it so should I call it as           `API.options(`/api/cloud_networks/${ems_id}?cloud_network_id=${cloudNetworkId}`)` and for create I would just do
`API.options(`/api/cloud_networks/${ems_id}`?

agrare:
@GilbertCherrie params_for_edit is going to be an instance method so you don't need to give it either of those as params because they are available as part of the cloud_network instance.

E.g. if you call `OPTIONS /api/cloud_networks/${cloudNetworkId}` the EMS is available to the object as `ext_management_system` and cloudNetworkId is just `id`

GilbertCherrie:
> @GilbertCherrie params_for_edit is going to be an instance method so you don't need to give it either of those as params because they are available as part of the cloud_network instance.
> 
> E.g. if you call `OPTIONS /api/cloud_networks/${cloudNetworkId}` the EMS is available to the object as `ext_management_system` and cloudNetworkId is just `id`

So I don't need to pass anything to params_for_edit and it would have access to ems and the id?

agrare:
> So I don't need to pass anything to params_for_edit and it would have access to ems and the id?

You're loading the record [[here]](https://github.com/ManageIQ/manageiq-api/pull/1058/files#diff-4c387b2b1fa70c078e35f28c11fa2dea10961698eb5db17d745d3c937e31a09bR27), that means that when you call the instance method `#params_for_edit` you are in the context of the instance aka you have access to its attributes.  The ems_id is one attribute which is equivalent to the `${ems_id}` that you were passing in.

GilbertCherrie:
> > So I don't need to pass anything to params_for_edit and it would have access to ems and the id?
> 
> You're loading the record [[here]](https://github.com/ManageIQ/manageiq-api/pull/1058/files#diff-4c387b2b1fa70c078e35f28c11fa2dea10961698eb5db17d745d3c937e31a09bR27), that means that when you call the instance method `#params_for_edit` you are in the context of the instance aka you have access to its attributes. The ems_id is one attribute which is equivalent to the `${ems_id}` that you were passing in.

I think I understand, but one last point of confusion for me is how does the api code differentiate between ems_id and id if the API call is simply done as OPTIONS /api/cloud_networks/${ems_id}. The {ems_id} would just be a number so how would it know if this is an ems_id or cloud_network_id?

agrare:
> api code differentiate between ems_id and id if the API call is simply done as OPTIONS /api/cloud_networks/${ems_id}.

When you call `OPTIONS /api/cloud_networks/:id` the API just sees `/api/cloud_networks/1` for example.  The convention is that number is the id of the record in the (in this case) cloud_networks table.  That is the equivalent of doing `CloudNetwork.find(1)` in a rails console (which actually is what you are doing here: https://github.com/ManageIQ/manageiq-api/pull/1058/files#diff-4c387b2b1fa70c078e35f28c11fa2dea10961698eb5db17d745d3c937e31a09bR27

Remember you aren't passing ems_id to this call, you're passing cloudNetworkId.

If you did `GET /api/cloud_volumes/12` that would be equivalent to `CloudVolume.find(12)`.

The API doesn't know anything about the ems_id.  You can get the ems_id by virtue of the fact that you're already working with a single record, and that record has an ems_id.

app/controllers/api/cloud_networks_controller.rb
```diff
@@ -9,5 +9,22 @@ def delete_resource(type, resource_id, _data = {})
       task_id = cloud_network.delete_cloud_network_queue(User.current_user.userid)
       action_result(true, "Deleting #{cloud_network.name}", :task_id => task_id)
     end
+
+    def options
+      return super unless params.keys.include_any?(%w[id ems_id])
+
+      if params.key?("ems_id")
+        ems = ExtManagementSystem.find(params["ems_id"])
```
agrare, line 17: Should use the `resource_search` method here: `resource_search(params["ems_id"], :ext_management_systems, ExtManagementSystem)`

app/controllers/api/cloud_networks_controller.rb
```diff
@@ -9,5 +9,22 @@ def delete_resource(type, resource_id, _data = {})
...
+        raise BadRequestError, "No Cloud Network support for - #{ems.class}" unless defined?(ems.class::CloudNetwork)
+
+        klass = ems.class::CloudNetwork
+        raise BadRequestError, "No DDF specified for - #{klass}" unless klass.respond_to?(:params_for_create)
+
+        render_options(:cloud_networks, :form_schema => CloudNetwork.class_by_ems(ems).params_for_create(ems))
+      else
+        cloud_network = CloudNetwork.find(params["id"])
```
agrare, line 25: Same thing here, `resource_search(params["id"], :cloud_networks, CloudNetwork)`

spec/requests/cloud_networks_spec.rb
```diff
@@ -97,4 +97,19 @@
...
+
+      allow(provider.class::CloudNetwork).to receive(:params_for_create).and_return('foo')
+
+      options(api_cloud_networks_url, :params => {:ems_id => provider.id})
+      options("#{api_cloud_networks_url}?ems_id=#{provider.id}")
+
+      expect(response.parsed_body['data']['form_schema']).to eq('foo')
+      expect(response).to have_http_status(:ok)
+    end
+  end
```
agrare, line 114: Should also add an `'OPTIONS /api/cloud_networks/:id` spec here

app/controllers/api/cloud_networks_controller.rb
GilbertCherrie: `ems = resource_search(params["ems_id"], :ems, ExtManagementSystem)`

Is this correct?

spec/requests/cloud_networks_spec.rb
GilbertCherrie: @agrare for this test case I currently have:
`
allow(provider.class::CloudNetwork).to receive(:params_for_edit).and_return('foo')

options(api_cloud_networks_url, :params => {})
options("#{api_cloud_networks_url}/#{provider.id}")

expect(response.parsed_body['data']['form_schema']).to eq('foo')
`
Not sure if I am setting this up wrong but is there any way I can get the cloud_network_id from provider because the way I have it now the test case is failing due to the options call not returning anything.

GilbertCherrie:
I believe failed tests are due to this PR: https://github.com/ManageIQ/manageiq-api/pull/1060 not being merged yet. Test cases pass locally with this PR included and should pass once it is merged.

agrare:
@GilbertCherrie just throw `Depends on: #1060` in the description of this PR

GilbertCherrie:
@agrare @Fryguy I saw this PR: https://github.com/ManageIQ/manageiq-api/pull/1060 get merged. If any of you have time would you be able to review and merge this PR as well?

agrare:
Seeing as this is a new API call, @Fryguy or @bdunne can you also review?

app/controllers/api/cloud_networks_controller.rb
```diff
@@ -9,5 +9,22 @@ def delete_resource(type, resource_id, _data = {})
       task_id = cloud_network.delete_cloud_network_queue(User.current_user.userid)
       action_result(true, "Deleting #{cloud_network.name}", :task_id => task_id)
     end
+
+    def options
+      return super unless params.keys.include_any?(%w[id ems_id])
+
+      if params.key?("ems_id")
+        ems = resource_search(params["ems_id"], :ext_management_systems, ExtManagementSystem)
+        raise BadRequestError, "No Cloud Network support for - #{ems.class}" unless defined?(ems.class::CloudNetwork)
```
agrare, line 18: I know this was just copied&pasted from elsewhere so not going to block this PR for it, but in general checking `supports?(:cloud_network)` on the manager would be better than check class names existing.

None of these features exist right now so I think it'd be better to add these and update the API in one PR set

app/controllers/api/cloud_networks_controller.rb
```diff
@@ -9,5 +9,22 @@ def delete_resource(type, resource_id, _data = {})
...
+
+        klass = ems.class::CloudNetwork
+        raise BadRequestError, "No DDF specified for - #{klass}" unless klass.respond_to?(:params_for_create)
```
agrare, line 21: We should be checking for `klass.supports?(:create)` instead of checking for `respond_to?(:params_for_create)` here

app/controllers/api/cloud_networks_controller.rb
```diff
@@ -9,5 +9,22 @@ def delete_resource(type, resource_id, _data = {})
...
     end
+
+    def options
+      return super unless params.keys.include_any?(%w[id ems_id])
+
+      if params.key?("ems_id")
+        ems = resource_search(params["ems_id"], :ext_management_systems, ExtManagementSystem)
+        raise BadRequestError, "No Cloud Network support for - #{ems.class}" unless defined?(ems.class::CloudNetwork)
+
+        klass = ems.class::CloudNetwork
```
agrare, line 20: `CloudNetwork.class_by_ems(ems)` is the preferred way of getting the class name

app/controllers/api/cloud_networks_controller.rb
```diff
@@ -9,5 +9,22 @@ def delete_resource(type, resource_id, _data = {})
...
+        raise BadRequestError, "No DDF specified for - #{klass}" unless klass.respond_to?(:params_for_create)
+
+        render_options(:cloud_networks, :form_schema => CloudNetwork.class_by_ems(ems).params_for_create(ems))
```
agrare, line 23: You already have the `klass` above so you can do `klass.params_for_create(ems)` here

agrare (commented):
Couple of minor things here

app/controllers/api/cloud_networks_controller.rb
GilbertCherrie: Which line is this referring to?

agrare: this one

```suggestion
        klass = CloudNetwork.class_by_ems(ems)
```

agrare:
@GilbertCherrie can you update the description of your UI PR to indicate that it depends on this one?

app/controllers/api/cloud_networks_controller.rb
```diff
@@ -15,12 +15,13 @@ def options
 
       if params.key?("ems_id")
         ems = resource_search(params["ems_id"], :ext_management_systems, ExtManagementSystem)
-        raise BadRequestError, "No Cloud Network support for - #{ems.class}" unless defined?(ems.class::CloudNetwork)
+        klass = CloudNetwork.class_by_ems(ems)
 
-        klass = ems.class::CloudNetwork
-        raise BadRequestError, "No DDF specified for - #{klass}" unless klass.respond_to?(:params_for_create)
+        raise BadRequestError, "No Cloud Network support for - #{ems.class}" unless klass.supports?(:cloud_network)
```
agrare, line 20: Sorry think I confused you, we don't have `cloud_networks` as a feature that we are tracking so you can't do `supports?(:cloud_network)` yet.

You can keep the `defined?(ems.class::CloudNetwork)` for now

agrare (approved):

app/controllers/api/cloud_networks_controller.rb
```diff
@@ -9,5 +9,23 @@ def delete_resource(type, resource_id, _data = {})
       task_id = cloud_network.delete_cloud_network_queue(User.current_user.userid)
       action_result(true, "Deleting #{cloud_network.name}", :task_id => task_id)
     end
+
+    def options
+      return super unless params.keys.include_any?(%w[id ems_id])
+
+      if params.key?("ems_id")
```
Fryguy, line 16: I think this might be cleaner as follows:

```ruby
def options
  if params.key?("ems_id")
    options_by_ems_id
  elsif params.key?("id")
    options_by_id
  else
    super
  end
end
```

Then break out these other 2 sections into their own private methods `options_by_ems_id` and `options_by_id`. 

WDYT?

Fryguy: (Maybe not for this PR, especially if this pattern was copy/pasted)

GilbertCherrie: I think splitting them up is cleaner, I can change it right now if you want it split up

app/controllers/api/cloud_networks_controller.rb
```diff
@@ -9,5 +9,23 @@ def delete_resource(type, resource_id, _data = {})
...
+        ems = resource_search(params["ems_id"], :ext_management_systems, ExtManagementSystem)
```
Fryguy, line 17: I can never remember, but does `resource_search` handle the RBAC?

app/controllers/api/cloud_networks_controller.rb
Fryguy: Go for it 😊

app/controllers/api/cloud_networks_controller.rb
agrare: Yes

Fryguy:
Overall LGTM - just have one question on the RBAC.

@NickLaMuro we also need your re-review.

app/controllers/api/cloud_networks_controller.rb
Fryguy: If the user cannot see it, do we get a nil back or does it instantly 404?

agrare: That I don't know

app/controllers/api/cloud_networks_controller.rb
```diff
@@ -9,5 +9,31 @@ def delete_resource(type, resource_id, _data = {})
       task_id = cloud_network.delete_cloud_network_queue(User.current_user.userid)
       action_result(true, "Deleting #{cloud_network.name}", :task_id => task_id)
     end
+
+    def options_by_ems_id
```
Fryguy, line 13: Can you make `options_by_ems_id` and `options_by_id` private? (Either prefix with `private` or put them in a `private` section at the bottom).

spec/requests/cloud_networks_spec.rb
```diff
@@ -97,4 +97,35 @@
...
+
+  describe 'OPTIONS /api/cloud_networks' do
+    it 'returns a DDF schema for add when available via OPTIONS' do
+      zone = FactoryBot.create(:zone, :name => "api_zone")
+      provider = FactoryBot.create(:ems_network, :zone => zone)
+
+      allow(provider.class::CloudNetwork).to receive(:params_for_create).and_return('foo')
+
+      options(api_cloud_networks_url, :params => {:ems_id => provider.id})
+      options("#{api_cloud_networks_url}?ems_id=#{provider.id}")
```
Fryguy, line 109: This feels like it should be 2 separate specs as opposed to calling one after the other.  I don't think the expects on lines 111-112 will pick up on anything set in the first call (or put another way, if you delete line 108, this test will still pass, so it's not actually being tested)

GilbertCherrie: I just tested it as 2 different specs, the spec with just line 108 failed. Do I keep it as 2 specs or leave it as 1 spec but remove line 108?

Fryguy: I would keep it as 2 specs, and then we have to figure out why the first line fails.  If this was a copy past, we'll have to figure out why this isn't working.

Offhand, I'm wondering if it's a String vs Symbol thing (i.e. `:ems_id` vs `"ems_id"`), since that's the key you are looking for, but I thought `params` handled that.

GilbertCherrie: @Fryguy I just checked the cloud_subnets_spec.rb which is where this line initially came from, 
`    options(api_cloud_subnets_url, :params => {:ems_id => provider.id})
      options("#{api_cloud_subnets_url}?ems_id=#{provider.id}")`
After repeating the same process of removing the second options call and only testing the first the test case fails, so I am wondering if possibly the options call is being used incorrectly. I also checked the string vs symbol issue and it did not fix the problem.

Fryguy: Interesting.  @agrare maybe you have some idea?

agrare: I have no idea why they call `options` twice but only check the response once

agrare: welp this is interesting

```
     def options
+      raise
       if params.key?("ems_id")
         options_by_ems_id
       elsif params.key?("id")
diff --git a/spec/requests/cloud_networks_spec.rb b/spec/requests/cloud_networks_spec.rb
index b0633efa..7387ec4f 100644
--- a/spec/requests/cloud_networks_spec.rb
+++ b/spec/requests/cloud_networks_spec.rb
@@ -106,7 +106,7 @@ RSpec.describe 'Cloud Networks API' do
       allow(provider.class::CloudNetwork).to receive(:params_for_create).and_return('foo')
 
       options(api_cloud_networks_url, :params => {:ems_id => provider.id})
-      options("#{api_cloud_networks_url}?ems_id=#{provider.id}")
+      #options("#{api_cloud_networks_url}?ems_id=#{provider.id}")
```

That first one isn't invoking the overridden options call at all?

agrare:
```
response.parsed_body
{"error"=>{"kind"=>"bad_request", "message"=>"809: unexpected token at 'ems_id=5000000000028'", "klass"=>"JSON::ParserError", 
```

GilbertCherrie: @agrare I have been trying out different ways of calling this and none seem to work:

`options(api_cloud_networks_url, :params => {:ems_id => provider.id})`

This line is returning a 400 error, but when I remove the ems_id and leave params as empty (:params => {}) then it returns a 200 response. So is the problem with the ems_id?

agrare: What's wrong with `options("#{api_cloud_networks_url}?ems_id=#{provider.id}")`

GilbertCherrie: That one works, but @Fryguy wanted me to split up the 2 options call into separate specs so one calls this as well:
`options(api_cloud_networks_url, :params => {:ems_id => provider.id})`

agrare: I'm wondering if that is even valid, I don't know where that came from or if it is expected to work.

Maybe a question for someone deeper into our API than me.

GilbertCherrie: I saw it in cloud_subnets_spec.rb so I used that as a starting point for my test cases. When I tried to put it as a separate  test case it failed and then I went back to cloud_subnets_spec.rb and tried the test case with only this options call and saw the same result of the test case failing. So I am wondering if we are either using `options(api_cloud_networks_url, :params => {:ems_id => provider.id})` incorrectly or if it is simply not expected to work.

agrare: > So I am wondering if we are either using options(api_cloud_networks_url, :params => {:ems_id => provider.id}) incorrectly or if it is simply not expected to work.

Yes that's exactly what I'm wondering as well

GilbertCherrie: @Fryguy would it be possible to leave it as just 1 test case with the working api call:
`options("#{api_cloud_networks_url}?ems_id=#{provider.id}")`?

Fryguy: Yeah, for now let's leave as just 1 working test case, and then can you open up a separate GitHub issue for the other one that we can remember to go back and investigate later?

GilbertCherrie: I made the change to the spec file and will open up the issue now.

Fryguy (approved):

NickLaMuro (approved):

Fryguy:
Backported to `morphy` in commit b6713cede8856ae07b95eb8017398123b97e8ea7.

```text
commit b6713cede8856ae07b95eb8017398123b97e8ea7
Date:   Fri Jul 30 11:36:39 2021 -0400

    Merge pull request #1058 from GilbertCherrie/add_options_call_for_cloud_network
    
    Added options call for cloud networks
    
    (cherry picked from commit f1fa1d853e2bb3432e7eb875a8de6e44d618b6c9)
```

PR #1059, opened 2021-07-22
[WIP] [POC][Api::Filter] Add AND filter priority

NickLaMuro:
This is a proof of concept trying to fix the current implementation of `AND`/`OR` filtering in the API.

Probably going to abandon this PR, as probably implementing a `IN` operator is a simpler solution for the use case here.  But pushing it up as something to consider refining in the future if something like this seems necessary and can't be solved in another way.

Before
------

For example:

```
filter[]=col1=foo
&filter[]=col2=bar
&filter[]=or+col2=baz
&filter[]=col3=bar
&filter[]=or+col3=baz
```

Currently evaluates to:

```sql
WHERE (col1 = 'foo' AND col2 = 'baz' AND col3 = 'baz')
   OR col2 = 'baz'
   OR col3 = 'baz'
```

Which is confusing and I don't think obvious to any user.

After
-----

Currently this implementation isn't much more useful as you only can combine all `OR` statements together.  The previous implementation still exists, but favoring `AND` instead of `OR` on the top level can be triggered by passing `filtering_options=and_priority` in the query params.

Effectively the above evaluates to:

```sql
WHERE col1 = 'foo'
  AND (col2 = 'bar' OR col2 = 'baz' OR col3 = 'bar' OR col3 = 'baz')
```

Instead of what is probably more useful is some sort of grouping by column:

```sql
WHERE col1 = 'foo'
  AND (col2 = 'bar' OR col2 = 'baz')
  AND (col3 = 'bar' OR col3 = 'baz')
```

### DIFF
--- a/app/controllers/api/base_controller/parameters.rb
+++ b/app/controllers/api/base_controller/parameters.rb
@@ -24,7 +24,7 @@ module Api
       #
       def filter_param(klass)
         return nil if params['filter'].blank?
-        Filter.parse(params["filter"], klass)
+        Filter.parse(params["filter"], klass, params["filter_options"])
       end
 
       def by_tag_param
--- a/lib/api/filter.rb
+++ b/lib/api/filter.rb
@@ -17,15 +17,16 @@ module Api
       "!~"  => {:default => "REGULAR EXPRESSION DOES NOT MATCH"},
     }.freeze
 
-    attr_reader :filters, :model, :and_expressions, :or_expressions
+    attr_reader :filters, :model, :options, :and_expressions, :or_expressions
 
-    def self.parse(filters, model)
-      new(filters, model).parse
+    def self.parse(filters, model, options=nil)
+      new(filters, model, options).parse
     end
 
-    def initialize(filters, model)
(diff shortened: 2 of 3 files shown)


NickLaMuro:
Closing in favor of https://github.com/ManageIQ/manageiq-api/pull/1061

kbrock:
I do think this PR has validity.

But in order to fully fix this, the parameters would need to be order dependent and the whole `and_expression` and `or_expression` that take the parameters out of context would need to be changed. For an expression with multiple operators that have different precedence, you can not reorder the operations. You probably also need a way to modify the order of operations (e.g. use of `()`)

For backwards compatibility, we'd possibly need 2 parsing modes. Seems like a great big headache.

I agree that a punt here is our best scenario. The other PR is much simpler and addresses our most common use case.

NickLaMuro:
What @kbrock said (for once, with more words than myself!).

Again, felt this PR had validity as showing a way we could handle different modes of filtering without changing the existing behavior, but the end result still didn't give what we wanted, so it makes more sense to close for now.

Can be re-opened if we find a use for it.

PR #1060, opened 2021-07-26
Allow OPTIONS on /collection/:id for DDF resource params

agrare:
Some cases require requesting options on an individual resource (e.g. DDF parameters for editing a specific resource).  These parameters might be different based on that particular record.

Currently DDF parameters for "create" are returned by calling `OPTIONS /collection`.

This enables DDF parameters for "update" by calling `OPTIONS /collection/:id`

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -10,7 +10,7 @@ Rails.application.routes.draw do
 
     Api::ApiConfig.collections.each do |collection_name, collection|
       # OPTIONS action for each collection
-      match collection_name.to_s, :controller => collection_name, :action => :options, :via => :options, :as => nil
+      match "#{collection_name}(/:id)", :controller => collection_name, :action => :options, :via => :options, :as => nil
 
       collection_name_pluralized, resource_name = Api::Routing.inflections_for_named_route_helpers(collection_name.to_s)
 


app/controllers/api/cloud_networks_controller.rb
```diff
@@ -9,5 +9,17 @@ def delete_resource(type, resource_id, _data = {})
...
+      return super unless %w[id ems_id].include_any?(params.keys)
+
+      if params.key?("ems_id")
+        ems = ExtManagementSystem.find(params["ems_id"])
+        render_options(:cloud_networks, :form_schema => CloudNetwork.class_by_ems(ems).params_for_create(ems))
+      else
+        cloud_network = CloudNetwork.find(params["id"])
+        render_options(:cloud_networks, :form_schema => cloud_network.params_for_edit)
+      end
+    end
```
agrare, line 23: NOTE: this is an example of how class and instance DDF params can be returned by calling `OPTIONS /api/cloud_networks` and `OPTIONS /api/cloud_networks/:id`

agrare:
Ref: https://github.com/ManageIQ/manageiq-api/pull/1058#issuecomment-885878884

app/controllers/api/cloud_networks_controller.rb
```diff
@@ -9,5 +9,17 @@ def delete_resource(type, resource_id, _data = {})
       task_id = cloud_network.delete_cloud_network_queue(User.current_user.userid)
       action_result(true, "Deleting #{cloud_network.name}", :task_id => task_id)
     end
+
+    def options
+      return super unless params.keys.include_any?(%w[id ems_id])
+
+      if params.key?("ems_id")
+        ems = ExtManagementSystem.find(params["ems_id"])
```
Fryguy, line 17: Does this need RBAC? (same with the find call below.

agrare: Yes need to convert over to resource_search just nailing down the type / collection_type params since they're not passed in

agrare: This is also just an example, if we're okay with the routes match I'll drop this commit from the PR since it has its own PR https://github.com/ManageIQ/manageiq-api/pull/1058

agrare:
Hey @Fryguy can you re-review?

Fryguy:
Adding lasker and morphy yes because we want to backport #1058

Fryguy:
only marked lasker because the other one is marked as lasker - figured we'd discuss at triage.

chessbyte:
Removed Lasker from both - I had labeled it Lasker before there was a Morphy label.

Fryguy:
Backported to `morphy` in commit 2e96909ad080d368a81496b7b681e0bf71558839.

```text
commit 2e96909ad080d368a81496b7b681e0bf71558839
Date:   Wed Jul 28 12:15:32 2021 -0400

    Merge pull request #1060 from agrare/allow_instance_options
    
    Allow OPTIONS on /collection/:id for DDF resource params
    
    (cherry picked from commit e9885a7b3316ba1ab13f4b33e9e9a71285092d95)
```

PR #1061, opened 2021-07-26
[Api::Filter] Allow filtering '=' via arrays

NickLaMuro:
Adds functionality to the API to allow filtering a particular column by one of many types of values.  This could be particularly useful when filtering via a `type` column that is an STI column.

The SQL (functionally) that would be generated from the filter would be the following:

```sql
SELECT *
  FROM miq_requests
 WHERE (type = "MiqProvisionRequest" OR type = "VmMigrateRequest")
```

This avoids having to make changes to `MiqExpression` that would allow for something similar to how `where(:type => [a,b,c])` works already, and also ensures that we are using something that can be converted into SQL on the backend.

Links
-----

* Alternative proposal PR that doesn't quite work:  https://github.com/ManageIQ/manageiq-api/pull/1059

QA
--

I tested this using the following script:

```ruby
# Usage:  ruby fetch_requests.rb

require 'json'
require 'net/http'
require 'pp'

# Modify these vars if you want to change dates and such
days_ago = 7
states   = %w[pending_approval approved denied]
types    = %w[
               MiqProvisionConfiguredSystemRequest
               MiqProvisionRequest
               OrchestrationStackRetireRequest
               PhysicalServerProvisionRequest
               PhysicalServerFirmwareUpdateRequest
               ServiceRetireRequest
               ServiceReconfigureRequest
               ServiceTemplateProvisionRequest
               VmCloudReconfigureRequest
               VmMigrateRequest
               VmReconfigureRequest
               VmRetireRequest
           ]


miq_uri          = URI("http://localhost:3000")
http             = Net::HTTP.new(miq_uri.host, miq_uri.port)
headers          = { "accept" => "application/json" }

auth_hdr         = headers.merge("authorization" => "Basic " + ["admin:smartvm"].pack('m0'))
token            = JSON.parse(http.get("/api/auth", auth_hdr).body)["auth_token"]

req_hdrs         = headers.merge('X-Auth-Token' => token.to_s)
params           = URI.encode_www_form :attributes        => ["job_plays", "stdout"],
                                       :format_attributes => "stdout=html"

                   # equals:  'created_on>2000-01-01 00:00:00'

params           = "filter[]=created_on%3E#{(Time.now - days_ago * 24 * 60 * 60).to_s.gsub(/ /, '+')}"
                   # equals   'approval_state =~ /pending_approval|approved|denied/'
params          += "&filter[]=approval_state=[#{states.join(",")}]"
                   # equals   'approval_state =~ /MiqProvisionRequest|OrchestrationStackRetireRequest|.../'
params          += "&filter[]=type=[#{types.join(",")}]"
                   # equals:  'reason =~ Auto'
params          += "&filter[]=reason=~Auto"


services_url     = "/api/requests?expand=resources&#{params}"
response         = http.get(services_url, req_hdrs)

pp JSON.parse(response.body)
```

And confirmed in the log that I was getting a proper query in the database using these filters:

```
[----] D, [2021-07-26T11:54:18.342990 #97861:3fd074e1a29c] DEBUG -- :    (0.3ms)  SELECT "miq_product_features"."identifier" FROM "miq_product_features" INNER JOIN "miq_roles_features" ON "miq_product_features"."id" = "miq_roles_features"."miq_product_feature_id" WHERE "miq_roles_features"."miq_user_role_id" = $1  [["miq_user_role_id", 1]]
[----] D, [2021-07-26T11:54:18.348058 #97861:3fd074e1a29c] DEBUG -- :   MiqRequest Load (0.3ms)  SELECT "miq_requests".* FROM "miq_requests" WHERE (("miq_requests"."created_on" > '2021-07-19 16:54:18' AND (("miq_requests"."approval_state" = 'pending_approval' OR "miq_requests"."approval_state" = 'approved') OR "miq_requests"."approval_state" = 'denied') AND ((((((((((("miq_requests"."type" = 'MiqProvisionConfiguredSystemRequest' OR "miq_requests"."type" = 'MiqProvisionRequest') OR "miq_requests"."type" = 'OrchestrationStackRetireRequest') OR "miq_requests"."type" = 'PhysicalServerProvisionRequest') OR "miq_requests"."type" = 'PhysicalServerFirmwareUpdateRequest') OR "miq_requests"."type" = 'ServiceRetireRequest') OR "miq_requests"."type" = 'ServiceReconfigureRequest') OR "miq_requests"."type" = 'ServiceTemplateProvisionRequest') OR "miq_requests"."type" = 'VmCloudReconfigureRequest') OR "miq_requests"."type" = 'VmMigrateRequest') OR "miq_requests"."type" = 'VmReconfigureRequest') OR "miq_requests"."type" = 'VmRetireRequest')))
[----] D, [2021-07-26T11:54:18.352262 #97861:3fd074e1a29c] DEBUG -- :   MiqRequest Inst Including Associations (0.1ms - 1rows)
[----] D, [2021-07-26T11:54:18.354745 #97861:3fd074e1a29c] DEBUG -- :   MiqApproval Load (0.2ms)  SELECT "miq_approvals".* FROM "miq_approvals" WHERE "miq_approvals"."miq_request_id" = $1 ORDER BY "miq_approvals"."id" ASC LIMIT $2  [["miq_request_id", 3], ["LIMIT", 1]]
[----] D, [2021-07-26T11:54:18.354951 #97861:3fd074e1a29c] DEBUG -- :   MiqApproval Inst Including Associations (0.1ms - 1rows)
[----] D, [2021-07-26T11:54:18.356607 #97861:3fd074e1a29c] DEBUG -- :   CACHE MiqRequest Load (0.0ms)  SELECT "miq_requests".* FROM "miq_requests" WHERE (("miq_requests"."created_on" > '2021-07-19 16:54:18' AND (("miq_requests"."approval_state" = 'pending_approval' OR "miq_requests"."approval_state" = 'approved') OR "miq_requests"."approval_state" = 'denied') AND ((((((((((("miq_requests"."type" = 'MiqProvisionConfiguredSystemRequest' OR "miq_requests"."type" = 'MiqProvisionRequest') OR "miq_requests"."type" = 'OrchestrationStackRetireRequest') OR "miq_requests"."type" = 'PhysicalServerProvisionRequest') OR "miq_requests"."type" = 'PhysicalServerFirmwareUpdateRequest') OR "miq_requests"."type" = 'ServiceRetireRequest') OR "miq_requests"."type" = 'ServiceReconfigureRequest') OR "miq_requests"."type" = 'ServiceTemplateProvisionRequest') OR "miq_requests"."type" = 'VmCloudReconfigureRequest') OR "miq_requests"."type" = 'VmMigrateRequest') OR "miq_requests"."type" = 'VmReconfigureRequest') OR "miq_requests"."type" = 'VmRetireRequest')))
[----] D, [2021-07-26T11:54:18.360629 #97861:3fd074e1a29c] DEBUG -- :   MiqRequest Inst Including Associations (0.1ms - 1rows)
```

### DIFF
--- a/lib/api/filter.rb
+++ b/lib/api/filter.rb
@@ -17,21 +17,20 @@ module Api
       "!~"  => {:default => "REGULAR EXPRESSION DOES NOT MATCH"},
     }.freeze
 
-    attr_reader :filters, :model
+    attr_reader :filters, :model, :and_expressions, :or_expressions
 
     def self.parse(filters, model)
       new(filters, model).parse
     end
 
     def initialize(filters, model)
-      @filters = filters
-      @model = model
+      @filters         = filters
+      @model           = model
+      @and_expressions = []
+      @or_expressions  = []
     end
 
     def parse
-      and_expressions = []
-      or_expressions = []
-
       filters.select(&:present?).each do |filter|
         parsed_filter = parse_filter(filter)
         *associations, attr = parsed_filter[:attr].split(".")
--- a/lib/api/filter.rb
+++ b/lib/api/filter.rb
@@ -43,9 +43,15 @@ module Api
         end
 
         associations.map! { |assoc| ".#{assoc}" }
+
         field = "#{model.name}#{associations.join}-#{attr}"
-        target = parsed_filter[:logical_or] ? or_expressions : and_expressions
-        target << {parsed_filter[:operator] => {"field" => field, "value" => parsed_filter[:value]}}
+        expr  = {parsed_filter[:operator] => {"field" => field, "value" => parsed_filter[:value]}}
+
+        if parsed_filter[:logical_or]
+          or_expressions << expr
+        else
+          and_expressions << expr
+        end
       end
 
       and_part = and_expressions.one? ? and_expressions.first : {"AND" => and_expressions}
--- a/lib/api/filter.rb
+++ b/lib/api/filter.rb
@@ -53,8 +53,6 @@ module Api
         end
       end
 
-      and_part = and_expressions.one? ? and_expressions.first : {"AND" => and_expressions}
-      composite_expression = or_expressions.empty? ? and_part : {"OR" => [and_part, *or_expressions]}
       MiqExpression.new(composite_expression).tap do |expression|
         raise BadRequestError, "Must filter on valid attributes for resource" unless expression.valid?
       end
@@ -117,6 +115,11 @@ module Api
       end
 
       {:logical_or => logical_or, :operator => method, :attr => filter_attr, :value => filter_value}
+    end
+
+    def composite_expression
+      and_part = and_expressions.one? ? and_expressions.first : {"AND" => and_expressions}
+      or_expressions.empty? ? and_part : {"OR" => [and_part, *or_expressions]}
     end
 
     def target_class(klass, reflections)
--- a/lib/api/filter.rb
+++ b/lib/api/filter.rb
@@ -6,7 +6,7 @@ module Api
       ">="  => {:default => ">="},
       "<"   => {:default => "<", :datetime => "BEFORE"},
       ">"   => {:default => ">", :datetime => "AFTER"},
-      "="   => {:default => "=", :datetime => "IS", :regex => "REGULAR EXPRESSION MATCHES", :string_set => "includes all", :null => "IS NULL"},
+      "="   => {:default => "=", :datetime => "IS", :regex => "REGULAR EXPRESSION MATCHES", :string_set => "includes all", :array => "=", :null => "IS NULL"},
 
       # string-only matching, use quotes
       "=="  => {:default => "="},
@@ -44,7 +44,12 @@ module Api
         associations.map! { |assoc| ".#{assoc}" }
 
         field = "#{model.name}#{associations.join}-#{attr}"
-        expr  = {parsed_filter[:operator] => {"field" => field, "value" => parsed_filter[:value]}}
+        op    = parsed_filter[:operator]
+        expr  = if parsed_filter[:value].kind_of?(Array)
+                  {"OR" => parsed_filter[:value].map { |val| single_expression(field, op, val) }}
+                else
+                  single_expression(field, op, parsed_filter[:value])
+                end
 
         if parsed_filter[:logical_or]
           or_expressions << expr
@@ -85,6 +90,13 @@ module Api
       str_method = is_regex ? methods[:regex] : methods[:default]
 
       filter_value, method = case filter_value
+                             when /^\[(.*)\]$/
+                               unless methods[:array]
+                                 raise BadRequestError, "Unsupported operator for arrays: #{operator}"
+                               end
+
+                               array_value = $1.split(",")
+                               [array_value, methods[:array]]
                              when /^'(.*)'$/, /^"(.*)"$/
                                unquoted_filter_value = $1
                                if column_type(model, filter_attr) == :string_set && methods[:string_set]
@@ -120,6 +132,10 @@ module Api
     def composite_expression
       and_part = and_expressions.one? ? and_expressions.first : {"AND" => and_expressions}
       or_expressions.empty? ? and_part : {"OR" => [and_part, *or_expressions]}
+    end
+
+    def single_expression(field, operator, value)
+      {operator => {"field" => field, "value" => value}}
     end
 
     def target_class(klass, reflections)
--- a/spec/lib/api/filter_spec.rb
+++ b/spec/lib/api/filter_spec.rb
@@ -45,6 +45,20 @@ RSpec.describe Api::Filter do
       expect(actual.exp).to eq(expected)
     end
 
+    it "supports multiple possible values using arrays with =" do
+      filters  = ["name=[foo,bar,baz]"]
+      actual   = described_class.parse(filters, Vm)
+      expected = {
+        "OR" => [
+          {"=" => {"field" => "Vm-name", "value" => "foo"}},
+          {"=" => {"field" => "Vm-name", "value" => "bar"}},
+          {"=" => {"field" => "Vm-name", "value" => "baz"}}
+        ]
+      }
+
+      expect(actual.exp).to eq(expected)
+    end
+
     it "supports NULL/nil equality test via =" do
       filters = ["retired=NULL"]
 


NickLaMuro:
@miq-bot assign @Fryguy 

Please review and let us know what you think.  This is to help Melody with a form conversion, and doing this avoids having to bring everything back and filter in Ruby, but also avoids needing to touch `MiqExpression` to allow for this change (higher risk).


cc @kbrock @MelsHyrule

lib/api/filter.rb
```diff
@@ -114,6 +128,15 @@ def parse_filter(filter)
       {:logical_or => logical_or, :operator => method, :attr => filter_attr, :value => filter_value}
     end
 
+    def composite_expression
```
kbrock, line 131: not a big fan. I liked parse only having local variables.

can you extract this logic while keeping the variables out of the global scope?
`composite_expression(and_expressions, or_expressions)`

lib/api/filter.rb
```diff
@@ -82,6 +90,12 @@ def parse_filter(filter)
       str_method = is_regex ? methods[:regex] : methods[:default]
 
       filter_value, method = case filter_value
+                             when /^\[(.*)\]$/
+                               unless methods[:array]
+                                 raise BadRequestError, "Unsupported operator for arrays: #{operator}"
+                               end
+
+                               [$1.split(","), methods[:array]]
```
kbrock, line 98: I like how `unquoted_filter_value` introduced a variable close to the regular expression. So it is very clear the `$1` reference.

Although, with that case, it is nicer because the value can be used asis.

Can you do something similar here?

`array_contents = $1.split(",")`

lib/api/filter.rb
```diff
@@ -6,7 +6,7 @@ class Filter
       ">="  => {:default => ">="},
       "<"   => {:default => "<", :datetime => "BEFORE"},
       ">"   => {:default => ">", :datetime => "AFTER"},
-      "="   => {:default => "=", :datetime => "IS", :regex => "REGULAR EXPRESSION MATCHES", :string_set => "includes all", :null => "IS NULL"},
+      "="   => {:default => "=", :datetime => "IS", :regex => "REGULAR EXPRESSION MATCHES", :string_set => "includes all", :array => "=", :null => "IS NULL"},
```
kbrock, line 9: The `:array => value` is basically a boolean here. I think all the cases are just going to use the original operator.

Does it make sense to use the values `AND` or `OR`. Those are the only real options I can see changing across `<` and others.

I'd imagine that all of these would use the same operator, but maybe `>` would require and `AND`

Otherwise, it is just acting as a boolean.

lib/api/filter.rb
```diff
@@ -43,13 +42,22 @@ def parse
...
+
+        field    = "#{model.name}#{associations.join}-#{attr}"
+        operator = parsed_filter[:operator]
+        expr     = if parsed_filter[:value].kind_of?(Array)
+                     {"OR" => parsed_filter[:value].map { |v| single_expression(field, operator, v) }}
+                   else
+                     single_expression(field, operator, parsed_filter[:value])
+                   end
+
+        if parsed_filter[:logical_or]
```
kbrock, line 54: this is so much nicer than the previous code.

lib/api/filter.rb
```diff
@@ -43,13 +42,22 @@ def parse
         end
 
         associations.map! { |assoc| ".#{assoc}" }
-        field = "#{model.name}#{associations.join}-#{attr}"
-        target = parsed_filter[:logical_or] ? or_expressions : and_expressions
-        target << {parsed_filter[:operator] => {"field" => field, "value" => parsed_filter[:value]}}
+
+        field    = "#{model.name}#{associations.join}-#{attr}"
+        operator = parsed_filter[:operator]
```
kbrock, line 47: nit: can we just use `parsed_filter[:operator]` in 2 places rather than introducing a variable?
(that way `field` will be an unchanged line as well)

kbrock (commented):
This looks very nice Nick.

I'm amused that I want to get rid of a double use variable and introduce a single use one. Those do not have a strong reaction from me.

Really don't like the introduction of the two global variables.
We can discuss if you want, but not a hill I'd die on.

lib/api/filter.rb
NickLaMuro: Humbly disagree and don't plan to change.  This is pretty idiomatic Ruby to use instance variables (not "global") like this and I think it makes definitions and callers less verbose and wordy.

For example, I picked `puma` as an example, and the first class I looked at does something pretty much like this with the first method I found:

https://github.com/puma/puma/blob/master/lib/puma/client.rb#L96-L98

In this case, the `@ready` var is called directly instead of using an `attr_reader`, but I think the general idea is the same.

* * *

So unless I get push back from others, I don't plan to change.

lib/api/filter.rb
NickLaMuro: Yeah, I am fine with that idea.  Was going with terse, but I can be a bit verbose if you prefer.

lib/api/filter.rb
NickLaMuro: I don't think I follow completely with your explanation here.  Can you elaborate?

* * *

That said, I was trying to keep this as small as possible to start, and iterate on it if we felt it was necessary.  Unless you think what you are suggesting might change how the end user interacts with it, I might hold of on supporting other operators with this for now.

lib/api/filter.rb
NickLaMuro: Agreed, I pulled a decent about of the commits from https://github.com/ManageIQ/manageiq-api/pull/1059 because I thought this needed to be done regardless if that purpose of that PR got in.

lib/api/filter.rb
NickLaMuro: Honestly, the lines were just getting very long with that didn't want to keep typing out `parsed_filter` when building these lines (where it made sense).

I could change it to `op` so that the `field = ...` line stays the same length though (effectively keeping it unchanged as well).

lib/api/filter.rb
kbrock: lol I wrote up something long here saying to ignore it.
my idea makes sense to me but not enough to elaborate.
Maybe some day over tea. Though we'd have better stuff to discuss like the weather.

lib/api/filter.rb
kbrock: changing to `op` may freak out rubocop. keep as is.

NickLaMuro: @kbrock uhhh... already changed, and `rubocop` seems cool with it.

lib/api/filter.rb
kbrock: lol. you've made changes like this before. it is your style.

the or expressions and and expressions is only valid within the context of `parse`
I also am not that fond of how the parsing is being performed, so elevating the local variables to globals further sets it in stone.

Can you be bribed to pass those 2 variables into the local private method `composite_expression` and do away with these ~~global~~ class variables?

NickLaMuro: >  ~~global~~ class variables

Instance variables... they are _**instance**_ variables...

kbrock: ugh. ^ what he said

lib/api/filter.rb
kbrock: thank you rubocop for being kind to my friend

Fryguy:
Backported to `morphy` in commit d596a80313bc6e6a462d09d53ce6b6b301999e12.

```text
commit d596a80313bc6e6a462d09d53ce6b6b301999e12
Date:   Fri Jul 30 10:27:03 2021 -0400

    Merge pull request #1061 from NickLaMuro/array_based_filtering
    
    [Api::Filter] Allow filtering '=' via arrays
    
    (cherry picked from commit b6e5172323285a9dc6c2caf5242f81285b50646b)
```

Issue #1064, opened 2021-08-05
Clear cloud network fields when choose is selected


GilbertCherrie:
The fields on the cloud network form should be cleared when choose is selected.

<img width="1059" alt="Screen Shot 2021-08-05 at 4 20 20 PM" src="https://user-images.githubusercontent.com/32444791/128416307-9d1f6aea-854c-460e-9c86-79f80c5ddcb6.png">

PR #1065, opened 2021-08-05
Added option to return no fields when no network manager is selected for cloud network form

GilbertCherrie:
Fixes: https://github.com/ManageIQ/manageiq-api/issues/1064

I tried to fix this issue on the UI code but after trying different methods it does not seem possible. It resulted in the UI updating faster than the api would return its data causing the cloud tenants field to have incorrect data. Another UI fix was possible but resulted in log errors. This was the only solution found that works and produces no errors.

Before:
Screen Shot 2021-08-05 at 4 20 20 PM

After:
Screen Shot 2021-08-05 at 4 21 13 PM

@miq-bot add_reviewer @Fryguy 
@miq-bot add-label bug

### DIFF
--- a/app/controllers/api/cloud_networks_controller.rb
+++ b/app/controllers/api/cloud_networks_controller.rb
@@ -11,14 +11,18 @@ module Api
     end
 
     private def options_by_ems_id
-      ems = resource_search(params["ems_id"], :ext_management_systems, ExtManagementSystem)
-      klass = CloudNetwork.class_by_ems(ems)
+      if params["ems_id"] == "nil"
+        render_options(:cloud_networks, :form_schema => {:fields => []})
+      else
+        ems = resource_search(params["ems_id"], :ext_management_systems, ExtManagementSystem)
+        klass = CloudNetwork.class_by_ems(ems)
 
-      raise BadRequestError, "No Cloud Network support for - #{ems.class}" unless defined?(ems.class::CloudNetwork)
+        raise BadRequestError, "No Cloud Network support for - #{ems.class}" unless defined?(ems.class::CloudNetwork)
 
-      raise BadRequestError, "No DDF specified for - #{klass}" unless klass.supports?(:create)
+        raise BadRequestError, "No DDF specified for - #{klass}" unless klass.supports?(:create)
 
-      render_options(:cloud_networks, :form_schema => klass.params_for_create(ems))
+        render_options(:cloud_networks, :form_schema => klass.params_for_create(ems))
+      end
     end
 
     private def options_by_id
--- a/spec/requests/cloud_networks_spec.rb
+++ b/spec/requests/cloud_networks_spec.rb
@@ -127,4 +127,13 @@ RSpec.describe 'Cloud Networks API' do
       expect(response).to have_http_status(:ok)
     end
   end
+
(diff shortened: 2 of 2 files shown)


Fryguy:
This feels like a hack and doesn't really belong on the API side.  Is this not something you can do on the form side?  cc @agrare Thoughts?

agrare:
Yeah this doesn't seem right at all.  If there is no network manager selected yet why ask for DDF parameters?

GilbertCherrie:
@kavyanekkalapu and I tried to fix this on the form side but the only solution we found produced a console error.

GilbertCherrie:
> Yeah this doesn't seem right at all. If there is no network manager selected yet why ask for DDF parameters?

The fields only show up after selecting a network manager and then selecting Choose again. On initial load when Choose is selected there is no fields there.

agrare:
Just read the issue, I thought this was on initial page load not when a network manager was "unselected"

Yeah definitely not an API concern should be fixed in the UI.

kavyanekkalapu:
@Fryguy @agrare Like @GilbertCherrie mentioned, We tried every possible way to fix from ui side and couldn't fix it, if we fix it in api like this pr, would it cause any further issues?

agrare:
This introduces a totally new concept to our API (the idea of an ID being equal to the string `"nil"`) as a workaround for an issue on the UI.

I just checked the provider create page and it has the same issue.  Will we have to add this to every single API that has DDF forms also?

I don't understand the UI code very well, but issuing an API call to return an empty array is the only way to clear out a set of selected fields on a UI screen.

kavyanekkalapu:
Yes same issue is present un subnets page and providers page, if this pr approved, we are thinking of making similar fix other two places.

Fryguy:
Yeah I'm a big no on this one - there has got to be a way to react to the event selection of the manager drop down and instead of issuing an API call, just hide all of the fields

Fryguy:
Is the problem that DDF itself just doesn't have a way to query "nothing"?

Another way to thing about this is, is there a "reset" button on the form and if so, can we trigger a reset when the dropdown selects the nil entry?

Fryguy:
Maybe @rvsia or @Hyperkid123 know a way to do this with DDF?

tl;dr There is an initial drop down that drives the rest of the form and determines what fields to show by querying the API to get the form details.  However, when the user selects the `<Choose>` entry, there's no API call technically, so how can we clear the form?

rvsia:
@Fryguy what about to just hide the rest of the form using [condition](https://data-driven-forms.org/schema/is) when there is no selected value?

GilbertCherrie:
I managed to figure out another new method of fixing this issue in the UI without using any new api code. This is my PR: https://github.com/ManageIQ/manageiq-ui-classic/pull/7815

I will hopefully get it tested/reviewed tomorrow but this new solution should be viable.

kavyanekkalapu:
> @Fryguy what about to just hide the rest of the form using [condition](https://data-driven-forms.org/schema/is) when there is no selected value?

We tried all this solutions. it is working on selecting `choose` and selecting provider, but bottom fields are not refreshing when we switch between providers. For now closing this pr. Tomorrow, Gilbert and i will see if we can find any other solutions.

PR #1070, opened 2021-09-09
Convert cloud volume operations to SupportsFeatureMixin

chessbyte:
Part of https://github.com/ManageIQ/manageiq/pull/21393

### DIFF
--- a/app/controllers/api/cloud_volumes_controller.rb
+++ b/app/controllers/api/cloud_volumes_controller.rb
@@ -7,8 +7,7 @@ module Api
 
       klass = CloudVolume.class_by_ems(ext_management_system)
 
-      validate = klass.validate_create_volume(ext_management_system)
-      raise validate[:message] unless validate[:available]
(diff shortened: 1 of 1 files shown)


app/controllers/api/cloud_volumes_controller.rb
```diff
@@ -7,8 +7,7 @@ def create_resource(_type, _id = nil, data = {})
 
       klass = CloudVolume.class_by_ems(ext_management_system)
 
-      validate = klass.validate_create_volume(ext_management_system)
-      raise validate[:message] unless validate[:available]
+      raise BadRequestError, ext_management_system.unsupported_reason(:cloud_volume_create) unless ext_management_system.supports?(:cloud_volume_create)
```
chessbyte, line 10: @agrare @Fryguy Lines 54-61 of this file look very fragile and error-prone.  Wondering if it should be reworked in this PR or a future one.

agrare (approved):

PR #1072, opened 2021-09-10
Fix retirement api for instances

GilbertCherrie:
Fixes the instances retirement form API.

Steps to reproduce bug:
1) Navigate to Compute -> Cloud -> Instances
Screen Shot 2021-09-14 at 10 19 27 AM

2) Select an instance then click Lifecycle and Set Retirement Dates
Screen Shot 2021-09-14 at 10 30 05 AM

3) Select Time Delay from Now from the first drop down menu.
Screen Shot 2021-09-14 at 10 30 54 AM

4) Enter any amount of hours, days, weeks or months and click submit.
Screen Shot 2021-09-14 at 10 31 13 AM

@miq-bot add-label bug
@miq-bot add_reviewer @Fryguy
@miq-bot add_reviewer @bdunne

### DIFF
--- a/config/api.yml
+++ b/config/api.yml
@@ -2075,6 +2075,10 @@
         :identifier: instance_reset
       - :name: set_ownership
         :identifier: vm_ownership
+      - :name: retire
+        :identifier: instance_retire_now
+      - :name: request_retire
+        :identifier: instance_retire
     :resource_actions:
       :get:
       - :name: read
@@ -2100,6 +2104,10 @@
         :identifier: instance_reset
       - :name: set_ownership
         :identifier: vm_ownership
+      - :name: retire
+        :identifier: instance_retire_now
+      - :name: request_retire
+        :identifier: instance_retire
     :snapshots_subcollection_actions:
       :get:
       - :name: read


config/api.yml
```diff
@@ -2075,6 +2075,12 @@
         :identifier: instance_reset
       - :name: set_ownership
         :identifier: vm_ownership
+      - :name: retire
+        :identifier:
+        - instance_retire_now
```
Fryguy, line 2080: Since there's only a single identifier this can just be inlined (same for all of the other sections in this diff as well)

```suggestion
        :identifier: instance_retire_now
```

Fryguy:
I'm not 100% sure if these are the right identifiers - maybe @jrafanie or @agrare knows?

jrafanie:
> I'm not 100% sure if these are the right identifiers - maybe @jrafanie or @agrare knows?

these seem correct...  vm/instance each have `*_retire` and `*_retire_now`.

jrafanie:
> > I'm not 100% sure if these are the right identifiers - maybe @jrafanie or @agrare knows?
> 
> these seem correct... vm/instance each have `*_retire` and `*_retire_now`.

Note, I don't know why instance vs. vm retirement is any different.  If you're able to see the "vm/instance" with rbac and you have retire permissions, you should be able to retire it.  If they wanted to exclude you from vm or service retirement, maybe they should not let you see the vm/service.  I don't know why it's so fine grained.

But this is outside the scope of this PR.  It's a pet peeve of mine that we duplicate the same "action" permission to include the category of the target object instead of letting RBAC handle the object categories and let product features be distinct actions.

Fryguy:
I guess what I don't understand , and maybe it's a naming problem, but IMO, retire and request_retire are basically the same thing, just one is now and the other is delayed, so why are there 2 different permissions?  To your point, that is also too fine grained.  if you can retire you can retire 🤷

Fryguy:
Backported to `morphy` in commit de4be6c323877cad04562b0773e31666321f8d48.

```text
commit de4be6c323877cad04562b0773e31666321f8d48
Date:   Wed Sep 22 10:20:59 2021 -0400

    Merge pull request #1072 from GilbertCherrie/fix_instances_retirement
    
    Fix retirement api for instances
    
    (cherry picked from commit f971f1bd42ebef07babf0199dcacf5740e3a2eef)
```

PR #1076, opened 2021-09-17
properly pass audit information for tasks

kbrock:
these now properly pass the `userid` instead of the `user.id`

they were not showing up in the tasks view because the tasks were not assigned to the correct user


kbrock:
I am considering adding a separate method for a standard delete. not doing as part of this as it is only a logging change

NickLaMuro (approved):

PR #1077, opened 2021-09-21
[WIP] refactor crud

kbrock:
Many of our API calls submit a request (via a task) to the ems to create, delete, or update a resource.

These all follow the same pattern of seeing if the ems supports the feature, sending off a task, and reporting the task status to the end user.

This started with network routers, but ended up working across most (all?) api controllers that generate an ems task/request.

### DIFF
--- a/app/controllers/api/auth_key_pairs_controller.rb
+++ b/app/controllers/api/auth_key_pairs_controller.rb
@@ -1,26 +1,14 @@
 module Api
   class AuthKeyPairsController < BaseController
-    def create_resource(_type, _id = nil, data = {})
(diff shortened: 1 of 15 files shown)


kbrock:
status update:

- this is not necessary for our network router changes
- need to change the specs to not be provider class specific. the provider specific test can be moved to the provider repos
- need to push these changes across the create, update, delete methods. (wip commit just shows where there is potential)

kbrock:
Next: think I'll pull out delete from here

kbrock:
closed in favor of #1083

PR #1079, opened 2021-09-23
Network routes options

kbrock:
requires:
- [ ] https://github.com/ManageIQ/manageiq-providers-openstack/pull/731

Introduce OPTIONS predicate for network routes

issues:

- if they ask for options for a resource that is not valid, is that a blank response or an error?
- Is there a better way to write these tests that do not depend upon openstack the provider?
- does it make sense to move something like this into the base `options`? (not sure if it is too opinionated around an ems telling us how to create and a resource telling us how to update)


kbrock:
@miq-bot cross_repo_tests including ManageIQ/manageiq-providers-openstack#731

kbrock:
@miq-bot cross_repo_tests ManageIQ/manageiq-providers-openstack#731

kbrock:
reduced the number of changed lines in the spec to resolve a merge conflict with another PR

kbrock:
@bdunne @GilbertCherrie This is ready. no longer depends upon changes in other repos.

kbrock:
robocop: I feel more comfortable with parens in those 2 calls even without args

kbrock:
push: using built in matchers. who knew?

spec/support/api/helpers.rb
```diff
@@ -265,6 +265,39 @@ def expect_forbidden_request
...
+            end
+
+          if model.kind_of?(Class) # typically create
+            expect(model).to(receive_this_message)
+          else
+            allow_any_instance_of(model.class).to(receive_this_message)
+          end
+        end
+
+        # when testing requests, ensure the model supports a certain attribute
```
kbrock, line 288: pulling this out into core

spec/support/api/helpers.rb
```diff
@@ -265,6 +265,39 @@ def expect_forbidden_request
         def select_attributes(attrlist)
           attrlist.sort.select { |attr| !::Api.encrypted_attribute?(attr) }
         end
+
+        # when testing options requests, ensure the base model responds in a certain way
+        # this is used for OPTIONS for DDF
+        def stub_options_support(model, action = :update, fields = [])
```
kbrock, line 271: this is api specific

spec/support/api/helpers.rb
kbrock: pulled into core

kbrock: (this class does not depend upon that type of checks, so it has been removed)

kbrock:
forced push:
 fixes for spec description and ensuring return status ok

PR #1080, opened 2021-09-24
Convert AuthKey operations to SupportsFeatureMixin

chessbyte:
Goes with https://github.com/ManageIQ/manageiq/pull/21457

### DIFF
--- a/app/controllers/api/auth_key_pairs_controller.rb
+++ b/app/controllers/api/auth_key_pairs_controller.rb
@@ -17,7 +17,7 @@ module Api
     def delete_resource(type, id, _data = {})
       delete_action_handler do
         key_pair = resource_search(id, type, collection_class(type))
-        raise "Delete not supported for #{key_pair.name}" unless key_pair.respond_to?(:delete_key_pair_queue)
+        raise "Delete not supported for #{key_pair.name}" unless key_pair.supports?(:delete)
 
         task_id = key_pair.delete_key_pair_queue(current_user.userid)
         action_result(true, "Deleting #{key_pair.name}", :task_id => task_id)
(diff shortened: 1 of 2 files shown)


chessbyte:
the API spec is failing [here](https://app.travis-ci.com/github/ManageIQ/manageiq-cross_repo-tests/jobs/540091764).  I need some help understanding why this worked before and is now broken.  @kbrock maybe you can take a look.

chessbyte:
Spec is failing because it is validating against an instance of `ManageIQ::Providers::CloudManager::AuthKeyPair` which, by definition, does not support delete whereas an instance of `ManageIQ::Providers::Amazon::CloudManager::AuthKeyPair` does.  @agrare ideas?

kbrock:
elsewhere I was using something like:

```ruby
      expect(akp.class).to receive(:supports?).with(:delete).and_return(true)
      # or
      expect_any_instance_of(akp.resource.class).to receive(:supports?).with(:key_pair_delete).and_return(true)
```

Currently you are using the second one / the ems version but as you know, I'm trying to nudge us towards the first

kbrock:
@chessbyte check out https://github.com/ManageIQ/manageiq-api/pull/1079

In helpers I added methods to help us with models support.
If you feel this needs to be used in all repos, I can move into manageiq core.

```ruby
      stub_action_supports(akp.class, :delete)
      # or
      stub_action_supports(akp.resource, :key_pair_delete)
```

chessbyte:
@kbrock thanks for the suggestions.  Ended up using
```ruby
expect_any_instance_of(akp.class).to receive(:supports?).with(:delete).and_return(true)
```

agrare (approved):

PR #1083, opened 2021-10-05
Condense delete_action_handler

kbrock:
### Overview

Most of the delete api actions are pretty uniform.

I put this together to hopefully consolidate this code.
But I also wanted to highlight the differences so we could have a discussion on how we think the api should work.

This is focused on delete, but that is because this was the simplest case.
Update should basically be identical, but I kept it to delete to cut down on some of the changes.

### supports

We have a few different ways of handling `supports?`. In some cases we do not do it at all.

I had thought about making


chessbyte:
> I can go towards this if it is requested but due to the various ways that supports is implemented, it didn't work out as easily as I thought it would.

Maybe it makes more sense to first finish the transition from AvailabilityMixin to SupportsFeatureMixin tracked [here](https://github.com/ManageIQ/manageiq/issues/21179).  Wondering if that will make such refactorings easier.

kbrock:
> > I can go towards this if it is requested but due to the various ways that supports is implemented, it didn't work out as easily as I thought it would.
> 
> Maybe it makes more sense to first finish the transition from `AvailabilityMixin` to `SupportsFeatureMixin` tracked [here](https://github.com/ManageIQ/manageiq/issues/21179). Wondering if that will make such refactorings easier.

I'm not impartial.
It was a bit of work to modify the API. And I'd prefer not to loose it. This code will bit rot fast.
I would imagine it conflicts with some of the `AvaiabilityMixin` changes.
It is also the reason why I only focused on `:delete` which I felt was the most stable but highlighted what one direction we could take.


My thought is if we setup code in a way that encouraged CRUD, and encourages changes to the non-provider classes, then when we transitioned over to `SupportsFeatureMixin`, the direction would be more obvious, the number of provider changes would be minimized, and the current special cases would go away.

Fryguy:
cc @agrare

app/controllers/api/base_controller/generic.rb
```diff
@@ -172,16 +191,10 @@ def add_subcollection_data_to_resource(resource, type)
...
-          resource.destroy!
-          result = action_result(true, "#{type} id: #{id} deleting")
-          add_href_to_result(result, type, id)
-          log_result(result)
-          result
-        end
+      # ??? do we actually use this for anything ???
+      def delete_resource_action(_type, resource, _data)
+        resource.destroy!
+        {}
```
Fryguy, line 197: Is the subclass supposed to `super` to this?  Maybe this should just 

```ruby
raise NotImplementedError, "must be implemented in a subclass"
```

kbrock:
> wondering if the code should call the model instance for all delete requests - and it figures out whether to delegate or not. but probably this is too much to tackle in this PR

@chessbyte We could have it in the model or the controller. I originally was thinking we could have 2 different api controller hierarchies: direct destroy and defer to provider implementations. There is a third less common implementation that is a direct destroy, but it is thrown on the queue.

Also noticing that different models have very different ways to say "don't delete me". Some raise exceptions and other require you to explore `read_only?` flags. Not sure if the `:supports` could overlap this or if it is conflating the concerns here. 

Our edge cases like trying to delete a model that is not valid (deleting your own user record). Some return 200, success=false while others return different status codes. We need to get more similarity if we want to converge on common code, but I'm weary to make this more consistent due to the testing effort here.

TL;DR: The common trends are really starting to surface. the refresh and update actions will just be a ✂️  effort.

kbrock:
**UPDATE:** split changes that cause spec differences into one commit and another that just does the conversion. Hope this will allow you to see the api changes. It is mostly (all?) messaging wording changes.

I may push changes to the 3 controllers that are being a little stuborn.

But this should be it for now.

just pushed rubocop changes. Nice that codeclimate is already asking me to change the `refresh_resource`. soon... soon...

kbrock:
WIP: so there is a dynamically generated  method name coming into play here. I need to rework what is the generic method name

frustrating b/c already spent too much time in here and what I thought was close moved a little further away

kbrock:
PUSH:
- removed special firmware model_ident
- changed verbiage around delete tags  to match all other deletes (Deleting at beginning not end)
- all TODO items have the words `TODO: `, and some `NOTE` entries too
-

Issue #1074, opened 2021-09-10
Slow API requests part deux


jrafanie:
Followup to #805 

```
Vms controller:

Baseline:
4.2s 348 KB
http://localhost:3000/api/vms?offset=0&limit=1000&expand=resources&attributes=id,href,vendor,type,name,hostname,hostnames,availability_zone_id,ems_id,operating_system.product_name,storage.name,created_on,updated_on,guid,power_state,state_changed_on,tenant.name,ipaddresses,tags,v_owning_datacenter,last_compliance_status,hardware.cpu_sockets,hardware.memory_mb,active,archived,hardware.allocated_disk_storage,cloud,service.id,service.guid

/api/vms
|   ms | queries | query (ms) | rows |
| ---: |    ---: |       ---: | ---: |
| 4147 |    2047 |       1538 |  900 |
| 4481 |    2047 |       1645 |  900 |
| 4163 |    2047 |       1541 |  900 |
| 4501 |    2047 |       1611 |  900 |


Drop from api request: tags:
3.6s 344 KB
http://localhost:3000/api/vms?offset=0&limit=1000&expand=resources&attributes=id,href,vendor,type,name,hostname,hostnames,availability_zone_id,ems_id,operating_system.product_name,storage.name,created_on,updated_on,guid,power_state,state_changed_on,tenant.name,ipaddresses,v_owning_datacenter,last_compliance_status,hardware.cpu_sockets,hardware.memory_mb,active,archived,hardware.allocated_disk_storage,cloud,service.id,service.guid

/api/vms
|   ms | queries | query (ms) | rows |
| ---: |    ---: |       ---: | ---: |
| 3551 |    1581 |       1304 |  900 |
| 3943 |    1581 |       1391 |  900 |
| 3563 |    1581 |       1286 |  900 |
| 3932 |    1581 |       1380 |  900 |

Drop from api request: service.guid and service.id
2.6s 344 KB
http://localhost:3000/api/vms?offset=0&limit=1000&expand=resources&attributes=id,href,vendor,type,name,hostname,hostnames,availability_zone_id,ems_id,operating_system.product_name,storage.name,created_on,updated_on,guid,power_state,state_changed_on,tenant.name,ipaddresses,v_owning_datacenter,last_compliance_status,hardware.cpu_sockets,hardware.memory_mb,active,archived,hardware.allocated_disk_storage,cloud

/api/vms
|   ms | queries | query (ms) | rows |
[...]

PR #1084, opened 2021-10-06
[WIP] [POC] [Performance] Add includes for direct relations

NickLaMuro:
This change adds includes for direct relationships that are passed in as attributes so they are loaded as part of the main query.

There is a quick RBac check just before allowing adding that attribute to determine if it participates in rbac, but otherwise it works just like an attribute.

The reason this didn't get included previously is relations like `tags` on `Vm` and others are lumped into the `virtual_attributes_for` grouping for the API, despite not being a virtual attribute (for "reasons"), and since it is not a `virtual_includes` nor is it an attribute with a `dot` notation, it doesn't get included with the rest.

Benchmarks
----------

Tested using:

```console
$ bundle exec miqperf benchmark -a -c 3 "/api/vms?offset=0&limit=1000&expand=resources&attributes=id,href,name,tags"
```

### Before

| 2392 |     478 |      557.4 | 2036 |
| 1178 |     476 |      462.9 |  472 |
|  963 |     476 |      408.3 |  472 |

### After

| 1772 |      14 |         41 | 2037 |
|  452 |      11 |       48.9 |  472 |
|  245 |      11 |       24.5 |  472 |

Links
-----

* Partially addresses performance issues from:
  - https://github.com/ManageIQ/manageiq-api/issues/1074

QA
--

I think it makes sense to throw this patch in a running environment and determine if it will make a difference in idle performance as well as if it doesn't mess with general performance of the API.  I don't have an environment for this, so this is a call for help! :bow:

### DIFF
--- a/app/controllers/api/base_controller/renderer.rb
+++ b/app/controllers/api/base_controller/renderer.rb
@@ -585,6 +585,10 @@ module Api
         attrs = virtual_attributes_for(klass) do |type, attr_name, attr_base|
           if klass.virtual_includes(attr_name) && !klass.attribute_supported_by_sql?(attr_name) && attr_base.blank?
             attr_name
+          elsif klass.respond_to?(:reflect_on_association) && klass.reflect_on_association(attr_name)
+            next if attr_base_uses_rbac?(attr_base)
+
+            attr_name
           else
             next if attr_base.blank?
             next if virtual_attribute_accessor(type, attr_name)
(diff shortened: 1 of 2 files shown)


kbrock:
Nick: sorry if you get a bunch of chatter on this. I will close and reopen my own if it looks like we have a lot of changes here/want to rebase/continue.

@jrafanie @Fryguy This looks like a no braine. Do we want to resurrect?
It basically stalled because it looked intimidating to test.

PR #1086, opened 2021-10-21
[WIP] Cloud Volume Snapshot Subcollection

mengxiwuu:
Related PR on [instance](https://github.com/ManageIQ/manageiq/pull/13729/files). 

- Currently have the API.get() function set up
-  Still need to work on API.post()

Screen Shot 2021-11-03 at 11 02 06 AM

### DIFF
--- a/app/controllers/api/subcollections/cloud_volume_snapshots.rb
+++ b/app/controllers/api/subcollections/cloud_volume_snapshots.rb
@@ -1,0 +1,64 @@
+module Api
+  module Subcollections
+    module CloudVolumeSnapshots
+      def snapshots_query_resource(object)
+        object.cloud_volume_snapshots
+      end
+
+      def cloud_volume_snapshots_create_resource(parent, _type, _id, data)
+        raise parent.unsupported_reason(:cloud_volume_snapshot_create) unless parent.supports?(:cloud_volume_snapshot_create)
+
+        message = "Creating cloud volume snapshot #{data["name"]} for #{cloud_volume_snapshot_ident(parent)}"
+        task_id = queue_object_action(
+          parent,
+          message,
+          :method_name => "create_cloud_volume_snapshot",
+          :args        => [data["name"], data["description"], data.fetch("memory", false)]
+        )
+
+        action_result(true, message, :task_id => task_id)
+      rescue => e
+        action_result(false, e.to_s)
+      end
+
+      def delete_resource_cloud_volume_snapshots(parent, type, id, _data)
+        cloud_volume_snapshot = resource_search(id, type)
+        begin
+          raise parent.unsupported_reason(:remove_cloud_volume_snapshot) unless parent.supports?(:remove_cloud_volume_snapshot)
+
+          message = "Deleting cloud volume snapshot #{cloud_volume_snapshot.name} for #{model_ident(parent, type)}"
+          task_id = queue_object_action(parent, message, :method_name => "remove_cloud_volume_snapshot", :args => [id])
+          action_result(true, message, :task_id => task_id)
+        rescue => e
+          action_result(false, e.to_s)
+        end
+      end
+      alias cloud_volume_snapshots_delete_resource delete_resource_cloud_volume_snapshots
+
+      def cloud_volume_snapshots_revert_resource(parent, type, id, _data)
+        raise parent.unsupported_reason(:revert_to_cloud_volume_snapshot) unless parent.supports?(:revert_to_cloud_volume_snapshot)
+
+        cloud_volume_snapshot = resource_search(id, type)
+
+        message = "Reverting to cloud volume snapshot #{cloud_volume_snapshot.name} for #{cloud_volume_snapshot_ident(parent)}"
+        task_id = queue_object_action(parent, message, :method_name => "revert_to_cloud_volume_snapshot", :args => [id])
+        action_result(true, message, :task_id => task_id)
+      rescue => e
+        action_result(false, e.to_s)
+      end
+
+      def cloud_volume_snapshots_subcollection_options(parent)
+        raise BadRequestError, "No DDF specified for cloud volume snapshots in #{parent}" unless parent.respond_to?(:params_for_create_cloud_volume_snapshot)
+
+        {:cloud_volume_snapshot_form_schema => parent.params_for_create_cloud_volume_snapshot}
+      end
+
+      private
+
+      def cloud_volume_snapshot_ident(parent)
+        parent_ident = collection_config[@req.collection].description.singularize
+        "#{parent_ident} id:#{parent.id} name:'#{parent.name}'"
+      end
+    end
+  end
+end
--- a/app/controllers/api/subcollections/cloud_volume_snapshots.rb
+++ b/app/controllers/api/subcollections/cloud_volume_snapshots.rb
@@ -1,7 +1,7 @@
 module Api
   module Subcollections
     module CloudVolumeSnapshots
-      def snapshots_query_resource(object)
+      def cloud_volume_snapshots_query_resource(object)
         object.cloud_volume_snapshots
       end
 
--- a/config/api.yml
+++ b/config/api.yml
@@ -847,6 +847,7 @@
     :verbs: *gpppd
     :subcollections:
     - :tags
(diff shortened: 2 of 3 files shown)


app/controllers/api/cloud_volumes_controller.rb
```diff
@@ -1,6 +1,8 @@
 module Api
   class CloudVolumesController < BaseController
     include Subcollections::Tags
+    include Subcollections::Snapshots
```
kbrock, line 4: again, we have a model called Snapshots, which is different from CloudVolumeSnapshot.

While I think it would be great if we could have '/snapshot' as a url, I really don't want us to call this subcollection snapshots since this is the wrong name/will be confusing if we want vms/snapshots

kbrock (commented):
just the rename of the subcollection

app/controllers/api/cloud_volumes_controller.rb
mengxiwuu: Already posted on the [issue](https://github.com/ManageIQ/manageiq-api/issues/1094) raised.

config/api.yml
```diff
@@ -884,6 +885,25 @@
         :identifier: cloud_volume_tag
       - :name: unassign
         :identifier: cloud_volume_tag
+    :snapshots_subcollection_actions:
```
kbrock, line 888:
```suggestion
    :cloud_volume_snapshots_subcollection_actions:
```

config/api.yml
```diff
@@ -847,6 +847,7 @@
     :verbs: *gpppd
     :subcollections:
     - :tags
+    - :snapshots
```
kbrock, line 850:
```suggestion
    - :cloud_volume_snapshots
```

This is going to link to the wrong class

kbrock (commented):
Currently, `Api::Href` pulls the subcollection name out of the url and that is fed into all our infrastructure. 

We would need to change `Api::RequestAdapter` or `Api::Href` to read from `api.yml` to convert the `snapshots` to use type `cloud_volume_snapshots` otherwise we will run into this mess here and with `collection_class(type)` and other issues.

I again suggest we use a url that uses the fully qualified name `cloud_volume_snapshots` rather than `snapshots`

app/controllers/api/subcollections/snapshots.rb
```diff
@@ -2,7 +2,7 @@ module Api
   module Subcollections
     module Snapshots
       def snapshots_query_resource(object)
-        object.snapshots
+        object.send(snapshots_model_method)
```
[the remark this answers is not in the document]
mengxiwuu: Sure! I already opened a [new issue](https://github.com/ManageIQ/manageiq-api/issues/1094) for this PR and document the suggest change from `snapshots` to `cloud_volume_snapshots` here

app/controllers/api/subcollections/snapshots.rb
```diff
@@ -59,6 +59,10 @@ def snapshot_ident(parent)
         parent_ident = collection_config[@req.collection].description.singularize
         "#{parent_ident} id:#{parent.id} name:'#{parent.name}'"
       end
+
```
kbrock, line 62: remove these changes

config/api.yml
```diff
@@ -884,6 +885,25 @@
...
+    :snapshots_subresource_actions:
+      :get:
+      - :name: read
+        :identifier: cloud_volume_snapshot_view
+      :post:
+      - :name: delete
+        :identifier: cloud_volume_snapshot_delete
+      :delete:
+      - :name: delete
+        :identifier: cloud_volume_snapshot_delete
```
kbrock, line 906: remove the references to snapshot. we want cloud_volume_snapshots

spec/requests/snapshots_spec.rb
```diff
@@ -610,4 +610,100 @@
...
+        expect(response.parsed_body).to include(expected)
+        expect(response).to have_http_status(:ok)
+      end
+
+      it "will not list snapshots unless authorized" do
+        api_basic_authorize
+        cloud_volume = FactoryBot.create(:cloud_volume)
+        FactoryBot.create(:cloud_volume_snapshot, :cloud_volume => cloud_volume)
+
+        get(api_cloud_volume_snapshots_url(nil, cloud_volume))
```
kbrock, line 639:
```suggestion
        get(api_cloud_volume_cloud_volume_snapshots_url(nil, cloud_volume))
```

app/controllers/api/cloud_volumes_controller.rb
```diff
@@ -49,5 +51,16 @@ def options
...
+    private
+
+    def set_collection_klasses
+      @collection_klasses ||= {}
+      @collection_klasses[:snapshots] = CloudVolumeSnapshot
+    end
+
+    def snapshots_model_method
+      :cloud_volume_snapshots
+    end
```
kbrock, line 64: not needed, we are going to reference cloud_volume_snapshot instead of having the code jump to the wrong conclusion

kbrock (changes requested):
I do not see `Subcollections::CloudVolumeSnapshot` defined.
I thought it was before.

you probably need to add this class and it should look very similar to the snapshot one.

app/controllers/api/subcollections/snapshots.rb
```diff
@@ -2,7 +2,7 @@ module Api
   module Subcollections
     module Snapshots
       def snapshots_query_resource(object)
-        object.snapshots
+        object.cloud_volume_snapshots
```
kbrock, line 5:
```
cp app/controllers/api/subcollections/snapshots.rb app/controllers/api/subcollections/cloud_volume_snapshots.rb
git checkout master app/controllers/api/subcollections/snapshots.rb # revert
```

So there will be a cloud volume snapshots subcollection that deals with cloud volume snapshots, and the original (untouched) snapshots controller that deals with vm snapshots

config/api.yml
```diff
@@ -884,6 +885,25 @@
         :identifier: cloud_volume_tag
       - :name: unassign
         :identifier: cloud_volume_tag
+    :cloud_volume_snapshots_subcollection_actions:
```
kbrock, line 888: looking good

spec/requests/snapshots_spec.rb
```diff
@@ -610,4 +610,100 @@
...
+          "results" => [
+            a_hash_including(
+              "success"   => true,
+              "message"   => "Creating snapshot Alice's snapshot for Cloud Volume id:#{cloud_volume.id} name:'Alice's VM'",
+              "task_id"   => anything,
+              "task_href" => a_string_matching(api_tasks_url)
+            )
+          ]
+        }
+        expect(response.parsed_body).to include(expected)
```
kbrock, line 686: probably want to use `expect_single_action_result`

kbrock (commented):
Good job. We're getting there

app/controllers/api/cloud_volumes_controller.rb
```diff
@@ -1,6 +1,8 @@
 module Api
   class CloudVolumesController < BaseController
     include Subcollections::Tags
+    include Subcollections::CloudVolumeSnapshot
+    before_action :set_collection_klasses
```
kbrock, line 5: this is also no longer necessary

app/controllers/api/subcollections/snapshots.rb
mengxiwuu: Hi @kbrock! Thank you so much for the help! I keep getting this error in my local environment when running test cases. Do you know where this error might coming from: 
Screen Shot 2021-12-15 at 4 41 27 PM

kbrock (commented):
just a few thoughts

app/controllers/api/subcollections/cloud_volume_snapshots.rb
```diff
@@ -0,0 +1,64 @@
+module Api
+  module Subcollections
+    module CloudVolumeSnapshots
+      def snapshots_query_resource(object)
```
kbrock, line 4:
```suggestion
      def cloud_volume_snapshots_query_resource(object)
```

kbrock:
closing in favor of https://github.com/ManageIQ/manageiq-api/pull/1134

PR #1095, opened 2021-11-05
Remove send from chargeback api

kbrock:
### Tangent from

I am reworking the api to make the flow of the actions more clear.
The `grep send` pointed here and it looked like this `send` was unnecessary and a tad bit confusing

### Goal

This is not a delete PR (sorry LJ). It is a simplification PR that will probably be more code.

The first commit is the one I care about. it replaces `send` with `case`.
The other commits are just trying to remove unnecessary dynamic keys in the hash.

Please view it commit by commit.

🚩 Half a dozen commits for 2 dozen lines of code, with a total delta of about 0 deserves a ref flag.

### DIFF
--- a/app/controllers/api/mixins/chargeback_assignment.rb
+++ b/app/controllers/api/mixins/chargeback_assignment.rb
@@ -119,8 +119,16 @@ module Api
       end
 
       def target(parameter_record, assignment_type, rate_type)
-        target_assignment_method = "#{assignment_type}_target_assignment"
-        send(target_assignment_method, parameter_record[assignment_type.to_s], assignment_type, rate_type)
+        case assignment_type.to_sym
+        when :tag
+          tag_target_assignment(parameter_record[assignment_type.to_s], assignment_type, rate_type)
+        when :label
+          label_target_assignment(parameter_record[assignment_type.to_s], assignment_type, rate_type)
+        when :resource
+          resource_target_assignment(parameter_record[assignment_type.to_s], assignment_type, rate_type)
+        else
+          raise BadRequestError, "Unknown assignment_type of #{assignment_type}"
+        end
       end
 
       def convert_assignment_key_from(parameter_key)
--- a/app/controllers/api/mixins/chargeback_assignment.rb
+++ b/app/controllers/api/mixins/chargeback_assignment.rb
@@ -118,19 +118,6 @@ module Api
         target_from(record['href'], assignment_type, rate_type)
       end
 
-      def target(parameter_record, assignment_type, rate_type)
-        case assignment_type.to_sym
-        when :tag
-          tag_target_assignment(parameter_record[assignment_type.to_s], assignment_type, rate_type)
-        when :label
-          label_target_assignment(parameter_record[assignment_type.to_s], assignment_type, rate_type)
-        when :resource
-          resource_target_assignment(parameter_record[assignment_type.to_s], assignment_type, rate_type)
-        else
-          raise BadRequestError, "Unknown assignment_type of #{assignment_type}"
-        end
-      end
-
       def convert_assignment_key_from(parameter_key)
         parameter_key == 'resource' ? :object : parameter_key.to_sym
       end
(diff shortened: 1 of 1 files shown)


kbrock:
@jrafanie if there are any commits that are confusing, let me know. I tried to make all of them very simple/obvious

We have a few magic naming conventions in the api, that is why I'm trying to remove the send.

app/controllers/api/mixins/chargeback_assignment.rb
```diff
@@ -104,35 +100,37 @@ def chargeback_rate(parameter_record)
...
         rate = chargeback_rate(parameter_record)
 
-        {:cb_rate => rate, convert_assignment_key_from(assignment_type) => target(parameter_record, assignment_type, rate_type)}
+        case assignment_type.to_sym
+        when :tag
+          tag_target_assignment(parameter_record, rate, rate_type)
+        when :label
+          label_target_assignment(parameter_record, rate, rate_type)
+        when :resource
+          resource_target_assignment(parameter_record, rate, rate_type)
```
Fryguy, line 132: I think you should add an else raise, which sort of happened before via NoMethodError.  Not sure what the right exception here would be

app/controllers/api/mixins/chargeback_assignment.rb
```diff
@@ -104,35 +100,37 @@ def chargeback_rate(parameter_record)
...
       def convert_assignment_key_from(parameter_key)
         parameter_key == 'resource' ? :object : parameter_key.to_sym
       end
 
       def rate_assignment(parameter_record, assignment_type, rate_type)
-        parameter_record['label'] = parameter_record.delete('resource') if assignment_type == :label
-
         rate = chargeback_rate(parameter_record)
 
-        {:cb_rate => rate, convert_assignment_key_from(assignment_type) => target(parameter_record, assignment_type, rate_type)}
```
jrafanie, line 135: was this the only caller of the `target` method?

jrafanie: Note, I don't like how the old code refers to a local `target` variable while we also had that method.  It's minor but there's no reason to use the same name from a different scope.  I'm glad that's going away.

app/controllers/api/mixins/chargeback_assignment.rb
```diff
@@ -104,35 +100,37 @@ def chargeback_rate(parameter_record)
         @chargeback_rate[rate_id] ||= ChargebackRate.find(rate_id)
       end
 
-      def tag_target_assignment(record, _assignment_type, rate_type)
-        target = tag_assigment(record, rate_type)
-        [target, rate_type == "Storage" ? 'storage' : record['assignment_prefix']]
```
jrafanie, line 109: the format of the return from this method changed.  Were all the callers updated?

jrafanie:
No major problems.  Just questions.  Were tests covering this change?  It seems like a good overall change to remove the send.

app/controllers/api/mixins/chargeback_assignment.rb
kbrock: after these changes, there are no callers for target `ag --ruby '\btarget\b'`

app/controllers/api/mixins/chargeback_assignment.rb
kbrock: these only have one caller that is within this file:
tag_target_assignment, label_target_assignment, resource_target_assignment

app/controllers/api/mixins/chargeback_assignment.rb
kbrock: done

kbrock: there is a BadRequestError for another issue with assignment_type, so I stayed with that one

kbrock:
push: added the else clause to the case statement

I changed this in the first commit, so it ended up conflicting with every commit. Pretty sure I resolved them correctly.

Fryguy (approved):

kbrock:
> Were tests covering this change

@jrafanie I just reconfirmed that introducing bugs into the 3 `*_target_assignment` methods causes quite a few tests fail. I had done this intentionally while developing to make sure something had my back.

tests can be found in [shared_examples/assignments.rb](https://github.com/ManageIQ/manageiq-api/blob/master/spec/support/shared_examples/assigments.rb)

jrafanie (approved):

PR #1096, opened 2021-11-05
Remove unnecessary send

kbrock:
### Overview

Trying to make the api more clear, especially around `send("#{type}_#{action}_action")'

### This PR

- Removing `send` from the API.
- collapsed an extra layer of indirection in update_collection.

### DIFF
--- a/app/controllers/api/base_controller/manager.rb
+++ b/app/controllers/api/base_controller/manager.rb
@@ -20,9 +20,9 @@ module Api
               "Unimplemented Action #{action} for #{type} resources" unless respond_to?(target)
 
         if id
-          get_and_update_one_collection(@req.subcollection?, target, type, id)
+          update_one_collection(@req.subcollection?, target, type, id, @req.resource)
         else
-          get_and_update_multiple_collections(@req.subcollection?, target, type)
+          update_multiple_collections(@req.subcollection?, target, type, @req.resources)
         end
       end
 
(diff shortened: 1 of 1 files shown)


app/controllers/api/base_controller/manager.rb
```diff
@@ -2,8 +2,10 @@ module Api
   class BaseController
     module Manager
       def update_collection(type, id)
-        if @req.method == :put || @req.method == :patch
-          return send("#{@req.method}_resource", type, id)
+        if @req.method == :put
+          return edit_resource(type, id, @req.json_body)
```
Fryguy, line 6: Should this be `put_resource`?

```suggestion
          return put_resource(type, id, @req.json_body)
```

Fryguy: Oh this is the collapsing of the redirection you mentioned - didn't realize this is what you meant.

app/controllers/api/base_controller/manager.rb
```diff
@@ -2,8 +2,10 @@ module Api
...
+        elsif @req.method == :patch
+          return patch_resource(type, id)
         end
```
Fryguy, line 9: Would it be cleaner to move to a case statement?

```suggestion
        case @req.method
        when :put
          return edit_resource(type, id, @req.json_body)
        when :patch
          return patch_resource(type, id)
        end
```

PR #1097, opened 2021-11-09
Condense enqueue_ems_action

kbrock:
Blocked by / built on top of:
- [x] https://github.com/ManageIQ/manageiq-api/pull/1103 (first 2 commits)

### Overview

We have a pattern for actions that request the ems to perform an operation.
So I converted all `queue_object_action` calls over to `enqueue_ems_action`

Suggest reviewing by commit (unsure if commits will show in proper order but they are contained)

### Before
- `enqueue_action` mostly works for ems operations
- some ems operations get a user and others do not

### After
- fixed queue messaging when auto deriving a description.
- `enqueue_action` is generic and `enqueue_ems_action` is ems specific.
- user is now sent to all ems operations.
- introduced `:supports` parameter to easily enforce `supports?`.
- everything has been converted across except for subcollections and generic objects.
 

### Review

Only the first 2 commits do anything, and are minor changes at that.
The subsequent commits just leverage `enqueue_ems_action` (and delete code)

---

### Code effect

```diff
 class CloudTemplatesController < BaseController
   def check_compliance_resource(type, id, _data = nil)
-    api_action(type, id) do |klass|
-      container_image = resource_search(id, type, klass)
-      api_log_info("Checking compliance of #{container_image_ident(container_image)}")
-      request_compliance_check(container_image)
-   end
+   enqueue_ems_action(type, id, "Check Compliance for", :method_name => "check_compliance", :supports => true)
  end
-
-  private
-
-  def cloud_template_ident(cloud_template)
-    "CloudTemplate id:#{cloud_template.id} name:'#{cloud_template.name}'"
-  end
-
-  def request_compliance_check(cloud_template)
-    desc = "#{cloud_template_ident(cloud_template)} check compliance requested"
-    raise BadRequestError, "Checking Compliance for #{cloud_template_ident(cloud_template)}: #{model.unsupported_reason(:check_compliance)}" unless model.supports?(:check_compliance)
-
-    task_id = queue_object_action(cloud_template, desc, :method_name => "check_compliance")
-    action_result(true, desc, :task_id => task_id)
-  rescue StandardError => err
-    action_result(false, err.to_s)
-  end
 end
```

Of course, not all are as compact as this, but this is a common and best case scenario.

### DIFF
--- a/app/controllers/api/base_controller/action.rb
+++ b/app/controllers/api/base_controller/action.rb
@@ -46,6 +46,7 @@ module Api
       # @option options :method_name   - method name for the queue
       # @option options :args          - args for the queue method
       # @option options :role          - role for queue (defaults: ems_operations)
+      # @option options :supports      - check that this method_name is supported by the model
       def enqueue_action(type, id, action_phrase = nil, options = {})
         if action_phrase.kind_of?(Hash)
           options = action_phrase
@@ -53,7 +54,11 @@ module Api
         end
         action_phrase ||= "Performing #{options[:method_name]} for "
 
+        supports = options.delete(:supports)
+        supports = options[:method_name] if supports == true
+
         api_resource(type, id, action_phrase) do |model|
+          ensure_supports(type, model, options[:method_name], supports) if supports
           yield(model) if block_given?
           desc = "#{action_phrase} #{model_ident(model, type)}"
           {:task_id => queue_object_action(model, desc, options)}
--- a/app/controllers/api/base_controller/action.rb
+++ b/app/controllers/api/base_controller/action.rb
@@ -98,14 +98,6 @@ module Api
 
         MiqTask.generic_action_with_callback(task_options, queue_options)
       end
-
-      def queue_options(method, role = nil)
-        {
-          :method_name => method,
-          :role        => role,
-          :user        => true
-        }
-      end
     end
   end
 end
--- a/app/controllers/api/base_controller/action.rb
+++ b/app/controllers/api/base_controller/action.rb
@@ -49,10 +49,10 @@ module Api
       def enqueue_action(type, id, action_phrase = nil, options = {})
         if action_phrase.kind_of?(Hash)
           options = action_phrase
-          action_phrase ||= "Performing #{args[:method_name]} for "
+          action_phrase = nil
         end
+        action_phrase ||= "Performing #{options[:method_name]} for "
 
-        options[:role] ||= "ems_operations"
         api_resource(type, id, action_phrase) do |model|
           yield(model) if block_given?
           desc = "#{action_phrase} #{model_ident(model, type)}"
@@ -60,10 +60,20 @@ module Api
         end
       end
 
+      def enqueue_ems_action(type, id, action_phrase = nil, options = {}, &block)
+        if action_phrase.kind_of?(Hash)
+          options = action_phrase
+          action_phrase = nil
+        end
+        options.reverse_merge!(:role => "ems_operations", :user => true)
+        enqueue_action(type, id, action_phrase, options, &block)
+      end
+
       def queue_object_action(object, summary, options)
+        user = User.current_user
         task_options = {
           :action => summary,
-          :userid => User.current_user.userid
+          :userid => user.userid
         }
 
         queue_options = {
@@ -74,7 +84,11 @@ module Api
           :role        => options[:role] || nil,
         }
 
-        queue_options.merge!(options[:user]) if options.key?(:user)
+        if options[:user]
+          queue_options[:user_id]   = user.id
+          queue_options[:group_id]  = user.current_group.id
+          queue_options[:tenant_id] = user.current_tenant.id
+        end
         queue_options[:zone] = object.my_zone if %w(ems_operations smartstate).include?(options[:role])
 
         MiqTask.generic_action_with_callback(task_options, queue_options)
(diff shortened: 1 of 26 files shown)


app/controllers/api/base_controller/action.rb
```diff
@@ -31,9 +31,16 @@ def api_resource(type, id, action_phrase)
         log_result(result)
         result
       rescue ActiveRecord::RecordNotFound, ForbiddenError, BadRequestError, NotFoundError => err
-        single_resource? ? raise : action_result(false, err.to_s)
+        raise if single_resource?
+        result = action_result(false, err.to_s)
+        add_href_to_result(result, type, id)
+        log_result(result)
+        result
```
Fryguy, line 38: offhand this looks like it could be DRYed up with the code in the other rescue block.

app/controllers/api/base_controller/action.rb
```diff
@@ -47,24 +54,36 @@ def api_resource(type, id, action_phrase)
...
+          ensure_supports(type, model, options[:method_name], supports) if supports
           yield(model) if block_given?
           desc = "#{action_phrase} #{model_ident(model, type)}"
           {:task_id => queue_object_action(model, desc, options)}
         end
       end
 
+      def enqueue_ems_action(type, id, action_phrase = nil, options = {}, &block)
+        hsh = action_phrase.kind_of?(Hash) ? action_phrase : options
+        hsh.reverse_merge!(:role => "ems_operations", :user => true)
```
Fryguy, line 78:
```suggestion
        hash = action_phrase.kind_of?(Hash) ? action_phrase : options
        hash.reverse_merge!(:role => "ems_operations", :user => true)
```

app/controllers/api/vms_controller.rb
```diff
@@ -169,66 +104,32 @@ def add_event_resource(type, id = nil, data = nil)
...
-      raise BadRequestError, "Must specify a new_name" if new_name.blank?
-
-      api_action(type, id) do |klass|
-        vm = resource_search(id, type, klass)
-        api_log_info("Renaming #{vm_ident(vm)} to #{new_name}")
 
-        result = validate_vm_for_action(vm, "rename")
-        result = rename_vm(vm, new_name) if result[:success]
-        result
+      # loosing that we are renaming to #{new_name}
```
Fryguy, line 119: Saying we're losing something in a comment is confusing, because a reader doesn't know what we are losing it relative to (relative to the previous code or relative to something earlier in the current code).  Instead, I think a TODO about what are our intentions for the code is easier to understand.

```suggestion
      # TODO: Find a way to get `new_name` into the response text
```

kbrock: the consolidation is going to loos stuff

app/controllers/api/vms_controller.rb
```diff
@@ -169,66 +104,31 @@ def add_event_resource(type, id = nil, data = nil)
...
-
-      api_action(type, id) do |klass|
-        vm = resource_search(id, type, klass)
-        api_log_info("Renaming #{vm_ident(vm)} to #{new_name}")
 
-        result = validate_vm_for_action(vm, "rename")
-        result = rename_vm(vm, new_name) if result[:success]
-        result
+      enqueue_ems_action(type, id, "Renaming", :method_name => "rename") do
+        raise BadRequestError, "Must specify a new_name" if new_name.blank?
```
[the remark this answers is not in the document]
kbrock: There are a few instances of these.

I had debated this.

For a request with bad parameters , this will still look up the vm.
But we still want the block to log the result, add the resource link, and depending if it is a single resource, raise the exception or just use a failed action response hash.;

I can dive in and see if there is a dry way to handle this.

kbrock:
@Fryguy ok, reduced the duplicate code.

I want to continue going with reducing the `api_action` work next (probably another 🍪  worth)

So let me know your thoughts on this one and I can plan accordingly

kbrock:
codeclimate: 45 closed, one opened

app/controllers/api/vms_controller.rb
kbrock: yikes. this method should not have been refactored

kbrock: Also, I solved this in the PR I just extracted. At least for the case where we are not queuing

kbrock:
PR is good, blocked by dependent PR that was pulled out out

Don't want to review that code too many times

app/controllers/api/base_controller/action.rb
```diff
@@ -77,23 +88,15 @@ def queue_object_action(object, summary, options)
           :role        => options[:role] || nil,
         }
 
-        queue_options.merge!(options[:user]) if options.key?(:user)
+        if options.key?(:user)
```
Fryguy, line 91: Since we require `:user => true` via the interface and technically that value can be overridden...we should check the actual value.  (i.e. `{:user => false}.key?(:user) == true`

```suggestion
        if options[:user]
```

Fryguy:
Overall LGTM - only one comment.  I'd like @agrare to also review before merge.

agrare (approved):
:+1: from me looks awesome

app/controllers/api/base_controller/action.rb
kbrock: done

kbrock:
push: fixed edge case in deriving `action_phrase` from `enqueue_ems_action`

PR #1098, opened 2021-11-09
Use standard create/update/delete for supports? checks

agrare:
Rather than using "special" supports feature names we should prefer standard create/update/delete where possible

Depends on:
- [ ] https://github.com/ManageIQ/manageiq-providers-openstack/pull/754
- [ ] https://github.com/ManageIQ/manageiq-providers-amazon/pull/727
- [ ] https://github.com/ManageIQ/manageiq-providers-ibm_cloud/pull/303
- [ ] https://github.com/ManageIQ/manageiq-ui-classic/pull/7954
- [ ] https://github.com/ManageIQ/manageiq/pull/21550

### DIFF
--- a/app/controllers/api/host_aggregates_controller.rb
+++ b/app/controllers/api/host_aggregates_controller.rb
@@ -3,12 +3,11 @@ module Api
     include Subcollections::Tags
 
     def create_resource(_type, _id, data = {})
-      ext_management_system = resource_search(data['ems_id'], :providers, collection_class(:providers))
-      data.delete('ems_id')
+      ext_management_system = resource_search(data.delete('ems_id'), :providers, collection_class(:providers))
+      klass = ext_management_system.class_by_ems('HostAggregate')
+      raise BadRequestError, klass.unsupported_reason(:create) unless klass.supports?(:create)
 
-      raise "Creation of Host Aggregates is not supported for this provider" unless ext_management_system.supports?(:create_host_aggregate)
-
-      task_id = ext_management_system.create_host_aggregate_queue(session[:userid], data.symbolize_keys)
+      task_id = klass.create_aggregate_queue(session[:userid], ext_management_system, data.symbolize_keys)
       action_result(true, "Creating Host Aggregate #{data['name']} for Provider: #{ext_management_system.name}", :task_id => task_id)
     rescue => err
       action_result(false, err.to_s)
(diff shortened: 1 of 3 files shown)


agrare:
@miq-bot cross-repo-tests /all, manageiq#21550, manageiq-ui-classic#7954, manageiq-providers-ibm_cloud#303, manageiq-providers-amazon#727, manageiq-providers-openstack#754

Fryguy:
@agrare Is this going to collide with #1097 ?  If so, should we get that one in first, since it's probably harder to refactor than this one? cc @kbrock

agrare:
@Fryguy real quick search doesn't look like this will conflict with that one, but no issue holding off on this until that one is merged I have a bit of work to do to get this whole PR set green anyway

agrare:
Okay cross-repo tests are green, taking all of these out of WIP

kbrock:
the other PR is 9 commits. So would be nice to get that in first ;)
I agree, it doesn't look like there should be much collision.

kbrock (approved):
looks great

app/controllers/api/host_aggregates_controller.rb
```diff
@@ -3,12 +3,11 @@ class HostAggregatesController < BaseController
     include Subcollections::Tags
 
     def create_resource(_type, _id, data = {})
-      ext_management_system = resource_search(data['ems_id'], :providers, collection_class(:providers))
-      data.delete('ems_id')
+      ext_management_system = resource_search(data.delete('ems_id'), :providers, collection_class(:providers))
+      klass = ext_management_system.class_by_ems('HostAggregate')
+      raise BadRequestError, klass.unsupported_reason(:create) unless klass.supports?(:create)
```
kbrock, line 8: YAY. all three of these are nice to move over to supports

kbrock (approved):
looks great

kbrock:
@agrare  My only question is on the ui classic one. I'm guessing we want to merge this and then introduce changes from https://github.com/ManageIQ/manageiq/pull/21553 ?

Or are you thinking that you want to get the subclass supports into there and then refactor that PR?

agrare:
> Or are you thinking that you want to get the subclass supports into there and then refactor that PR?

I started out not wanting to introduce too many other changes into that PR so I worked with what I had but I'd love to get the `subclasses_supporting` PR in first then use that instead since it looks really close/already-ready

chessbyte:
I guess the natural followup to this PR set is to change the _queue method name as well.

Something like

```
raise BadRequestError, klass.unsupported_reason(:create) unless klass.supports?(:create)
task_id = klass.create_queue(session[:userid], ext_management_system, data)
```

instead of

```
raise BadRequestError, klass.unsupported_reason(:create) unless klass.supports?(:create)
task_id = klass.create_key_pair_queue(session[:userid], ext_management_system, data)
```

PR #1102, opened 2021-11-18
Add a GET /api/metrics endpoint

agrare:
There already exists an /api/metric_rollups endpoint but no way to query for realtime metrics.  This also renames the product feature for metric_rollups to match the API identifier.

Followed #4 for consistency

Depends on:
- [x] https://github.com/ManageIQ/manageiq/pull/21568

### DIFF
--- a/app/controllers/api/subcollections/metrics.rb
+++ b/app/controllers/api/subcollections/metrics.rb
@@ -1,0 +1,17 @@
+module Api
+  module Subcollections
+    module Metrics
+      RESOURCE_TYPES = {
+        'vms' => 'VmOrTemplate'
+      }.freeze
+
+      def metrics_query_resource(object)
+        params[:resource_type] = RESOURCE_TYPES[@req.collection] || object.class.to_s
+        params[:resource_ids] ||= [object.id]
+
+        metrics_service = MetricsService.new(params)
+        metrics_service.query_metrics
+      end
+    end
+  end
+end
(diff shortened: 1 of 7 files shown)


agrare:
@miq-bot cross-repo-tests manageiq#21568, manageiq-ui-classic

config/api.yml
```diff
@@ -2237,12 +2237,32 @@
     :klass: ChargebackRateDetailMeasure
   :metric_rollups:
     :description: Metric Rollups
-    :identifier: metrics
+    :identifier: metric_rollups
```
Fryguy, line 2240: Are we suggestion there will be 2 separate features for RBAC?  I'm wondering if it's just simple to have one feature that applies to both metrics and rollups.

chessbyte: `less is more` ™️

app/controllers/api/subcollections/metrics.rb
```diff
@@ -0,0 +1,17 @@
+module Api
+  module Subcollections
+    module Metrics
+      RESOURCE_TYPES = {
+        'vms' => 'VmOrTemplate'
```
agrare, line 5: NOTE other collections have metrics also (e.g. hosts and container_groups), going to tackle that in a follow-up

Fryguy:
@agrare Can you update the OP with example request and response for the sprint review?

PR #1103, opened 2021-11-18
Add errors hrefs back into responses

kbrock:
https://github.com/ManageIQ/manageiq-api/pull/1097

This feels separate enough that I wanted to pull out into its own PR.

### before before:

- many api errors had href in the responses. 

### before:

- more api errors had no href in the responses. (many that were transitioned to `api_resource` lost this).
- `api_resource` helper method forced the response message to be determined before looking up the resource. Turned out to not be flexible enough.

### after:

- api errors (all `api_resource` methods) have href in the responses.
- api messages can be customized with `api_resource` by optionally returning a result hash instead of just the result hash options.
-  converted a few other methods from awkward href code to use `api_result`.

### DIFF
--- a/app/controllers/api/base_controller/action.rb
+++ b/app/controllers/api/base_controller/action.rb
@@ -26,7 +26,11 @@ module Api
           api_log_info("#{action_phrase} #{type.to_s.titleize} id: #{id}")
           resource = resource_search(id, type, klass)
           result_options = yield(resource)
-          action_result(true, "#{action_phrase} #{model_ident(resource, type)}", result_options)
+          if result_options.key?(:success) # full action hash (finer grained messaging)
+            result_options
+          else # result_options is action_hash (preferred)
+            action_result(true, "#{action_phrase} #{model_ident(resource, type)}", result_options)
+          end
         rescue ActiveRecord::RecordNotFound, ForbiddenError, BadRequestError, NotFoundError => err
           single_resource? ? raise : action_result(false, err.to_s)
         rescue => err
--- a/app/controllers/api/base_controller/action.rb
+++ b/app/controllers/api/base_controller/action.rb
@@ -3,12 +3,12 @@ module Api
     module Action
       private
 
-      def api_action(type, id, options = {})
+      def api_action(type, id)
         klass = collection_class(type)
 
-        result = yield(klass) if block_given?
+        result = yield(klass)
 
-        add_href_to_result(result, type, id) unless options[:skip_href]
+        add_href_to_result(result, type, id) unless result[:href]
         log_result(result)
         result
       end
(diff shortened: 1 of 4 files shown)


kbrock:
code climate:

This method is getting more complicated.
But it is/will handle probably 80% of the api requests, so a number of edge cases are showing up.
I think the 15 lines are straight forward but I did write it so...

app/controllers/api/base_controller/action.rb
```diff
@@ -3,12 +3,12 @@ class BaseController
...
       private
 
-      def api_action(type, id, options = {})
+      def api_action(type, id)
         klass = collection_class(type)
 
-        result = yield(klass) if block_given?
+        result = yield(klass)
 
-        add_href_to_result(result, type, id) unless options[:skip_href]
```
kbrock, line 11: `options[:skip_href]` was never used.

But we sometimes put in the parent href into the response. So putting in this caveat allows us to use api_action in more instances (in the parent cases, we were basically rewriting this function)

app/controllers/api/base_controller/action.rb
```diff
@@ -3,12 +3,12 @@ class BaseController
     module Action
       private
 
-      def api_action(type, id, options = {})
```
kbrock, line 6: we never pass options.

app/controllers/api/base_controller/action.rb
```diff
@@ -19,21 +19,21 @@ def api_action(type, id, options = {})
...
-      rescue ActiveRecord::RecordNotFound, ForbiddenError, BadRequestError, NotFoundError => err
-        single_resource? ? raise : action_result(false, err.to_s)
-      rescue => err
-        action_result(false, err.to_s)
+          api_log_info("#{action_phrase} #{type.to_s.titleize} id: #{id}")
+          resource = resource_search(id, type, klass)
+          result_options = yield(resource)
+          # result_options is either action_hash options (preferred) or full action hash (finer grained messaging)
+          result = action_result(true, "#{action_phrase} #{model_ident(resource, type)}", result_options) unless result_options.key?(:success)
+          result
```
Fryguy, line 31: what happens to result in the opposite case (when result_options does have a :success key).  I think it would be nil, and then what does the caller do?

Fryguy: Should this actually return result_options, since that's what's returned form the yield?

kbrock: good catch

kbrock: we don't call this new code path yet, that is how it slipped through tests

kbrock:
@Fryguy I made that change. Good catch.

Is this looking good to you know?

kbrock:
push: no change

removed extra line from the second commit. final code is the same but wanted the path here to be more obvious

PR #1105, opened 2021-11-23
[WIP] Creation of host initiator group

gregoryb-at-ibm:


app/controllers/api/host_initiator_groups_controller.rb
```diff
@@ -0,0 +1,17 @@
...
+  class HostInitiatorGroupsController < BaseController
+    def refresh_resource(type, id, _data = nil)
+      enqueue_action(type, id, "Refreshing", :method_name => :refresh_ems)
+    end
+
+    def create_resource(type, _id = nil, data = {})
+      raise BadRequestError, "ems_id not defined for #{type} resource" if data['ems_id'].blank?
+
+      ext_management_system = ExtManagementSystem.find(data['ems_id'])
+      task_id = HostInitiatorGroup.create_host_initiator_group_queue(session[:userid], ext_management_system, data)
```
agrare, line 11: @gregoryb-at-ibm you should check that `ext_management_system.class_by_ems(:HostInitiatorGroup).supports?(:create)`

agrare: `AuthKeyPairsController` is a good example, https://github.com/ManageIQ/manageiq-api/blob/master/app/controllers/api/auth_key_pairs_controller.rb

config/api.yml
```diff
@@ -1927,6 +1951,7 @@
       :post:
       - :name: refresh
         :identifier: host_initiator_refresh
+
```
agrare, line 1954: Looks like an unrelated change

spec/requests/host_initiator_groups_spec.rb
```diff
@@ -0,0 +1,112 @@
...
+
+    it "creates new Host Initiator Group" do
+      api_basic_authorize(collection_action_identifier(:host_initiator_groups, :create))
+      provider = FactoryBot.create(:ems_autosde, :name => 'Autosde')
+      request = {
+        "action"   => "create",
+        "resource" => {
+          "ems_id"              => provider.id,
+          "name"                => "test_host_initiator_group",
+          "physical_storage_id" => "1",
```
agrare, line 28: Prefer not to hard-code IDs in specs, can you `physical_storage = FactoryBot.create(:physical_storage)` and use `physical_storage.id.to_s` instead?

spec/requests/host_initiator_groups_spec.rb
```diff
@@ -0,0 +1,112 @@
...
+        }
+      }
+
+      post(api_host_initiator_groups_url, :params => request)
+
+      expect_multiple_action_result(1, :success => true, :message => "Creating Host Initiator Group test_host_initiator_group for Provider: #{provider.name}", :task => true)
+    end
```
agrare, line 35: Can you also add a spec testing 403 forbidden without the appropriate role?

app/controllers/api/host_initiator_groups_controller.rb
agrare: And once that is in place can you add a spec that this fails for a provider that doesn't support creation of host_initiator_groups?  (You can use `stub_supports`)

agrare:
LGTM, @bdunne ?

gregoryb-at-ibm:
> LGTM, @bdunne ?

Hi @bdunne,
Could you take a look please?

gregoryb-at-ibm:
Ping @bdunne

PR #1106, opened 2021-12-01
resource_search derive klass

kbrock:
update resource_search in 2 (commits) ways:

1. Optionally pass a search key so searching by `name` instead of `id` does not need to reimplement `resource_search`.
2. Make `klass` optional for `resource_search` since it is basically just `collection_class(type)` anyway.

There are a number of `resource_search` instances that explicitly pass a different value into `collection_class` than `type`. I held off of updating these. They changes seemed to work fine but I'm trying to keep this PR as straight forward as possible.

### DIFF
--- a/app/controllers/api/automate_domains_controller.rb
+++ b/app/controllers/api/automate_domains_controller.rb
@@ -67,7 +67,7 @@ module Api
       "Automate Domain id:#{domain.id} name:'#{domain.name}'"
     end
 
-    def resource_search(id, type, klass)
+    def resource_search(id, type, klass = nil)
       if id.to_s =~ /\A\d+\z/
         super
       else
--- a/app/controllers/api/automate_domains_controller.rb
+++ b/app/controllers/api/automate_domains_controller.rb
@@ -67,17 +67,9 @@ module Api
       "Automate Domain id:#{domain.id} name:'#{domain.name}'"
     end
 
-    def resource_search(id, type, klass = nil)
-      if id.to_s =~ /\A\d+\z/
-        super
-      else
-        begin
-          domain = collection_class(:automate_domains).find_by!(:name => id)
-        rescue
-          raise NotFoundError, "Couldn't find #{klass} with 'name'=#{id}"
-        end
-        super(domain.id, type, klass)
-      end
+    def resource_search(id, type, klass = nil, key_id = nil)
+      key_id = "name" if id && !id.integer?
+      super
     end
 
     def current_tenant
--- a/app/controllers/api/base_controller/generic.rb
+++ b/app/controllers/api/base_controller/generic.rb
@@ -245,9 +245,10 @@ module Api
         end
       end
 
-      def validate_id(id, type, klass)
-        return if collection_config.resource_identifier(type) != "id"
-        raise NotFoundError, "Invalid #{klass} id #{id} specified" unless id.kind_of?(Integer) || id =~ /\A\d+\z/
+      def validate_id(id, key_id, klass)
+        if id.nil? || (key_id == "id" && !id.integer?)
+          raise BadRequestError, "Invalid #{klass} #{key_id} #{id || "nil"} specified"
+        end
       end
     end
   end
--- a/app/controllers/api/base_controller/generic.rb
+++ b/app/controllers/api/base_controller/generic.rb
@@ -31,7 +31,7 @@ module Api
           data_spec = data.collect { |key, val| "#{key}=#{val}" }.join(", ")
           raise NotFoundError, "Invalid #{type} resource specified - #{data_spec}"
         end
-        resource = resource_search(id, type, collection_class(type))
+        resource = resource_search(id, type)
         opts = {
           :name                  => type.to_s,
           :is_subcollection      => false,
@@ -43,8 +43,7 @@ module Api
       end
 
       def edit_resource(type, id, data)
-        klass = collection_class(type)
-        resource = resource_search(id, type, klass)
+        resource = resource_search(id, type)
         resource.update!(data.except(*ID_ATTRS))
         resource
       end
@@ -76,10 +75,9 @@ module Api
       end
 
       def retire_resource(type, id, data = nil)
-        klass = collection_class(type)
         if id
           msg = "Retiring #{type} id #{id}"
-          resource = resource_search(id, type, klass)
+          resource = resource_search(id, type)
           if data && data["date"]
             opts = {}
             opts[:date] = data["date"]
@@ -107,7 +105,7 @@ module Api
         end
 
         api_log_info("Invoking #{action} on #{type} id #{id}")
-        resource = resource_search(id, type, collection_class(type))
+        resource = resource_search(id, type)
         unless resource_custom_action_names(resource).include?(action)
           raise BadRequestError, "Unsupported Custom Action #{action} for the #{type} resource specified"
         end
--- a/app/controllers/api/base_controller/renderer.rb
+++ b/app/controllers/api/base_controller/renderer.rb
@@ -125,7 +125,8 @@ module Api
 
       private
 
-      def resource_search(id, type, klass)
+      def resource_search(id, type, klass = nil)
+        klass  ||= collection_class(type)
         validate_id(id, type, klass)
         key_id = collection_config.resource_identifier(type)
         target =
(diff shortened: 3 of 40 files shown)


app/controllers/api/base_controller/generic.rb
```diff
@@ -253,9 +251,9 @@ def model_ident(model, type = nil)
         end
       end
 
-      def validate_id(id, type, klass)
-        return if collection_config.resource_identifier(type) != "id"
-        raise NotFoundError, "Invalid #{klass} id #{id} specified" unless id.kind_of?(Integer) || id =~ /\A\d+\z/
+      def validate_id(id, key_id, klass)
+        return if key_id != "id"
+        raise NotFoundError, "Invalid #{klass} id #{id} specified" unless id.kind_of?(Integer) || id.to_s =~ /\A\d+\z/
```
kbrock, line 256: if id is nil, then this raises ruby warnings. `to_s` removes the warning and makes it always fail (which is what we want)

app/controllers/api/automate_domains_controller.rb
```diff
@@ -67,17 +67,8 @@ def automate_domain_ident(domain)
...
-      else
-        begin
-          domain = collection_class(:automate_domains).find_by!(:name => id)
-        rescue
-          raise NotFoundError, "Couldn't find #{klass} with 'name'=#{id}"
-        end
-        super(domain.id, type, klass)
-      end
+    def resource_search(id, type, klass = nil, key_id = nil)
+      super(id, type, klass, id.to_s =~ /\A\d+\z/ ? :id : :name)
```
Fryguy, line 71:
```suggestion
      super(id, type, klass, id.to_s.integer? ? :id : :name)
```

Fryguy: i see the old code was that way though, so no need to change for this PR.

app/controllers/api/base_controller/generic.rb
Fryguy: No need to check for kind_of?(Integer) anymore, then

kbrock: I was using to_s to fix the `nil`, but I guess it fixes the integers as well.

kbrock: thanks for the pointer to `integer?`. Didn't realize rails overrode that for 'integer` and `string`.

app/controllers/api/automate_domains_controller.rb
kbrock: done. thanks

kbrock:
update: integer instead of regex checks, swapped :id/:name to "id"/"name", changed validation so a nil name will throw better validation error.

app/controllers/api/generic_object_definitions_controller.rb
```diff
@@ -136,17 +137,11 @@ def generic_objects_request?
...
-    def resource_search(id, type, klass)
-      if id.to_s =~ /\A\d+\z/
-        super
-      else
-        go_def = klass.find_by!(:name => id)
-        filter_resource(go_def, type, klass)
-      end
+    def resource_search(id, type, klass = nil, _key_id = nil)
+      super(id, type, klass, id&.integer? ? "id" : "name")
     end
```
Fryguy, line 145: Wondering if this is clearer...up to you:

```suggestion
    def resource_search(id, type, klass = nil, key_id = nil)
      key_id = id&.integer? ? "id" : "name"
      super
    end
```

app/controllers/api/service_orders_controller.rb
```diff
@@ -26,13 +26,13 @@ def clear_resource(type, id, _data)
...
-      service_order = resource_search(id, type, collection_class(type))
+      service_order = resource_search(id, type)
       service_order.checkout
       service_order
     end
 
-    def validate_id(id, type, klass)
-      id == USER_CART_ID || super(id, type, klass)
+    def validate_id(id, key_id, klass)
+      id == USER_CART_ID || super(id, key_id, klass)
```
Fryguy, line 35:
```suggestion
      id == USER_CART_ID || super
```

app/controllers/api/base_controller/generic.rb
```diff
@@ -253,9 +251,10 @@ def model_ident(model, type = nil)
         end
       end
 
-      def validate_id(id, type, klass)
-        return if collection_config.resource_identifier(type) != "id"
-        raise NotFoundError, "Invalid #{klass} id #{id} specified" unless id.kind_of?(Integer) || id =~ /\A\d+\z/
+      def validate_id(id, key_id, klass)
+        if key_id == "id" && !id&.integer? || id.nil?
```
Fryguy, line 255: This is very confusing mixin && and || without the precendence...perhaps:

```suggestion
        if key_id == "id" && (id.nil? || !id.integer?)
```

app/controllers/api/base_controller/generic.rb
```diff
@@ -253,9 +251,10 @@ def model_ident(model, type = nil)
...
+          raise NotFoundError, "Invalid #{klass} #{key_id} #{id} specified"
```
Fryguy, line 256:
```suggestion
          raise NotFoundError, "Invalid #{klass} #{key_id} #{id.inspect} specified"
```

app/controllers/api/base_controller/renderer.rb
```diff
@@ -125,9 +125,10 @@ def virtual_attribute_accessor(type, attr)
 
       private
 
-      def resource_search(id, type, klass)
-        validate_id(id, type, klass)
-        key_id = collection_config.resource_identifier(type)
+      def resource_search(id, type, klass = nil, key_id = nil)
+        klass  ||= collection_class(type)
+        key_id ||= collection_config.resource_identifier(type)
```
Fryguy, line 130: Instead of making key_id a parameter, which is derived in every case except 1, perhaps we should make it a subclassable method, similar to validate_id, which is also only overridden in one case.  In fact, maybe it makes sense to move the "key_id" concept into validate_id, perhaps having that method _return_ the `key_id`

app/controllers/api/base_controller/generic.rb
kbrock: the original code path went through with id if they specified nothing.
So it was not as lenient, and it properly stated that they did not have an id.

If we remove the 'id.nil?` then , then id.nil? case will go through with a name.
I did modify the validation to say that name requires a value as well, so the error should be thrown, but I wanted the text.

I'll try adding parens or something and see if it becomes more clear.

app/controllers/api/base_controller/renderer.rb
kbrock: Yea, I think we only have 2 or 3 cases where we pass in a string or an id.

I liked passing in id here because the method that calls this has the knowledge that a string name is being used to look up. So it consolidates the logic.

I like the idea of deferring this decision.

app/controllers/api/generic_object_definitions_controller.rb
kbrock: I link I like the old way but not a hill... done

kbrock:
I'm being indecisive:

Currently, we swap the id passed into search and then override the search to handle that change.
It seems to make more sense to have the searching method state the intent in the first place.

```ruby
    def fetch_generic_object_definition(type, id, data)
      # current code
      resource_search(id || data['name'], type)
      # a
      id&.integer? ? resource_search(id, type) : resource_search(data['name'], type, nil, "name")
      # b
      resource_search(id || data['name'], type, nil, id ? nil : "name")
      # c
      if id.nil?
        resource_search(data['name'], type, nil, "name")
      else
        resource_search(id, type)
      end
    end

    # only needed for first case
    def resource_search(id, type, klass = nil, key_id = nil)
      key_id = "name" unless id&.integer?
      super
    end

app/controllers/api/base_controller/generic.rb
kbrock: I didn't like the double quotes coming across, I did an `|| 'nil'`

app/controllers/api/base_controller/renderer.rb
kbrock: The over arching goal is to set this up in a way that `api_resource` can just call `resource_search` without any special cases and it would work.

Currently, `resource_search` for these cases require `data['name']`. This is not available so that is why there is all this kludging.

Maybe once klass (and possibly key_id) is gone, then that will open us up for passing in the `data` or `name` parameter and they will all fall into the common pattern.

kbrock:
push:
- changed NotFound to BadRequest when the id is wrong, and changed the accompanying specs
- used `super` when overriding search behavior from id to name

PR #1108, opened 2021-12-08
Resource search klass

kbrock:
### Summary

No longer passing the `klass` into `resource_search` where it is unnecessary.
(and at a latter time probably don't ever pass `klass` into `resource_search`)

### Details

Reminder, `resource_search` looks up the `klass` from the type if it is not passed in:

```ruby
def resource_search(id, type, klass = nil)
  klass ||= collection_class(type)
```

This PR no longer looks up the `klass` before calling `resource_search`:

```diff
     def edit_resource(type, id, data)
-      auth = resource_search(id, type, collection_class(type))
+      auth = resource_search(id, type)
```

### Permutations

This has one commit per permutation type:

1. Don't lookup a `klass` by the same `type` that is passed it into `resource_search`
2. Don't specify the `klass` when it correlates to a controller's `type`.
3. Use `type=:providers` when looking up an ems.
    `resource_search(id, :ext_management-systems) ==> Unknown type`
4. Use plural `type` when looking up resources.
     NOTE: `resource_search` uses `type` in the following ways:
    - lookup the `klass`: before: passing in a bad type, required passing in a `klass`. after: Passing in right `type` now finds the `klass`.
    - looking up the `id`: before: passing in a bad type failed the lookup, but defaulted to `"id"` so it mostly works. after: passing in the right `type` finds the id (yes, typically `"id"`).
    - look for a `find_#{type}`. This is rare and does not exist for `features`, `providers`, and `providers`. But the only ones that do exist use a plural `type`.
5. Don't pass a hard coded class when the `type` resolves to that same class.

### Testing

I added print statements for every line changed:
- verified that the `type` was the expected value
- verified each change had a test on it
- verified class returned from the lookup was the expected value.

I had only found 2 areas of the code that were not tested (and not changed).
The sub-collections had a strange nuance, so I did not changing that as well

### DIFF
--- a/app/controllers/api/base_controller/renderer.rb
+++ b/app/controllers/api/base_controller/renderer.rb
@@ -561,7 +561,7 @@ module Api
         type = @req.collection.to_sym
         base_klass = collection_class(type)
 
-        ems = resource_search(ems_id, :ext_management_systems, ExtManagementSystem)
+        ems = resource_search(ems_id, :providers)
         klass = ems.class_by_ems(base_klass.name)
         raise BadRequestError, "No #{type.to_s.titleize} support for - #{ems.name}" unless klass
         raise BadRequestError, klass.unsupported_reason(:create) unless klass.supports?(:create)
--- a/app/controllers/api/authentications_controller.rb
+++ b/app/controllers/api/authentications_controller.rb
@@ -1,7 +1,7 @@
 module Api
   class AuthenticationsController < BaseController
     def edit_resource(type, id, data)
-      auth = resource_search(id, type, collection_class(:authentications))
+      auth = resource_search(id, type)
       raise "Update not supported for #{authentication_ident(auth)}" unless auth.respond_to?(:update_in_provider_queue)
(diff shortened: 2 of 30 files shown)


kbrock:
code climate: removed 6 added same 6

I agree, it is a simple refactoring `s/api_action/api_resource/g`,
but I wanted to group similar changes in this PR rather than let it sprawl.

app/controllers/api/base_controller/renderer.rb
```diff
@@ -561,7 +561,7 @@ def render_create_resource_options(ems_id)
         type = @req.collection.to_sym
         base_klass = collection_class(type)
 
-        ems = resource_search(ems_id, :ext_management_systems, ExtManagementSystem)
+        ems = resource_search(ems_id, :providers)
```
kbrock, line 564: pretty sure I added this code and did it before I knew better

kbrock:
no changes. force pushed for a rebase only

kbrock:
@Fryguy I tried to pick the more straightforward changes for this PR. Do you need me to split up to make this easier to merge?

PR #1112, opened 2021-12-15
Add VM set_description action

agrare:
```
POST /api/vms/:id -d {"action": "set_description", "new_description": "test"}
{
  "success": true,
  "message": "VM id:2 name:'VM_NAME' setting description to test",
  "task_id": "82",
  "task_href": "http://localhost:3000/api/tasks/82",
  "href": "http://localhost:3000/api/vms/2"
}
```

https://github.com/ManageIQ/manageiq/issues/21561

### DIFF
--- a/app/controllers/api/vms_controller.rb
+++ b/app/controllers/api/vms_controller.rb
@@ -131,6 +131,22 @@ module Api
     end
     central_admin :rename_resource, :rename
 
+    def set_description_resource(type, id, data = {})
+      raise BadRequestError, "Must specify an id for renaming a #{type} resource" unless id
+
+      new_description = data&.dig("new_description")&.strip
+      raise BadRequestError, "Must specify a new_description" if new_description.blank?
+
+      api_action(type, id) do |klass|
+        vm = resource_search(id, type, klass)
+        api_log_info("Setting description on #{vm_ident(vm)} to #{new_description}")
+
+        result = validate_vm_for_action(vm, "set_description")
+        result = set_description_vm(vm, new_description) if result[:success]
+        result
+      end
+    end
+
     def shutdown_guest_resource(type, id = nil, _data = nil)
       enqueue_ems_action(type, id, "Shutting Down", :method_name => "shutdown_guest", :supports => true)
     end
@@ -298,6 +314,14 @@ module Api
     def rename_vm(vm, new_name)
       desc = "#{vm_ident(vm)} renaming to #{new_name}"
       task_id = vm.rename_queue(User.current_user.userid, new_name)
+      action_result(true, desc, :task_id => task_id)
+    rescue => err
+      action_result(false, err.to_s)
+    end
+
+    def set_description_vm(vm, new_description)
+      desc = "#{vm_ident(vm)} setting description to #{new_description}"
+      task_id = vm.set_description_queue(User.current_user.userid, new_description)
       action_result(true, desc, :task_id => task_id)
     rescue => err
       action_result(false, err.to_s)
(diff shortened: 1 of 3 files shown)


config/api.yml
```diff
@@ -4744,6 +4744,8 @@
         :identifier: vm_guest_shutdown
       - :name: reboot_guest
         :identifier: vm_guest_restart
+      - :name: set_description
+        :identifier: vm_rename
```
agrare, line 4748: Thought the `vm_rename` feature would work for changing the description, but LMK if you want a new product feature for this

agrare:
Seeing a weird failure where an unauthorized user is getting 404 instead of 403 when a URL doesn't exist, digging into that a big now

kbrock:
> Seeing a weird failure where an unauthorized user is getting 404 instead of 403 when a URL doesn't exist, digging into that a big now

ooh, I get it.
you look up a vm that you should not be able to see, and they give a 404.
rather than seeing it and then trying and it saying you can't do that to a vm.

Fryguy:
That might have been an intentional choice... 403 implies the existence of something, whereas 404 doesn't.  In a multi-tenant scenario, from a security perspective 404 is better, otherwise Tenant A can interrogate the URLs and determine things like which ids are not theirs but belong to another tenant.  Then they can get information like number of VMs under management.  It's minor information, but we figured better to give no information at all.

agrare:
> That might have been an intentional choice... 403 implies the existence of something, whereas 404 doesn't.

This same exact test is in 8 other actions and they all expect forbidden not 404, e.g. https://github.com/ManageIQ/manageiq-api/blob/bfeef003506bf14118901d2ecaebfe49439b75ee/spec/requests/vms_spec.rb#L1485-L1491

app/controllers/api/vms_controller.rb
```diff
@@ -302,5 +318,13 @@ def rename_vm(vm, new_name)
...
       action_result(false, err.to_s)
     end
+
+    def set_description_vm(vm, new_description)
+      desc = "#{vm_ident(vm)} setting description to #{new_description}"
+      task_id = vm.set_description_queue(User.current_user.userid, new_description)
+      action_result(true, desc, :task_id => task_id)
+    rescue => err
+      action_result(false, err.to_s)
+    end
```
kbrock, line 328: Maybe something like this?

```suggestion
    def set_description_resource(type, id, data = {})
      new_description = data&.dig("new_description")&.strip
      api_resource(type, id, "Describing", :supports => :update) do |vm|
        raise BadRequestError, "Must specify a new_description" if new_name.blank?

        {:task_id => vm.set_description_queue(User.current_user.userid, new_description)}
      end
    end
```

spec/requests/vms_spec.rb
```diff
@@ -1621,6 +1621,64 @@ def query_match_regexp(*tables)
...
+          a_hash_including(
+            "message" => a_string_matching(/#{vm2.id}.* setting description to test2/i),
+            "success" => true,
+            "href"    => api_vm_url(nil, vm2)
+          )
+        )
+      }
+
+      expect(response.parsed_body).to include(expected)
+      expect(response).to have_http_status(:ok)
```
kbrock, line 1678: maybe use `expect_multiple_action_result` instead?

app/controllers/api/vms_controller.rb
kbrock: oops. put this in the wrong spot. think you get the general idea

kbrock:
@agrare do you want me to redo this PR in the other style?

agrare:
@kbrock sure that'd be great thanks

PR #1115, opened 2021-12-21
consolidate api.yml put/delete/patch with post

kbrock:
Currently, we are not reading the `put/edit` or `patch/edit` portion of `api.yml`. Instead, we use the `post/edit`
Deleting the unused put and patch from api.yml makes it clearer

Currently we are reading `post/delete` for POST to collections and `delete/delete` DELETE to collections.
We are using `post/delete` for POST to resources.

Consolidating the privileges into just `post/delete` which will be used for both POST and DELETE. The hope is that a single entry for api.yml makes the rules simpler and will help make it more consistent.

### DIFF
--- a/config/api.yml
+++ b/config/api.yml
@@ -56,9 +56,6 @@
         :identifier: miq_action_edit
       - :name: delete
         :identifier: miq_action_delete
-      :delete:
-      - :name: delete
-        :identifier: miq_action_delete
     :resource_actions:
       :get:
       - :name: read
@@ -66,15 +63,6 @@
       :post:
       - :name: edit
         :identifier: miq_action_edit
-      - :name: delete
-        :identifier: miq_action_delete
-      :patch:
-      - :name: edit
-        :identifier: miq_action_edit
-      :put:
-      - :name: edit
-        :identifier: miq_action_edit
-      :delete:
       - :name: delete
         :identifier: miq_action_delete
   :alert_actions:
@@ -130,15 +118,6 @@
         :identifier: alert_definition_profile_edit
       - :name: unassign
         :identifier: alert_definition_profile_edit
-      :patch:
-      - :name: edit
-        :identifier: alert_definition_profile_edit
-      :put:
-      - :name: edit
-        :identifier: alert_definition_profile_edit
-      :delete:
-      - :name: delete
-        :identifier: alert_definition_profile_delete
     :alert_definitions_subcollection_actions:
       :post:
       - :name: assign
@@ -170,9 +149,6 @@
       :post:
       - :name: edit
         :identifier: alert_definition_edit
-      - :name: delete
-        :identifier: alert_definition_delete
-      :delete:
       - :name: delete
         :identifier: alert_definition_delete
   :alerts:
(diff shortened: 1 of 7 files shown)


Fryguy:
Do we also have to then change the verbs (e.g. `*gpppd`) for each of these?  I'm assuming the verbs still exist, its just the permissions definition comes from post?

---

@agrare would also like your review here.

kbrock:
> Do we also have to then change the verbs (e.g. `*gpppd`) for each of these? I'm assuming the verbs still exist, its just the permissions definition comes from post?

Yes, the external API is the same. The privileges for `PUT` and `PATCH` look at `post/edit` both before and after, so they have not changed in the least. We just deleted the sections from `api.yml` that they were previously not being read.

The `DELETE` is changed. Before some `post/delete` were allowed and `delete/delete` were not, and vice versa. After, they both are both allowed or neither is allowed.

It looks like a few of the discovery methods are not not happy / failing tests.
I'll look into that and see if it is a simple addition or if we need to throw this PR away.

kbrock:
pushed update:

In specs, when looking up privileges, it sometimes looked for delete/delete, put/edit, patch/edit.
I could have changed a few hundred lines of specs but opted to modify the helper to ignore the http method parameter instead.
Holler if you want me to modify the specs to fix it. (easy to do but adds a lot of churn and may not be as backport friendly)

Some specs discovered the methods available, and delete/delete is no longer returned. This is a change to the API. I opted to modify the specs but can look into modifying the spec description generator instead.

spec/support/api/helpers.rb
```diff
@@ -50,6 +50,8 @@ def entity_action_identifier(type, entity, action, method = :post)
         end
 
         def action_identifier(type, action, selection = :resource_actions, method = :post)
+          method = action == :read ? :get : :post
```
Fryguy, line 53: Does this effectively ignore the passed in `method` then?

Fryguy: Oh I misread your comment, so yes it does ignore it.  In that case, let's maybe change the method signature to say:

```diff
-         def action_identifier(type, action, selection = :resource_actions, method = :post)
+         def action_identifier(type, action, selection = :resource_actions, _method = :post)
```

kbrock: yea. I debated fixing all the callers or here.
Would like your thoughts on whether it is better to "do it right" and have a lot of change/harder to back port or do it quick.

kbrock: heh. your second comment just showed up. just pushed that change.

kbrock:
push:
- change tenant_group spec to pass symbols
- changed helper signature to signify that the `_method` is ignored.

config/api.yml
```diff
@@ -68,15 +65,6 @@
         :identifier: miq_action_edit
       - :name: delete
         :identifier: miq_action_delete
-      :patch:
-      - :name: edit
-        :identifier: miq_action_edit
-      :put:
-      - :name: edit
-        :identifier: miq_action_edit
```
agrare, line 76: @kbrock are these product features still used when you POST with an action of edit?

kbrock: we have never used put/edit, or patch/edit, we always use post/edit.
For all these examples, there should remain a post/edit. If not, then I introduced a bug

kbrock: let me try again

Yes. We use `miq_action_edit` when we `PUT/edit` and `POST/edit`.
We are currently, and have always, used `POST/edit` privileges for both.

agrare (approved):

Issue #1117, opened 2022-01-18
Template API Issue with child_resources


GilbertCherrie:
When trying to edit a template using the edit API call to a /api/templates/{recordId} it returns an error:

<img width="1129" alt="Error" src="https://user-images.githubusercontent.com/32444791/149977422-7c1c3a33-8555-468c-b2da-df63c9b3e1e3.png">

Data being sent in API call:
<img width="305" alt="Screen Shot 2022-01-18 at 11 26 56 AM" src="https://user-images.githubusercontent.com/32444791/149977522-f37303df-d3df-4830-a025-c4baf9825507.png">

This same call works for /api/vms/ but not templates.

Fryguy:
So, as discussed, the templates controller doesn't accept the parent_resource and child_resources, but I also didn't think it should blow up, since the templates controller explicitly removes everything except name and description.  Do you have a stack trace?

Fryguy:
@kbrock Can you help @GilbertCherrie here?  it's almost like the values aren't being removed?

GilbertCherrie:
@Fryguy I guess I wasn't too clear but on my local API code I changed the code so that it doesn't remove any values. This error is the result of that. I just thought the edit should still work even if all the logic is missing in the api code for the parent and children.

GilbertCherrie:
I fixed the issue and will get a PR out. After reviewing the vm controller file it turns out the vm is not actually sending in the parent and child resources to the update call. These are only used in different functions that set the parent and child relationships. So after adding in the logic from the vm controller file, the template controller file seems to be fixed now.

PR #1118, opened 2022-01-19
Fixed Templates and VM API

GilbertCherrie:
Fixes: https://github.com/ManageIQ/manageiq-api/issues/1117

Fixed the edit template API call and the VM controller "No Parent" option.

### DIFF
--- a/app/controllers/api/mixins/genealogy.rb
+++ b/app/controllers/api/mixins/genealogy.rb
@@ -1,0 +1,51 @@
+module Api
+  module Mixins
+    module Genealogy
+      RELATIONSHIP_COLLECTIONS = %w[vms templates].freeze
+      VALID_EDIT_ATTRS = %w[description name child_resources parent_resource].freeze
+
+      def edit_resource_with_genealogy(type, id, data)
+        attrs = validate_edit_data(data)
+        parent, children = build_parent_children(data)
+        resource_search(id, type).tap do |resource|
+          resource.replace_children(children)
+          resource.set_parent(parent)
+          resource.update!(attrs)
+        end
+      end
+
+      private
+
+      def validate_edit_data(data)
+        invalid_keys = data.keys - VALID_EDIT_ATTRS - valid_custom_attrs
+        raise BadRequestError, "Cannot edit values #{invalid_keys.join(', ')}" if invalid_keys.present?
+
+        data.except('parent_resource', 'child_resources')
+      end
+
+      def valid_custom_attrs
+        Vm.virtual_attribute_names.select { |name| name =~ /custom_\d/ }
+      end
+
+      def build_parent_children(data)
+        children = if data.key?('child_resources')
+                     data['child_resources'].collect do |child|
+                       fetch_parent_child_relationship(child['href'])
+                     end
+                   end
+
+        parent_href = data.dig('parent_resource', 'href')
+        parent = fetch_parent_child_relationship(parent_href) if parent_href
+
+        [parent, Array.wrap(children)]
+      end
+
+      def fetch_parent_child_relationship(href)
+        href = Href.new(href)
+        raise "Invalid relationship type #{href.subject}" unless RELATIONSHIP_COLLECTIONS.include?(href.subject)
+
+        resource_search(href.subject_id, href.subject)
+      end
+    end
+  end
+end
--- a/app/controllers/api/templates_controller.rb
+++ b/app/controllers/api/templates_controller.rb
@@ -1,11 +1,14 @@
 module Api
   class TemplatesController < BaseController
+    include Api::Mixins::Genealogy
     include Subcollections::Policies
     include Subcollections::PolicyProfiles
     include Subcollections::Tags
 
-    def edit_resource(type, id = nil, data = {})
-      super(type, id, data.extract!('name', 'description'))
+    def edit_resource(type, id, data)
+      edit_resource_with_genealogy(type, id, data)
+    rescue => err
+      raise BadRequestError, "Cannot edit Template - #{err}"
     end
 
     def check_compliance_resource(type, id, _data = nil)
--- a/app/controllers/api/vms_controller.rb
+++ b/app/controllers/api/vms_controller.rb
@@ -2,6 +2,7 @@ module Api
   class VmsController < BaseController
     extend Api::Mixins::CentralAdmin
     include Api::Mixins::PolicySimulation
+    include Api::Mixins::Genealogy
     include Subcollections::Accounts
     include Subcollections::Cdroms
     include Subcollections::Compliances
@@ -15,9 +16,6 @@ module Api
     include Subcollections::Snapshots
     include Subcollections::Software
     include Subcollections::Tags
-
-    RELATIONSHIP_COLLECTIONS = %w[vms templates].freeze
-    VALID_EDIT_ATTRS = %w[description name child_resources parent_resource].freeze
 
     def start_resource(type, id = nil, _data = nil)
       enqueue_ems_action(type, id, "Starting", :method_name => "start", :supports => true)
@@ -47,13 +45,7 @@ module Api
     end
 
     def edit_resource(type, id, data)
-      attrs = validate_edit_data(data)
-      parent, children = build_parent_children(data)
-      resource_search(id, type).tap do |vm|
-        vm.replace_children(children)
-        vm.set_parent(parent)
-        vm.update!(attrs)
-      end
+      edit_resource_with_genealogy(type, id, data)
     rescue => err
       raise BadRequestError, "Cannot edit VM - #{err}"
     end
@@ -210,36 +202,6 @@ module Api
 
     def miq_server_message(miq_server)
       miq_server ? "Set miq_server id:#{miq_server.id}" : "Removed miq_server"
-    end
-
-    def validate_edit_data(data)
-      invalid_keys = data.keys - VALID_EDIT_ATTRS - valid_custom_attrs
-      raise BadRequestError, "Cannot edit values #{invalid_keys.join(', ')}" if invalid_keys.present?
-      data.except('parent_resource', 'child_resources')
-    end
-
-    def build_parent_children(data)
-      children = if data.key?('child_resources')
-                   data['child_resources'].collect do |child|
-                     fetch_relationship(child['href'])
-                   end
-                 end
-
-      parent = if data.key?('parent_resource')
-                 fetch_relationship(data['parent_resource']['href'])
-               end
-
-      [parent, Array(children)]
-    end
-
-    def fetch_relationship(href)
-      href = Href.new(href)
-      raise "Invalid relationship type #{href.subject}" unless RELATIONSHIP_COLLECTIONS.include?(href.subject)
-      resource_search(href.subject_id, href.subject)
-    end
-
-    def valid_custom_attrs
-      Vm.virtual_attribute_names.select { |name| name =~ /custom_\d/ }
     end
 
     def vm_ident(vm)
--- a/spec/requests/templates_spec.rb
+++ b/spec/requests/templates_spec.rb
@@ -28,9 +28,46 @@ RSpec.describe "Templates API" do
   end
 
   context "editing a template" do
-    let!(:template) { FactoryBot.create(:template, :name => 'foo', :description => 'bar') }
+    let!(:template)     { FactoryBot.create(:template, :name => 'foo', :description => 'bar') }
+    let(:ems)           { FactoryBot.create(:ems_vmware) }
+    let(:host)          { FactoryBot.create(:host) }
+    let(:template_openstack)       { FactoryBot.create(:template_openstack, :host => host, :ems_id => ems.id, :raw_power_state => "ACTIVE") }
+    let(:template_openstack1)      { FactoryBot.create(:template_openstack, :host => host, :ems_id => ems.id, :raw_power_state => "ACTIVE") }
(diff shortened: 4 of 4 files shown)


app/controllers/api/vms_controller.rb
```diff
@@ -226,7 +227,11 @@ def build_parent_children(data)
                  end
 
       parent = if data.key?('parent_resource')
-                 fetch_relationship(data['parent_resource']['href'])
+                 if data['parent_resource'].nil?
+                  nil
+                 else
+                   fetch_relationship(data['parent_resource']['href'])
+                 end
```
kbrock, line 234: think `dig` is your friend:

```ruby
  href = data.dig('parent_resource', 'href')
  parent = fetch_relationship(href) if href
```

app/controllers/api/templates_controller.rb
```diff
@@ -4,8 +4,51 @@ class TemplatesController < BaseController
...
+
+      parent = if data.key?('parent_resource')
+                 if data['parent_resource'].nil?
+                   nil
+                 else
+                   fetch_relationship(data['parent_resource']['href'])
+                 end
+               end
+
+      [parent, Array(children)]
```
kbrock, line 44: I think you want to use `Array.wrap()` here:

```suggestion
      [parent, Array.wrap(children)]
```

not sure if you want `Array(children)` or `Array.wrap(children)`
If `children = [child]`, did you want `[[child]]` or `[child]` coming back?
Also note `Array.wrap(nil) # => nil`

app/controllers/api/templates_controller.rb
```diff
@@ -4,8 +4,51 @@ class TemplatesController < BaseController
...
+                   end
+                 end
+
+      parent = if data.key?('parent_resource')
+                 if data['parent_resource'].nil?
+                   nil
+                 else
+                   fetch_relationship(data['parent_resource']['href'])
+                 end
+               end
```
kbrock, line 42:
```suggestion
            parent_href = data.dig('parent_resource', 'href')
            parent = fetch_relationship(parent_href) if parent_href
```

kbrock (commented):
this looks good.

If you are just copy pasting, let me know and I can merge and then put my suggestions into another PR.

GilbertCherrie:
> this looks good.
> 
> If you are just copy pasting, let me know and I can merge and then put my suggestions into another PR.

Yeah this was mostly copy paste. Not sure if this needs test cases or not and I want to finalize my UI PR before I take this off WIP.

GilbertCherrie:
@miq-bot add-label bug

spec/requests/templates_spec.rb
```diff
@@ -28,10 +28,48 @@
...
 
   context "editing a template" do
-    let!(:template) { FactoryBot.create(:template, :name => 'foo', :description => 'bar') }
+    let!(:template)     { FactoryBot.create(:template, :name => 'foo', :description => 'bar') }
+    let(:zone)          { FactoryBot.create(:zone, :name => "api_zone") }
+    let(:ems)           { FactoryBot.create(:ems_vmware, :zone => zone) }
+    let(:host)          { FactoryBot.create(:host) }
+    let(:template_openstack)       { FactoryBot.create(:template_openstack, :host => host, :ems_id => ems.id, :raw_power_state => "ACTIVE") }
+    let(:template_openstack1)      { FactoryBot.create(:template_openstack, :host => host, :ems_id => ems.id, :raw_power_state => "ACTIVE") }
+    let(:template_openstack2)      { FactoryBot.create(:template_openstack, :host => host, :ems_id => ems.id, :raw_power_state => "ACTIVE") }
```
kbrock, line 37:
```suggestion
    let(:template_openstack)       { FactoryBot.create(:template_openstack, :host => host, :ems => ems, :raw_power_state => "ACTIVE") }
    let(:template_openstack1)      { FactoryBot.create(:template_openstack, :host => host, :ems => ems, :raw_power_state => "ACTIVE") }
    let(:template_openstack2)      { FactoryBot.create(:template_openstack, :host => host, :ems => ems, :raw_power_state => "ACTIVE") }
```

spec/requests/templates_spec.rb
```diff
@@ -28,10 +28,48 @@
   end
 
   context "editing a template" do
-    let!(:template) { FactoryBot.create(:template, :name => 'foo', :description => 'bar') }
+    let!(:template)     { FactoryBot.create(:template, :name => 'foo', :description => 'bar') }
+    let(:zone)          { FactoryBot.create(:zone, :name => "api_zone") }
+    let(:ems)           { FactoryBot.create(:ems_vmware, :zone => zone) }
```
kbrock, line 33: don't think zone is needed
```suggestion
    let(:ems)           { FactoryBot.create(:ems_vmware) }
```

if it is, then just ignore this suggestion

spec/requests/templates_spec.rb
```diff
@@ -28,10 +28,48 @@
...
+    let(:host)          { FactoryBot.create(:host) }
+    let(:template_openstack)       { FactoryBot.create(:template_openstack, :host => host, :ems_id => ems.id, :raw_power_state => "ACTIVE") }
+    let(:template_openstack1)      { FactoryBot.create(:template_openstack, :host => host, :ems_id => ems.id, :raw_power_state => "ACTIVE") }
+    let(:template_openstack2)      { FactoryBot.create(:template_openstack, :host => host, :ems_id => ems.id, :raw_power_state => "ACTIVE") }
+    let(:new_vms) { FactoryBot.create_list(:template_openstack, 2) }
     before { api_basic_authorize(action_identifier(:templates, :edit)) }
     subject { send(req, api_template_url(nil, template), :params => params) }
 
+    describe "POST /api/templates/:c_id" do
+      it 'can edit a VM with an appropriate role' do
```
kbrock, line 43:
```suggestion
      it 'can edit a template with an appropriate role' do
```

kbrock (commented):
just those nits.

this looks great

Fryguy:
If it's copy-paste, maybe it's better to make this a mixin named GenealogyMixin and included in both vms and templates.  Or we can do that in a follow up PR.

GilbertCherrie:
> If it's copy-paste, maybe it's better to make this a mixin named GenealogyMixin and included in both vms and templates. Or we can do that in a follow up PR.

It is mostly copy paste. Do you want it as a mixix or is this PR good for now?

spec/requests/templates_spec.rb
GilbertCherrie: This change caused an error but I have implemented the other changes fine.

kbrock:
Mixin sounds like a good route. Do you know how to do things like this?

GilbertCherrie:
yeah I looked at some examples. I can give it a try.

GilbertCherrie:
@Fryguy @kbrock I moved all the code to a mixin file. Is this good to merge now?

app/controllers/api/mixins/genealogy.rb
```diff
@@ -0,0 +1,58 @@
+module Api
+  module Mixins
+    module Genealogy
+      RELATIONSHIP_COLLECTIONS = %w[vms templates].freeze
+      VALID_EDIT_ATTRS = %w[description name child_resources parent_resource].freeze
```
Fryguy, line 5: This is interesting...I'm wondering if we leave the valid attrs in the individual models and only "add" child_resources  and parent_resource...I'm ok with this being in a follow up though, since it might be fairly complicated?

app/controllers/api/mixins/genealogy.rb
```diff
@@ -0,0 +1,58 @@
...
+      def tag_specified(id, data)
+        if id.to_i > 0
+          klass  = collection_class(:tags)
+          tagobj = klass.find(id)
+          return tag_path_to_spec(tagobj.name).merge(:id => tagobj.id)
+        end
+        parse_tag(data)
+      end
+
+      def assign_genealogy(type, id, data)
```
Fryguy, line 16: Not sure I like this name because it doesn't only assign genealogy...it also updates the rest of the attributes.  What about:

```suggestion
      def edit_resource_with_genealogy(type, id, data)
```

Fryguy: Alternatively we can just have `edit_resource` and call super, but I'm concerned that might get "lost" in the mixin and confuse people.  Curious what @kbrock thinks

app/controllers/api/mixins/genealogy.rb
```diff
@@ -0,0 +1,58 @@
+module Api
+  module Mixins
+    module Genealogy
+      RELATIONSHIP_COLLECTIONS = %w[vms templates].freeze
+      VALID_EDIT_ATTRS = %w[description name child_resources parent_resource].freeze
+
+      def tag_specified(id, data)
```
Fryguy, line 7: I think all of these methods should be private so they are not exposed accidentally to the controller/router.

app/controllers/api/mixins/genealogy.rb
Fryguy: I think the way to do this "right" is to have a VmsAndTemplates mixin.  In there we define `VALID_EDIT_ATTRS = %w[description name].freeze`, and remove it from here, and then also move `valid_custom_attrs` into VmsAndTemplate Mixin.  Finally, in the VmsAndTemplates mixin, you add in an include of Api::Mixins::Genealogy (alternatively just inline it right into the VmsAndTemplates mixin).

Then we add here: `VALID_GENEALOGY_ATTRS = %w[child_resources parent_resource].freeze`

Then down below we do 

```
        def validate_edit_data(data)
-         invalid_keys = data.keys - VALID_EDIT_ATTRS - valid_custom_attrs
+        invalid_keys = data.keys - VALID_EDIT_ATTRS - VALID_GENEALOGY_ATTRS - valid_custom_attrs
```

Something like that.

Fryguy: I can push a commit with my idea if that's what we want to do

app/controllers/api/mixins/genealogy.rb
```diff
@@ -0,0 +1,58 @@
...
+                     end
+                   end
+
+        parent_href = data.dig('parent_resource', 'href')
+        parent = fetch_relationship(parent_href) if parent_href
+
+        [parent, Array.wrap(children)]
+      end
+
+      def fetch_relationship(href)
```
Fryguy, line 50: This might be too generically named now that it's in a mixin...not sure.

app/controllers/api/mixins/genealogy.rb
Fryguy, line 7: Why is this method in this mixin?  Seems unrelated to genealogy

GilbertCherrie: Oh I copy and pasted this mixin template from another file, forgot to delete that.

app/controllers/api/mixins/genealogy.rb
GilbertCherrie: I think that sounds good. If you want you can push a commit I can also try to do this on my end if you are busy.

GilbertCherrie: @Fryguy I pushed a commit with all the other suggestions, if you want I can also try implementing this one as well.

GilbertCherrie: @Fryguy I tried to implement this and it just broke my code. I'm assuming I did something wrong and not exactly sure which methods need to go in which mixin.

app/controllers/api/templates_controller.rb
```diff
@@ -1,11 +1,16 @@
 module Api
   class TemplatesController < BaseController
+    include Api::Mixins::Genealogy
     include Subcollections::Policies
     include Subcollections::PolicyProfiles
     include Subcollections::Tags
 
-    def edit_resource(type, id = nil, data = {})
-      super(type, id, data.extract!('name', 'description'))
+    VALID_EDIT_ATTRS = %w[description custom_1 name child_resources parent_resource].freeze
```
Fryguy, line 8: custom_1 should not be needed here because vm_custom_attrs will take care of all of the custom_# columns

Fryguy (approved):
LGTM - @kbrock can you also review?  We're going to defer some of the more complicated refactorings fro my other comment, because for now this works and moves @GilbertCherrie forward.  I'll be creating a follow up PR with the refactorings so we can discss that separately.

@kbrock, I'm not sure if you'd want more specs than this or if this is sufficient.

app/controllers/api/mixins/genealogy.rb
kbrock: your name looks good. I also like the way Gilbert's new code reads.
Good suggestion

app/controllers/api/mixins/genealogy.rb
kbrock: I do like the idea of having the decision of what keys are valid as part of the individual controller. nice call

app/controllers/api/mixins/genealogy.rb
```diff
@@ -0,0 +1,51 @@
...
+      VALID_EDIT_ATTRS = %w[description name child_resources parent_resource].freeze
+
+      def edit_resource_with_genealogy(type, id, data)
+        attrs = validate_edit_data(data)
+        parent, children = build_parent_children(data)
+        resource_search(id, type).tap do |resource|
+          resource.replace_children(children)
+          resource.set_parent(parent)
+          resource.update!(attrs)
+        end
```
kbrock, line 14: in the followup, think this is the part that is genealogy. adding the children.
having the concept of valid params and the like seems part of the controller
having the concepts of adding children seems part of the mixin

PR #1120, opened 2022-01-19
Add VM set_description action

kbrock:
Alternative to #1112

ManageIQ/manageiq#21561

```
POST /api/vms/:id -d {"action": "set_description", "new_description": "test"}
{
  "success": true,
  "message": "VM id:2 name:'VM_NAME' setting description to test",
  "task_id": "82",
  "task_href": "http://localhost:3000/api/tasks/82",
  "href": "http://localhost:3000/api/vms/2"
}
```

@agrare sorry. Just had to tweak the way you wrote `set_resource_description`

### DIFF
--- a/app/controllers/api/vms_controller.rb
+++ b/app/controllers/api/vms_controller.rb
@@ -114,6 +114,16 @@ module Api
     end
     central_admin :rename_resource, :rename
 
+    def set_description_resource(type, id, data = {})
+      new_description = data&.dig("new_description")&.strip
+      api_resource(type, id, "Setting description for", :supports => :set_description) do |vm|
+        raise BadRequestError, "Must specify a new_description" if new_description.blank?
+
+        task_id = vm.set_description_queue(User.current_userid, new_description)
+        action_result(true, "Setting description for #{model_ident(vm, type)} to #{new_description}", :task_id => task_id)
+      end
+    end
+
     def shutdown_guest_resource(type, id = nil, _data = nil)
       enqueue_ems_action(type, id, "Shutting Down", :method_name => "shutdown_guest", :supports => true)
     end
--- a/config/api.yml
+++ b/config/api.yml
@@ -4569,6 +4569,8 @@
         :identifier:
         - vm_retire_now
         - sui_vm_retire
+      - :name: set_description
+        :identifier: vm_edit
       - :name: set_owner
         :identifier: vm_edit
       - :name: set_ownership
@@ -4650,6 +4652,8 @@
       - :name: delete
         :identifier: vm_delete
       - :name: set_owner
+        :identifier: vm_edit
+      - :name: set_description
         :identifier: vm_edit
       - :name: set_ownership
         :identifier: vm_ownership
--- a/spec/requests/vms_spec.rb
+++ b/spec/requests/vms_spec.rb
@@ -1607,6 +1607,49 @@ describe "Vms API" do
     end
   end
 
+  context "Vm set_description action" do
+    it "to an invalid vm" do
+      api_basic_authorize action_identifier(:vms, :set_description)
+
+      post(invalid_vm_url, :params => gen_request(:set_description, "new_description" => "test"))
+
+      expect(response).to have_http_status(:not_found)
+    end
+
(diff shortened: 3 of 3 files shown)


spec/requests/vms_spec.rb
```diff
@@ -1621,6 +1621,50 @@ def query_match_regexp(*tables)
...
+    end
+
+    it "to an invalid vm without appropriate role" do
+      api_basic_authorize
+
+      post(invalid_vm_url, :params => gen_request(:set_description, "new_description" => "test"))
+
+      # this is not found because you can't find the vm to set the description
+      # rather than you can find the vm but you can't do the operation to i
+      expect(response).to have_http_status(:not_found)
```
agrare, line 1640: How is this action different from e.g. VM Retire which returns `:forbidden` for this exact test: https://github.com/ManageIQ/manageiq-api/blob/1372260a328c9b16339bfd0e019c0b8a5fc0fe59/spec/requests/vms_spec.rb#L1339-L1345

I don't have an issue with returning 404 instead of 403 I just think we need to be consistent

kbrock:
@agrare agreed
As we are standardizing the infrastructure, we are going towards 404.

While I prefer 403, we can not go that direction. Our rbac is on the query so nothing comes back - we can't tell the difference between 403/404.

I have a number of outstanding branches/PRs consolidating code and those move us to towards this 404 consistency

app/controllers/api/vms_controller.rb
```diff
@@ -131,6 +131,17 @@ def rename_resource(type, id, data = {})
...
     central_admin :rename_resource, :rename
 
+    def set_description_resource(type, id, data = {})
+      new_description = data&.dig("new_description")&.strip
+      api_resource(type, id, "Setting description for") do |vm|
+        ensure_supports(type, vm, :set_description)
+        raise BadRequestError, "Setting description requires a new_description" if new_description.blank?
+
+        task_id = vm.set_description_queue(User.current_user.userid, new_description)
+        action_result(true, "Setting description for #{model_ident(vm, type)} to #{new_description}", :task_id => task_id)
```
kbrock, line 141: do we need to tell the user what they passed in, or can we just use the default description of `"Setting description for #{model_ident(vm, type)}"` (which will get automatically added for us).

kbrock:
I just rewrote this and realized that this PR already existed

update:
- rebased
- added a few spec lines

config/api.yml
```diff
@@ -4534,6 +4534,8 @@
         :identifier: vm_guest_shutdown
       - :name: reboot_guest
         :identifier: vm_guest_restart
+      - :name: set_description
+        :identifier: vm_rename
```
Fryguy, line 4538: not sure about this identifier...I'm wondering if `vm_edit` is better.

kbrock:
update:

- change privs from vm_rename to vm_edit

jeffibm:
Hey @kbrock, I was trying to use the API like this - 

`http://localhost:3000/api/vms/2983?action=set_description&new_description=test description`

is this the right way?

Also, this is how the component is being rendered.
```
= react('VmEditForm', {:recordId    => @record.id&.to_s,
                         :emsId       => @record.ems_id&.to_s,
                         :showTitle   => !@edit[:explorer],
                         :isTemplate  => @record.kind_of?(::MiqTemplate),
                         :displayName => model_for_vm(@record).display_name})
```

we have 2 kinds of id in here, 
1. `@record.id `(ManageIQ::Providers::redhat-new::InfraManager::Vm) and
2. `@record.ems_id`

I am guessing we need to use the `@record.id ` in the API post request.

kbrock:
Hey Jeffrey,

Use `@record.id`.
The rest looks good to me.

From here, the actual description will not change but rather a message will get sent to the provider to change it at the source.

Think you'll get a task I'd, and when the task completes, it will tell you success but in fact the success is the queuing to the provider.

If you have a provider setup, then it will flow back, otherwise you won't be able to test it on your machine.

kbrock:
@jeffibm Think the issue is this needs to be a POST and not a GET

`http://localhost:3000/api/vms/2983?action=set_description&new_description=test description`
(where `2983` is `@record.id`)

PR #1122, opened 2022-01-21
put rbac around ems lookup

kbrock:
We need to run regular rbac around looking up resources for the user

Found a few lookups that didn't have rbac around them so added it
Don't think there is a way to exploit this other than learning that an ems exists given the `ems.id`.

### DIFF
--- a/app/controllers/api/auth_key_pairs_controller.rb
+++ b/app/controllers/api/auth_key_pairs_controller.rb
@@ -1,7 +1,7 @@
 module Api
   class AuthKeyPairsController < BaseController
     def create_resource(_type, _id = nil, data = {})
-      ext_management_system = ExtManagementSystem.find(data['ems_id'])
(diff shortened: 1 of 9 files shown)


Fryguy:
What happens if a non-accessible ems_id is passed?  Does it return 403/404 or will it return nil and then the next line blows up with a nil error?

kbrock:
> What happens if a non-accessible ems_id is passed? Does it return 403/404 or will it return nil and then the next line blows up with a nil error?

Pretty sure we will return a 404.
If you notice, 

> What happens if a non-accessible ems_id is passed? Does it return 403/404 or will it return nil and then the next line blows up with a nil error?

It will raise a 404 because the ems that they specified is not found (it was hidden from them)

Fryguy:
@kbrock Should this go back to morphy?

PR #1125, opened 2022-01-31
add queue_name for ems_operations

kbrock:
### Overview

In the API, we queue ems operations to go to the ems operations worker.

Most of the places in our code we use `Model.method_queue` and the method handles the details.
A few places in the api use `enqueue_ems_action` to directly put the message on the queue.

In most places, we are getting away from this code and creating an enqueue method for each action (i.e.: `def action_queue`)

### Problem

For ems operations, `enqueue_ems_action` was not setting the correct `queue_name`.

### Before

For 57 cases, we add partial queue items:

```ruby
{
  :role       => "ems_operations",
  :zone       => object.zone,
  :queue_name => nil
}
```

(`grep enqueue_ems_action` to find them)

### After

For these cases, we add the `queue_name`.

```ruby
{
  :role       => "ems_operations",
  :zone       => object.zone,
  :queue_name => object.ems.queue_name_for_ems_operations
}
```

### Before

Service start and stop requests were going to the ems operations worker. This new code blew up since the model did not have an ems defined.

### After

Service start and stop requests are going to the generic worker.

### Extra

Some of our `ContainerImage` specs has a nil ems and were blowing up. So I added the `&.`

### DIFF
--- a/app/controllers/api/base_controller/action.rb
+++ b/app/controllers/api/base_controller/action.rb
@@ -129,6 +129,9 @@ module Api
           queue_options[:tenant_id] = user.current_tenant.id
         end
         queue_options[:zone] = object.my_zone if %w(ems_operations smartstate).include?(options[:role])
+        if !queue_options.key?(:queue_name) && options[:role] == "ems_operations"
+          queue_options[:queue_name] = object.ext_management_system&.queue_name_for_ems_operations
+        end
 
         MiqTask.generic_action_with_callback(task_options, queue_options)
       end
--- a/app/controllers/api/services_controller.rb
+++ b/app/controllers/api/services_controller.rb
@@ -72,15 +72,15 @@ module Api
     end
 
     def start_resource(type, id = nil, _data = nil)
-      enqueue_ems_action(type, id, "Starting", :method_name => "start")
+      enqueue_action(type, id, "Starting", :method_name => "start")
     end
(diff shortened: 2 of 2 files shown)


kbrock:
Looked into the failure here and a little confused by c216a1ed9370328bf2305d16ac00dd77fe4aa132

It says that starting and stopping a service should be done by the ems worker.

@agrare Does this sound right?

agrare:
I've known about service operations for about the last 2 minutes, but no that doesn't look right @kbrock  These direct service operations appear to work only on records in the database not against an EMS

Fryguy:
If I recall, service power operations effectively forward to each of the underlying resources.  I'm not sure though if they put a queue item, which puts _another_ queue item, or if they try to do it directly.

kbrock:
Update:

- no longer sending service operations to ems ops

(was a forced push only for a rebase)

kbrock:
> I've known about service power operations for about the last 2 minutes, but no that doesn't look right @kbrock These direct service operations appear to work only on records in the database not against an EMS
> - @agrare 

I think we are all in agreement here.
Changing these to just go to a normal queue rather than going to the ems operations queue.

Fryguy (approved):
LGTM but want @agrare to review

agrare (approved):

PR #1127, opened 2022-02-05
Switch to GitHub Actions

Fryguy:
Switch to GitHub Actions


.github/workflows/ci.yaml
```diff
@@ -0,0 +1,47 @@
...
+          POSTGRESQL_USER: root
+          POSTGRESQL_PASSWORD: smartvm
+          POSTGRESQL_DATABASE: vmdb_test
+        options: --health-cmd pg_isready --health-interval 2s --health-timeout 5s --health-retries 5
+        ports:
+        - 5432:5432
+      memcached:
+        image: manageiq/memcached:1.5
+        ports:
+        - 11211:11211
```
Fryguy, line 26: Not sure these are right, plus I need health things, but we can do that in a follow up.

PR #1128, opened 2022-02-05
[WIP] Remove memcached service

Fryguy:
I'm not sure why we need this...disabling memcached locally still passes all of tests.  @jrafanie Do you know?

WIP because I want to understand if we need this or not.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -20,10 +20,6 @@ jobs:
         options: --health-cmd pg_isready --health-interval 2s --health-timeout 5s --health-retries 5
         ports:
         - 5432:5432
-      memcached:
-        image: manageiq/memcached:1.5
-        ports:
-        - 11211:11211
     env:
       PGHOST: localhost
       PGPASSWORD: smartvm


kbrock:
I thought it may be used for sessions for tests. but looks like it didn't
It is probably just copy/paste from manageiq `ci.yml`

chessbyte:
> I thought it was for session too, but I don't see it. I also don't see memcached in core.

IMHO, memcached should be in the same repo where Dalli client is

```sh
$ ag dalli | grep Gemfile | less
manageiq-api/Gemfile.lock:547:    dalli (2.7.11)
manageiq-api/Gemfile.lock:1271:  dalli (~> 2.7, >= 2.7.8)
manageiq-providers-ibm_cloud/Gemfile.lock:545:    dalli (2.7.11)
manageiq-providers-ibm_cloud/Gemfile.lock:1268:  dalli (~> 2.7, >= 2.7.8)
manageiq-ui-classic/Gemfile.lock:555:    dalli (2.7.11)
manageiq-ui-classic/Gemfile.lock:1316:  dalli (~> 2.7, >= 2.7.8)
manageiq-providers-openstack/Gemfile.lock:545:    dalli (2.7.11)
manageiq-providers-openstack/Gemfile.lock:1268:  dalli (~> 2.7, >= 2.7.8)
manageiq/Gemfile:34:gem "connection_pool",                                       :require => false # For Dalli
manageiq/Gemfile:36:gem "dalli",                            "~>3.0",             :require => false
manageiq/Gemfile.lock:1041:    dalli (2.7.11)
manageiq/Gemfile.lock:1268:  dalli (~> 2.7, >= 2.7.8)
```

chessbyte:
> IMHO, memcached should be in the same repo where Dalli client is

Having said that, isn't the only code using memcached in the UI?  Maybe both Classic and Service?

Fryguy:
@chessbyte Think your local repos are old...dalli is only in core: https://github.com/search?q=org%3AManageIQ+dalli&type=code

Fryguy:
Here's the cache specs: https://github.com/ManageIQ/manageiq-api/blob/master/spec/requests/authentication_spec.rb#L243

jrafanie:
> Anything that does auth will need memcached because that's where we store the session tokens. There are specs here that specifically test sql, memory and memcache session stores, and somehow they all pass without memcached

Maybe our stubbing isn't working and it's still using the default of memory for test?
https://github.com/ManageIQ/manageiq/blob/master/config/settings/test.yml#L2

kbrock:
If possible, our tests should not use memcached, nor external providers, ldap, ...

I think LJ nailed it on the head

> Maybe our stubbing isn't working and it's still using the default of memory for test?

We use `memory` for testing environments

Lets merge this

jrafanie:
> If possible, our tests should not use memcached, nor external providers, ldap, ...
> 
> I think LJ nailed it on the head
> 
> > Maybe our stubbing isn't working and it's still using the default of memory for test?
> 
> We use `memory` for testing environments

Just to confirm a theory...

from 7 days ago... https://app.travis-ci.com/github/ManageIQ/manageiq-api/jobs/558169983

```
$ bundle exec rake
** Using session_store: ActionDispatch::Session::MemoryStore
** ManageIQ master, codename: Najdorf
[Coveralls] Set up the SimpleCov formatter.
[Coveralls] Using SimpleCov's 'rails' settings.
** Using session_store: ActionDispatch::Session::MemoryStore # <==== 🤦  
** ManageIQ master, codename: Najdorf
Randomized with seed 47213
..........................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................
```

jrafanie (approved):
Based on our findings below, we use memory store in test so I'm 👍 with removing this for now and if we want to test memcached store in the future, we can put this back then.

jrafanie:
> Wait...before we merge...I was pretty certain we _wanted_ to test cache and sql in addition to memory because there were nuances. Maybe I'm wrong, but I'd really like to understand this aspect.

We should have tests just for the differences. 

I don't remember all of the differences but we split large payloads to get around memcached size constraints and threading support not guaranteed for memory unlike the "production" memcached/sql stores.

>> If possible, our tests should not use memcached, nor external providers, ldap, ...

> Ideally our tests should all use real things as much as possible. We have to fake it in places where we can't do this, obviously, like providers, but IMO, if we can use real systems we should, because it tests all the subtleties and nuances. For example, there's a 1MB limit in memcached which cannot be tested with memory without pretending.
>
>I'm fine with memory being the "base" for a majority of the other tests, but there should be a handful of tests explicitly testing all 3 of memory, sql, and cache.

👍 

I think that all makes sense but given we don't actually test memcached/sql in this repo, dropping memcached here is not really changing anything.  We can come back and add tests for this later when we actually use it.

Fryguy:
Maybe I'm not understanding...according to these we test all 3 of memory, sql and memcache here: https://github.com/ManageIQ/manageiq-api/blob/4d3d91787806f420fa675242387e8df671a50c11/spec/requests/authentication_spec.rb#L243 and here https://github.com/ManageIQ/manageiq-api/blob/d819e489a326ef26b5fb1d7f2f1c494c03cc8eba/spec/requests/users_spec.rb#L627

What I'm asking is if that is broken, or stubbed, or something else?  I can't figure it out.  I expected this to fail the memcache test when I remove the memcache service.

Fryguy:
Verified locally that if I disable memcached then the session store specs fail, which is what I suspected should happen.  Not sure when this was fixed, but closing now.

PR #1133, opened 2022-02-17
Use api helper for edit_resource

kbrock:
Part of the journey to move us to using `api_resource` which bundles together calls to `api_action`, `search_resource`, `action_result`

Extracted (but no dependencies)

- [x] #1135
- [x] #1159 
- [ ]

### DIFF
--- a/app/controllers/api/cloud_volumes_controller.rb
+++ b/app/controllers/api/cloud_volumes_controller.rb
@@ -10,16 +10,9 @@ module Api
     end
 
     def edit_resource(type, id, data = {})
-      raise BadRequestError, "Must specify an id for editing a #{type} resource" unless id
-
-      cloud_volume = resource_search(id, type)
-
-      raise BadRequestError, cloud_volume.unsupported_reason(:update) unless cloud_volume.supports?(:update)
-
-      task_id = cloud_volume.update_volume_queue(User.current_user, data)
-      action_result(true, "Updating #{cloud_volume.name}", :task_id => task_id)
-    rescue => err
-      action_result(false, err.to_s)
+      api_resource(type, id, "Updating", :supports => :update) do |cloud_volume|
+        {:task_id => cloud_volume.update_volume_queue(User.current_userid, data.deep_symbolize_keys)}
+      end
     end
 
     def safe_delete_resource(type, id, _data = {})
--- a/app/controllers/api/container_images_controller.rb
+++ b/app/controllers/api/container_images_controller.rb
@@ -2,31 +2,18 @@ module Api
   class ContainerImagesController < BaseController
     include Subcollections::CustomAttributes
 
-    def scan_resource(type, image_id, _payload)
-      raise BadRequestError, "Must specify an id for scanning a #{type} resource" unless image_id
-      api_action(type, image_id) do |klass|
-        image = resource_search(image_id, type, klass)
-        begin
-          task = image.scan
-          if task.present?
-            action_result(true, "#{container_image_ident(image)} scanning", :task_id => task.id)
-          else
-            action_result(false, "#{container_image_ident(image)} failed to start scanning")
-          end
-        rescue => err
-          action_result(false, err.to_s)
+    def scan_resource(type, id, _payload)
+      api_resource(type, id, "Scanning") do |image|
+        if (task = image.scan)
+          {:task_id => task.id}
+        else
+          action_result(false, "Failed Scanning #{model_ident(image, type)}")
         end
       end
     end
 
     def check_compliance_resource(type, id, _data = nil)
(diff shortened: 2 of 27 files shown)


app/controllers/api/cloud_volumes_controller.rb
```diff
@@ -9,16 +9,10 @@ def create_resource(type, _id = nil, data = {})
...
-      cloud_volume = resource_search(id, type)
-
-      raise BadRequestError, cloud_volume.unsupported_reason(:update) unless cloud_volume.supports?(:update)
-
-      task_id = cloud_volume.update_volume_queue(User.current_user, data)
-      action_result(true, "Updating #{cloud_volume.name}", :task_id => task_id)
-    rescue => err
-      action_result(false, err.to_s)
+      api_resource(type, id, "Updating") do |cloud_volume|
+        ensure_supports(type, cloud_volume, :update)
```
Fryguy, line 13: Does it make sense to push down the supports check?  I'm pictureing something like:

```suggestion
      api_resource(type, id, "Updating", :supports => :update) do |cloud_volume|
```

kbrock: Ok, so there is a special case when the action is not the same as the supports action name. I didn't know how to handle this but turns out this is a rare case anyway.
```
      enqueue_ems_action(type, id, "Scanning", :method_name => "scan", :supports => :smartstate_analysis, :role => "smartstate")
```

In our case, I think we want to use `:supports => type` for most (if not all) of these cases.

I can put together a PR with the supports use case.

kbrock: @Fryguy Created #1135 - Is this what you had in mind?

app/controllers/api/container_images_controller.rb
```diff
@@ -2,31 +2,18 @@ module Api
...
-          if task.present?
-            action_result(true, "#{container_image_ident(image)} scanning", :task_id => task.id)
-          else
-            action_result(false, "#{container_image_ident(image)} failed to start scanning")
-          end
-        rescue => err
-          action_result(false, err.to_s)
+    def scan_resource(type, id, _payload)
+      api_resource(type, id, "Scanning") do |image|
+        if (task = image.scan)
```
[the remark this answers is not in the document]
agrare: I _think_ this ends up just being `MiqEvent.raise_evm_event(event_target, event, inputs, options)` in https://github.com/ManageIQ/manageiq-providers-kubernetes/blob/master/app/models/manageiq/providers/kubernetes//container_manager.rb#L1017-L1019

agrare (approved):
LGTM, @Fryguy ?

app/controllers/api/host_aggregates_controller.rb
```diff
@@ -26,37 +20,25 @@ def delete_resource_main_action(_type, host_aggregate, _data = {})
...
+        {:task_id => host_aggregate.remove_host_queue(User.current_userid, new_host)}
+      end
     end
 
     private
 
-    def host_aggregate_ident(host_aggregate)
-      "Host Aggregate id:#{host_aggregate.id} name: '#{host_aggregate.name}'"
+    def search_host(host_id)
+        raise BadRequestError, "Must specify a host_id" if data["host_id"].nil?
```
agrare, line 39: Spoke too soon before CI finished :laughing: should be 
```suggestion
        raise BadRequestError, "Must specify a host_id" if host_id.nil?
```

app/controllers/api/host_aggregates_controller.rb
```diff
@@ -26,37 +20,25 @@ def delete_resource_main_action(_type, host_aggregate, _data = {})
...
+
+        resource_search(data["host_id"], :hosts)
```
agrare, line 41: And

```suggestion
        resource_search(host_id, :hosts)
```

Fryguy:
Both rubocops should be done too.

app/controllers/api/container_images_controller.rb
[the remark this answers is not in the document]
kbrock: Do we want to change this or is it working good enough for now?
(my goal was to refactor more than to change, but if there is a way to make this more standard, then that is good too)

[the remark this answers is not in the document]
agrare: Not saying we should do this now just pointing it out this is good for now

app/controllers/api/host_aggregates_controller.rb
kbrock: that "parameter is not being used" sure catches a lot of errors for me.

kbrock:
update:
- fixed conflicts

PR #1134, opened 2022-02-18
Add support for cloud volume snapshot subcollection

kbrock:
Add Cloud Volume Snapshots to the api:

- Moved the `_queue` calls to the core repo.
- Using `CloudVolumeSnapshot.supports :crud` (instead of `CloudVolume.supports :create_snapshot`)
- Changed methods to use the `raw_` naming convention

Related PRs:

- [x] https://github.com/ManageIQ/manageiq/pull/21801
- [x] https://github.com/ManageIQ/manageiq-providers-openstack/pull/787
- [x] https://github.com/ManageIQ/manageiq-providers-amazon/pull/755

Continuation of #1086

### DIFF
--- a/app/controllers/api/subcollections/cloud_volume_snapshots.rb
+++ b/app/controllers/api/subcollections/cloud_volume_snapshots.rb
@@ -1,0 +1,40 @@
+module Api
+  module Subcollections
+    module CloudVolumeSnapshots
+      def cloud_volume_snapshots_query_resource(object)
+        object.cloud_volume_snapshots
+      end
+
+      def cloud_volume_snapshots_create_resource(parent, type, _id, data)
+        api_action(type, nil) do
+          klass = parent.ext_management_system.class_by_ems(:CloudVolumeSnapshot)
+          ensure_supports(type, klass, :create)
+          message = "Creating cloud volume snapshot #{data["name"]} for #{model_ident(parent, :cloud_volumes)}"
+          task_id = parent.create_volume_snapshot_queue(User.current_userid, data.symbolize_keys)
+
+          action_result(true, message, :task_id => task_id)
+        end
+      end
+
+      def update_resource_cloud_volume_snapshots(_parent, type, id, data)
+        api_resource(type, id, "Updating", :supports => :update) do |cloud_volume_snapshot|
+          {:task_id => cloud_volume_snapshot.update_cloud_volume_queue(User.current_userid, data.symbolize_keys)}
+        end
+      end
+      alias cloud_volume_snapshots_update_resource update_resource_cloud_volume_snapshots
+
+      def delete_resource_cloud_volume_snapshots(_parent, type, id, _data)
+        api_resource(type, id, "Deleting", :supports => :delete) do |cloud_volume_snapshot|
+          {:task_id => cloud_volume_snapshot.delete_cloud_volume_queue(User.current_userid)}
+        end
+      end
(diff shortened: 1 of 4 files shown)


kavyanekkalapu:
@kbrock Yes UI pr is ready. As soon as api is ready, we can make that change.

config/api.yml
```diff
@@ -746,6 +746,12 @@
         :identifier: cloud_tenant_tag
       - :name: unassign
         :identifier: cloud_tenant_tag
+  :cloud_volume_snapshots:
+    :description: Cloud Volume Snapshots
+    :options:
+    - :subcollection
+    :verbs: *gpd
```
Fryguy, line 753: Should this be `gpppd`?

app/controllers/api/subcollections/cloud_volume_snapshots.rb
```diff
@@ -0,0 +1,57 @@
+module Api
+  module Subcollections
+    module CloudVolumeSnapshots
+      def cloud_volume_snapshots_query_resource(object)
+        object.cloud_volume_snapshots
+      end
+
+      def cloud_volume_snapshots_create_resource(parent, _type, _id, data)
+        raise parent.unsupported_reason(:snapshot_create) unless parent.supports?(:snapshot_create)
```
Fryguy, line 9: Should this be using the new refactorings?

Fryguy:
API wise I had a couple of questions, but I think @agrare will understand the CloudVolume specifics better than me.

app/controllers/api/subcollections/cloud_volume_snapshots.rb
kbrock: We have the non sub-collections case nailed down.
This cases are so similar, but there are slight changes around the model passed to supports and the way the resources were looked up.

It is similar enough that creating a separate method feels wrong.
They are different enough that having a method support both got ugly fast

For now I was just punting, but it does look like a few methods could be simplified.

kbrock:
WIP: introducing `CloudVolumeSnapshot.supports :create`

agrare:
@kbrock is there a corresponding core PR that adds the create_snapshot queue method (https://github.com/ManageIQ/manageiq-api/pull/1134/files#diff-fbe61b55150e084b1ac3b9f4f21233ed9ef9841b40f359384bbb8b8ebb25cfa1R15) ?  The only methods I see are `create_volume_snapshot` (not queue) on amazon and openstack CloudVolume subclasses

app/controllers/api/subcollections/cloud_volume_snapshots.rb
kbrock: Sorry I didn't share with you sooner.

app/controllers/api/subcollections/cloud_volume_snapshots.rb
```diff
@@ -0,0 +1,57 @@
...
+          :method_name => "create_snapshot",
+          :args        => [data["name"], data["description"], data.fetch("memory", false)]
+        )
+
+        action_result(true, message, :task_id => task_id)
+      rescue => e
+        action_result(false, e.to_s)
+      end
+
+      def delete_resource_cloud_volume_snapshots(parent, _type, id, _data)
```
kbrock, line 24:
```suggestion
      def delete_resource_cloud_volume_snapshots(parent, type, id, _data)
```

kbrock:
update:
- using `_raw`/ core queue

agrare (approved):

PR #1135, opened 2022-02-18
Move supports to api_resource

kbrock:
extracted from #1133

## Description

`api_resource` works for many of our actions.
Having it handle `supports` in an easy way can be a good win.
It also makes it easy to grep for code that does not implement supports.

For most of these `enqueue_action` and `api_resource` cases, we are going for having `supports` handled automatically. Putting this as part of the framework makes it easy to grep code for missing supports checks.

## Example

`PhysicalStoragesController` shows how to use `api_resource(..., :supports => :delete)`

The existing code base may not use that much but a number of examples will probably get updated in #1133

## Minor wrinkle

All of our cases have the `supports` value the same as the action except for one: the action is `scan` while `supports :smartstate_analysis`. This one case will have a potentially unclear error case for developers.

```ruby
enqueue_ems_action(type, id, "Scanning", :method_name => "scan", :supports => :smartstate_analysis, :role => "smartstate")
```

## Aside

I kinda wish I did not introduce the `delete_resource_main_action and just stuck with `delete_resource_action`. That has a half dozen cases where we could have leveraged `:supports => :delete`:

```ruby
    def delete_resource_main_action(type, key_pair, _data)
      ensure_supports(type, key_pair, :delete)
      {:task_id => key_pair.delete_key_pair_queue(User.current_userid)}
    end

    def delete_resource_action(type, key_pair, _data)
      api_resource(type, id, "Deleting", :supports => :delete) do |key_pair|
        {:task_id => key_pair.delete_key_pair_queue(User.current_userid)}
      end
    end
```

### DIFF
--- a/app/controllers/api/base_controller/action.rb
+++ b/app/controllers/api/base_controller/action.rb
@@ -45,13 +45,14 @@ module Api
       #
       # - enforces id exists
       # - constructs action_result for successes and failures
-      def api_resource(type, id, action_phrase)
+      def api_resource(type, id, action_phrase, supports: false)
         api_action(type, id) do
           id ||= @req.collection_id
           raise BadRequestError, "#{action_phrase} #{type.to_s.titleize} requires an id" unless id
 
           api_log_info("#{action_phrase} #{type.to_s.titleize} id: #{id}")
           resource = resource_search(id, type)
+          ensure_supports(type, resource, supports) if supports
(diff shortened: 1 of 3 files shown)


kbrock:
push: fixed cop and renamed a variable to fix syntax error in `PhysicalResource#delete_action`

Fryguy:
LGTM - @agrare Please also review.

Fryguy (approved):

agrare (approved):

PR #1141, opened 2022-03-24
[WIP] Attach detach cloud volume api

MelsHyrule:
Fixes:
- [ ] https://github.com/ManageIQ/manageiq-ui-classic/issues/7603

dependent upon:
- [x] #1147 
- [x] #1149 

Required for upgrading [_attach.html.haml](https://github.com/ManageIQ/manageiq-ui-classic/blob/master/app/views/vm_common/_attach.html.haml#L33) and [_detach.html.haml](https://github.com/ManageIQ/manageiq-ui-classic/blob/master/app/views/vm_common/_detach.html.haml#L25)

Introduces the attach_resource and detach_resource endpoints for cloud volumes.

### DIFF
--- a/app/controllers/api/base_controller/renderer.rb
+++ b/app/controllers/api/base_controller/renderer.rb
@@ -545,23 +545,24 @@ module Api
       end
 
       def render_options(resource, data = {})
-        p "mels pre render options 1" ## none of this is being hit
-        p "resource"
-        p resource
         klass = collection_class(resource)
-        p "klass"
-        p klass
-        p "mels collection class"
-        p klass
         render :json => OptionsSerializer.new(klass, data).serialize
       end
 
-      def render_update_resource_options(id)
+      def render_update_resource_options(id, action)
         type = @req.collection.to_sym
         resource = resource_search(id, type)
         raise BadRequestError, resource.unsupported_reason(:update) unless resource.supports?(:update)
 
-        render_options(type, :form_schema => resource.params_for_update)
+        if action == "attach"
+          if resource.try(:params_for_attach) ## TODO can we make this one line?
+            render_options(type, :form_schema => resource.params_for_attach)  
+          else
+            render_options(type, :form_schema => {})  
+          end
+        else
+          render_options(type, :form_schema => resource.params_for_update)
+        end
       end
 
       def render_create_resource_options(ems_id)
@@ -574,35 +575,6 @@ module Api
         raise BadRequestError, klass.unsupported_reason(:create) unless klass.supports?(:create)
 
         render_options(type, :form_schema => klass.params_for_create(ems))
-      end
-
-      def render_attach_resource_options_mels(ems_id)
-        p "-------------------------- mels attach options "
-        type = @req.collection.to_sym
-        base_klass = collection_class(type)
-  
-        ## render_create_resource_options
-        ems = resource_search(ems_id, :providers) # TODO catch nil cases
-        klass = ems.class_by_ems(base_klass.name)
-        raise BadRequestError, "No #{type.to_s.titleize} support for - #{ems.name}" unless klass
-        raise BadRequestError, klass.unsupported_reason(:create) unless klass.supports?(:create)
-  
-        ## render_update_resource_options
-        # type = @req.collection.to_sym # we do it above
-        resource = resource_search(ems_id, type)
-        raise BadRequestError, resource.unsupported_reason(:update) unless resource.supports?(:update)
-  
-        ## for hardcoding faster | storageManager: 5 | record id: 445 |
-        p "content"
-        p klass
-        p type
-        p ems
-        p "--------"
-        p type
-        p resource
-        p "---------------------- mels render options" ## this is the last thing that does get hit
-        render_options(type, :form_schema => klass.mels())
-        # render_options(type, :form_schema => resource.mels())
       end
 
       # This is a helper method used by both .determine_include_for_find and
--- a/app/controllers/api/base_controller/renderer.rb
+++ b/app/controllers/api/base_controller/renderer.rb
@@ -565,7 +565,7 @@ module Api
         if resource.try(:params_for_attach) ## TODO can we make this one line?
           render_options(type, :form_schema => resource.params_for_attach)  
         else
-          render_options(type, :form_schema => {})  
+          render_options(type, :form_schema => {}) ## {:fields => []}
         end
       end
 
--- a/app/controllers/api/base_controller/renderer.rb
+++ b/app/controllers/api/base_controller/renderer.rb
@@ -545,7 +545,14 @@ module Api
       end
 
       def render_options(resource, data = {})
+        p "mels pre render options 1" ## none of this is being hit
+        p "resource"
+        p resource
         klass = collection_class(resource)
+        p "klass"
+        p klass
+        p "mels collection class"
+        p klass
         render :json => OptionsSerializer.new(klass, data).serialize
       end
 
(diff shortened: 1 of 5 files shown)


app/controllers/api/cloud_volumes_controller.rb
```diff
@@ -47,5 +47,25 @@ def create_backup_resource(type, id, data)
         {:task_id => cloud_volume.backup_create_queue(User.current_userid, data)}
       end
     end
+
+    def attach_resource(type, id, data = {})
+      raise BadRequestError, "Must specify a vm_id" if data["vm_id"].nil?
+
+      api_resource(type, id, "Attaching Resource", :supports => :attach_volume) do |cloud_volume|
```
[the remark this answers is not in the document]
MelsHyrule: should we update `{:task_id => cloud_volume.detach_volume_queue(User.current_userid, data["vm_id"])}` to be instead ```{:task_id => cloud_volume.detach_volume_queue(User.current_userid, resource_search(data["vm_id"], :vms))}``` ?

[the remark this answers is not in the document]
agrare: `cloud_volume.attach_volume_queue` takes an `:ems_ref` not an `:id` or `Vm` object, so you should lookup the VM like @kbrock's suggestion, then pass in `vm.ems_ref`

app/controllers/api/cloud_volumes_controller.rb
```diff
@@ -51,7 +51,26 @@ def create_backup_resource(type, id, data)
     def restore_backup_resource(type, id, data)
       api_resource(type, id, "Restoring backup for", :supports => :backup_restore) do |cloud_volume|
         {:task_id => cloud_volume.backup_restore_queue(User.current_userid, data.symbolize_keys)}
+    end
```
agrare, line 54: Missing an `end` right above this one from the conflict resolution

spec/requests/cloud_volumes_spec.rb
```diff
@@ -192,6 +192,38 @@
...
+      expect(response).to have_http_status(:ok)
+    end
+
+    it 'detaches Cloud Volume from an instance' do
+      zone = FactoryBot.create(:zone)
+      ems = FactoryBot.create(:ems_autosde, :zone => zone)
+      hw = FactoryBot.create(:hardware)
+      vm = FactoryBot.create(:vm_vmware, :hardware => hw)
+      cloud_volume = FactoryBot.create(:cloud_volume_autosde, :ext_management_system => ems)
+      disk = FactoryBot.create(:disk, :hardware => hw, :backing => cloud_volume)
```
agrare, line 217: If you don't need to use `disk` later on you just need the record you could do 
```suggestion
      FactoryBot.create(:disk, :hardware => hw, :backing => cloud_volume)
```
And then rubocop won't yell at you

app/controllers/api/cloud_volumes_controller.rb
```diff
@@ -53,5 +53,27 @@ def restore_backup_resource(type, id, data)
         {:task_id => cloud_volume.backup_restore_queue(User.current_userid, data.symbolize_keys)}
       end
     end
+
+    def attach_resource(type, id, data = {})
+      api_resource(type, id, "Attaching Resource to", :supports => :attach_volume) do |cloud_volume|
+        raise BadRequestError, "Must specify a vm_id" if data["vm_id"].blank?
+
+        vm = resource_search(data["vm_id"], :vms)
+        {:task_id => cloud_volume.attach_volume_queue(User.current_userid, vm.ems_ref)}
```
agrare, line 62: @MelsHyrule we'll also need an optional `device` parameter here:
```suggestion
        {:task_id => cloud_volume.attach_volume_queue(User.current_userid, vm.ems_ref, data["device"].presence)}
```
should do the trick

spec/requests/cloud_volumes_spec.rb
```diff
@@ -192,6 +192,38 @@
...
     end
+
+    it 'attaches Cloud Volume to an instance' do
+      zone = FactoryBot.create(:zone)
+      ems = FactoryBot.create(:ems_autosde, :zone => zone)
+      vm = FactoryBot.create(:vm_vmware)
+      cloud_volume = FactoryBot.create(:cloud_volume_autosde, :ext_management_system => ems)
+
+      api_basic_authorize(action_identifier(:cloud_volumes, :attach, :resource_actions, :post))
+      stub_supports(cloud_volume.class, :attach)
```
agrare, line 203:
```suggestion
      stub_supports(cloud_volume.class, :attach_volume)
```

spec/requests/cloud_volumes_spec.rb
```diff
@@ -192,6 +192,38 @@
...
+    it 'detaches Cloud Volume from an instance' do
+      zone = FactoryBot.create(:zone)
+      ems = FactoryBot.create(:ems_autosde, :zone => zone)
+      hw = FactoryBot.create(:hardware)
+      vm = FactoryBot.create(:vm_vmware, :hardware => hw)
+      cloud_volume = FactoryBot.create(:cloud_volume_autosde, :ext_management_system => ems)
+      FactoryBot.create(:disk, :hardware => hw, :backing => cloud_volume)
+
+      api_basic_authorize(action_identifier(:cloud_volumes, :detach, :resource_actions, :post))
+      stub_supports(cloud_volume.class, :detach)
```
agrare, line 220:
```suggestion
      stub_supports(cloud_volume.class, :detach_volume)
```

app/controllers/api/cloud_volumes_controller.rb
```diff
@@ -34,14 +34,48 @@ def delete_resource_main_action(_type, cloud_volume, _data)
...
         super
       end
     end
 
+    def render_attach_resource_options_mels(ems_id)
+      p "-------------------------- mels attach options "
+      type = @req.collection.to_sym
+      base_klass = collection_class(type)
+
+      ems = resource_search(ems_id, :providers)
```
kbrock, line 56: ems may be nil

app/controllers/api/base_controller/renderer.rb
```diff
@@ -549,12 +549,20 @@ def render_options(resource, data = {})
...
       end
 
-      def render_update_resource_options(id)
+      def render_update_resource_options(id, action)
         type = @req.collection.to_sym
         resource = resource_search(id, type)
         raise BadRequestError, resource.unsupported_reason(:update) unless resource.supports?(:update)
 
-        render_options(type, :form_schema => resource.params_for_update)
+        if action == "attach"
```
agrare, line 557: I don't think overriding the `render_update_resource_options` method is the way to go here

For example won't we not get here if we are looking for the "attach" action but the resource doesn't support "update" a totally different action? https://github.com/ManageIQ/manageiq-api/pull/1141/files#diff-48fb0a597ebc41e5c8c4fde22100d5577a08a07b84761152f69eaa2f580c4c37R555

Plus "update" is in the name, then we also have it handle other options

app/controllers/api/cloud_volumes_controller.rb
```diff
@@ -34,7 +34,8 @@ def delete_resource_main_action(_type, cloud_volume, _data)
 
     def options
       if (id = params["id"])
-        render_update_resource_options(id)
+        action = params["option_action"] || "update"
+        render_update_resource_options(id, action)
```
agrare, line 38: Could we `send("render_#{action}_resource_options(id)")` instead?  Then we would localize the methods to the controller that has those actions

agrare:
@MelsHyrule I was good with tackling the `params_for_attach` in a separate PR (https://github.com/ManageIQ/manageiq-ui-classic/pull/8205#discussion_r841804705) I think what you had prior to trying to add the OPTIONS call was good if you want to move forward with that and I think @kbrock is tackling a general way of dealing with action options in his other PR

agrare:
@MelsHyrule merged https://github.com/ManageIQ/manageiq-api/pull/1147 if you want to build on that

MelsHyrule:
With changes from https://github.com/ManageIQ/manageiq-api/pull/1147 this PR is no longer needed, closing now.

PR #1146, opened 2022-04-08
Custom button set reorder API

akhilkr128:
Adding Api for [task](https://github.com/ManageIQ/manageiq-ui-classic/issues/7901)

Dependent PR:
- [x] https://github.com/ManageIQ/manageiq/pull/21814

## Successful API call

Screenshot 2022-04-06 at 10 36 18 AM

## Call returns error

Screenshot 2022-04-06 at 10 36 02 AM

### DIFF
--- a/app/controllers/api/custom_button_sets_controller.rb
+++ b/app/controllers/api/custom_button_sets_controller.rb
@@ -7,5 +7,12 @@ module Api
     def edit_resource(type, id, data)
       super(type, id, data.deep_symbolize_keys)
     end
+
+    def reorder_resource(type, id, data)
+      api_action(type, id) do
+        CustomButtonSet.reorder_group_index(data["ids"])
+        action_result(true, "Button Group Reorder saved")
+      end
+    end
   end
 end
--- a/config/api.yml
+++ b/config/api.yml
@@ -1247,6 +1247,8 @@
         :identifier: ab_group_edit
       - :name: delete
         :identifier: ab_group_delete
+      - :name: reorder
+        :identifier: ab_group_reorder
     :resource_actions:
       :get:
       - :name: read
--- a/spec/requests/custom_button_sets_spec.rb
+++ b/spec/requests/custom_button_sets_spec.rb
@@ -108,6 +108,25 @@ RSpec.describe 'CustomButtonSets API' do
       expect(response.parsed_body).to include(expected)
       expect(cb_set.reload.set_data[:button_icon]).to eq("ff ff-closed")
     end
+
+    it 'can reorder custom button set by ids' do
+      api_basic_authorize collection_action_identifier(:custom_button_sets, :reorder)
+
+      request = {
+        'action'   => 'reorder',
+        'resource' => {
+          'ids' => [cb_set.id, cb_set2.id]
+        }
+      }
+
+      post(api_custom_button_sets_url, :params => request)
+
+      expected = {
+        'results' => [{"href" => "#{api_custom_button_sets_url}/", "message" => 'Button Group Reorder saved', "success" => true}]
+      }
+      expect(response).to have_http_status(:ok)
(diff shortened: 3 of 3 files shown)


akhilkr128:
@miq-bot cross-repo-test https://github.com/ManageIQ/manageiq/pull/21814

akhilkr128:
@miq-bot cross-repo-test https://github.com/ManageIQ/manageiq/pull/21814

config/api.yml
```diff
@@ -1210,6 +1210,8 @@
         :identifier: ab_group_edit
       - :name: delete
         :identifier: ab_group_delete
+      - :name: order
```
Fryguy, line 1213: I'm thinking the action name should be `reorder`.  To me, `order`, sounds like you're going to purchase something.

app/controllers/api/custom_button_sets_controller.rb
```diff
@@ -7,5 +7,12 @@ def create_resource(type, id, data)
     def edit_resource(type, id, data)
       super(type, id, data.deep_symbolize_keys)
     end
+
+    def order_resource(_, _, data)
+      CustomButtonSet.update_group_index(data["ids"])
```
Fryguy, line 12: I think we need an RBAC check here, or does that happen automatically?  cc @kbrock

Fryguy: @akhilkr128 Do you know what RBAC was used in the old UI?

app/controllers/api/custom_button_sets_controller.rb
```diff
@@ -7,5 +7,12 @@ def create_resource(type, id, data)
     def edit_resource(type, id, data)
       super(type, id, data.deep_symbolize_keys)
     end
+
+    def order_resource(_, _, data)
```
Fryguy, line 11: @akhilkr128 As mentioned elsewhere, I'm thinking this should be `reorder`.  I'm concerned the term `order` might get confused with things like ordering services.

config/api.yml
akhilkr128: @Fryguy Do you think `sort` will be a better alternative?

app/controllers/api/custom_button_sets_controller.rb
akhilkr128: We need an admin role, correct me if I am wrong. 
see the [yml](https://github.com/ManageIQ/manageiq/blob/619680b323bc4305e23508cb40228225fdc5b85e/db/fixtures/miq_product_features.yml#L2583)

akhilkr128:
@miq-bot cross-repo-test https://github.com/ManageIQ/manageiq/pull/21814

app/controllers/api/custom_button_sets_controller.rb
```diff
@@ -7,5 +7,12 @@ def create_resource(type, id, data)
     def edit_resource(type, id, data)
       super(type, id, data.deep_symbolize_keys)
     end
+
+    def reorder_resource(_, _, data)
+      CustomButtonSet.reorder_group_index(data["ids"])
+      action_result(true, "Button Group Reorder saved")
+    rescue => error
+      action_result(false, error.to_s)
+    end
```
kbrock, line 16: Probably not necessary but I'd feel more comfortable using the `api_action` pattern

It does little things different error reports if groups of calls are bundled together.

```suggestion
    def reorder_resource(type, id, data)
      api_action(type, id) do
        CustomButtonSet.reorder_group_index(data["ids"])
        action_result(true, "Button Group Reorder saved")
      end
    end
```

kbrock (commented):
this looks good.
I merged the core change so this should be 🍏 now.
just the one change and then we can merge

akhilkr128:
@miq-bot cross-repo-test https://github.com/ManageIQ/manageiq/pull/21814

PR #1147, opened 2022-04-12
allow OPTIONS and DDF for custom actions

kbrock:
This allows us to provide DDF forms for custom actions
In this case, we are providing DDF for CloudVolume#attach


app/controllers/api/base_controller/renderer.rb
```diff
@@ -549,6 +549,18 @@ def render_options(resource, data = {})
...
 
+      # TODO: not sure if we should keep a separate update
+      #       or have 2 copy/paste methods
+      #       changing all params_for_update seems a bit much
+      def render_resource_options(id, action = :update)
+        type = @req.collection.to_sym
+        resource = resource_search(id, type)
+        raise BadRequestError, resource.unsupported_reason(action) unless resource.supports?(action)
+
+        schema = action == :update ? resource.params_for_update : resource.params_for_action(action)
```
agrare, line 560: > resource.params_for_action(action)

This is going to end up with a method and a giant `case/when` statement in the providers, might it be better to have a `params_for_#{action}` method?

kbrock: we can do this. was hoping to avoid the dynamic method generation but you are probably right

app/controllers/api/base_provider_controller.rb
```diff
@@ -0,0 +1,19 @@
...
+      elsif (ems_id = params["ems_id"])
+        render_create_resource_options(ems_id)
+      # from base controller:
+      elsif params.key?(:subcollection)
+        # TODO: add subcollection support
+        # (not sure if this still makes sense)
+        render :json => ""
+      else
+        render_options(@req.collection)
+      end
```
agrare, line 16: Would it make sense to `else super` ?

lib/api/request_adapter.rb
```diff
@@ -58,6 +58,8 @@ def json_body
       @json_body ||= begin
                        body = @request.body.read if @request.body
                        body.blank? || body == "null" ? {} : JSON.parse(body)
+                     rescue JSON::ParserError
+                       @request.params
```
agrare, line 62: This might turn out to be necessary, but curious that in the POST case the body comes in as JSON but in the OPTIONS case it is key=val seems like it might be something else causing the body not to be coming in as JSON

app/controllers/api/base_controller/renderer.rb
agrare: Yeah 100% know what you mean, it is guarded by the `resource.supports?(action)` so we aren't just `.send()`'ing anything, and we discussed checking the collection actions from api.yml that should mitigate any user provided input issues

lib/api/request_adapter.rb
kbrock: think options are not expecting parameters and that is why this is getting messed up

kbrock:
push:
- rebase
- added priv checks for options

spec/requests/cloud_volumes_spec.rb
```diff
@@ -177,6 +177,29 @@
...
+    end
+
+    it 'returns no DDF schema for non supported actions via OPTIONS' do
+      zone = FactoryBot.create(:zone)
+      provider = FactoryBot.create(:ems_autosde, :zone => zone)
+      cloud_volume = FactoryBot.create(:cloud_volume_autosde, :ext_management_system => provider)
+
+      options(api_cloud_volume_url(nil, cloud_volume), :params => {"option_action" => "bogus"})
+
+      expect(response).to have_http_status(:bad_request)
```
kbrock, line 201: do we want this case to just be blank `form_schema => {}` or do we want to keep an error here?
The previous code seemed to just return a blank result when in doubt.

lib/api/request_adapter.rb
```diff
@@ -55,10 +61,13 @@ def hide?(thing)
...
-      @json_body ||= begin
-                       body = @request.body.read if @request.body
-                       body.blank? || body == "null" ? {} : JSON.parse(body)
-                     end
+      @json_body ||=
+        begin
+          body = @request.body.read if @request.body
+          body.blank? || body == "null" ? {} : JSON.parse(body)
+        rescue JSON::ParserError
+          @request.params
```
kbrock, line 69: @agrare you ok with this change?

It is strange that http `OPTIONS` does not have a JSON `body` while `GET` and `POST` seem to handle the parameters just fine.

lib/api/request_adapter.rb
```diff
@@ -25,6 +25,12 @@ def action
                   end
     end
 
+    # currently only custom option actions are returned
+    # default options of :create and :update are not handled
+    def option_action
+      @params["option_action"]
+    end
```
kbrock, line 32: not thrilled with this.

It is nice to have this only defined in one spot.
To support the validation check, we would either need to expose `@req.params` or this field.
Seemed the better of the two options.

app/controllers/api/generic_object_definitions_controller.rb
```diff
@@ -111,7 +111,7 @@ def self.allowed_types
     end
 
     def options
-      render_options(:generic_object_definitions, build_generic_object_definition_options)
+      render_options(@req.collection, build_generic_object_definition_options)
```
kbrock, line 114: Sorry.
these changes were not necessary here

I was looking to generalize the handling of `options` and these 3 changes came out here.
I decided this to be as far as we needed to go.

lib/api/request_adapter.rb
agrare: I was hoping we'd be able to find out why it is different since it seems strange, e.g. is it a default Content-Type thing?

app/controllers/api/base_provider_controller.rb
```diff
@@ -0,0 +1,14 @@
...
+    def options
+      if (id = params["id"])
+        action = (@req.option_action || :update).to_sym
+        render_resource_options(id, action)
+      elsif (ems_id = params["ems_id"])
+        render_create_resource_options(ems_id)
+      else
+        super
+      end
+    end
```
kbrock, line 12: alternatively, we could make this change in the base controller.

I had higher ambitions for the `BaseProviderController` but this seemed good for now.

lib/api/request_adapter.rb
kbrock: ok. not sure what was up with rack, but it is strange that they are reporting that there is a body. A body is not allowed from the spec. It could be in rack or a bug in our http request code

app/controllers/api/generic_object_definitions_controller.rb
agrare: @kbrock was this necessary?

PR #1149, opened 2022-04-18
Attach detach cloud volume form (simplified)

MelsHyrule:
Required for upgrading [_attach.html.haml](https://github.com/ManageIQ/manageiq-ui-classic/blob/master/app/views/vm_common/_attach.html.haml#L33) and [_detach.html.haml](https://github.com/ManageIQ/manageiq-ui-classic/blob/master/app/views/vm_common/_detach.html.haml#L25) to react as per [7603](https://github.com/ManageIQ/manageiq-ui-classic/issues/7603).

Introduces the attach_resource and detach_resource endpoints for cloud volumes. 

This PR is a more simplified version of work done on [1141](https://github.com/ManageIQ/manageiq-api/pull/1141). Here we have not yet introduced the work done to make `options` more generic. That logic needs further testing and for the purposes of this PR falls slightly out of scope. 

@miq-bot add-reviewer @kbrock 
@miq-bot add-reviewer @agrare 
@miq-bot add-reviewer @kavyanekkalapu
@miq-bot assign @kavyanekkalapu

### DIFF
--- a/app/controllers/api/cloud_volumes_controller.rb
+++ b/app/controllers/api/cloud_volumes_controller.rb
@@ -55,5 +55,27 @@ module Api
         {:task_id => cloud_volume.backup_restore_queue(User.current_userid, backup['ems_ref'])}
       end
     end
+
+    def attach_resource(type, id, data = {})
+      api_resource(type, id, "Attaching Resource to", :supports => :attach_volume) do |cloud_volume|
+        raise BadRequestError, "Must specify a vm_id" if data["vm_id"].blank?
+
+        vm = resource_search(data["vm_id"], :vms)
+        {:task_id => cloud_volume.attach_volume_queue(User.current_userid, vm.ems_ref, data["device"].presence)}
+      end
+    rescue => err
+      action_result(false, err.to_s)
+    end
+
+    def detach_resource(type, id, data = {})
+      api_resource(type, id, "Detaching Resource from", :supports => :detach_volume) do |cloud_volume|
+        raise BadRequestError, "Must specify a vm_id" if data["vm_id"].blank?
+
+        vm = resource_search(data["vm_id"], :vms)
+        {:task_id => cloud_volume.detach_volume_queue(User.current_userid, vm.ems_ref)}
+      end
+    rescue => err
+      action_result(false, err.to_s)
+    end
   end
 end
--- a/config/api.yml
+++ b/config/api.yml
@@ -808,6 +808,10 @@
         :identifier: cloud_volume_backup_create
       - :name: restore_backup
         :identifier: cloud_volume_backup_restore
+      - :name: attach_volume
+        :identifier: cloud_volume_edit
+      - :name: detach_volume
+        :identifier: cloud_volume_edit
     :tags_subcollection_actions:
       :post:
       - :name: assign
--- a/spec/requests/cloud_volumes_spec.rb
+++ b/spec/requests/cloud_volumes_spec.rb
@@ -208,6 +208,68 @@ describe "Cloud Volumes API" do
 
       expect(response).to have_http_status(:ok)
     end
+
+    it 'attaches Cloud Volume to an instance' do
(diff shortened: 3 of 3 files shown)


app/controllers/api/base_controller/renderer.rb
```diff
@@ -569,6 +569,18 @@ def render_create_resource_options(ems_id)
...
+      def render_attach_resource_options(id)
+        type = @req.collection.to_sym
+        resource = resource_search(id, type)
+        raise BadRequestError, resource.unsupported_reason(:attach_volume) unless resource.supports?(:attach_volume)
+
+        if resource.try(:params_for_attach) ## TODO can we make this one line?
+          render_options(type, :form_schema => resource.params_for_attach)  
+        else
+          render_options(type, :form_schema => {}) ## {:fields => []}
+        end
```
Fryguy, line 581:
```suggestion
        render_options(type, :form_schema => resource.try(:params_for_attach) || {})
```

Fryguy: I think that should get you to a single line as per your TODO.

app/controllers/api/cloud_volumes_controller.rb
```diff
@@ -34,7 +34,8 @@ def delete_resource_main_action(_type, cloud_volume, _data)
 
     def options
       if (id = params["id"])
-        render_update_resource_options(id)
+        action = params["option_action"] || "update"
+        send("render_#{action}_resource_options", id)
```
Fryguy, line 38: I'm concerned here that someone might pass a strange option_action, which could bypass security in the future.   Not sure if it's better to whitelist attach and detach or just have separate calls out to render_attach_resource_options render_update_resource_options.

app/controllers/api/base_controller/renderer.rb
```diff
@@ -569,6 +569,18 @@ def render_create_resource_options(ems_id)
...
+      end
+
```
agrare, line 583: Hey @MelsHyrule I thought for this simplified version we were going to drop the OPTIONS / params_for_attach and take care of that in a follow-up

MelsHyrule: This function slipped through the cracks when i was simplifying, it's not being used or called so ill remove it now

app/controllers/api/cloud_volumes_controller.rb
agrare, line 38: Don't think you're using these options either right now

app/controllers/api/base_controller/renderer.rb
MelsHyrule: Ill try this change in my other PR where I keep render_options https://github.com/ManageIQ/manageiq-api/pull/1141/files#diff-48fb0a597ebc41e5c8c4fde22100d5577a08a07b84761152f69eaa2f580c4c37R565-R569

app/controllers/api/cloud_volumes_controller.rb
MelsHyrule: This was something that slipped through from the other PR in the split, it will not be merged here so we can look into it more in the other PR https://github.com/ManageIQ/manageiq-api/pull/1141/files#diff-ff1fde5895f3426d678a7f5731cc08c27f3602cda6453a66677502b4731f5173R37-R38

config/api.yml
```diff
@@ -801,6 +801,10 @@
         :identifier: cloud_volume_backup_create
       - :name: restore_backup
         :identifier: cloud_volume_backup_restore
+      - :name: attach
```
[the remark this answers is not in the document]
agrare: @kbrock it is in a sub-module (not sure why) https://github.com/ManageIQ/manageiq/blob/master/app/models/cloud_volume/operations.rb#L4-L7

agrare:
LGTM, @kbrock ?

agrare (approved):

PR #1150, opened 2022-04-20
Fix cloud object store container spec file

GilbertCherrie:
Fixed incorrect value in the cloud object store container spec file. The emsType value is not used in the provider or api code so it is removed from the spec tests.

### DIFF
--- a/spec/requests/cloud_object_store_containers_spec.rb
+++ b/spec/requests/cloud_object_store_containers_spec.rb
@@ -58,7 +58,6 @@ describe "Cloud Object Store Containers API" do
       submit_data = {
         :ems_id          => provider.id,
         :name            => 'foo',
-        :emsType         => provider.type,
         :cloud_tenant_id => cloud_tenant.id,
       }
       post(api_cloud_object_store_containers_url, :params => submit_data)


GilbertCherrie:
@miq-bot add_reviewer @Fryguy 
@miq-bot assign @Fryguy 
@miq-bot add-label bug

agrare:
Is this also an issue here then? https://github.com/ManageIQ/manageiq-ui-classic/blob/master/app/javascript/components/cloud-object-store-container-form/index.jsx#L47

GilbertCherrie:
> Is this also an issue here then? https://github.com/ManageIQ/manageiq-ui-classic/blob/master/app/javascript/components/cloud-object-store-container-form/index.jsx#L47

Yeah it looks like it although I don't know if this value is needed since there seems to be no difference in this code as oppossed to changing it to ems_type. Both hit the no credentials defined so I'm assuming they are the same.

GilbertCherrie:
@agrare I also tried it without submitting the emsType value and it also seemed to work. Do you think this is safe to delete from here and the ui?

kbrock:
so this option ends up being passes as is to the providers.
I have not found anywhere that actually uses this value in our code.
This value will not mean anything to the providers since it is a manageiq specific value

- [openstack/cloud_object_store_container.rb:41](https://github.com/ManageIQ/manageiq-providers-openstack/blob/master/app/models/manageiq/providers/openstack/storage_manager/swift_manager/cloud_object_store_container.rb#L41) `options` sent generally, but the value could not mean anything to them.
- [cloud_object_store_container.rb:76](https://github.com/ManageIQ/manageiq-providers-amazon/blob/master/app/models/manageiq/providers/amazon/storage_manager/s3/cloud_object_store_container.rb#L76) `options` not generically sent to amazon.

GilbertCherrie:
> so this option ends up being passes as is to the providers. I have not found anywhere that actually uses this value in our code. This value will not mean anything to the providers since it is a manageiq specific value
> 
> * [openstack/cloud_object_store_container.rb:41](https://github.com/ManageIQ/manageiq-providers-openstack/blob/master/app/models/manageiq/providers/openstack/storage_manager/swift_manager/cloud_object_store_container.rb#L41) `options` sent generally, but the value could not mean anything to them.
> * [cloud_object_store_container.rb:76](https://github.com/ManageIQ/manageiq-providers-amazon/blob/master/app/models/manageiq/providers/amazon/storage_manager/s3/cloud_object_store_container.rb#L76) `options` not generically sent to amazon.

@kbrock I didn't see this value being used anywhere. I only had it initially because the old form used it in its API call. After some basic testing and looking through code it seems safe to remove.

agrare:
:+1: yeah I don't see this value being used or even useful since we are sending in the ems_id we don't need the type

PR #1151, opened 2022-04-22
Add API controller for physical server profiles

zitnik:
This PR adds API support for physical server profiles coming from the **Cisco Intersight provider**.

Beside the basic GET operations, it supports assignment, deployment, and unassignment of servers:

```
POST /api/physical_server_profiles/
{
    "action": "assign_server",
    "resources": [
        {
            "id": "5",
            "server_id": "90"
        }
    ]
}
```

```
POST /api/physical_server_profiles/
{
    "action": "deploy_server",
    "resources": [
        {
            "id": "5"
        }
    ]
}
```

```
POST /api/physical_server_profiles/
{
    "action": "unassign_server",
    "resources": [
        {
            "id": "5"
        }
    ]
}
```

See also related PRs in the [MiQ core repository (21830)](https://github.com/ManageIQ/manageiq/pull/21830) and the [Cisco Intersight provider repository (42)](https://github.com/ManageIQ/manageiq-providers-cisco_intersight/pull/42).

### DIFF
--- a/app/controllers/api/physical_server_profiles_controller.rb
+++ b/app/controllers/api/physical_server_profiles_controller.rb
@@ -1,0 +1,13 @@
+module Api
+  class PhysicalServerProfilesController < BaseController
+    include Subcollections::EventStreams
+
+
+
+    private
+
+    def ensure_resource_exists(type, id)
+      raise NotFoundError, "#{type} with id:#{id} not found" unless collection_class(type).exists?(id)
+    end
+  end
+end
--- a/app/controllers/api/physical_server_profiles_controller.rb
+++ b/app/controllers/api/physical_server_profiles_controller.rb
@@ -3,9 +3,9 @@ module Api
     include Subcollections::EventStreams
 
     def assign_server_resource(type, id, data)
-      enqueue_ems_action(type, id, :method_name => :assign_server, :args => [data["server_id"]]) do
-        ensure_resource_exists(:physical_servers, data["server_id"])
-      end
+      # Make sure the requested server exists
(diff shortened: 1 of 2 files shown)


app/controllers/api/physical_server_profiles_controller.rb
```diff
@@ -0,0 +1,25 @@
...
+    end
+
+    def unassign_server_resource(type, id, _data)
+      enqueue_ems_action(type, id, :method_name => :unassign_server)
+    end
+
+    private
+
+    def ensure_resource_exists(type, id)
+      raise NotFoundError, "#{type} with id:#{id} not found" unless collection_class(type).exists?(id)
```
kbrock, line 22: can we use `resource_search(type, id)` instead?
That will apply Rbac when available

kbrock:
Is this a top level entity, or does is it a sub-collection of physical servers?

zitnik:
> Is this a top level entity, or does is it a sub-collection of physical servers?

It's a top level entity. We made it as such because server profiles in Cisco Intersight exist independently of servers.

app/controllers/api/physical_server_profiles_controller.rb
gberginc: @kbrock I've used `resource_search` to handle server checking.

zitnik:
The spec tests should pass after this feature definition fix is merged into core: [https://github.com/ManageIQ/manageiq/pull/21852](https://github.com/ManageIQ/manageiq/pull/21852)

PR #1158, opened 2022-05-03
Cloud database OPTIONS spec test

nasark:
@miq-bot assign @agrare 
@miq-bot assign @kbrock 

### DIFF
--- a/spec/requests/cloud_databases_spec.rb
+++ b/spec/requests/cloud_databases_spec.rb
@@ -1,4 +1,6 @@
 RSpec.describe 'Cloud Databases API' do
+  include Spec::Support::SupportsHelper
+
   context 'cloud databases index' do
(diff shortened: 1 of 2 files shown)


spec/requests/cloud_databases_spec.rb
```diff
@@ -21,7 +23,7 @@
 
   describe 'cloud database actions' do
     let(:ems) do
-      FactoryBot.create(:ems_amazon)
+      FactoryBot.create(:ems_ibm_cloud_vpc)
```
agrare, line 26: Does the specific type of ems matter here?  You're stubbing the params so I think we should be able to leave this as simply `:ems_cloud`?  If we can it is best to use generic types over specific provider types

agrare (approved):

PR #1160, opened 2022-05-03
Condense api_action to api_request for automate_domains

kbrock:
extracted from https://github.com/ManageIQ/manageiq-api/pull/1133

puts a few checks around Automate Domain


kbrock:
WIP:
- can we pull the create/update into `supports`?

kbrock:
UN-WIP: moving the `supports` was too much for now.
So keeping the logic the same, and focusing on cleaning it

PR #1162, opened 2022-05-10
Fix cloud_volume_snapshat_create_queue

kbrock:
this has moved from the parent create volume snapshot to this object

related to https://github.com/ManageIQ/manageiq/pull/21855


agrare (approved):

agrare:
Hey @kbrock what is the order of these PRs?  It looks like this gets merged first then you can remove https://github.com/ManageIQ/manageiq/pull/21855/files#diff-032695b775947e51924fc224994c640a3b326f6ca7455ec0465397380e065b35R195-R198 ?

PR #1163, opened 2022-05-25
implemented validation storage system button

TonySchneider:
# Description
We implemented a 'validate-storage' rest API on the AutoSDE side to have the ability to validate the storage system credentials before we send the creation task to the queue. With this feature, we're saving time and can change the credentials (if there are provided some invalid parameters) without waiting for the response of the queued task.
I implemented the validate button on the attaching storage system page. So we can press the button and validate that the provided parameters are valid. If they are valid, AutoSDE saved this validation request and when we'll press the submit button, the queue task will not validate another time but just take the object that has been created by the previous validation request.

# Illustration video
validate-button

# Links
https://github.com/ManageIQ/manageiq/pull/21882
https://github.com/ManageIQ/manageiq-ui-classic/pull/8279
https://github.com/ManageIQ/manageiq-providers-autosde/pull/149


TonySchneider:
@miq-bot cross-repo-tests https://github.com/ManageIQ/manageiq-ui-classic/pull/8279

app/controllers/api/physical_storages_controller.rb
```diff
@@ -29,5 +29,11 @@ def delete_resource_action(type, id = nil, _data = nil)
         {:task_id => physical_storage.delete_physical_storage_queue(User.current_user)}
       end
     end
+
+    def validate_resource(type, _id = nil, data = {})
+      create_ems_resource(type, data, :supports => true) do |ems, klass|
```
agrare, line 34: @TonySchneider I don't think `create_ems_resource` is what you're looking for here.  This is a convenience method for the create action where you're trying to validate.

You probably want to use `api_action` and then `ensure_supports(type, klass, :validate)`

TonySchneider: Yeah, I see since I actually don't need the full_action_results method to be called, and also the action should be different... got it.
Will handle it.

TonySchneider:
@miq-bot cross-repo-tests https://github.com/ManageIQ/manageiq-ui-classic/pull/8279, https://github.com/ManageIQ/manageiq-providers-autosde/pull/150

TonySchneider:
@agrare changed the validation function according to your requirements.

agrare:
@TonySchneider can you add some spec tests for this new API?  Tests should cover invalid parameters, missing user permissions, class that doesn't support validate, and then the success case.

app/controllers/api/physical_storages_controller.rb
```diff
@@ -29,5 +29,21 @@ def delete_resource_action(type, id = nil, _data = nil)
...
+        ems = resource_search(data['ems_id'], :providers)
+
+        raise BadRequestError, "Couldn't find a provider by the provider id - '#{data['ems_id']}'" unless ems
+
+        klass = ems.class_by_ems(physical_storage.name.split(':').last)
+        ensure_supports(type, klass, :validate)
+
+        task_id = physical_storage.validate_storage_queue(User.current_userid, ems, data)
+        action_result(true, "Validating #{physical_storage.name}", :task_id => task_id)
+      end
```
kbrock, line 46: This looks too similar to `create_ems_resource` (thank you for that)

I am introducing `api_ems_resource` via #1167 for your use.

```suggestion
      api_ems_resource(type, data, "Validating", data, :supports => :validate) do |ems, klass|
        {:task_id => physical_storage.validate_storage_queue(User.current_userid, ems, data)}
      end
```

kbrock (commented):
sorry to introduce this in your way.

This looks like a great PR. thanks

app/controllers/api/physical_storages_controller.rb
TonySchneider: thanks, Keenan, great solution.
I'll test it and push.

TonySchneider: @kbrock I've changed the code to use api_ems_resource.
Please, before you merge your code to master, read my comment:
https://github.com/ManageIQ/manageiq-api/pull/1167/files#r901095007
Thanks

TonySchneider:
@kbrock I've changed the code to use api_ems_resource.
Please, before you merge your code to master, read my comment:
https://github.com/ManageIQ/manageiq-api/pull/1167/files#r901095007
Thanks

kbrock:
@TonySchneider this is looking good.
Thank you for pointing to the bug and I'm hoping this is all set for you.

app/controllers/api/physical_storages_controller.rb
```diff
@@ -29,5 +29,11 @@ def delete_resource_action(type, id = nil, _data = nil)
         {:task_id => physical_storage.delete_physical_storage_queue(User.current_user)}
       end
     end
+
+    def validate_resource(type, _id = nil, data = {})
+      api_ems_resource(type, data, "Validating", :supports => :validate) do |ems, physical_storage|
+        {:task_id => physical_storage.validate_storage_queue(User.current_userid, ems, data)}
```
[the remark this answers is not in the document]
TonySchneider: Yeah, I do pass it to 'api_ems_resource'

TonySchneider:
Thanks @kbrock

@agrare Is this PR can be merged?

agrare:
Code looks good @TonySchneider but still needs spec tests

TonySchneider:
@agrare 
pushed the spec tests for validate method. tell me if it is enough.

agrare:
This looks good, I think additional tests which use a generic EMS rather than ems_autosde and `stub_supports` to check when validate is supported and not supported would be good additions but you can do these in a follow-up

PR #1164, opened 2022-06-08
Add cloud volume clone feature

jdenver23:
Depends on:
- [x] https://github.com/ManageIQ/manageiq/pull/21923

### DIFF
--- a/spec/requests/cloud_volumes_spec.rb
+++ b/spec/requests/cloud_volumes_spec.rb
@@ -218,6 +218,7 @@ describe "Cloud Volumes API" do
   end
 
   describe 'create backup' do
+    let(:invalid_cloud_volume_url) { api_cloud_volume_url(nil, ApplicationRecord.id_in_region(999_999, ApplicationRecord.my_region_number)) }
     it 'it can create cloud volume backup through POST' do
       zone = FactoryBot.create(:zone)
       provider = FactoryBot.create(:ems_autosde, :zone => zone)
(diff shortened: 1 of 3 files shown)


agrare:
This looks good, we should add some spec tests

agrare:
For some example check out https://github.com/ManageIQ/manageiq-api/pull/1149/files#diff-a3c50450d3d70a77c2bddd76092b5ba015fba5015545be02eea66f7850ba52f1R212-R272 and https://github.com/ManageIQ/manageiq-api/pull/1120/files#diff-be7b7bd8c66aeebb9922d56d103c2c05f65c7429964cf5be395e630206644204R1619-R1625

jaywcarman (approved):
LGTM! Nice work @jdenver23

spec/requests/cloud_volumes_spec.rb
```diff
@@ -218,6 +218,7 @@
   end
 
   describe 'create backup' do
+    let(:invalid_cloud_volume_url) { api_cloud_volume_url(nil, ApplicationRecord.id_in_region(999_999, ApplicationRecord.my_region_number)) }
```
agrare, line 221: I think this should move into the clone context down here https://github.com/ManageIQ/manageiq-api/pull/1164/files#diff-a3c50450d3d70a77c2bddd76092b5ba015fba5015545be02eea66f7850ba52f1R327

agrare: Oh weird I just noticed the attach/detach specs are in the 'create backup' describe block.  Can you move the clone volume tests into their own describe block and out of the 'create backup' one?

jdenver23: Will do!

spec/requests/cloud_volumes_spec.rb
```diff
@@ -309,5 +309,47 @@
...
+        post(api_cloud_volume_url(nil, cloud_volume), :params => {:action => "clone"})
+        expected = {
+          "success" => false,
+          "message" => a_string_including("Clone for Cloud Volume id: #{cloud_volume.id} name: '': Feature not available\/supported")
+        }
+        expect(response.parsed_body).to include(expected)
+        expect(response).to have_http_status(:bad_request)
+      end
+
+      context "Cloud Volume clone action" do
```
agrare, line 343: Don't think this needs to be its own context since it isn't describing any different situation from the main describe block

agrare (approved):

PR #1167, opened 2022-06-17
extract api_ems_resource from create_ems_resource

kbrock:
Just like api_resource can be used by many actions,
it looks like there is a similar use case before resources are created.

This is split apart for #1163

### DIFF
--- a/app/controllers/api/base_controller/action.rb
+++ b/app/controllers/api/base_controller/action.rb
@@ -16,13 +16,13 @@ module Api
         result
       end
 
-      # wrapper around api_action for creating a resource
-      # alternative to api_resource when creating a resource
+      # wrapper around api_action for acting on an ems resource
+      # alternative to api_resource when working on an ems.
       #
       # NOTE: would have been nice to only pass ems_id, and name
       #       That way the external interface would be not this method's concern
       #       but passing data also allows asserting that an id is not passed
-      def create_ems_resource(type, data, supports: false)
+      def api_ems_resource(type, data, action_phrase, supports: false)
         api_action(type, nil) do
           assert_id_not_specified(data, type)
           ems_id, model_name = data['ems_id'], data['name']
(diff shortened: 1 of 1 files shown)


Fryguy (approved):

app/controllers/api/base_controller/action.rb
```diff
@@ -35,12 +35,17 @@ def create_ems_resource(type, data, supports: false)
           klass = ems.class_by_ems(base_klass)
           raise "#{type} not supported by Provider #{ems.name}" unless klass
 
-          ensure_supports(type, klass, :create) if supports
+          ensure_supports(type, klass, supports) if supports
 
-          full_action_results(yield(ems, klass)) { "Creating #{type.to_s.singularize.titleize} #{model_name} for Provider #{ems.name}" }
+          full_action_results(yield(ems, klass)) { "#{action} #{type.to_s.singularize.titleize} #{model_name} for Provider #{ems.name}" }
```
Fryguy, line 40: Bug?  Should this be action_phrase? Seems rubocop caught this too...maybe this needs a spec?

app/controllers/api/base_controller/action.rb
```diff
@@ -16,13 +16,13 @@ def api_action(type, id)
...
-      # wrapper around api_action for creating a resource
-      # alternative to api_resource when creating a resource
+      # wrapper around api_action for acting on an ems resource
+      # alternative to api_resource when working on an ems.
       #
       # NOTE: would have been nice to only pass ems_id, and name
       #       That way the external interface would be not this method's concern
       #       but passing data also allows asserting that an id is not passed
-      def create_ems_resource(type, data, supports: false)
+      def api_ems_resource(type, data, action_phrase, supports: false)
```
TonySchneider, line 25: (If you do not mind, I'll add some comments since I've checked it on my side)

As I see, you don't use the parameter action_phrase inside this function.. so when you're calling this function in 'create_ems_resource' it sends nil to argument 'data', and there is an extra argument.

So just need to change the #{action} to #{action_phrase} in api_ems_resource and remove the 'nil' parameter when we call this function in 'create_ems_resource'.

kbrock: thank you.
I had seen the `id` there before, so I added it into the api but then realized that you did not need the id (it was `nil`) so I removed it from one spot but forgot to remove from another.

Let me know if this works closer to your goals

agrare (approved):

PR #1168, opened 2022-06-27
Handle params_for_create that take parameters

kbrock:
Some providers define options that take an ems.
This tends to be for populating drop downs.

Others do not need them.

/cc @MelsHyrule


kbrock:
This fixed a problem @MelsHyrule was having trying to fetch the `OPTIONS` for Amazon::CloudDatabase

agrare (approved):

agrare:
@kbrock is this bug present in najdorf?

kbrock:
@agrare while yes this bug is in Najdorf, we only need this for converting edit cloud database pages to react

PR #1169, opened 2022-06-28
Add missing conf button for volume mappings

tzurvaza:
If we are entering `volume-mapping` from the provider **dashboard** (meaning for a specific EMS) the **configuration button is missing**:
## Before
image
image

## After
image

- Now that we have the configuration button for a specific ems we want it to be selected automaticly and disabled if we define a new `volume-mapping` for a specific ems:
image

## Links
https://github.com/ManageIQ/manageiq-ui-classic/pull/8337

### DIFF
--- a/app/controllers/api/volume_mappings_controller.rb
+++ b/app/controllers/api/volume_mappings_controller.rb
@@ -1,5 +1,6 @@
 module Api
-  class VolumeMappingsController < BaseProviderController
+  class VolumeMappingsController < BaseController
+
     def refresh_resource(type, id, _data = nil)
       enqueue_ems_action(type, id, "Refreshing", :method_name => :refresh_ems)
     end
--- a/app/controllers/api/volume_mappings_controller.rb
+++ b/app/controllers/api/volume_mappings_controller.rb
@@ -1,6 +1,5 @@
 module Api
   class VolumeMappingsController < BaseController
-
     def refresh_resource(type, id, _data = nil)
       enqueue_ems_action(type, id, "Refreshing", :method_name => :refresh_ems)
     end
--- a/config/api.yml
+++ b/config/api.yml
@@ -4764,7 +4764,7 @@
     :identifier: volume_mapping
     :options:
     - :collection
-    :verbs: *gp
+    :verbs: *gpd
     :klass: VolumeMapping
     :subcollections:
     :collection_actions:


app/controllers/api/volume_mappings_controller.rb
```diff
@@ -1,5 +1,5 @@
 module Api
-  class VolumeMappingsController < BaseProviderController
+  class VolumeMappingsController < BaseController
```
agrare, line 2: Hey @tzurvaza just curious why change back to the BaseController here?

agrare: Interesting that is surprising that it is causing issues as it only adds `OPTIONS` allowing for you to use DDF params defined in your provider cc @kbrock

tzurvaza: When I use `BaseProviderController` I got error (on the `OPTIONS` request): 
```
{
  "message": "undefined method `params_for_create' for #<Class:0x0000559a422769d8>",
  "klass": "NoMethodError"
}
```

tzurvaza: @agrare It's been a long time without an update and I want to push the changes, any update here?

agrare: Was waiting for resolution to https://github.com/ManageIQ/manageiq/pull/2197

agrare: Okay if we aren't subclassing these we don't need params_for_create so i'm good with this for now

kbrock: Sorry I did not reply.
This was added so we could get OPTIONS and ddf working.
Since it only has the one method (to add OPTIONS support), I'm confused why it would need to be removed

kbrock: followup: #1168 fixed a few issues with `params_for_create`. Seems like that PR would have fixed this issue.

agrare: I don't think it would have because the issue wasn't the parameters the issue was `::VolumeMapping` has no method `params_for_create`

kbrock: future me: we don't need params here because all children implement the same form so no DDF is needed

PR #1170, opened 2022-06-28
adding cloud_volume_type subcollection by provider

iv1111:


app/controllers/api/providers_controller.rb
```diff
@@ -33,6 +33,7 @@ class ProvidersController < BaseController
     include Subcollections::SecurityPolicyRules
     include Subcollections::Tags
     include Subcollections::Vms
+    include Subcollections::CloudVolumeTypes
```
agrare, line 36: @iv1111 can you sort this alphabetically under `CloudTenants` above?

agrare:
Looking good @iv1111 one comment and can you add subcollection spec tests similar to https://github.com/ManageIQ/manageiq-api/blob/master/spec/requests/providers_spec.rb#L1350-L1405 ?

spec/requests/providers_spec.rb
```diff
@@ -1451,6 +1451,23 @@ def gen_import_request
     end
   end
 
+  context 'cloud volume types subcollection' do
+    before do
+      @provider = FactoryBot.create(:ems_openstack)
```
agrare, line 1456: We try to not use specific provider types unless absolutely necessary.  Recommend:
```suggestion
      @provider = FactoryBot.create(:ems_storage)
```

agrare (approved):

Issue #1129, opened 2022-02-05
Sporadic test failure with authentication_spec


Fryguy:
```
$ SPEC_OPTS="--seed 32579" be rspec spec/requests/authentication_spec.rb
** Using session_store: ActionDispatch::Session::MemoryStore
** ManageIQ master, codename: Najdorf

Randomized with seed 32579
.................................F.............F............F.....

Failures:

  1) Authentication API Token Based Authentication when using a cache session store Tokens for Web Sockets gets a UI based token_ttl when requesting token for web sockets
     Failure/Error: expect(response.parsed_body["token_ttl"]).to eq(::Settings.session.timeout.to_i_with_method)

       expected: 3600
            got: nil

       (compared using ==)
     # ./spec/requests/authentication_spec.rb:371:in `block (6 levels) in <main>'
     # /Users/jfrey/.gem/ruby/2.7.4/gems/webmock-3.14.0/lib/webmock/rspec.rb:37:in `block (2 levels) in <main>'

  2) Authentication API Token Based Authentication when using a sql session store Tokens for Web Sockets gets a UI based token_ttl when requesting token for web sockets
     Failure/Error: expect(response.parsed_body["token_ttl"]).to eq(::Settings.session.timeout.to_i_with_method)

       expected: 3600
            got: nil

       (compared using ==)
     # ./spec/requests/authentication_spec.rb:371:in `block (6 levels) in <main>'
     # /Users/jfrey/.gem/ruby/2.7.4/gems/webmock-3.14.0/lib/webmock/rspec.rb:37:in `block (2 levels) in <main>'

  3) Authentication API Token Based Authentication when using a memory session store Tokens for Web Sockets gets a UI based token_ttl when requesting token for web sockets
     Failure/Error: expect(response.parsed_body["token_ttl"]).to eq(::Settings.session.timeout.to_i_with_method)

       expected: 3600
            got: nil

       (compared using ==)
     # ./spec/requests/authentication_spec.rb:371:in `block (6 levels) in <main>'
     # /Users/jfrey/.gem/ruby/2.7.4/gems/webmock-3.14.0/lib/webmock/rspec.rb:37:in `block (2 levels) in <main>'

Finished in 18.43 seconds (files took 5.24 seconds to load)
66 examples, 3 failures

Failed examples:

[...]

PR #1171, opened 2022-06-29
clear out cache for session specs

kbrock:
Tries to address #1129
Since we are changing the session_store, clear the token cache in the specs

This is copy/pasted from spec/requests/users_spec.rb see e5bc44c6e9efbc7
Thanks @jrafanie for this idea

### DIFF
--- a/spec/requests/authentication_spec.rb
+++ b/spec/requests/authentication_spec.rb
@@ -240,10 +240,18 @@ describe "Authentication API" do
   end
 
   context "Token Based Authentication" do
-    %w(sql memory cache).each do |session_store|
-      context "when using a #{session_store} session store" do
-        before { stub_settings_merge(:server => {:session_store => session_store}) }
+    before do
+      # Keep this outside of the loop below with the `let` providing the value
+      # for the stub, otherwise it will not work as expected
+      stub_settings_merge(:server => {:session_store => session_store_value})
+      TokenStore.token_caches.clear
+    end
 
+    after { TokenStore.token_caches.clear }
+
+    %w[cache sql memory].each do |session_store|
+      context "when using a #{session_store} session store" do
+        let(:session_store_value) { session_store }
         it "gets a token based identifier" do
           api_basic_authorize
 


spec/requests/authentication_spec.rb
```diff
@@ -240,10 +240,17 @@
...
+      # Keep this outside of the loop below with the `let` providing the value
+      # for the stub, otherwise it will not work as expected
+      stub_settings_merge(:server => {:session_store => session_store_value})      
+      TokenStore.token_caches.clear
+    end
 
+    after { TokenStore.token_caches.clear }
+
+    %w[sql memory cache].each do |session_store|
+      context "when using a #{session_store} session store" do
```
jrafanie, line 253: don't we still have to stub the session_store once we know the type (within this block)?

`stub_settings_merge(:server => {:session_store => session_store}) }`

spec/requests/authentication_spec.rb
```diff
@@ -240,10 +240,17 @@
   end
 
   context "Token Based Authentication" do
-    %w(sql memory cache).each do |session_store|
-      context "when using a #{session_store} session store" do
-        before { stub_settings_merge(:server => {:session_store => session_store}) }
+    before do
+      # Keep this outside of the loop below with the `let` providing the value
+      # for the stub, otherwise it will not work as expected
+      stub_settings_merge(:server => {:session_store => session_store_value})
```
jrafanie, line 246: copy / paste bug?  I don't see `session_store_value` defined here

spec/requests/authentication_spec.rb
jrafanie: I think we need to stub it each time we enter this loop

spec/requests/authentication_spec.rb
```diff
@@ -240,10 +240,17 @@
...
 
   context "Token Based Authentication" do
-    %w(sql memory cache).each do |session_store|
-      context "when using a #{session_store} session store" do
-        before { stub_settings_merge(:server => {:session_store => session_store}) }
+    before do
+      # Keep this outside of the loop below with the `let` providing the value
+      # for the stub, otherwise it will not work as expected
+      stub_settings_merge(:server => {:session_store => session_store_value})      
+      TokenStore.token_caches.clear
```
Fryguy, line 247: Can we put this in `EvmSpecHelper.clear_caches` instead?  That is called after every single spec.  Or is that not possible since that code is in core?

Fryguy:
Can you explain in a bit more detail why exactly this solves #1129?  It's hard to understand how they are related and the OP nor the commit message details the "why"

jrafanie:
I wish we could reliably recreate the failure.  I hacked it locally by forcing the token to appear invalid as I detailed in the [possible scenarios](https://github.com/ManageIQ/manageiq-api/issues/1129#issuecomment-1170389924) and as expected I hit this "sporadic failure".

Fryguy:
I see...thanks for the details.  IMO, we should move this cache clearing higher, either as piggy backing on the `EvmSpecHelper.clear_caches` from core or if it can't live in core then an `after(:suite)` call in the spec_helper (which is how `EvmSpecHelper.clear_caches` actually works).

kbrock:
There is:

```
  private_class_method def self.settings_restore
    Vmdb::Settings.reset_settings_constant(settings_load(@settings_backup))
  end
```

Do you want to tie this into there?
Feels like that method would know too much (and possibly grow really big) if it is responsible to undoing everything that could have long term effects of changed settings.
Many (most?) of our settings tend to set something that is cached.

spec/requests/authentication_spec.rb
Fryguy: Oh I was more thinking that, in general, any global caches should be cleared between specs...you never know if something else in the future will change to suddenly start using this token cache (or any other global cache).  I'm fine with this being tactical though.

Fryguy:
> Do you want to tie this into there?

No, we don't need that because all settings are cleared between every test as part of EvmSpecHelper.clear_caches: https://github.com/ManageIQ/manageiq/blob/master/spec/support/evm_spec_helper.rb#L49  called here: https://github.com/ManageIQ/manageiq/blob/master/spec/spec_helper.rb#L62-L64.  This is why I was suggesting adding `TokenStore.token_caches.clear` to `EvmSpecHelper.clear_caches`

PR #1176, opened 2022-07-26
Fix supports related API spec failures

agrare:
Fix a couple of spec failures related to recent supports feature changes

### DIFF
--- a/spec/requests/service_templates_spec.rb
+++ b/spec/requests/service_templates_spec.rb
@@ -604,33 +604,54 @@ describe "Service Templates API" do
       end
     end
 
-    context "with an unorderable template" do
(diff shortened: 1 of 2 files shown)


spec/requests/service_templates_spec.rb
```diff
@@ -604,33 +604,54 @@
       end
     end
 
-    context "with an unorderable template" do
```
agrare, line 607: NOTE I created two contexts here because the old context said it was unorderable but then expected it to be orderable but not displayed.

I created a context where it does not have the orderable trait and expect the `Service template does not belong to a service catalog` failure, then a second where it is orderable but not displayed and expect the `Service template is not configured to be displayed` error (which was the original expectation).

Fryguy (approved):

PR #1177, opened 2022-08-08
Reset session cache to fix sporadic ttl failures

kbrock:
related to https://github.com/ManageIQ/manageiq-api/issues/1129

We have been trying to track down this issue for a while.
This always works locally for me, and typically for api specs,
but this fails often for cross repo specs

Since the user token service is configured with the environment, and we
are altering the variables passed into the token service, I want to clear
out that service.

Changing the ttl values for non-auth tests, so I don't see the need to
clear across every test.


kbrock:
push:
- clearing out the cache both before tests (so tests run) and after tests (so these tests aren't screwing up other tests)

kbrock:
@miq-bot cross_repo_test include /core manageiq-api

any cross repo tests on api seems to make this fail

spec/requests/authentication_spec.rb
```diff
@@ -4,6 +4,10 @@
 describe "Authentication API" do
   ENTRYPOINT_KEYS = %w(name description version versions identity collections)
 
+  # some tests change token_ttl that affects authorization tokens.
+  before { Api::Environment.clear_caches }
```
Fryguy, line 8: These should probably be cleared on every test - the caller shouldn't have to know that `clear_caches` happens to clear the token_ttl.

That being said, I'm fine with this for now to fix these issues.

Fryguy:
Not sure why you need cross repo since these changes are localized.  Merging.

kbrock:
This does not seem to solve any problems, but it still looks related to the actual issue.
Will dig deeper

PR #1181, opened 2022-09-16
Adding vm resize option

alizapeikes:
Part of https://github.com/ManageIQ/manageiq-ui-classic/issues/7603

Depends on:
https://github.com/ManageIQ/manageiq-ui-classic/pull/8452

Connected to:
https://github.com/ManageIQ/manageiq/pull/22122
https://github.com/ManageIQ/manageiq-providers-ibm_cloud/pull/420
https://github.com/ManageIQ/manageiq-providers-openstack/pull/828

### DIFF
--- a/app/controllers/api/vms_controller.rb
+++ b/app/controllers/api/vms_controller.rb
@@ -1,5 +1,5 @@
 module Api
-  class VmsController < BaseController
+  class VmsController < BaseProviderController
     extend Api::Mixins::CentralAdmin
     include Api::Mixins::PolicySimulation
     include Api::Mixins::Genealogy
@@ -214,6 +214,16 @@ module Api
       api_resource(type, id, "Disassociating resource from", :supports => :disassociate_floating_ip) do |vm|
         {:task_id => vm.disassociate_floating_ip_queue(User.current_userid, data["floating_ip"])}
       end
+    end
+
+    def resize_resource(type, id, data)
+      api_resource(type, id, "Resizing Resource", :supports => :resize) do |vm|
+        raise BadRequestError, "Must specify new resize value/s" if data["resizeValues"].blank?
+
+        {:task_id => vm.resize_queue(User.current_userid, data)}
+      end
+    rescue => err
+      action_result(false, err.to_s)
     end
 
     private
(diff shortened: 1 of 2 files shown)


agrare:
The test failure looks legitimate, you'll have to add the new feature to https://github.com/ManageIQ/manageiq/blob/master/db/fixtures/miq_product_features.yml

jaywcarman:
> The test failure looks legitimate, you'll have to add the new feature to https://github.com/ManageIQ/manageiq/blob/master/db/fixtures/miq_product_features.yml

Looks like the vmware and ovirt providers support a `vm_reconfigure` feature. Is this (or should it be) the same as `vm_resize` or something different?

agrare:
> > The test failure looks legitimate, you'll have to add the new feature to https://github.com/ManageIQ/manageiq/blob/master/db/fixtures/miq_product_features.yml
>
> Looks like the vmware and ovirt providers support a vm_reconfigure feature. Is this (or should it be) the same as vm_resize or something different?

Reconfigure is a little different in that it is able to change cpus and memory (which is similar to changing a flavor) but it also can add/remove disks, nics, cdroms, etc...

Reconfigure is also only available on Infra type VMs and resize is more for Cloud type VMs

agrare (approved):

app/controllers/api/vms_controller.rb
```diff
@@ -216,6 +216,16 @@ def disassociate_resource(type, id, data)
       end
     end
 
+    def resize_resource(type, id, data)
+      api_resource(type, id, "Resizing Resource", :supports => :resize) do |vm|
+        raise BadRequestError, "Must specify new resize value/s" if data["resizeValues"].blank?
```
Fryguy, line 221: Not for this PR but it looks like none of these errors are translated, which we probably should do since it's a user facing message.

agrare: @Fryguy is this a general API issue?

```
$ grep -r BadRequestError app/controllers/ | wc -l
252
$ grep -r BadRequestError app/controllers/ | grep '_(' | wc -l
1
```

PR #1184, opened 2022-10-20
Apply gettext for string internationalization - vms_controller

alizapeikes:
As per @Fryguy suggestion:
#https://github.com/ManageIQ/manageiq-api/pull/1181#discussion_r996154644


app/controllers/api/vms_controller.rb
```diff
@@ -47,7 +47,7 @@ def shelve_offload_resource(type, id = nil, _data = nil)
     def edit_resource(type, id, data)
       edit_resource_with_genealogy(type, id, data)
     rescue => err
-      raise BadRequestError, "Cannot edit VM - #{err}"
+      raise BadRequestError, _("Cannot edit VM - #{err}")
```
Fryguy, line 50: This error can't be interpolated this way for i18n purposes and instead should use placeholders (see the i18n guide here: https://github.com/ManageIQ/guides/blob/master/i18n.md#string-interpolation)

Fryguy: Also below wherever there is interpolation

Fryguy (approved):
LGTM - would like a second set of eyes.  @jrafanie ?

jrafanie (approved):
LGTM

PR #1192, opened 2022-12-16
[WIP] Add zeitwerk code loader

jrafanie:


app/controllers/api/base_controller.rb
```diff
@@ -7,16 +12,16 @@ class BaseController < ActionController::API
...
+    include Parameters
+    include Parser
+    include Manager
+    include Action
+    include Logger
+    include Normalizer
+    include Renderer
+    include Results
+    include Generic
+    include Authentication
```
Fryguy, line 24: 😍  This is what we've been waiting for!

lib/manageiq-api.rb
```diff
@@ -1,2 +1,25 @@
-require "manageiq/api/engine"
-require "manageiq/api/version"
+
+require "zeitwerk"
+loader = Zeitwerk::Loader.for_gem(warn_on_extra_files: false)
+loader.ignore(__FILE__)
+
+# Add non-standard root directory to the path
+loader.push_dir("#{__dir__}/../app/controllers")
```
Fryguy, line 7: I would expect any rails engine to have this (as well as things like models, etc - do we have to do this in every plugin?

lib/manageiq-api.rb
```diff
@@ -1,2 +1,25 @@
...
+
+# This tells the loader to not expect Services namespace
+loader.collapse("#{__dir__}/services")
+
+loader.ignore("#{__dir__}/api/api_config.rb")  # Dynamically loads all Vmdb::Plugins - assumes we're mounted in manageiq
+
+# These inflectors teach the loader our naming convention here
+loader.inflector.inflect(
+  "manageiq"  => "ManageIQ",
```
Fryguy, line 16: Similarly, this is a pain to do everywhere - wonder if there's a way to inherit this from core (perhaps when core is finally converted we can come back and remove the redundant stuff)

manageiq-api.gemspec
```diff
@@ -22,6 +22,10 @@ Gem::Specification.new do |spec|
   # 2.9.0 regressed serialization
   # See: https://github.com/rails/jbuilder/issues/461
   spec.add_dependency "jbuilder", "~> 2.5", "!= 2.9.0"
+  spec.add_dependency "multi_json"
+  spec.add_dependency "rails"
+  spec.add_dependency "responders"
+  spec.add_dependency "zeitwerk"
```
Fryguy, line 28: Also here - not loving all the extra requirements (I realize this is WIP, so not really complaining yet)

lib/manageiq-api.rb
jrafanie: I'll look.  Thanks.

lib/manageiq-api.rb
jrafanie: yeah, that's why I wanted to try out a few repos to see what's common or different and to see if there's a better way to do it.

manageiq-api.gemspec
jrafanie: yeah, some were missing dependencies that are only obvious when I tested zeitwerk by enabling eager loading for nearly everything that was fairly easy to autoload.

I can look at extracting the missing dependencies and get them merged on master first.

jrafanie:
I'm closing this one.  Rails engines shouldn't be loading with zeitwerk independently because they get plugged into applications which are responsible for loading code.

If we want to eager load or do this in standalone gems, we can do some of these changes as we get closer to using zeitwerk.

PR #1201, opened 2023-02-13
Adrian/740029 add delete and update to host initiator groups

Lauferin:
see issue https://github.com/ManageIQ/manageiq-ui-classic/issues/8640


agrare:
@Lauferin the spec failures appear to be related to your changes

Lauferin:
> @Lauferin the spec failures appear to be related to your changes

I've fixed one of the problems. The other two don't seem to be related to my changes, and they're not appearing in this page anymore (although I did get an email about them).

agrare:
The cloud volume spec failure should be fixed by https://github.com/ManageIQ/manageiq-api/pull/1208 I'll kick this after that gets merged

agrare:
@Lauferin okay we're back to just a HostInitiator spec failing

Lauferin:
> @Lauferin okay we're back to just a HostInitiator spec failing

I changed something there but I don't know if it's working now because I'm again receiving failures from cloud_volume and something else.

PR #1202, opened 2023-02-13
Host verify_credentials

kbrock:
- parent PR: https://github.com/ManageIQ/manageiq-ui-classic/pull/8608
- depends upon: https://github.com/ManageIQ/manageiq/pull/22336

Exposes `Host#verify_credentials` for the api

This allows the user to verify credentials for a host.

/cc @MelsHyrule

### DIFF
--- a/app/controllers/api/hosts_controller.rb
+++ b/app/controllers/api/hosts_controller.rb
@@ -1,6 +1,7 @@
 module Api
   class HostsController < BaseController
     CREDENTIALS_ATTR = "credentials".freeze
+    AUTH_ATTR = "authentications".freeze
     AUTH_TYPE_ATTR = "auth_type".freeze
     DEFAULT_AUTH_TYPE = "default".freeze
 
@@ -11,21 +12,37 @@ module Api
     include Subcollections::Tags
 
     def edit_resource(type, id, data = {})
+      # TODO: drop 'credentials' parameter field when ui-classic hosts is in react
       credentials = data.delete(CREDENTIALS_ATTR)
+      authentications = data.delete(AUTH_ATTR)
       raise BadRequestError, "Cannot update non-credentials attributes of host resource" if data.any?
       resource_search(id, type).tap do |host|
+        # begin legacy ui-classic
         all_credentials = Array.wrap(credentials).each_with_object({}) do |creds, hash|
           auth_type = creds.delete(AUTH_TYPE_ATTR) || DEFAULT_AUTH_TYPE
           creds.symbolize_keys!
           creds.reverse_merge!(:userid => host.authentication_userid(auth_type))
           hash[auth_type.to_sym] = creds
         end
+        # end legacy ui-classic. if they provided the newer authentications, it will overwrite
+        all_credentials, _ = symbolize_password_keys!(authentications) if authentications
         host.update_authentication(all_credentials) if all_credentials.present?
       end
     end
 
     def check_compliance_resource(type, id, _data = nil)
       enqueue_ems_action(type, id, "Check Compliance for", :method_name => "check_compliance", :supports => true)
+    end
+
+    private
+
+    # takes credentials from params and converts into something for update_authentications
+    def symbolize_password_keys!(authentications)
+      auth_type = authentications.keys.first
+      # symbolize userid, password
+      authentications[auth_type].symbolize_keys!
+
+      return authentications, auth_type
     end
   end
 end
--- a/app/controllers/api/hosts_controller.rb
+++ b/app/controllers/api/hosts_controller.rb
@@ -30,6 +30,14 @@ module Api
       end
     end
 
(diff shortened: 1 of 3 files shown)


kbrock:
update:
- fixed spacing for hashes in 2 spec files

MelsHyrule:
In app/controllers/api/hosts_controller.rb `class HostsController < BaseController` to `BaseProviderController`.

spec/requests/providers_spec.rb
```diff
@@ -464,6 +464,22 @@ def have_endpoint_attributes(expected_hash)
     end
   end
 
+  describe "Providers validate" do
```
Fryguy, line 467: The action is `verify_credentials`, so should this say:?
```suggestion
  describe "Providers verify_credentials" do
```

kbrock:
update:
- rebased (fixes build)

app/controllers/api/hosts_controller.rb
```diff
@@ -11,21 +12,43 @@ class HostsController < BaseController
     include Subcollections::Tags
 
     def edit_resource(type, id, data = {})
+      # TODO: drop 'credentials' parameter field when ui-classic hosts is in react
       credentials = data.delete(CREDENTIALS_ATTR)
+      authentications = data.delete(AUTH_ATTR)
       raise BadRequestError, "Cannot update non-credentials attributes of host resource" if data.any?
       resource_search(id, type).tap do |host|
-        all_credentials = Array.wrap(credentials).each_with_object({}) do |creds, hash|
+        all_credentials = Array.wrap(credentials || []).each_with_object({}) do |creds, hash|
```
agrare, line 20: is the `|| []` necessary?
```
>> Array.wrap(nil)
=> []
```

app/controllers/api/hosts_controller.rb
```diff
@@ -11,21 +12,43 @@ class HostsController < BaseController
...
           auth_type = creds.delete(AUTH_TYPE_ATTR) || DEFAULT_AUTH_TYPE
           creds.symbolize_keys!
           creds.reverse_merge!(:userid => host.authentication_userid(auth_type))
           hash[auth_type.to_sym] = creds
         end
+        all_credentials, _ = authentications_params_to_authentications(authentications) if authentications
```
agrare, line 26: Is this overwriting `all_credentials` from up on line 20?  Hard to follow how authentications / credentials are used.

app/controllers/api/hosts_controller.rb
```diff
@@ -11,21 +12,43 @@ class HostsController < BaseController
...
       enqueue_ems_action(type, id, "Check Compliance for", :method_name => "check_compliance", :supports => true)
     end
+
+    private
+
+    # takes credentials from params and converts into something for update_authentications
+    def authentications_params_to_authentications(authentications)
+      auth_type = authentications.keys.first
+      # symbolize userid, password
+      authentications[auth_type].symbolize_keys!
```
agrare, line 49: Should we return a new hash rather than modify the original?  If you have to modify the parameter should this be a `!` method?

app/controllers/api/hosts_controller.rb
kbrock: thanks. I added comments.

The old way passed credentials.
the new way passes authentications

Yes, if they follow the new way, then we are ignoring the old way.
In theory, they will not pass the old way and the new way

kbrock: once we swap out ui classic, we can drop the old way

kbrock:
update:
- comment around credentials vs authentications
- rename symbolize_password_keys! (twice. sorry for double push)
- fixed `Array.wrap( || [])`

app/controllers/api/providers_controller.rb
```diff
@@ -102,15 +102,17 @@ def change_password_resource(type, id, data = {})
...
-      task_id = klass.verify_credentials_task(current_user.userid, zone_name, data)
-      action_result(true, 'Credentials sent for verification', :task_id => task_id)
-    rescue => err
-      action_result(false, err.to_s)
+    # NOTE: the ui's zone drop down is currently passing the zone_id
+    def verify_credentials_resource(type, id = nil, data = {})
+      api_action(type, id) do |klass|
+        ems_klass = fetch_provider_klass(klass, data)
+        zone_name = data.delete('zone_name')
+        zone_id = data.delete('zone_id')
```
agrare, line 110: Funny I was just looking into this when I saw my `zone` was blank for verifying credentials.  This seems like its own bug that we should fix separately so we can backport what do you think?

kbrock:
update:

- extracted provider authentication changes
- fixed issue with a pr refactor

app/controllers/api/providers_controller.rb
kbrock: see #1211

agrare (approved):

PR #1205, opened 2023-02-22
Implement cloud snapshot collector, list page, create & delete methods

TonySchneider:
We have developed an integration with a new resource type - "cloud volume snapshot" on the Autosde side. Now, we can provide a list of snapshots from the attached physical storage, create a new one based on a particular volume object and delete it - by using the Autosde rest endpoint (The Autosde GEM is already updated with these new features).
Following these new features, I have added to our (MIQ Autosde provider) collector and refresher to pull snapshots from Autosde and show them in the right place - Cloud Volume Snapshots. I also added the counter component to the provider dashboard.

List page:
image

Snapshot summary page:
image

Dashboard:
image

Create snapshot from cloud volume:
1.
image
2.
image

Delete button exists as well at the snapshots page
# related PRs:
- https://github.com/ManageIQ/manageiq-ui-classic/pull/8647
- https://github.com/ManageIQ/manageiq-providers-autosde/pull/210
- https://github.com/ManageIQ/manageiq/pull/22359

### DIFF
--- a/config/api.yml
+++ b/config/api.yml
@@ -758,10 +758,26 @@
         :identifier: cloud_tenant_tag
   :cloud_volume_snapshots:
     :description: Cloud Volume Snapshots
+    :identifier: cloud_volume_snapshots
     :options:
-    - :subcollection
-    :verbs: *gpppd
+    - :collection
+    :verbs: *gpd
     :klass: CloudVolumeSnapshot
+    :subcollections:
+    :collection_actions:
+      :get:
+      - :name: read
+        :identifier: cloud_volume_snapshot_view
+      :post:
+      - :name: delete
+        :identifier: cloud_volume_snapshot_delete
+    :resource_actions:
+      :get:
+      - :name: read
+        :identifier: cloud_volume_snapshot_view
+      :post:
+      - :name: delete
+        :identifier: cloud_volume_snapshot_delete
   :cloud_volume_types:
     :description: Cloud Volume Types
     :identifier: cloud_volume_type
--- a/config/api.yml
+++ b/config/api.yml
@@ -758,10 +758,10 @@
         :identifier: cloud_tenant_tag
   :cloud_volume_snapshots:
     :description: Cloud Volume Snapshots
-    :identifier: cloud_volume_snapshots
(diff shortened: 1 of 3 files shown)


TonySchneider:
@miq-bot cross-repo-tests https://github.com/ManageIQ/manageiq/pull/22359, https://github.com/ManageIQ/manageiq-providers-autosde/pull/210, https://github.com/ManageIQ/manageiq-ui-classic/pull/8647

agrare:
@TonySchneider hey so maybe I'm missing it, but how is the snapshot being created?  Is it over the API through the cloud_volume#snapshot_create action or via cloud_volume_snapshot#create?  Or is it not over the API at all and being done in the UI controller?

TonySchneider:
Hi @agrare , I have used the existing API - cloud_volume#snapshot_create since we need an existing volume for this method so when you want to create a snapshot for a specific volume, you should do this through the cloud volume page

agrare:
@TonySchneider please resolve the api.yaml failing spec https://github.com/ManageIQ/manageiq-api/actions/runs/4428242306/jobs/7767029454#step:7:104

TonySchneider:
@agrare fixed

config/api.yml
```diff
@@ -758,10 +758,26 @@
         :identifier: cloud_tenant_tag
   :cloud_volume_snapshots:
     :description: Cloud Volume Snapshots
+    :identifier: cloud_volume_snapshot
     :options:
-    - :subcollection
-    :verbs: *gpppd
```
agrare, line 763: Openstack supports updating cloud_volume_snapshots, why delete PUT/PATCH ?

TonySchneider: Currently, Autosde doesn't support updating snapshots, but I think I can leave it as it was, I'll check it

agrare:
@TonySchneider can you add tests covering these new APIs to the `spec/requests/cloud_volume_snapshots_spec.rb` spec?

TonySchneider:
aligned the branch with the master, will add the tests to `spec/requests/cloud_volume_snapshots_spec.rb`

PR #1207, opened 2023-02-24
Fix issues after core changes for partial match passwords

Fryguy:
@jrafanie Please review. Follow up to https://github.com/ManageIQ/manageiq/pull/22372

### DIFF
--- a/lib/api.rb
+++ b/lib/api.rb
@@ -11,7 +11,8 @@ module Api
   UnsupportedMediaTypeError = Class.new(ApiError)
 
   def self.encrypted_attribute?(attr)
-    Environment.encrypted_attributes.include?(attr.to_s) || attr.to_s.include?('password')
+    !Environment.encrypted_attributes_whitelist.include?(attr.to_s) &&
+      Environment.encrypted_attributes.any? { |a| attr.to_s.include?(a.to_s) }
   end
 
   def self.time_attribute?(attr)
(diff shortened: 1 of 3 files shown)


lib/api.rb
```diff
@@ -11,7 +11,7 @@ module Api
   UnsupportedMediaTypeError = Class.new(ApiError)
 
   def self.encrypted_attribute?(attr)
-    Environment.encrypted_attributes.include?(attr.to_s) || attr.to_s.include?('password')
+    attr != "verify_ssl" && Environment.encrypted_attributes.any? { |a| attr.to_s.include?(a.to_s) }
```
Fryguy, line 14: verify being a "weird" case is annoying.  I think I might change the UI to give a different name like password_verify, so we can get rid of verify from the filter list.

kbrock (approved):
Fix looks good
The indentation rubocops look like a good suggestion
Think the negation suggestion is suspect thought

jrafanie (approved):
LGTM, wish we didn't need these lists but it makes sense here since they're relevant to the API.

jrafanie:
LOL, @kbrock we tried to merge at the exact same time.... jinx!

Fryguy:
Backported to `petrosian` in commit 12dd450d657ccfdccd92c2750b15e6ae8bd549bb.

```text
commit 12dd450d657ccfdccd92c2750b15e6ae8bd549bb
Date:   Mon Feb 27 10:08:22 2023 -0500

    Merge pull request #1207 from Fryguy/fix_filter_params
    
    Fix issues after core changes for partial match passwords
    
    (cherry picked from commit a7790eb6d2c1718b5f019e8b14acddc4c046f962)
```

PR #1208, opened 2023-02-28
Fix cloud_volume clone supports spec

agrare:
The spec was using stub_supports_not on the class, not the specific cloud_volume which meant when autosde added clone support the spec began to fail.

We should be using a generic factory not a provider specific factory, and we should stub_supports_not on the object not just the class.

### DIFF
--- a/spec/requests/cloud_volumes_spec.rb
+++ b/spec/requests/cloud_volumes_spec.rb
@@ -326,8 +326,8 @@ describe "Cloud Volumes API" do
       end
 
       it 'clone raises an error if the cloud volume does not support clone' do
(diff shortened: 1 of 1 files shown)


agrare:
cc @kbrock

Fryguy:
Backported to `petrosian` in commit a4c08e329a805973eae43a04dfb20d87916174f2.

```text
commit a4c08e329a805973eae43a04dfb20d87916174f2
Date:   Tue Feb 28 10:41:54 2023 -0500

    Merge pull request #1208 from agrare/fix_cloud_volume_clone_spec
    
    Fix cloud_volume clone supports spec
    
    (cherry picked from commit f60a1629efff5ddbbfe2ae24cd0b22b6dac713e0)
```

PR #1210, opened 2023-03-08
use stub_settings_merge in specs

kbrock:
best to not alter settings between tests but rather stub settings.
These can lead to sporadic test failures.


kbrock:
update:
- deleted extra paren

kbrock:
@miq-bot cross-repo-tests manageiq-api

Yes, redundant, but it tends to fail in the cross repo test even when working locally

kbrock:
sigh. this does not fix the cross repo issue.

Something about the way settings runs in cross repo makes these tests return nothing for the `ttl` settings.

PR #1212, opened 2023-03-13
Fix specs requiring memcached

Fryguy:
When memcached was not available, the specs would fail, but not in the way that was expected. Before this commit, fetching a value would just return nil, and while the spec would fail, it was misleading. Now, the spec will fail early if memcached is not available and it is clear why the spec fails.

Additionally, it was found that the closures in users_spec were invalid, so while the spec was written to test sql, memory, and cache, due to the closures all 3 sets of tests were actually running cache. This was more evident when the memcached checks were added because both sql and memory were both failing in the abscence of memcached. This commit fixes those closures so they run accurately.

@jrafanie or @kbrock Please review.  This is much easier to review with whitespace turned off: https://github.com/ManageIQ/manageiq-api/pull/1212/files?w=1

Also note that this is why cross-repo tests continually fail on api - it's because we don't have memcached running there.  Separate PR incoming for that: https://github.com/ManageIQ/manageiq-cross_repo/pull/98

### DIFF
--- a/spec/requests/authentication_spec.rb
+++ b/spec/requests/authentication_spec.rb
@@ -243,10 +243,14 @@ describe "Authentication API" do
     end
   end
 
-  context "Token Based Authentication" do
+  describe "Token Based Authentication" do
     %w(sql memory cache).each do |session_store|
       context "when using a #{session_store} session store" do
-        before { stub_settings_merge(:server => {:session_store => session_store}) }
+        before do
+          MiqMemcached.client({}).get("test") if session_store == "cache"
+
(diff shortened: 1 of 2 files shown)


kbrock (approved):
YAY. great find.

:shipit:

agrare:
:heart: can confirm this fixes cross-repo testing with the API https://github.com/ManageIQ/manageiq-cross_repo-tests/actions/runs/4409589128/jobs/7726064878

kbrock:
followup:
this only runs tests against mem-cached
in the future, will need to stub `ManageIQ::Session.store`

PR #1214, opened 2023-03-13
Add workflows/workflow_instances controllers

agrare:
Depends on:
- [x] https://github.com/ManageIQ/manageiq/pull/22316
- [x] https://github.com/ManageIQ/manageiq/pull/22429

### DIFF
--- a/spec/requests/workflow_instances_spec.rb
+++ b/spec/requests/workflow_instances_spec.rb
@@ -1,0 +1,66 @@
+describe "WorkflowInstances API" do
+  context "GET /api/workflow_instances" do
+    context "without an appropriate role" do
+      before { api_basic_authorize }
+
+      it "returns forbidden" do
+        FactoryBot.create(:workflow_instance)
+
+        get(api_workflow_instances_url)
+
+        expect(response).to have_http_status(:forbidden)
+      end
+    end
+
+    context "with appropriate role" do
+      before { api_basic_authorize(collection_action_identifier(:workflow_instances, :read, :get)) }
+
+      it "returns all Workflow Instances" do
+        workflow_instance = FactoryBot.create(:workflow_instance)
+
+        get(api_workflow_instances_url)
+
+        expected = {
+          "name"      => "workflow_instances",
+          "resources" => [{"href" => api_workflow_instance_url(nil, workflow_instance)}]
+        }
+        expect(response).to have_http_status(:ok)
+        expect(response.parsed_body).to include(expected)
+      end
+    end
+  end
+
+  context "GET /api/workflow_instances/:id" do
+    context "without an appropriate role" do
+      before { api_basic_authorize }
+
+      it "returns forbidden" do
+        workflow_instance = FactoryBot.create(:workflow_instance)
+
+        get(api_workflow_instance_url(nil, workflow_instance))
+
+        expect(response).to have_http_status(:forbidden)
+      end
+    end
+
+    context "with appropriate role" do
+      before { api_basic_authorize(action_identifier(:workflow_instances, :read, :resource_actions, :get)) }
+
+      it "returns a single Workflow" do
+        workflow_instance = FactoryBot.create(:workflow_instance)
+
+        get(api_workflow_instance_url(nil, workflow_instance))
+
+        expected = {
+          "id"               => workflow_instance.id.to_s,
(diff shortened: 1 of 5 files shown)


agrare:
I dropped the product features out of the core pr so this is going to fail until we get those in

agrare:
@miq-bot cross-repo-tests https://github.com/ManageIQ/manageiq/pull/22429

spec/requests/workflow_instances_spec.rb
```diff
@@ -0,0 +1,66 @@
...
+
+      it "returns a single Workflow" do
+        workflow_instance = FactoryBot.create(:workflow_instance)
+
+        get(api_workflow_instance_url(nil, workflow_instance))
+
+        expected = {
+          "id"               => workflow_instance.id.to_s,
+          "context"          => workflow_instance.context,
+          "credentials"      => workflow_instance.credentials,
```
Fryguy, line 57: Maybe for a follow up pr, but might be nice to show that this does _not_ expose the password

Fryguy (approved):

spec/requests/workflow_instances_spec.rb
agrare: Possibly, I think we want to keep actual secrets out of this payload entirely though so this would only have the JSON path reference. It would be helpful to see what credential value it is using is.
Maybe if there is no JSON path we blank out the value in case they put a value directly in there or we prevent non-references from even being added

jeffibm:
Hey, This is how I get the API response - 

http://localhost:3000/api/workflow_instances
```
{
    "name": "workflow_instances",
    "count": 0,
    "subcount": 0,
    "pages": 0,
    "resources": [],
    "links": {
        "self": "http://localhost:3000/api/workflow_instances?offset=0",
        "first": "http://localhost:3000/api/workflow_instances?offset=0",
        "last": "http://localhost:3000/api/workflow_instances?offset=0"
    }
}
```


http://localhost:3000/api/workflows
```
{
    "name": "workflows",
    "count": 0,
    "subcount": 0,
    "pages": 0,
    "resources": [],
    "links": {
        "self": "http://localhost:3000/api/workflows?offset=0",
        "first": "http://localhost:3000/api/workflows?offset=0",
        "last": "http://localhost:3000/api/workflows?offset=0"
    }
}
```

agrare:
Yes that looks correct for an empty collection.
If you want to create an example workflow you can use `>> Workflow.create_from_json!(File.read("../floe/examples/workflow.asl"))` (assumes you have the https://github.com/ManageIQ/floe` gem locally)

Fryguy:
@jeffibm Once you get a workflow in your DB, you can then expand the resources and pull back whatever fields you need, like name.

jeffibm:
> Yes that looks correct for an empty collection.
> If you want to create an example workflow you can use `>> Workflow.create_from_json!(File.read("../floe/examples/workflow.asl"))` (assumes you have the https://github.com/ManageIQ/floe gem locally)

Hey @agrare , this is returning nil for me
```
> ManageIQ::Providers::Workflows::AutomationManager.first
  ManageIQ::Providers::Workflows::AutomationManager Load (0.5ms)  SELECT "ext_management_systems".* FROM "ext_management_systems" WHERE "ext_management_systems"."type" = $1 ORDER BY "ext_management_systems"."id" ASC LIMIT $2  [["type", "ManageIQ::Providers::Workflows::AutomationManager"], ["LIMIT", 1]]
  ManageIQ::Providers::Workflows::AutomationManager Inst Including Associations (0.0ms - 0rows)
 => nil 

```

jeffibm:
so I tried something like this and got
```
json = File.read("../floe/examples/workflow.asl")
json = JSON.parse(json) if json.kind_of?(String)
name = json["Comment"]
workflows_automation_manager = ManageIQ::Providers::AutomationManager.second
workflows_automation_manager.workflows.create!(:name => name, :workflow_content => json)
```
image

jeffibm:
> @jeffibm it looks like you didn't run `db:seed` to create the Workflows AutomationManager.

@agrare, I am still getting the same error..
did `bin/setup`
did `bundle exec rails db:seed`

Fryguy:
Note for future readers - we ended up removing the workflows and workflows_instances specific endpoints in favor of reusing the existing configuration_scripts endpoint

PR #1215, opened 2023-03-15
change `StorageServicesController` base class to `BaseProviderController`

OrGur1987:
changed `StorageServicesController` base class to `BaseProviderController` so we could generate params for create from the model.

part of issue:
- [ ] https://github.com/ManageIQ/manageiq-ui-classic/issues/8701


OrGur1987:
@bdunne 
hi, are the errors related to this change?

kbrock:
kicking

PR #1216, opened 2023-03-23
storage service edit

OrGur1987:
belongs to issue: https://github.com/ManageIQ/manageiq-ui-classic/issues/8728

depends upon:
- [x] manageiq/manageiq#22425

adds editing/update option to storage service controller and app yaml

### DIFF
--- a/config/api.yml
+++ b/config/api.yml
@@ -4237,7 +4237,7 @@
     :identifier: storage_service
     :options:
     - :collection
-    :verbs: *gpd
+    :verbs: *gpppd
     :klass: StorageService
     :subcollections:
     :collection_actions:
@@ -4247,6 +4247,8 @@
       :post:
       - :name: create
         :identifier: storage_service_new
+      - :name: edit
+        :identifier: storage_service_edit
       - :name: delete
         :identifier: storage_service_delete
       - :name: refresh
--- a/spec/requests/storage_services_spec.rb
+++ b/spec/requests/storage_services_spec.rb
@@ -1,5 +1,8 @@
 describe "Storage Services API" do
   include Spec::Support::SupportsHelper
+
+  let(:provider) { FactoryBot.create(:ems_autosde) }
+
   context "GET /api/storage_services" do
     it "returns all storage_services" do
       storage_service = FactoryBot.create(:storage_service)
@@ -35,7 +38,6 @@ describe "Storage Services API" do
   context "POST /api/storage_services" do
     it "creates new Storage Service" do
       api_basic_authorize(collection_action_identifier(:storage_services, :create))
-      provider = FactoryBot.create(:ems_autosde)
       request = {
         "action"   => "create",
         "resource" => {
@@ -51,8 +53,7 @@ describe "Storage Services API" do
   end
 
   it "deletes a single Storage Service" do
-    provider = FactoryBot.create(:ems_autosde, :name => 'Autosde')
-    service = FactoryBot.create("ManageIQ::Providers::Autosde::StorageManager::StorageService", :name => 'test_service', :ext_management_system => provider)
+    service = FactoryBot.create(:storage_service, :name => 'test_service', :ext_management_system => provider)
     api_basic_authorize('storage_service_delete')
 
     stub_supports(StorageService, :delete)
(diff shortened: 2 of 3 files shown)


OrGur1987:
@bdunne hi, this is ready for review
thanks

config/api.yml
```diff
@@ -4231,6 +4231,8 @@
       :post:
       - :name: create
         :identifier: storage_service_new
+      - :name: edit
+        :identifier: physical_storage_edit
```
kbrock, line 4235: did you want `storage_service_edit` here?

kbrock (changes requested):
this looks great. just that one change

kbrock:
Also, is there a way to add a spec here?

OrGur1987:
> Also, is there a way to add a spec here?

should I use `spec/models/cloud_volume_spec.rb` as an example?
the specs there test queuing methods

OrGur1987:
> > Also, is there a way to add a spec here?
> 
> should I use `spec/models/cloud_volume_spec.rb` as an example? the specs there test queuing methods

@kbrock

kbrock:
@OrGur1987 good point. you can't test that the records have been modified because this is just a request for the provider.

Yea, those tests are as good as any.

OrGur1987:
> @OrGur1987 good point. you can't test that the records have been modified because this is just a request for the provider.
> 
> Yes, base your tests upon those.

added an edit spec to `spec/requests/storage_services_spec.rb`.

spec/requests/storage_services_spec.rb
```diff
@@ -79,4 +79,19 @@
 
     expect(response).to have_http_status(:ok)
   end
+
+  context 'Storage Services edit action' do
+    it "PUT /api/storage_services/:id'" do
+      require 'byebug'
+      byebug
```
kbrock, line 86: oops. how did this get in there? ;)

spec/requests/storage_services_spec.rb
```diff
@@ -79,4 +79,19 @@
...
+      provider = FactoryBot.create(:ems_autosde, :name => 'Autosde')
+      storage_service = FactoryBot.create("ManageIQ::Providers::Autosde::StorageManager::StorageService", :name => 'test_service1', :ext_management_system => provider)
```
kbrock, line 88:
```suggestion
      storage_service = FactoryBot.create(:autosde_storage_manager, :ext_management_system => provider)
```

kbrock:
@OrGur1987 This looks great. Just one or two tweaks and we can get it in right away.

Thanks for doing this

spec/requests/storage_services_spec.rb
OrGur1987: extracted `provider` to `let()`.
extracting `storage_service` didn't quite work out with the `delete` and `get` specs.. 

in the `edit` spec, for some reason when I use `FactoryBot.create(:storage_service, ... )` the service does not support `update`, but it does when I use `FactoryBot.create("ManageIQ::Providers::Autosde::StorageManager::StorageService",...)`.

do u have any idea why? maybe it references the core repo `storage_service` model and not the provider's one?
it supposedly doesn't support `create` and `delete` as well, but those spec do pass..

```
    83:   context 'Storage Services edit action' do
    84:     it "PUT /api/storage_services/:id'" do
    85:       storage_service1 = FactoryBot.create("ManageIQ::Providers::Autosde::StorageManager::StorageService", :name => 'test_service1', :ext_management_system => provider)
    86:       storage_service2 = FactoryBot.create(:storage_service, :name => 'test_service1', :ext_management_system => provider)
    87: 
    88:       require 'byebug'
    89:       byebug
    90: 
=>  91:       api_basic_authorize('storage_service_edit')
    92:       put(api_storage_service_url(nil, storage_service))
    93: 
    94: 
    95:       expect(response.parsed_body["message"]).to include("Updating")
    96:       expect(response).to have_http_status(:ok)
    97:     end
    98:   end
    99: 
   100: end
(byebug) storage_service1.supports_update?
true
(byebug) storage_service2.supports_update?
false
(byebug) storage_service2.supports_create?
false
(byebug) storage_service2.supports_delete?
false
(byebug) storage_service1.supports_delete?
true

```

OrGur1987:
> @OrGur1987 This looks great. Just one or two tweaks and we can get it in right away.
> 
> Thanks for doing this

@kbrock is spec ok like that?

OrGur1987:
@kbrock  
kind reminder :)

spec/requests/storage_services_spec.rb
kbrock: sorry.

We tend to use `stub_supports` to ensure that you get the expected `supports?`.
In the far far future, we would like to change all these tests to use a generic providers with a bunch of expectations on them.

kbrock:
@OrGur1987 thanks for the ping. I'll leave this tab open and hope it does not slip through the cracks again.

Could you see my notes around the class of the `storage_service2`?
Also, don't set any attributes (i.e.: name) in the provider if it will work without it set.

OrGur1987:
> @OrGur1987 thanks for the ping. I'll leave this tab open and hope it does not slip through the cracks again.
> 
> Could you see my notes around the class of the `storage_service2`? Also, don't set any attributes (i.e.: name) in the provider if it will work without it set.
> 
> Not sure why the factory is a sticking point for me

@kbrock 
`stub_supports` fixed it thnx :)
removed the name from the provider

PR #1217, opened 2023-03-23
Add test for authentications

kbrock:
This is already well tested and is mostly for an example of how to use this for the front end

(we are going away from passing the "credentials" hash and passing in "authentications" now


kbrock:
update:
- fixed align keys problem. think bot did not try again.

PR #1218, opened 2023-04-10
storage service edit mode: check which attached resources are compliant to selected capabilities

OrGur1987:
issue: https://github.com/ManageIQ/manageiq-ui-classic/issues/8749

added button in service edit mode which enables user to check what currently attached resources will comply with the selected capabilities.

this PR adds the action to the api controller and to api.yml

- [x] note that it is based on the yet to be merged #1216, and should be rebased on master after that is merged.

### DIFF
--- a/app/controllers/api/storage_services_controller.rb
+++ b/app/controllers/api/storage_services_controller.rb
@@ -21,5 +21,12 @@ module Api
         {:task_id => storage_service.update_storage_service_queue(User.current_userid, data)}
       end
     end
+
+    def check_compliant_resources_resource(type, id = nil, data = {})
+      data["_id"] = id
+      api_ems_resource(type, data, "Checking Compliant Resources", :supports => :check_compliant_resources) do |ems, storage_service|
+        {:task_id => storage_service.check_compliant_resources_queue(User.current_userid, ems, data)}
+      end
+    end
   end
 end
--- a/config/api.yml
+++ b/config/api.yml
@@ -4255,6 +4255,7 @@
     :identifier: storage_service
     :options:
     - :collection
+    - :custom_actions
     :verbs: *gpppd
     :klass: StorageService
     :subcollections:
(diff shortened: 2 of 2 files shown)


OrGur1987:
@miq-bot cross-repo-tests ManageIQ/manageiq#22450

OrGur1987:
cross repo passed:
https://github.com/ManageIQ/manageiq-cross_repo-tests/pull/774

OrGur1987:
@miq-bot cross-repo-tests https://github.com/ManageIQ/manageiq/pull/22450

OrGur1987:
@bdunne @kbrock 
ready for review, thanks

OrGur1987:
@miq-bot cross-repo-tests ManageIQ/manageiq#22450

OrGur1987:
hi @agrare, 
@jeffibm merged the ui for this issue

app/controllers/api/storage_services_controller.rb
```diff
@@ -21,5 +21,12 @@ def edit_resource(type, id, data = {})
         {:task_id => storage_service.update_storage_service_queue(User.current_userid, data)}
       end
     end
+
+    def check_compliant_resources_resource(type, _id = nil, data = {})
+      data["_id"] = _id
```
kbrock, line 26: definitely want to use `id` and not `_id`

Also, do we want `data["id"]` or is that underscore correct?

OrGur1987: no problem changing the method param name to `id`, but changing the `data` hash key to `id` raises the following error in `api_ems_resource`: `Resource id or href should not be specified for creating a new storage_services`.

perhaps I shouldn't have used `api_ems_resource` if it's intended for creating new records only?
I used it because it appeared in the existing `validate` action that I based upon:
https://github.com/ManageIQ/manageiq-api/blob/e1980865bb2895ba3e72a9567ea2b3699583b9ba/app/controllers/api/physical_storages_controller.rb#L26
 
is there an alternative?

if `data["_id"]` can stay that way, I prefer using `api_ems_resource` since things are working right now..

thanks

PR #1219, opened 2023-04-20
Use zeitwerk/bundler friendly namespaced gem 'require'

jrafanie:
We already have lib/manageiq/api.rb which will be required by bundler.  There is no need for lib/manageiq-api.rb, especially because it doesn't follow rails/zeitwerk conventions for namespaces.

See explanation from https://github.com/ManageIQ/manageiq-consumption/pull/205


kbrock:
We do not `require "manageiq-api"` anywhere in the other repos. :shipit:

PR #1220, opened 2023-04-24
Use standard create/update/delete features for Authentications

agrare:
Move the `/api/authentications` endpoint towards using standard create/update/delete supports features rather than checking for specific methods.

The goal is to use the standard `create_ems_resource` and `ems_resource` actions here

Depends on:
- [x] https://github.com/ManageIQ/manageiq/pull/22894
- [x] https://github.com/ManageIQ/manageiq-providers-awx/pull/24

### DIFF
--- a/app/controllers/api/authentications_controller.rb
+++ b/app/controllers/api/authentications_controller.rb
@@ -1,22 +1,25 @@
 module Api
   class AuthenticationsController < BaseController
     def edit_resource(type, id, data)
-      api_resource(type, id, "Updating") do |auth|
-        ensure_respond_to(type, auth, :update, :update_in_provider_queue)
+      api_resource(type, id, "Updating", :supports => :update) do |auth|
         {:task_id => auth.update_in_provider_queue(data.deep_symbolize_keys)}
       end
     end
 
-    def create_resource(_type, _id, data)
+    def create_resource(type, _id, data)
       manager_resource, attrs = validate_auth_attrs(data)
-      task_id = AuthenticationService.create_authentication_task(manager_resource, attrs)
+
+      klass = ::Authentication.descendant_get(attrs['type'])
+      ensure_supports(type, klass, :create)
+
+      task_id = klass.create_in_provider_queue(manager_resource.id, attrs.deep_symbolize_keys)
       action_result(true, 'Creating Authentication', :task_id => task_id)
     rescue => err
       action_result(false, err.to_s)
     end
 
     def delete_resource_main_action(type, auth, _data)
-      ensure_respond_to(type, auth, :delete, :delete_in_provider_queue)
+      ensure_supports(type, auth, :delete)
       {:task_id => auth.delete_in_provider_queue}
     end
 
--- a/app/controllers/api/subcollections/authentications.rb
+++ b/app/controllers/api/subcollections/authentications.rb
@@ -5,8 +5,11 @@ module Api
         object.respond_to?(:authentications) ? object.authentications : []
       end
 
-      def authentications_create_resource(parent, _type, _id, data)
-        task_id = AuthenticationService.create_authentication_task(parent.manager, data)
+      def authentications_create_resource(parent, type, _id, data)
+        klass = ::Authentication.descendant_get(data['type'])
+        ensure_supports(type, klass, :create)
+
+        task_id = klass.create_in_provider_queue(parent.manager.id, data.deep_symbolize_keys)
         action_result(true, 'Creating Authentication', :task_id => task_id)
       rescue => err
         action_result(false, err.to_s)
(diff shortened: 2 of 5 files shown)


app/controllers/api/authentications_controller.rb
```diff
@@ -2,21 +2,25 @@ module Api
   class AuthenticationsController < BaseController
     def edit_resource(type, id, data)
       api_resource(type, id, "Updating") do |auth|
-        ensure_respond_to(type, auth, :update, :update_in_provider_queue)
+        ensure_supports(type, auth, :update)
```
kbrock, line 5:
```ruby
api_resource(type, id, "Updating", :supports => :update)
```

(or maybe even `:supports => true`)

app/controllers/api/authentications_controller.rb
```diff
@@ -2,21 +2,25 @@ module Api
...
       end
     end
 
-    def create_resource(_type, _id, data)
+    def create_resource(type, _id, data)
       manager_resource, attrs = validate_auth_attrs(data)
-      task_id = AuthenticationService.create_authentication_task(manager_resource, attrs)
+
+      klass = ::Authentication.descendant_get(attrs['type'])
+      ensure_supports(type, klass, :create)
```
kbrock, line 14: Darn, this is very close to the `api_resource` model.

almost like we want to define:
```ruby
def resource_search(id, type)
  ::Authentication.descendant_get(type)
end

def create_resource(type, _id, data)
  api_resource(type, nil, "Creating", :supports => :create) do |klass|
    manager_resource, attrs = validate_auth_attrs(data)
    {:task_id =} klass.create_in_provider_queue(manager_resource.id, attrs.deep_symbolize_keys)}
  end
end
```

But may need to fall back to `api_action`.

Now that I think about it, I had refactored this method a number of times.
Never got it quite right.

---

This is a definite improvement.

kbrock: ooh, right

I kept splitting apart the `validate_auth_attrs`:


```ruby

def create_resource(_type, _id, data)
  type, id = data_resource(data)
  # maybe attrs from data_resource ?
  api_resource(type, id, "Creating", :supports => :create) do |auth, klass|
    {:task_id => klass.create_in_provider_queue(auth.id, attrs.deep_symbolize_keys)}
  end
end

def data_resource(data)
  href = Href.new(data['manager_resource']['href'])
  raise 'invalid manager_resource href specified' unless href.subject && href.subject_id
  [href.subject, href.subject_id]
end
```

nvr mind. there is a reason I never created a PR for this.

app/controllers/api/authentications_controller.rb
```diff
@@ -2,21 +2,25 @@ module Api
...
+
+      task_id = klass.create_in_provider_queue(manager_resource.id, attrs.deep_symbolize_keys)
       action_result(true, 'Creating Authentication', :task_id => task_id)
     rescue => err
       action_result(false, err.to_s)
     end
 
     def delete_resource_main_action(type, auth, _data)
-      ensure_respond_to(type, auth, :delete, :delete_in_provider_queue)
+      ensure_supports(type, auth, :delete)
```
kbrock, line 23: ASIDE: every time I see the necessary `ensure_supports` here I wish I had not implemented `delete_resource_main_action` but had left it at `delete_resource_action`

kbrock:
the tests turned out very nicely.

I think getting to `api_action` or `api_resource` is a great start.

I think I have refactored this a number of times but got derailed with so many changes across the other controllers. Lets do as much as you want here but get it in.

The changes to the tests really show that you are on the right track.

kbrock:
@agrare un-WIP if you ready to merge this one.

There was only one suggestion for `api_resource`, but I'm ok merging without that change.

agrare:
Hey thanks for the nudge @kbrock let me rebase this and see if it still passes specs

kbrock:
kicking. I want this in

Fryguy:
Not all authentications are tied to a provider, so will this change prevent any operations on those?

agrare:
> Not all authentications are tied to a provider, so will this change prevent any operations on those? I'm asking because I thought this one was "special" for exactly this reason.

This still doesn't require authentications to be tied to a provider, previously we used only the presence of a `create_in_provider_queue` method existing on the class, this just changes that to use `supports :create/:update` instead.

agrare:
@kbrock @Fryguy I still need to update the providers to ensure they have the correct supports set, realizing now that is why I had it WIP before https://github.com/ManageIQ/manageiq-api/pull/1220#issuecomment-1697820938

kbrock:
> I still need to update the providers to ensure they have the correct supports set

can you restate this?
What supports (I see `:delete`) and where do those need to go?

kbrock:
re: WIP

I think https://github.com/ManageIQ/manageiq/pull/22894 is the only change needed. It does not require a change in the class listed. (though I did add a caveat that adding to each of the child leaf nodes is an option)

PR #1222, opened 2023-05-03
remove unneeded responders

kbrock:
we are not using respond_with, so no need to setup responders gem with respond_to

this removes the dependency on `responders` gem

part of https://github.com/ManageIQ/manageiq/pull/22490

### DIFF
--- a/app/controllers/api/base_controller.rb
+++ b/app/controllers/api/base_controller.rb
@@ -31,8 +31,6 @@ module Api
     before_action :ensure_pagination, :only => :index
     after_action :log_api_response
 
-    respond_to :json
-
     # Order *Must* be from most generic to most specific
     rescue_from(StandardError)                  { |e| api_error(:internal_server_error, e) }
     rescue_from(NoMethodError)                  { |e| api_error(:internal_server_error, e) }


kbrock:
@miq-bot cross-repo-test /all including ManageIQ/manageiq#22490

app/controllers/api/base_controller.rb
```diff
@@ -31,8 +31,6 @@ class BaseController < ActionController::API
     before_action :ensure_pagination, :only => :index
     after_action :log_api_response
 
-    respond_to :json
-
```
jrafanie, line 35: Wow, that was easy.  Was it too easy?  Are you sure it's not used?  🤯

kbrock: From what I read, the purpose of this method is to setup the metadata so `respond_with` works. But we don't use `respond_with`...

Where I got my information:

https://github.com/heartcombo/responders/blob/956cd874a997083d350951fde4c6cc4944f64e80/lib/action_controller/respond_with.rb#L17-L38

jrafanie: It hurts my soul that there is no gemspec deletion for repsonders needed here.  I guess the dependency wasn't moved over here when the repo was extracted.  Oh well.

LGTM though.

jrafanie (approved):

PR #1225, opened 2023-05-16
add test to document how to pass parameters into service template orders

kbrock:
There was a question on how dialog values (specifically the password) gets
passed into a service order

I ran this scenario to determine how to get fields into a service order.
This is the first step in documenting the API

- Shows 3 ways to get the dialog values into the api.
- Shows where the dialog field are added
- Shows how to get the dialog values out


kbrock:
update (x2):
- cops

PR #1226, opened 2023-05-19
drop Vmdb::GlobalMethod#column_type

kbrock:
part of https://github.com/ManageIQ/manageiq/pull/22520

- Don't traverse MiqExpression::Field associations, but use MiqExpression to do it.
- Use `column_type` in `MiqExpression::Field` instead of the `Vmdb::GlobalMethod#column_type`

Side effects: Api used to only have the input parameter to determine what to do with parameters. Now it has an MiqExpression::Field with all the metadata and introspection. There is a lot of potential here.

### DIFF
--- a/lib/api/filter.rb
+++ b/lib/api/filter.rb
@@ -33,8 +33,7 @@ module Api
     def parse
       filters.select(&:present?).each do |filter|
         parsed_filter = parse_filter(filter)
-        *associations, attr = parsed_filter[:attr].split(".")
-        field = MiqExpression::Field.new(model, associations, attr)
+        field = parsed_filter[:field]
         if field.associations.size > 1
           raise BadRequestError, "Filtering of attributes with more than one association away is not supported"
         end
@@ -83,6 +82,8 @@ module Api
       filter_attr, _, filter_value = filter.partition(operator)
       filter_attr.strip!
       filter_value.strip!
+      *associations, attr = filter_attr.split(".")
+      filter_field = MiqExpression::Field.new(model, associations, attr)
 
       is_regex = filter_value =~ /%|\*/ && methods[:regex]
       str_method = is_regex ? methods[:regex] : methods[:default]
@@ -97,7 +98,7 @@ module Api
                                [array_value, methods[:array]]
                              when /^'(.*)'$/, /^"(.*)"$/
                                unquoted_filter_value = $1
-                               if column_type(model, filter_attr) == :string_set && methods[:string_set]
+                               if filter_field.column_type == :string_set && methods[:string_set]
                                  [unquoted_filter_value, methods[:string_set]]
                                else
                                  [unquoted_filter_value, str_method]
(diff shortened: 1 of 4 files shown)


lib/api/filter.rb
```diff
@@ -138,16 +132,10 @@ def single_expression(field, operator, value)
...
-      else
-        target_class(klass.reflections_with_virtual[reflections.first.to_sym].klass, reflections[1..-1])
-      end
-    end
-
-    def virtual_or_physical_attribute?(klass, attribute)
-      klass.attribute_method?(attribute) || klass.virtual_attribute?(attribute)
+    def split_filter_string(str, operator)
+      filter_attr, _op, filter_value = str.partition(operator)
+      *associations, attr_name = filter_attr.strip.split(".")
```
Fryguy, line 137: I'm not sure if it's possible, but are there cases where the filter_attr will _not_ have a `.`?  The reason I ask is

```ruby
*associations, attr_name = "x".split(".")
associations # => []
attr_name    # => "x"
```

when compared to

```ruby
*associations, attr_name = "x.y".split(".")
associations # => ["x"]
attr_name    # => "y"
```

That is, the trailing object is put in attr_name, regardless of the preceding associations, and I wasn't sure if the trailing thing was _always_ an actual attr_name. (Hope that made sense)

lib/api/filter.rb
```diff
@@ -33,22 +33,19 @@ def initialize(filters, model)
...
+          raise BadRequestError, "attribute #{field} does not exist"
         end
 
-        associations.map! { |assoc| ".#{assoc}" }
-
-        field = "#{model.name}#{associations.join}-#{attr}"
         op    = parsed_filter[:operator]
         expr  = if parsed_filter[:value].kind_of?(Array)
-                  {"OR" => parsed_filter[:value].map { |val| single_expression(field, op, val) }}
+                  {"OR" => parsed_filter[:value].map { |val| single_expression(field.to_s, op, val) }}
```
Fryguy, line 46: Does single_expression always need a string as the input?  If so, can we do the .to_s within the single_expression method, or could that cause problems?

lib/api/filter.rb
```diff
@@ -33,22 +33,19 @@ def initialize(filters, model)
...
-        *associations, attr = parsed_filter[:attr].split(".")
-        if associations.size > 1
+        field = parsed_filter[:field]
+        if field.associations.size > 1
           raise BadRequestError, "Filtering of attributes with more than one association away is not supported"
         end
-        unless virtual_or_physical_attribute?(target_class(model, associations), attr)
-          raise BadRequestError, "attribute #{attr} does not exist"
+        unless field.valid?
+          raise BadRequestError, "attribute #{field} does not exist"
```
Fryguy, line 41: Does `field.to_s` print the right thing here, or will be print out like `#<MiqExpression::FIeld @...>`

lib/api/filter.rb
Fryguy: This code got moved, but the question still stands.

kbrock: Yes, good question. Most of the time the `filter_attr` will not have a `.`.
This is the base case and most common case for our tests: e.g.: https://github.com/ManageIQ/manageiq-api/blob/column_type/spec/lib/api/filter_spec.rb#L3

```
# model = "Vm"
str = "name"
associations, attr_name = [], "name

str = "host.name" 
associations, attr_name = ["host"], "name"
```

Fryguy: Ok perfect then this code is good

PR #1228, opened 2023-05-31
[WIP] Use AS::Dependencies interlock with classic autoloader

jrafanie:
Hide the details about inspecting the autoloader by using the core method which will use the interlock with classic and no interlock for zeitwerk.

Depends on: https://github.com/ManageIQ/manageiq/pull/22539
Related: https://github.com/ManageIQ/manageiq/issues/22000

### DIFF
--- a/app/controllers/api/providers_controller.rb
+++ b/app/controllers/api/providers_controller.rb
@@ -141,13 +141,13 @@ module Api
     end
 
     def permitted_subclasses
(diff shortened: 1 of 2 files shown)


jrafanie:
I'll rekick this once the dependent PR is merged

jrafanie:
Core PR now patches directly so this is not needed.

PR #1231, opened 2023-06-20
Add support for edit configuration_script_payloads

agrare:
Workflows have a credentials jsonb column which allows users to map credentials to values being requested by the workflow.  This means we have to allow for a user to edit configuration_script_payloads to do this mapping.

`PATCH /api/configuration_script_payloads/:id {"credentials":{"api_user":{"credential_ref": "manageiq_api", "credential_field": "userid"}, "api_password":{"credential_ref": "manageiq_api", "credential_field": "password"}}`

Follow-up:
* https://github.com/ManageIQ/manageiq-providers-workflows/pull/40

TODO:
- [x] Fail the request if any of the credentials can't be found
- [x] Add specs for authentications owned by other user/tenant

### DIFF
--- a/app/controllers/api/configuration_script_payloads_controller.rb
+++ b/app/controllers/api/configuration_script_payloads_controller.rb
@@ -11,6 +11,31 @@ module Api
       unpermitted_params = data.keys.map(&:to_s) - allowed_params
       raise BadRequestError, _("Invalid parameters: %{params}" % {:params => unpermitted_params.join(", ")}) if unpermitted_params.any?
 
+      # If a credentials payload is provided, map any requested authentication
+      # records to the configuration_script_payload via the
+      # authentications_configuration_script_payloads join table.
+      unless data["credentials"].nil?
+        # Credentials can be a static string or a payload with an external
+        # Authentication record referenced by credential_ref and credential_field.
+        credential_refs = data["credentials"].values.select { |val| val.kind_of?(Hash) }.pluck("credential_ref")
+        # Lookup the Authentication record by ems_ref in the parent manager's
+        # list of authentications.
+        credentials     = resource.manager&.authentications&.where(:ems_ref => credential_refs) || []
+        # Filter the collection based on the current user's RBAC roles.
+        credentials, _  = collection_filterer(credentials, "authentications", ::Authentication)
+        # If any requested authentications were unable to be found, either due
+        # to a bad credential_ref or due to RBAC then raise a 400 BadRequestError.
+        missing_credential_refs = credential_refs - credentials.pluck(:ems_ref)
+        if missing_credential_refs.any?
+          raise BadRequestError,
+                _("Could not find credentials %{missing_credential_refs}") %
+                {:missing_credential_refs => missing_credential_refs}
+        end
+        # Reset the authentications collection with the current set of credentials.
+        # This will also remove any credential references not in the new payload.
+        resource.authentications = credentials
+      end
+
       resource.update!(data.except(*ID_ATTRS))
       resource
     end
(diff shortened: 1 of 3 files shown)


Fryguy:
Does this allow the user to change things other than the credentials? If it's coming from git, we probably shouldn't allow it, except those specific fields/associations.

agrare:
@Fryguy updated to only allow updates to the name, payload, and payload_type if there is no linked configuration_script_source, and only updates to the description and credentials otherwise.

app/controllers/api/configuration_script_payloads_controller.rb
```diff
@@ -1,5 +1,18 @@
 module Api
   class ConfigurationScriptPayloadsController < BaseController
     include Subcollections::Authentications
+
+    def edit_resource(type, id, data)
+      resource = resource_search(id, type)
+
+      allowed_params  = %w[description credentials]
```
Fryguy, line 8: Do we also verify that the user has RBAC permission to the credentials they are trying to set?

config/api.yml
```diff
@@ -986,18 +986,28 @@
     :options:
     - :collection
     - :subcollection
-    :verbs: *g
+    :verbs:
+    - :get
+    - :put
+    - :post
+    - :patch
```
Fryguy, line 993: Minor but you should be able to use


```suggestion
    :verbs: *gppp
```

agrare: Do we have that?  I don't see it in the list of sets

app/controllers/api/configuration_script_payloads_controller.rb
agrare: That mapping to authentication records is done separately via the `configuration_script_payload/:id/authentications` subcollection, this is about credential references in the ASL payload are mapped to values in those authentications so I don't think RBAC would come into play here here, that would have already taken place when they mapped authentication records.

Example credentials payload:
`{"credentials":{"api_user":"$.manageiq_api.userid", "api_password":"$.manageiq_api.password"}}`

config/api.yml
Fryguy: Seems not, but then I'd prefer adding that one new set to keep the verbs section consistent.

agrare: Done

Fryguy:
As discussed, I think having the API for credential mapping be a 2 step process should be simplified.  That is right now a user must

1. Give permission to the workflow for a specific credential
2. Map fields from that credential into fields from the asl

Instead, in step 2 we already know, through the act of mapping, that they also want to give permission, so we can give permission implicitly under the covers.

So, given a payload of

```json
{
  "credentials": {
    "api_user": "$.manageiq_api.userid",
    "api_password": "$.manageiq_api.password"
  }
}
```

we would just give permission to manageiq_api directly if not already given (and ensure this is a credential they have access to in the first place), then update the credentials field accordingly.

---

Separately we discussed the payload itself.  A user could technically hardcode things like usernames, and then we wouldn't be able to tell apart references vs strings.  Additionally parsing strings is a potential source of error.  Instead a payload like the following would allow us to differentiate:

```json
{
  "credentials": {
    "api_user": "root",
    "api_password": {
      "credential_ref": "manageiq_api",
      "credential_field": "password"
    }
  }
}
```

agrare:
https://github.com/ManageIQ/manageiq-providers-workflows/pull/40 changes the expected credential payload format

app/controllers/api/configuration_script_payloads_controller.rb
agrare: Now that we're automatically updating the authentications join table we are running RBAC on the requested credentials below.

bdunne:
Backported to `quinteros` in commit cd280b117ab9f08e7f381f55c36228f4a940d0aa.

```text
commit cd280b117ab9f08e7f381f55c36228f4a940d0aa
Date:   Wed Aug 16 13:21:59 2023 -0400

    Merge pull request #1231 from agrare/add_edit_support_configuration_script_payloads
    
    Add support for edit configuration_script_payloads
    
    (cherry picked from commit 9a77ef21ecb166e529a77a9b51f6e299ac56d034)
```

PR #1232, opened 2023-07-12
dropping lifecycle_event table

kbrock:
part of https://github.com/ManageIQ/manageiq/issues/21204

It was added but never used

### DIFF
--- a/app/controllers/api/vms_controller.rb
+++ b/app/controllers/api/vms_controller.rb
@@ -67,14 +67,6 @@ module Api
       end
     end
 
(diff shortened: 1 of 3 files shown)


jrafanie (approved):
LGTM.  Do we need to be cautious with removing this in a patch release?  Do we need other things merged first?  I'm fine with whatever but I guess we need at least a minor bump.

Fryguy:
API isn't really semver, but this was a non functioning API anyway, so I think it's ok to put in a patch release.

kbrock:
Think this can be removed on its own and not lock step with the other lifecycle event code

PR #1233, opened 2023-07-12
Filter uses a regular expression

kbrock:
Overview
========

Instead of going through each character to find the operators,
use a regular expression that is coded to handle all the operators

Depends
=======

test helper

- https://github.com/ManageIQ/manageiq/pull/22615

Before
======

Loop through all characters in the filter
Loop through all operators
break when there is a match

After
=====

Use a regular expression to parse the filter

Notes
=====

This assumes a filter is made up of a name, operator, and value.
These assumptions seem to protect the code later on.

The original regular expression for the operator was: `/[!<>=][=~]=?/`
While easier to read, it had quite a few false positives.
The false positives were caught by the OPERATORS.key?, so not a big thing.
Possibly just using a long `/(a|b|c|d)/` may be a better route

I do wonder if we should be more greedy matching operator punctuation, and then
raising an error. The pedantic approach may make for a better developer
experience, but leaving it as it was for now.

### DIFF
--- a/lib/api/filter.rb
+++ b/lib/api/filter.rb
@@ -1,5 +1,6 @@
 module Api
   class Filter
+    OPERATOR_REGEXP = /^ *([^=!>< ]+) *(!==?|[=!]~|[<>=]=?) *(.*?) *$/.freeze
     OPERATORS = {
       "!="  => {:default => "!=", :regex => "REGULAR EXPRESSION DOES NOT MATCH", :null => "IS NOT NULL"},
       "<="  => {:default => "<="},
@@ -121,22 +122,13 @@ module Api
     end
 
     def split_filter_string(filter)
-      operator = nil
-      operators_from_longest_to_shortest = OPERATORS.keys.sort_by(&:size).reverse
-      filter.size.times do |i|
-        operator = operators_from_longest_to_shortest.detect do |o|
-          o == filter[(i..(i + o.size - 1))]
-        end
-        break if operator
-      end
+      filter_match = OPERATOR_REGEXP.match(filter)
+      filter_attr, operator, filter_value = filter_match&.captures
 
-      if operator.blank?
+      unless OPERATORS.key?(operator)
         raise BadRequestError, "Unknown operator specified in filter #{filter}"
       end
 
-      filter_attr, _op, filter_value = filter.partition(operator)
-      filter_value.strip!
-      filter_attr.strip!
       *associations, attr_name = filter_attr.split(".")
       [MiqExpression::Field.new(model, associations, attr_name), operator, filter_value]
     end
--- a/lib/api/filter.rb
+++ b/lib/api/filter.rb
@@ -68,7 +68,6 @@ module Api
       methods = OPERATORS[operator]
 
       is_regex = filter_value =~ /%|\*/ && methods[:regex]
-      str_method = is_regex ? methods[:regex] : methods[:default]
 
       filter_value, method = case filter_value
                              when /^\[(.*)\]$/
@@ -82,8 +81,10 @@ module Api
                                unquoted_filter_value = $1
                                if filter_field.column_type == :string_set && methods[:string_set]
                                  [unquoted_filter_value, methods[:string_set]]
+                               elsif is_regex
+                                 [unquoted_filter_value, methods[:regex]]
                                else
-                                 [unquoted_filter_value, str_method]
+                                 [unquoted_filter_value, methods[:default]]
                                end
                              when /^(NULL|nil)$/i
                                [nil, methods[:null] || methods[:default]]
--- a/spec/lib/api/filter_spec.rb
+++ b/spec/lib/api/filter_spec.rb
@@ -441,6 +441,7 @@ RSpec.describe Api::Filter do
 
     [
       'name^bb',
+      '=bb',
       'name',
     ].each do |str|
       it "complains about '#{str}'" do
--- a/spec/lib/api/filter_spec.rb
+++ b/spec/lib/api/filter_spec.rb
@@ -406,4 +406,52 @@ RSpec.describe Api::Filter do
       expect(actual.exp).to eq(expected)
     end
   end
+
+  describe "#split_filter_string (private)" do
+    let(:name_field) { MiqExpression::Field.new(Vm, [], "name") }
+
+    Api::Filter::OPERATORS.each_key do |operator|
+      it "matches operator #{operator}" do
+        expect(split_filter_string("name#{operator}x")).to eq([name_field, operator, "x"])
+        expect(split_filter_string("name#{operator}>x")).to eq([name_field, operator, ">x"])
+      end
+    end
+
+    it "parses parses associations" do
+      field = MiqExpression::Field.new(Vm, ["host"], "name")
+      expect(split_filter_string("host.name=x")).to eq([field, "=", "x"])
+    end
+
+    it "handles whitespace" do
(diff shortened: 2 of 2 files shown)


kbrock:
udpate:
- fix rubocops: spacing and add a `freeze`

lib/api/filter.rb
```diff
@@ -120,24 +122,16 @@ def single_expression(field, operator, value)
...
-        operator = operators_from_longest_to_shortest.detect do |o|
-          o == filter[(i..(i + o.size - 1))]
-        end
-        break if operator
-      end
+      filter_match = OPERATOR_REGEXP.match(filter)
+      filter_attr, operator, filter_value = filter_match[1..3] if filter_match
 
-      if operator.blank?
+      if operator.blank? || !OPERATORS.key?(operator)
```
Fryguy, line 128: Since a blank operator wouldn't be in the list anyway, can you do:

```suggestion
      unless OPERATORS.key?(operator)
```

lib/api/filter.rb
```diff
@@ -120,24 +122,16 @@ def single_expression(field, operator, value)
...
-      operator = nil
-      operators_from_longest_to_shortest = OPERATORS.keys.sort_by(&:size).reverse
-      filter.size.times do |i|
-        operator = operators_from_longest_to_shortest.detect do |o|
-          o == filter[(i..(i + o.size - 1))]
-        end
-        break if operator
-      end
+      filter_match = OPERATOR_REGEXP.match(filter)
+      filter_attr, operator, filter_value = filter_match[1..3] if filter_match
```
Fryguy, line 126:
```suggestion
      filter_attr, operator, filter_value = filter_match.captures if filter_match
```

lib/api/filter.rb
```diff
@@ -120,24 +122,16 @@ def single_expression(field, operator, value)
...
+      if operator.blank? || !OPERATORS.key?(operator)
         raise BadRequestError, "Unknown operator specified in filter #{filter}"
       end
 
-      filter_attr, _op, filter_value = filter.partition(operator)
-      filter_value.strip!
-      filter_attr.strip!
       *associations, attr_name = filter_attr.split(".")
       [MiqExpression::Field.new(model, associations, attr_name), operator, filter_value]
+      # [associations, attr_name, operator, filter_value]
```
Fryguy, line 134: I don't understand this comment.

spec/lib/api/filter_spec.rb
```diff
@@ -406,4 +406,52 @@
...
+      "name!=1"     => [[], "name", "!=", "1"],
+      "name=~1"     => [[], "name", "=~", "1"],
+      "name!~1"     => [[], "name", "!~", "1"],
+      "name!==1"    => [[], "name", "!==", "1"],
+      # funky formatting
+      " name = x "  => [[], "name", "=", "x"],
+      "name=  "     => [[], "name", "=", ""],
+      "name= ' ' "  => [[], "name", "=", "' '"],
+      "host.name=x" => [["host"], "name", "=", "x"],
+    }.each do |str, (associations, column, eoperator, evalue)|
```
Fryguy, line 431: eoperator and evalue?

spec/lib/api/filter_spec.rb
```diff
@@ -406,4 +406,52 @@
...
+      it "parses #{str}" do
+        field, operator, value = split_filter_string(str)
+
+        # currently, can not compare one MiqExpression::Target with another
```
Fryguy, line 435:
```suggestion
        # currently, cannot compare one MiqExpression::Target with another
```

On that note, does it make sense to implement `==` on MiqExpression::Target or is that incorrect?

spec/lib/api/filter_spec.rb
```diff
@@ -406,4 +406,52 @@
...
+    ].each do |str|
+      it "complains about '#{str}'" do
+        expect do
+          split_filter_string(str)
+        end.to raise_error(Api::BadRequestError, /Unknown operator specified/)
+      end
+    end
+
+    def split_filter_string(str)
+      described_class.new("", Vm).send(:split_filter_string, str)
```
Fryguy, line 454: we have `model` defined as a `let` above, but we hardcoded Vm here.  I think we don't need the let anoymore.

lib/api/filter.rb
Fryguy: Alternatively it might be nice to use named captures in the original regex with:

```ruby
    OPERATOR_REGEXP = /^ *(?<filter_attr>[^=!>< ]+) *(?<operator>!==?|[=!]~|[<>=]=?) *(?<filter_value>.*?) *$/.freeze
```

Then here you can use `values_at` or `named_captures` (or continue to just use `captures`)

lib/api/filter.rb
```diff
@@ -1,5 +1,6 @@
 module Api
   class Filter
+    OPERATOR_REGEXP = /^ *([^=!>< ]+) *(!==?|[=!]~|[<>=]=?) *(.*?) *$/.freeze
```
Fryguy, line 3: Can you write a spec that ensures that every key in the `OPERATORS` hash will be recognized by this regexp?  The reason I'm asking is that if someone adds a _new_ operator to the `OPERATORS` hash, we need to ensure that this Regexp also covers it.  I think the spec could look something like:

```ruby
OPERATORS.keys.each do |operator|
  it "OPERATORS_REGEXP handles the '#{operator}' operator" do
    expect("x#{operator}y").to match(OPERATORS_REGEXP)
  end
end
```

spec/lib/api/filter_spec.rb
```diff
@@ -406,4 +406,52 @@
...
     end
   end
+
+  describe "#split_filter_string (private)" do
+    let(:model) { Vm }
+
+    {
+      # all operators, and a few curve balls
+      "name=1"      => [[], "name", "=", "1"],
+      "name=>1"     => [[], "name", "=", ">1"],
```
Fryguy, line 416: interesting find!

spec/lib/api/filter_spec.rb
kbrock: yea. I had forgotten to link it

lib/api/filter.rb
kbrock: I dropped.

to make testing easier, I had changed this to the return value.
And I'm still wondering about it.

But to minimize change, probably best to just leave this method as is

spec/lib/api/filter_spec.rb
kbrock: the old tests needed expected_operator and operator.

Since adding `MiqExpression::Field#==`, the variables went away.

all set

lib/api/filter.rb
kbrock: nice! I changed to only depends upon `OPERATOR`

If the regexp is missing an entry, then `split_filter_string` will not parse it correctly.

lib/api/filter.rb
kbrock: I tried a few times. was having getting it to find the captures. may revisit at a later time.

kbrock:
udpate:
- depends upon https://github.com/ManageIQ/manageiq/pull/22615 - but it cleans up the code
- key? => includes
- dynamic tests using `OPERATORS`
- remove unneeded `let`

kbrock:
UN-WIP: rebuild now that dependency was merged

Fryguy:
This fell off my radar - will re-review

kbrock:
update:
- rebase master

PR #1239, opened 2023-09-05
Allow configuration_script_payloads#credentials to be shown

agrare:
The configuration_script_payloads#credentials jsonb column does not contain credentials but based on how the user names the keys (e.g. `"vcenter_password": {"credential_ref": "vcenter_credential", "credential_field": "password"}` the payload will be stripped which causes issues for the UI to display these.

```
>> pp ConfigurationScriptPayload.first.credentials
=> 
{"manageiq_api_user"=>{"credential_ref"=>"manageiq-api", "credential_field"=>"userid"},
 "manageiq_api_password"=>{"credential_ref"=>"manageiq-api", "credential_field"=>"password"}}
```

Problem:
```
GET /api/configuration_script_payloads/1
{
  "href": "http://localhost:3000/api/configuration_script_payloads/1",
  "id": "1",
  "manager_id": "2",
  ...
  "credentials": {
    "manageiq_api_user": {
      "credential_ref": "manageiq-api",
      "credential_field": "userid"
    }
  },
```

With this PR:
```
GET /api/configuration_script_payloads/1
{
  "href": "http://localhost:3000/api/configuration_script_payloads/1",
  "id": "1",
  "manager_id": "2",
  ...
  "credentials": {
    "manageiq_api_user": {
      "credential_ref": "manageiq-api",
      "credential_field": "userid"
    },
    "manageiq_api_password": {
      "credential_ref": "manageiq-api",
      "credential_field": "password"
    }
  },
```

### DIFF
--- a/app/controllers/api/configuration_script_payloads_controller.rb
+++ b/app/controllers/api/configuration_script_payloads_controller.rb
@@ -2,6 +2,18 @@ module Api
   class ConfigurationScriptPayloadsController < BaseController
     include Subcollections::Authentications
 
+    def api_resource_action_options
+      # ConfigurationScriptPayloads do not have any passwords stored directly
+      # in the record, they reference the Authentication model via the
+      # credentials jsonb mapping.  The names of these mappings are user defined
+      # and can include e.g. "api_password" => {"credential_ref" => ..} and this
+      # entire key would be removed from the payload.
+      #
+      # Since there aren't any encrypted attributes in this record it is safe
+      # to include encrypted attributes in the payload response.
+      %w[include_encrypted_attributes]
+    end
+
     def edit_resource(type, id, data)
       resource = resource_search(id, type)
 
--- a/spec/requests/configuration_script_payloads_spec.rb
+++ b/spec/requests/configuration_script_payloads_spec.rb
@@ -105,7 +105,22 @@ RSpec.describe 'Configuration Script Payloads API' do
         it "adds the authentication to the configuration_script_payload.authentications" do
           api_basic_authorize collection_action_identifier(:configuration_script_payloads, :edit, :post)
 
-          post(api_configuration_script_payloads_url, :params => {:action => 'edit', :resources => [{:id => script_payload.id, :name => 'foo', :credentials => {"my-cred" => {"credential_ref" => "my-credential", "credential_field" => "userid"}}}]})
+          expected_credentials = {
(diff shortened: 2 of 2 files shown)


app/controllers/api/configuration_script_payloads_controller.rb
```diff
@@ -2,6 +2,14 @@ module Api
...
     include Subcollections::Authentications
 
+    def api_resource_action_options
+      if @req.action == "read" && @req.collection_id && @req.subcollection.blank?
+        %w[include_encrypted_attributes]
+      else
+        super
+      end
+    end
+
```
agrare, line 12: NOTE this is overly broad as it disables all filters, would prefer to selectively allow specific columns while still filtering out actual v2 key values.

kbrock: we throw away strings (that should be encrypted) but keep hashes?

```ruby
def normalize_encrypted(value)
  value.kind_of?(String) ? nil : value
end
```

kbrock: @agrare It will not work if credentials are a mapping string

```json
      "Credentials": {
        "api_user.$": "$.api_user",
        "api_password.$": "$.api_password",
        "vcenter_user.$": "$.vcenter_user",
        "vcenter_password.$": "$.vcenter_password"
      },
```

/via https://github.com/ManageIQ/workflows-examples/blob/master/provision-vm-service/provision-vm.asl#L9C1-L14C9

But if it is a hash, then it may work:

```json
{"api_password": {"credential_ref": "manageiq_api", "credential_field": "userid"}}
```

kbrock:
Is it possible to have a spec with a record in there?
That way we can tweak how we want this return to work.

The thought is it will allow `$.` to work?
aws has a particular format for marking a field as dynamic. Do we want to follow this and add that specific data into the encryption helper?

agrare:
> Is it possible to have a spec with a record in there?

:+1: good idea, added

Fryguy:
Backported to `quinteros` in commit a430cfffe04bba6a51e7b214c4b89afef1003c32.

```text
commit a430cfffe04bba6a51e7b214c4b89afef1003c32
Date:   Fri Sep 22 09:45:24 2023 -0400

    Merge pull request #1239 from agrare/allow_configuration_script_payloads_credentials_to_be_shown
    
    Allow configuration_script_payloads#credentials to be shown
    
    (cherry picked from commit 39e5ed90a731afc22919cf17cb81f4dd2e17fb5d)
```

PR #1244, opened 2023-12-19
Drop legacy Host credentials payload, replaced by DDF params

agrare:
For a period of time we supported both the older `:credentials` payload as well as the newer DDF `authentications => {auth_type => {userid, password}}` style.

https://github.com/ManageIQ/manageiq-api/pull/1202

Now that the host edit form is in react (https://github.com/ManageIQ/manageiq-ui-classic/pull/8608) we no longer need to keep this old style around.

Depends on:
- [x] https://github.com/ManageIQ/manageiq/pull/22819

Follow-up:
* Update the manageiq-documentation/api/references to udpate the host examples

### DIFF
--- a/app/controllers/api/hosts_controller.rb
+++ b/app/controllers/api/hosts_controller.rb
@@ -1,6 +1,5 @@
 module Api
   class HostsController < BaseProviderController
-    CREDENTIALS_ATTR = "credentials".freeze
     AUTH_ATTR = "authentications".freeze
     AUTH_TYPE_ATTR = "auth_type".freeze
     DEFAULT_AUTH_TYPE = "default".freeze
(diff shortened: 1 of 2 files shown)


app/controllers/api/hosts_controller.rb
```diff
@@ -12,45 +11,28 @@ class HostsController < BaseProviderController
...
+
       resource_search(id, type).tap do |host|
-        # begin legacy ui-classic
-        all_credentials = Array.wrap(credentials).each_with_object({}) do |creds, hash|
-          auth_type = creds.delete(AUTH_TYPE_ATTR) || DEFAULT_AUTH_TYPE
-          creds.symbolize_keys!
-          creds.reverse_merge!(:userid => host.authentication_userid(auth_type))
-          hash[auth_type.to_sym] = creds
+        if authentications.present?
+          authentications.deep_symbolize_keys!
```
agrare, line 20: NOTE update_authentications only works with symbolized keys

app/controllers/api/hosts_controller.rb
```diff
@@ -12,45 +11,28 @@ class HostsController < BaseProviderController
...
-        host.update_authentication(all_credentials) if all_credentials.present?
       end
     end
 
     def verify_credentials_resource(type, id = nil, data = {})
       api_resource(type, id, "Verifying Credentials for") do |host|
-        remember_host = data["remember_host"] == "true"
-        authentications, auth_type = symbolize_password_keys!(data[AUTH_ATTR])
-        {:task_id => host.verify_credentials_task(User.current_userid, auth_type, :credentials => authentications, :remember_host => remember_host)}
+        auth_type = data["authentications"].keys.first
```
agrare, line 28: NOTE In the future I'd like to make `Host#verify_credentials` as close as possible to `ExtManagementSystem#verify_credentials` which takes just the data payload and no auth_type.

spec/requests/hosts_spec.rb
```diff
@@ -94,8 +45,8 @@
         params = {
           "action"    => "edit",
           "resources" => [
-            {:id => host1.id, :authentications => {"default" => {:username => "abc", :password => "abc123"}}},
-            {:id => host2.id, :authentications => {"default" => {:username => "def", :password => "def456"}}}
+            {:id => host1.id, "authentications" => {"default" => {"username" => "abc", "password" => "abc123"}}},
+            {:id => host2.id, "authentications" => {"default" => {"username" => "def", "password" => "def456"}}}
```
Fryguy, line 49: I think id should be a string here too, right?
```suggestion
            {"id" => host1.id, "authentications" => {"default" => {"username" => "abc", "password" => "abc123"}}},
            {"id" => host2.id, "authentications" => {"default" => {"username" => "def", "password" => "def456"}}}
```

spec/requests/hosts_spec.rb
```diff
@@ -62,23 +28,8 @@
...
-
-      it "can update passwords on multiple hosts by id (via credentials)" do
-        host1 = FactoryBot.create(:host_with_authentication)
-        host2 = FactoryBot.create(:host_with_authentication)
-        api_basic_authorize action_identifier(:hosts, :edit)
-        options = [
-          {:id => host1.id, :credentials => {:password => "abc123"}},
-          {:id => host2.id, :credentials => {:password => "def456"}}
+          {:href => api_host_url(nil, host1), "authentications" => {"default" => {"password" => "abc123"}}},
+          {:href => api_host_url(nil, host2), "authentications" => {"default" => {"password" => "def456"}}}
```
Fryguy, line 32: Similarly here for href, I think it should be a string key

Fryguy (approved):
LGTM, just one minor question in the specs.

spec/requests/hosts_spec.rb
agrare: Done

PR #1248, opened 2024-02-19
Update Resource Pool API to include type attribute

Guddetisandeep:
Dependent PR for enabling Resource Pool feature under Clouds section(Compute > Clouds > Resource Pools)

- manageiq Core changes
https://github.com/ManageIQ/manageiq/pull/22879

- manageiq-ui-classic changes
https://github.com/ManageIQ/manageiq-ui-classic/pull/8929

- manageiq-schema changes
https://github.com/ManageIQ/manageiq-schema/pull/718

- manageiq-api changes
https://github.com/ManageIQ/manageiq-api/pull/1248

### DIFF
--- a/app/controllers/api/resource_pools_controller.rb
+++ b/app/controllers/api/resource_pools_controller.rb
@@ -3,5 +3,29 @@ module Api
     include Subcollections::Policies
     include Subcollections::PolicyProfiles
     include Subcollections::Tags
+
+    # GET /api/resource_pools
+    def index
+      if params[:type].present?
+        resource_pools = fetch_resource_pools_by_type(params[:type])
+        res = collection_filterer(resource_pools, :resource_pools, ResourcePool).flatten
+        counts = Api::QueryCounts.new(ResourcePool.count, res.count, resource_pools.count)
+        render_collection(:resource_pools, res, :counts => counts, :expand_resources => @req.expand?(:resources), :name => 'resource_pools')
+      else
+        super
+      end
+    end
+
+    private
+
+    # Fetch resource pools by exact type
+    def fetch_resource_pools_by_type(type)
+      rp = type.safe_constantize unless type.is_a?(Class)
+      if rp && rp <= ResourcePool
+        rp.all
+      else
+        ResourcePool.none
+      end
+    end
   end
 end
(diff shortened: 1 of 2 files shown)


Guddetisandeep:
The query returns only the resource pools associated with cloud management when the `type` attribute is used with the specified class `(ManageIQ::Providers::CloudManager::ResourcePool)`
```
curl -s -X GET -H "X-Auth-Token: $(miq_auth_token)" -H "Accept: application/json" "http://localhost:3000/api/resource_pools?type=ManageIQ::Providers::CloudManager::ResourcePool" | jq .
{
  "name": "resource_pools",
  "count": 142,
  "subcount": 1,
  "subquery_count": 1,
  "pages": 1,
  "resources": [
    {
      "href": "http://localhost:3000/api/resource_pools/162"
    }
  ],
  "actions": [
    {
      "name": "query",
      "method": "post",
      "href": "http://localhost:3000/api/resource_pools"
    }
  ],
  "links": {
    "self": "http://localhost:3000/api/resource_pools?type=ManageIQ::Providers::CloudManager::ResourcePool&offset=0",
    "first": "http://localhost:3000/api/resource_pools?type=ManageIQ::Providers::CloudManager::ResourcePool&offset=0",
    "last": "http://localhost:3000/api/resource_pools?type=ManageIQ::Providers::CloudManager::ResourcePool&offset=0"
  }
}
```

app/controllers/api/resource_pools_controller.rb
```diff
@@ -3,5 +3,32 @@ class ResourcePoolsController < BaseController
...
+    else
+      super
+    end
+  end
+    
+    
+    private
+
+    # Fetch resource pools by exact type
+    def fetch_resource_pools_by_type(type)
```
Fryguy, line 23: This is not the right way to do this.  I would expect a type parmeter incoming that lists the type you need, and then it's a normal STI query.

Fryguy:
@Guddetisandeep - We found there's a better way to do this than using a type filter.  Instead, pass a `collection_class` parameter at the same level as the filter.

Fryguy (changes requested):
I think we can close this PR, and the callers can use `collection_class` instead, which should already work for ResourcePools.

Guddetisandeep:
Closing this PR using `collection_class` instead of the type filter.

PR #1249, opened 2024-02-23
Add MiqTask#deliver and execute

kbrock:
Provisioning RequestTasks:

- deliver to kick off the before and followup workflow
- execute will kick off the main meat

### DIFF
--- a/config/api.yml
+++ b/config/api.yml
@@ -3233,6 +3233,8 @@
       :post:
       - :name: cancel
         :identifier: miq_request_control
+      - :name: execute
+        :identifier: miq_request_control
     :subcollection_actions:
       :get:
       - :name: read
--- a/app/controllers/api/request_tasks_controller.rb
+++ b/app/controllers/api/request_tasks_controller.rb
@@ -1,5 +1,15 @@
 module Api
   class RequestTasksController < BaseController
     include Api::Mixins::ResourceCancel
+
+    # execute queues the work.
+    # this is typically called from within the workflow
+    def execute_resource(type, id, _data)
+      api_resource(type, id, "Executing") do |task|
+        raise BadRequestError, "Resource must be approved. state is #{task.state}" unless task.approved?
+
+        task.execute_queue
+      end
+    end
   end
 end


kbrock:
update:
- Rebased
- Fixed whitespace, and unused parameter

rubocop:
- Left `api_resource` with an actual block - follows the pattern better

config/api.yml
```diff
@@ -3233,6 +3233,10 @@
       :post:
       - :name: cancel
         :identifier: miq_request_control
+      - :name: deliver
+        :identifier: miq_request_control
+      - :name: execute
+        :identifier: miq_request_control
```
Fryguy, line 3239: I'm not sure here - This would give users with the ability to cancel the requests to also execute and deliver, which is really a system level task - I wonder if this should be super admin only or something 🤔

kbrock: @Fryguy did you just say `service_admin`?

kbrock: this will be using the authentication record that workflow/floe is passing. (api_user)

@agrare Do you typically pass superadmin to floe?
Do you feel that is what customers should do?

agrare: I had imagined that the user ordering the service would be the one posting to the API, an admin would create the workflows and map what credentials would be used but at order time it would be the unprivileged user account

kbrock: Most of these tasks will talk to the API and want to act as the requesting user.

But there are requests that need to be super admin.

This endpoint is basically saying "All the workflows have been run. All the variables have been figured out. Let's talk with vmware and make the actual request happen"

That should only be available from a super admin.
The current vmware specific workflow has:
- vmware admin auth
- miq requestor user auth

So it would make sense to have:
- miq admin auth (to make provider privileged requests)
- miq requestor user auth (to do stuff as user privileges)

Just not sure the best way to pass the admin auth

Fryguy: I'm lost now about who and when - let's discuss over voice and maybe draw it out.

kbrock:
WIP: testing ensuring it was approved

config/api.yml
```diff
@@ -3233,6 +3233,8 @@
       :post:
       - :name: cancel
         :identifier: miq_request_control
+      - :name: execute
+        :identifier: miq_request_control
```
Fryguy, line 3237: Continung from https://github.com/ManageIQ/manageiq-api/pull/1249#discussion_r1500946506

I still don't like this permission here...something feels off about it.  I admit that miq_request_control is used for cancel, but being able to cancel your own request is very different from executing it.

kbrock: added 'you can only execute this if it has been approved' feels like it lessens the request.

kbrock:
we are using a builtin instead

PR #1256, opened 2024-03-25
test sub-type search for basic resource listing (GET index)

kbrock:
alternative to https://github.com/ManageIQ/manageiq-api/pull/1248

Overview
========

A typical api request views all VMs: `/api/vms`
If a screen displays all of a specific resource (e.g.: all Cloud or Infra VMs), then `collection_class` can be used: e.g.: `/api/vms?collection_class=ManageIQ::Providers::CloudManager::Vm`

Change
======

There was concern that `collection_class` could only handle a leaf class (e.g.: `ManageIQ::Providers::Openstack::CloudManager::Vm`) and would not be able to handle a grouping (e.g.: `ManageIQ::Providers::CloudManager::Vm`).

Originally this was the functionality added to display these resources. But it turns out `collection_class` can handle any approperiate class.

So this test has been converted to a test only PR.

### DIFF
--- a/spec/requests/querying_spec.rb
+++ b/spec/requests/querying_spec.rb
@@ -1007,5 +1007,19 @@ describe "Querying" do
       expect_query_result(:vms, 2, 2)
       expect(response.parsed_body["resources"].collect { |vm| vm["name"] }).to match_array(%w(bb cc))
     end
+
+    it "finds child classes given a parent class" do
+      FactoryBot.create(:vm_vmware, :name => "aa")
+      FactoryBot.create(:vm_vmware_cloud, :name => "bb")
+      FactoryBot.create(:vm_amazon, :name => "cc")
+
+      # parent of amazon and vmware_cloud.
+      vmc = FactoryBot.build(:vm_cloud)
+
+      get api_vms_url, :params => { :expand => "resources", :collection_class => vmc.class.name }
+
+      expect_query_result(:vms, 2, 2)
+      expect(response.parsed_body["resources"].collect { |vm| vm["name"] }).to match_array(%w(bb cc))
+    end
   end
 end


kbrock:
@Guddetisandeep does this meet your needs?

Guddetisandeep:
> @Guddetisandeep does this meet your needs?

Yes, I am able to retrieve the `resource pools` by specifying the subclass as the type

Fryguy:
> @Fryguy thoughts?
> 
>     * Do you want all resources to handle a `type` parameter?

No.  However, there already are resources that handle a type parameter, such as providers, vms, etc.  I thought the code for those already would have a helper method and you wouldn't have to reimplement one as in this PR.  Can you check if one exists for those already and reuse it?  Otherwise I could see using your method in those other places.

>     * Is it ok to only support a single value for `type` parameter rather than a list of types?

That's a good question.  I think for now single type is fine, but multiple types might be very useful.  I know @agrare hit something recently where he needed multiple types (or perhaps it was an in-between class type like `Vm`).

Fryguy:
@kbrock How did you fix the brakeman warning?  The code doesn't look any different and yet brakeman didn't complain?

kbrock:
@Fryguy the new code makes it more obvious that the unprotected class is only available/in play 

```ruby
param_klass = params[:type]&.safe_constantize
klass = param_klass if param_klass && param_klass < klass
# not obvious if we are going to use `param_klass` in the code
```

```ruby
# @param base_type [String] api.yaml[type].klass
# @param child_type [String|Nil] @params[:type]
def collection_param_class(base_type, child_type)
  klass = collection_class(base_type)

  # lookup children for a given type
  param_klass = child_type&.safe_constantize ## <== unprotected class
  if param_klass && param_klass < klass      ## <== verifying unprotected class
    param_klass
  else
    klass
  end
end
# only the prote
```

The code didn't really change, but the new code made it more obvious that `param_klass` wasn't being used but `klass` is the important one. And it shows that there is a check on the class.

kbrock:
> there already are resources that handle a type parameter, such as providers, vms, etc. I thought the code for those already would have a helper method and you wouldn't have to reimplement one as in this PR. Can you check if one exists for those already and reuse it?
> -- @fryguy

This is for the list operator. The controllers do not override `index`, `collection_search`, or `find_collection`.

I did not find any use of `type` for filtering. We do use `class_by_ems` to find an ems specific type for resources, but pretty sure that is the only type like filtering that we use. 
tangent: we do reference `type` in `edit_resource`, but that is not `index`, where this PR focuses. And it is not applicable for `index`.

Just to make sure we are clear, This current implementation introduces `type` filtering for all list operations/`index`.

app/controllers/api/base_controller.rb
```diff
@@ -45,7 +45,7 @@ class BaseController < ActionController::API
     rescue_from(UnsupportedMediaTypeError)      { |e| api_error(:unsupported_media_type, e) }
 
     def index
-      klass = collection_class(@req.subject)
+      klass = collection_param_class(@req.subject, params[:type])
```
kbrock, line 48: call question: will params[:type] end up in the query?

(this may cancel the return values out)

kbrock:
please check out `collection_class` concept

Fryguy:
After discussion with @kbrock I said the wrong thing which led to confusion.  There is a concept called `collection_class` that allows scoping the calls to a particular subclass of the collection.  Separately users can pass `type` to the filter (which gets sent to the where clause), but that's not the one I was thinking of and I conflated the two.

kbrock:
@Fryguy please view the changed test.

The fact that this is greens suggests to me that `collection_class` meets our needs, as is.

Do we just want to change the `type` filter?

```
s/filter[]="type=(.*)"/collection_class=\1/
```

Fryguy:
@kbrock Looks good.  Can you update the OP on this PR since it's still referring to the previous code before the switch to just the test.

PR #1257, opened 2024-04-05
Use standard SupportsFeatureMixin

kbrock:
I'm reworking the way stub_supports works.
These two examples are not working well (and don't look valid to me)

stub_supports(klass, :feature)

- klass extends SupportsFeatureMixin
- pass the klass and not a symbol


agrare:
@kbrock there is another `stub_supports_not(:cloud_volume)` down on line https://github.com/ManageIQ/manageiq-api/pull/1257/files#diff-a3c50450d3d70a77c2bddd76092b5ba015fba5015545be02eea66f7850ba52f1R284

kbrock:
update:
- added one missing `stub_supports` change (good 👀 @agrare )

WIP:
- converting hardcoded classes in stubs to `FactoryBot` (again, good idea @agrare )

kbrock:
UN-WIP: changed `stub_supports` to not hardcode classes. (in the process, some of them changed from generic to specific classes)

PR #1259, opened 2024-04-16
Use pluggable credential types method call

nasark:
Originally credential types were hard coded in the `Authentication` class. With this change we can define credential types in the respective classes and then pull the info for the OPTIONS call from the new `Authentications.credential_types` method

Needed for:
- [ ] https://github.com/ManageIQ/manageiq/pull/22995

@miq-bot assign @agrare 
@miq-bot add_reviewer @agrare 
@miq-bot add_labels enhancement

### DIFF
--- a/spec/requests/authentications_spec.rb
+++ b/spec/requests/authentications_spec.rb
@@ -465,21 +465,26 @@ RSpec.describe 'Authentications API' do
       options(api_authentications_url)
 
       additional_options = {
-        'credential_types' => build_credential_options
+        "credential_types" => credential_types
       }
       expect_options_results(:authentications, additional_options)
     end
   end
 
-  def build_credential_options
-    Authentication::CREDENTIAL_TYPES.each_with_object({}) do |(description, klass), hash|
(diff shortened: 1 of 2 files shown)


spec/requests/authentications_spec.rb
```diff
@@ -465,21 +465,26 @@
...
+    credential_subclasses.each_with_object({}) do |klass, credential_hash|
+      unless credential_hash.key?(klass.credential_type.to_sym)
+        credential_hash[klass.credential_type.to_sym] = {}
+      end
+
+      if defined? klass::API_OPTIONS
+        klass::API_OPTIONS.tap do |options|
           options[:attributes].each do |_k, val|
             val[:type] = val[:type].to_s if val && val[:type]
           end
```
nasark, line 486: This was needed in the specs because the nested hashes in the response contain string keys

nasark:
@miq-bot cross-repo-test https://github.com/ManageIQ/manageiq/pull/22995, https://github.com/ManageIQ/manageiq-providers-workflows/pull/77, https://github.com/ManageIQ/manageiq-providers-embedded_terraform/pull/13

PR #1263, opened 2024-07-09
 Rails7 fixes

jrafanie:
These are backwards compatible changes to get rails 7 to work on this repository.

- [x] [Cross repo test with rails 7 core branch](https://github.com/ManageIQ/manageiq-cross_repo-tests/pull/883)

### DIFF
--- a/spec/requests/requests_spec.rb
+++ b/spec/requests/requests_spec.rb
@@ -263,14 +263,12 @@ RSpec.describe "Requests API" do
       request.add_tag(t.name, t.children.first.name)
 
       api_basic_authorize action_identifier(:requests, :read, :resource_actions, :get)
-      get api_request_url(nil, request), :params => { :attributes => "workflow,v_allowed_tags,v_workflow_class" }
+      get api_request_url(nil, request), :params => {:attributes => "workflow,v_allowed_tags"}
 
       expected_response = a_hash_including(
-        "id"               => request.id.to_s,
-        "workflow"         => a_hash_including("values"),
-        "v_allowed_tags"   => [a_hash_including("children")],
-        "v_workflow_class" => a_hash_including(
-          "instance_logger" => a_hash_including("klass" => request.workflow.class.to_s))
+        "id"             => request.id.to_s,
+        "workflow"       => a_hash_including("values"),
+        "v_allowed_tags" => [a_hash_including("children")]
       )
 
       expect(response.parsed_body).to match(expected_response)
@@ -314,11 +312,10 @@ RSpec.describe "Requests API" do
                                    :source_type => vm_template.class.name)
 
       api_basic_authorize action_identifier(:requests, :read, :resource_actions, :get)
-      get api_request_url(nil, request), :params => { :attributes => "workflow,v_allowed_tags,v_workflow_class" }
+      get api_request_url(nil, request), :params => {:attributes => "workflow,v_allowed_tags"}
 
       expected_response = a_hash_including(
-        "id"               => request.id.to_s,
-        "v_workflow_class" => {}
+        "id" => request.id.to_s
       )
 
       expect(response.parsed_body).to match(expected_response)
--- a/spec/requests/servers_spec.rb
+++ b/spec/requests/servers_spec.rb
@@ -110,7 +110,7 @@ RSpec.describe "Servers" do
       server = FactoryBot.create(:miq_server, :status => 'started')
 
       expect { post(api_server_url(nil, server), :params => gen_request(:delete)) }.to change(MiqServer, :count).by(0)
-      expect_single_action_result(:success => false, :message => 'Failed to destroy the record')
+      expect_single_action_result(:success => false, :message => 'Failed to destroy')
     end
 
     it "can delete a server with DELETE if the server is deletable" do


app/controllers/api/base_controller/normalizer.rb
```diff
@@ -56,6 +56,10 @@ def normalize_attr(attr, value)
           Float::INFINITY.to_s
         elsif Api.resource_attribute?(attr)
           normalize_resource(value)
+        elsif value == NilClass
+          {}
```
jrafanie, line 60: @kbrock do you have any idea why this was needed?  The test below was expecting a nil v_workflow_class to come back as `{}`. Not a string, not json `null`, but `{}`.

spec/requests/requests_spec.rb
```diff
@@ -269,8 +269,7 @@
         "id"               => request.id.to_s,
         "workflow"         => a_hash_including("values"),
         "v_allowed_tags"   => [a_hash_including("children")],
-        "v_workflow_class" => a_hash_including(
-          "instance_logger" => a_hash_including("klass" => request.workflow.class.to_s))
```
jrafanie, line 273: I have no idea why this passed on rails 6.1 🤣

spec/requests/servers_spec.rb
```diff
@@ -110,7 +110,7 @@
       server = FactoryBot.create(:miq_server, :status => 'started')
 
       expect { post(api_server_url(nil, server), :params => gen_request(:delete)) }.to change(MiqServer, :count).by(0)
-      expect_single_action_result(:success => false, :message => 'Failed to destroy the record')
+      expect_single_action_result(:success => false, :message => 'Failed to destroy')
```
jrafanie, line 113: Rails change the method here to add more record information so it's easier to use a simpler string check to satisfy this on rails 6.1 and 7.

spec/requests/requests_spec.rb
jrafanie: FYI, instance_logger is an instance variable coming from the Vmdb::Logging stuff.  Not sure why this was coming back previously.  Maybe json dump of the class object included the ivars.  That's why I'm changing it to return the Class#name as it's weird to store the Class object.

Fryguy: where is v_workflow_class defined?

jrafanie: https://github.com/ManageIQ/manageiq/blob/master/app/models/miq_request.rb#L51

jrafanie: The rescue nil in the [class method](https://github.com/ManageIQ/manageiq/blob/master/app/models/miq_request.rb#L339-L347) drives me mad and I think it's related to the NilClass `{}` above.  I really have no idea how any of this stuff works but this `instance_logger` in the test was completely wrong and thankfully rails 7 detected it.  Not sure why it worked in 6.1

kbrock: interestingly, I could only find `v_workflow_class` referenced in this spec. Not sure where else it is used. It is declared as a `String` and should not be a hash.

kbrock: Wow, I had so much trouble tracking down where this method was defined `delegate` is not very `grep` friendly. It was added in https://github.com/ManageIQ/manageiq/pull/13441

```ruby
  virtual_column  :v_workflow_class,     :type => :string,   :uses => :workflow
  delegate :class,                       :to => :workflow,   :prefix => :v_workflow
```

Interestingly, there is a `workflow_class` method that we probably should be using instead.
But more importantly, it is declared to return a `String` (which it probably should be), but it returns a `Class`. It is possible that our serialization of a `Class` may have changed.

It appears that the example is not properly creating a workflow, so `nil.class` is returning a `NilClass`.

---

I put a `byebug` into the class but getting all sorts of errors:

These all work in `rails c` for me, but fail when I'm running this specific spec.

```ruby
(byebug) MiqRequest.new.description
*** NameError Exception: uninitialized constant MiqRequest::TASK_DESCRIPTION

  default_value_for(:message)       { |r| "#{r.class::TASK_DESCRIPTION} - Request Created" }
(byebug) request.class
MiqProvisionRequest
(byebug) MiqProvisionRequest.new.description
nil
(byebug) request.v_workflow_class
*** NameError Exception: uninitialized constant MiqRequestWorkflow::DialogFieldValidation

  include DialogFieldValidation
          ^^^^^^^^^^^^^^^^^^^^^

(byebug) require "miq_request_workflow/dialog_field_validation"
(byebug) MiqProvisionRequest.new.v_workflow_class
(byebug) MiqProvisionRequest.new.workflow_class
^^^ all give the same error
```

I tried redefining the `DialogFieldValidation` to the long style (`module MiqProvisionWorkflow; module DialogFieldValidator; end; end` instead of `module parent::DialogFieldValidator`) but that did not help.

This is a tangent, but this feels a little chicken and egg.
If you want to define the child module, define the parent class (which references the child module)

kbrock: update:

The MiqProvisionRequest.workflow_class references a class, but `constantize` throws that exception listed above. Match that with a `rescue nil` and we get a `nil` coming out. `nil.class == NilClass`

jrafanie: yeah, is the test and code busted or just the test?  I can try to create a proper workflow to fix the test.

kbrock: Thnx Jason: once I updated all my repos, I was able to load this just fine.

One thing about the code, it compares the `workflow.class` and the `workflow_class`. So if it gets busted (goes from a value to nil), then it will always pass the test.

kbrock:
While I think we do want to understand why `NilClass` is not working the same as all other classes, and previous rails `NilClass` behavior, let us remember that this method under test is bogus and needs to be removed.

It is not referenced (don't think it ever has been), and I find it hard to believe that we want a method that returns an obscure hash. My bet is we wanted the name of the workflow class. TBH/ not sure why the front end would even need that.

jrafanie:
> While I think we do want to understand why `NilClass` is not working the same as all other classes, and previous rails `NilClass` behavior, let us remember that this method under test is bogus and needs to be removed.
> 
> It is not referenced (don't think it ever has been), and I find it hard to believe that we want a method that returns an obscure hash. My bet is we wanted the name of the workflow class. TBH/ not sure why the front end would even need that.

Ok, I'll drop the components of the tests that are checking `v_workflow_class`.  I agree that we probably want the class name... so I think the first hunk in the patch below seems like the correct thing to do... but I'm fine with doing it if we need to do it.

```diff
diff --git a/app/controllers/api/base_controller/normalizer.rb b/app/controllers/api/base_controller/normalizer.rb
index f2585e2c..ddefee7d 100644
--- a/app/controllers/api/base_controller/normalizer.rb
+++ b/app/controllers/api/base_controller/normalizer.rb
@@ -56,6 +56,10 @@ module Api
           Float::INFINITY.to_s
         elsif Api.resource_attribute?(attr)
           normalize_resource(value)
+        elsif value == NilClass
+          {}
+        elsif value.kind_of?(Class)
+          value.name
         else
           value
         end
diff --git a/spec/requests/requests_spec.rb b/spec/requests/requests_spec.rb
index f9a20d49..9111cc7c 100644
--- a/spec/requests/requests_spec.rb
+++ b/spec/requests/requests_spec.rb
@@ -269,8 +269,7 @@ RSpec.describe "Requests API" do
         "id"               => request.id.to_s,
         "workflow"         => a_hash_including("values"),
         "v_allowed_tags"   => [a_hash_including("children")],
-        "v_workflow_class" => a_hash_including(
-          "instance_logger" => a_hash_including("klass" => request.workflow.class.to_s))
+        "v_workflow_class" => request.workflow.class.to_s
       )

       expect(response.parsed_body).to match(expected_response)
@@ -318,7 +317,7 @@ RSpec.describe "Requests API" do

       expected_response = a_hash_including(
         "id"               => request.id.to_s,
-        "v_workflow_class" => {}
+        "v_workflow_class" => {} # Need to figure out why NilClass comes back as {} previously
       )

       expect(response.parsed_body).to match(expected_response)
```

jrafanie:
@miq-bot cross-repo-tests https://github.com/ManageIQ/manageiq/pull/22873

jrafanie:
Ok, I think I resolved all the issues.  I've removed the bogus test areas that asserted values for v_workflow_class as that's not functioning "correctly" and we don't have good reasons to change the bogus behavior, so we'll tackle that later if needed.

spec/requests/requests_spec.rb
jrafanie: I'm punting on this (see below).  We'll tackle this if we determine the correct behavior it should have.  Right now, it doesn't make sense why it's ok on rails 6.1 when it's clearly wrong.  It's possible no one is using it and not reported a bug.  Either way, we can handle it when it's a problem.

kbrock:
Glad you dropped this rather than trying to track down our serialization of classes (which was probably a non-implemented feature anyway)

PR #1265, opened 2024-08-15
Query metric rollups using start_date derived from the first rollup

jrafanie:
This resolves a time based sporadic failure.  This means it works for many hours of the day but fails with specific ones.

Previously, `Time.zone` is UTC, so `Time.zone.today` when run at August 15th at 12:05 AM UTC would return `Thu, 15 Aug 2024`.

If we created the rollups using `1.hour.ago`, the first one would be August 14th 11:05 PM.

If we our rollups were on August 14th and we're looking for August 15th hourly rollups, we find nothing.

This change ensures both the timestamp on the rollups and the start_date we query the rollups with come from the same time.

### DIFF
--- a/spec/requests/providers_spec.rb
+++ b/spec/requests/providers_spec.rb
@@ -1457,9 +1457,10 @@ describe "Providers API" do
     let(:provider) { FactoryBot.create(:ems_redhat) }
     let(:provider1) { FactoryBot.create(:ems_redhat) }
     let(:url) { api_provider_metric_rollups_url(nil, provider) }
+    let(:first_ts) { 1.hour.ago }
 
     before do
-      FactoryBot.create_list(:metric_rollup_cm_hr, 3, :resource => provider, :timestamp => 1.hour.ago)
+      FactoryBot.create_list(:metric_rollup_cm_hr, 3, :resource => provider, :timestamp => first_ts)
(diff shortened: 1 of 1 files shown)


Fryguy (approved):

Fryguy:
Cool I thought this was gonna involve some timecop shenanigans

Fryguy:
So is this technically a Rails 7 issue?  That is you said Time.zone now returns UTC, so did it not before?

jrafanie:
> So is this technically a Rails 7 issue? That is you said Time.zone previously returns UTC, so does it not now?

Sorry, "previously" meant "before this commit".  Our schedule runs for CI are at 12AM UTC so 1.hour.ago is the prior day.

It looks like it's been failing via scheduled but working on merges or manual runs because it's not in that 12-1AM UTC date range.

image

Fryguy:
Backported to `radjabov` in commit 28df3ee4464fc32fc5294ef728f2d89cd3292046.

```text
commit 28df3ee4464fc32fc5294ef728f2d89cd3292046
Date:   Fri Aug 16 09:19:56 2024 -0400

    Merge pull request #1265 from jrafanie/fix_sporadic_test_failures_due_to_date_wrapping
    
    Query metric rollups using start_date derived from the first rollup
    
    (cherry picked from commit c46f88165299e31dd7b73f3d94f90dec339c5b1b)
```

PR #1278, opened 2025-01-30
Expose attribute aliases in list of attributes

kbrock:
Blocked:
- [x] https://github.com/ManageIQ/manageiq/pull/23324

This enables fixes for 7.1
But this does not depend upon 7.1 and even in 7.0 is a more accurate implementation

### DIFF
--- a/app/controllers/api/pictures_controller.rb
+++ b/app/controllers/api/pictures_controller.rb
@@ -21,7 +21,7 @@ module Api
       if !@req.attributes.empty? || @additional_attributes
         @req.attributes | Array(@additional_attributes) | ID_ATTRS
       else
-        Picture.attribute_names - %w(content)
+        Picture.attribute_names | Picture.try(:attribute_aliases)&.keys - %w(content)
       end
     end
   end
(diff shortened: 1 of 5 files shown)


app/controllers/api/pictures_controller.rb
```diff
@@ -21,7 +21,7 @@ def attribute_selection
       if !@req.attributes.empty? || @additional_attributes
         @req.attributes | Array(@additional_attributes) | ID_ATTRS
       else
-        Picture.attribute_names - %w(content)
+        Picture.attribute_names | Picture.try(:attribute_aliases)&.keys - %w(content)
```
jrafanie, line 24: Same comment as https://github.com/ManageIQ/manageiq-automation_engine/pull/565/files#r1936234812

Fryguy:
I made a comment in the other PR, but I don't understand this.  Does this not return the same thing in Rails 7.1 as in 7.0? In 7.0 the `attribute_names` already has everything.

kbrock:
WIP: introducing more generic method for the concept of attributes and attribute aliases. After that is done, will change this code to leverage that.

kbrock:
update:
- use all_attribute_ames

un-WIP

jrafanie (approved):
LGTM

PR #1280, opened 2025-02-10
Test each subhash for included attributes, not subhash equality

jrafanie:
rails 7.1 adds id_value attribute alias, see:
https://www.github.com/rails/rails/pull/48930

We only care about the subhash includes some attributes, not that the subhashes match all attributes.


jrafanie:
Kicking tests

PR #1293, opened 2025-08-04
Fix API counts to honor RBAC

Fryguy:
The original idea behind the full count was to show that there are extra records that could be returned if different filters were applied. However, this does not need to apply to RBAC as users do not need to know about these records.

@kbrock Please review.

This does not solve https://github.com/ManageIQ/manageiq-ui-classic/issues/9448 directly, but is required as a first step to fix it.  It can be merged independently.


Fryguy:
@miq-bot cross-repo-test /all

kbrock (approved):
We really don't want `klass.count` - great find.

Of note, this looks like it is adding a query, but it is just moving it up

PR #1298, opened 2025-09-17
Snapshot Operations should queue_object_action with role=ems_operations

agrare:
Snapshot operations were being queued for the generic role when they should be queued on the ems_operations role

$ curl --user admin:smartvm -X POST -H "Content-Type: application/json" http://localhost:3000/api/vms/281/snapshots/3 -d '{"action": "delete"}'

image

Fixes https://github.com/ManageIQ/manageiq/issues/23588

### DIFF
--- a/app/controllers/api/subcollections/snapshots.rb
+++ b/app/controllers/api/subcollections/snapshots.rb
@@ -14,6 +14,7 @@ module Api
           parent,
           message,
           :method_name => "create_snapshot",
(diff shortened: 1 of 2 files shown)


Fryguy (approved):

agrare:
It looks like most ems_operations go through `enqueue_ems_action` which takes `type, id` instead of a parent object which is maybe why these were missed being part of a subcollection.

PR #1301, opened 2025-10-29
Remove unpermitted parameters warning

kbrock:
pulled out of #1300 - it is unrelated

We always pass a few parameters. these parameters come from query_parameters.
These were triggering a warning in our logging.

Rails adds parameters from routes and actionpack. like :c_id, :s_id, :action, :controller, :format

We only care about a few of these. So only outputting the ones we care about and ignoring the rest

```
[----] D, [2025-10-28T23:00:25.289277#13035:5200] DEBUG -- : Unpermitted parameters: :expand, :attributes, :c_id. [...]
```

/cc @jrafanie not sure if you have an opinion on this

### DIFF
--- a/app/controllers/api/base_controller/logger.rb
+++ b/app/controllers/api/base_controller/logger.rb
@@ -33,7 +33,7 @@ module Api
         return unless api_log_info?
         log_request("Request", @req.to_hash)
         unfiltered_params = request.query_parameters
-                                   .merge(params.permit(:action, :controller, :format).to_h)
+                                   .merge(params.slice(:action, :controller, :format).permit!)
                                    .merge("body" => @req.json_body)
         log_request("Parameters", @parameter_filter.filter(unfiltered_params))
         log_request_body


Fryguy (commented):
Not a fan of permitting everything and then slicing... Kind of goes against the right way to use action params. I preferred the original way.

That being said I thought params had a form of slice without the except 🤔

kbrock:
@Fryguy Yea, I don't like this either.

## Original code

```ruby
params.permit(:action, :controller, :format).to_h
```

Reads: Give me all parameters, and complain if there are any parameters besides `action`, `controller`, `format`.

Since there is almost always `c_id`, `s_id`. `format`, `offset`, `expand`, and `attributes`, it will always complain.

## I wanted

Please give me `action`, `controller`, and `format` from `ActionController` and `routes.rb`.

```ruby
params.slice(:action, :controller, :format)
```

This was an error.

## Closest I could come

```ruby
params.permit!.to_h.slice(:action, :controller, :format)
```

Punt on strong parameters. Just give me these 3 parameters..

app/controllers/api/base_controller/logger.rb
```diff
@@ -33,7 +33,7 @@ def log_api_request
         return unless api_log_info?
         log_request("Request", @req.to_hash)
         unfiltered_params = request.query_parameters
-                                   .merge(params.permit(:action, :controller, :format).to_h)
+                                   .merge(params.permit!.to_h.slice(:action, :controller, :format))
```
Fryguy: Think you can do


```suggestion
                                   .merge(params.slice(:action, :controller, :format).permit!)
```

Fryguy (approved):

kbrock:
@Fryguy So Post requests go through this. (Think Get requests do not - which is odd)
Most tests go through this code since it is called as a `before_action :log_api_request`, so introducing an error causes a bunch of specs to fail. So at least we have protection there.

I can't find a good way to test this.
Spent a day trying to add a spec and failing. Just can't find a way to green-red-green this.
Lets call this complete unless you have an idea how to test.

Fryguy:
as long as tests would fail if an error was in there then I think we're covered. I was concerned specifically with the `return unless api_log_info?` line breaking out of the method early in test and thus this code not getting exercised. But if you says it's covered, then I'm good with it.

PR #1310, opened 2026-02-18
Fix deprecation in specs

Fryguy:
Fixes the deprecation warning:

```
DEPRECATION WARNING: Skipping over leading brackets in parameter name "[\"api\",\"authentication_timeout\"]" is deprecated and will parse differently in Rails 8.1 or Rack 3.0. (called from call at /Users/jfrey/dev/manageiq/lib/request_started_on_middleware.rb:12)
```

Rails 8.0 can process the body of the delete properly now if told to process it `:as => :json`.

@jrafanie Please review.


jrafanie (approved):
😍

PR #1313, opened 2026-03-22
Support Host PowerOn/PowerOff and VM Delete Operation

shashank-gowda:
API support for Host Power On/Off and VM Delete operations

### DIFF
--- a/app/controllers/api/vms_controller.rb
+++ b/app/controllers/api/vms_controller.rb
@@ -54,6 +54,10 @@ module Api
       enqueue_ems_action(type, id, "Deleting", :method_name => "destroy")
     end
 
+    def terminate_resource(type, id = nil, _data = nil)
+      enqueue_ems_action(type, id, "Terminating", :method_name => "vm_destroy", :supports => true)
+    end
+
     def set_owner_resource(type, id = nil, data = nil)
       api_resource(type, id, "Setting owner of") do |vm|
         owner = data.blank? ? "" : data["owner"].strip
--- a/spec/requests/hosts_spec.rb
+++ b/spec/requests/hosts_spec.rb
@@ -1,4 +1,17 @@
 RSpec.describe "hosts API" do
+  include Spec::Support::SupportsHelper
+
+  let(:zone) { FactoryBot.create(:zone, :name => "api_zone") }
+  let(:ems) { FactoryBot.create(:ems_vmware, :zone => zone) }
+  let(:power_state) { "unknown" }
+  let(:host) { FactoryBot.create(:host, :ext_management_system => ems, :power_state => power_state) }
+  let(:host1) { FactoryBot.create(:host, :ext_management_system => ems) }
+  let(:host2) { FactoryBot.create(:host, :ext_management_system => ems) }
+  let(:host_url) { api_host_url(nil, host) }
+  let(:host1_url) { api_host_url(nil, host1) }
+  let(:host2_url) { api_host_url(nil, host2) }
+  let(:invalid_host_url) { api_host_url(nil, ApplicationRecord.id_in_region(999_999, ApplicationRecord.my_region_number)) }
+
   describe "editing a host's password" do
     context "with an appropriate role" do
       it "can edit the password on a host using new/react format" do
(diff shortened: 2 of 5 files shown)


agrare:
Hey @shashank-gowda can you squash all of the "Fix * Issues" commits into the first commit?

My philosophy, if multiple commits are helping to show the progress of what was done in a PR in a way that would be helpful for someone to understand later on then I keep them as separate commits.  If I'm just fixing something minor that doesn't actually show the progression of work (aka small test failures, rubocop warnings, syntax errors, etc... that were introduced in the same PR) then I amend/squash

spec/requests/hosts_spec.rb
```diff
@@ -266,5 +282,110 @@
     end
   end
 
+  describe "Host start action" do
+    it "starts an invalid host" do
```
agrare: The name of these tests are a little misleading, I would expect something like: `"starting an invalid host fails with Not Found"`

spec/requests/hosts_spec.rb
```diff
@@ -266,5 +282,110 @@
...
+      api_basic_authorize action_identifier(:hosts, :start)
+
+      post(invalid_host_url, :params => gen_request(:start))
+
+      expect(response).to have_http_status(:not_found)
+    end
+
+    it "starts an invalid host without appropriate role" do
```
agrare: Same here, `"starting a host without the appropriate role fails with Forbidden"`
Also, best to use a valid host so the only reason this would fail is because of the role

spec/requests/hosts_spec.rb
```diff
@@ -266,5 +282,110 @@
...
+      api_basic_authorize
+
+      post(invalid_host_url, :params => gen_request(:start))
+
+      expect(response).to have_http_status(:forbidden)
+    end
+
+    it "starts a powered on host" do
+      api_basic_authorize action_identifier(:hosts, :start)
+      update_power_state("on", host)
```
agrare: It is better to do this with a context and `let(:power_state)` instead of updating created records.  When running tests in parallel this is one way that sporadic test failures can be introduced.

If you do this above
```
let(:host) { FactoryBot.create(:host, :ext_management_system => ems, :power_state => power_state) }
let(:power_state) { "on" }
```

Then here you can do:
```
context "with a powered on host"
  let(:power_state) { "on" }

  it "fails due to invalid power state" do
    ...
```

app/controllers/api/vms_controller.rb
```diff
@@ -54,6 +54,10 @@ def delete_resource(type, id, _data = nil)
       enqueue_ems_action(type, id, "Deleting", :method_name => "destroy")
     end
 
+    def terminate_resource(type, id = nil, _data = nil)
+      enqueue_ems_action(type, id, "Terminating", :method_name => "vm_destroy", :supports => true)
+    end
```
agrare, line 59: Why call the action terminate and not destroy since that matches the name on the backend?

shashank-gowda: @agrare The existing delete_resource action maps to the backend destroy method. To avoid confusion between the two different operations, I used the term terminate for the action that invokes vm_destroy.

spec/requests/hosts_spec.rb
```diff
@@ -1,4 +1,16 @@
...
+  include Spec::Support::SupportsHelper
+
+  let(:zone) { FactoryBot.create(:zone, :name => "api_zone") }
+  let(:ems) { FactoryBot.create(:ems_vmware, :zone => zone) }
+  let(:host) { FactoryBot.create(:host, :ext_management_system => ems) }
+  let(:host1) { FactoryBot.create(:host, :ext_management_system => ems) }
+  let(:host2) { FactoryBot.create(:host, :ext_management_system => ems) }
+  let(:host_url) { api_host_url(nil, host) }
+  let(:host1_url) { api_host_url(nil, host1) }
+  let(:host2_url) { api_host_url(nil, host2) }
```
agrare, line 12: I don't see these used any longer they can be deleted

spec/requests/hosts_spec.rb
```diff
@@ -1,4 +1,16 @@
 RSpec.describe "hosts API" do
+  include Spec::Support::SupportsHelper
+
+  let(:zone) { FactoryBot.create(:zone, :name => "api_zone") }
+  let(:ems) { FactoryBot.create(:ems_vmware, :zone => zone) }
+  let(:host) { FactoryBot.create(:host, :ext_management_system => ems) }
+  let(:host1) { FactoryBot.create(:host, :ext_management_system => ems) }
+  let(:host2) { FactoryBot.create(:host, :ext_management_system => ems) }
```
agrare, line 9: And then we don't need these either

spec/requests/hosts_spec.rb
```diff
@@ -266,5 +278,131 @@
     end
   end
 
+  describe "Host start action" do
+    let(:host) { FactoryBot.create(:host, :ext_management_system => ems, :power_state => power_state) }
```
agrare: If you updated the `let(:host)` up on [[line 6]](https://github.com/ManageIQ/manageiq-api/pull/1313/changes#diff-3d401e551f3cbc5dfe8b3ae769b2dd7cbc1371be8f8c2a2a3c691a01fd5e211eR6) to include the `:power_state` then you wouldn't need to redefine this here, you would only need to set the `let(:power_state)` in the context (you'll need a default power_state above also)

spec/requests/hosts_spec.rb
```diff
@@ -266,5 +278,131 @@
...
+        expect(MiqQueue.where(:method_name => "start",
+                              :user_id     => @user.id,
+                              :group_id    => @user.current_group.id,
+                              :tenant_id   => @user.current_tenant.id).count).to eq(1)
+      end
+
+      it "starts multiple hosts" do
+        api_basic_authorize collection_action_identifier(:hosts, :start)
+        host1_with_power = FactoryBot.create(:host, :ext_management_system => ems, :power_state => "off")
+        host2_with_power = FactoryBot.create(:host, :ext_management_system => ems, :power_state => "off")
```
agrare: Why redeclare these here and [[here]](https://github.com/ManageIQ/manageiq-api/pull/1313/changes#diff-3d401e551f3cbc5dfe8b3ae769b2dd7cbc1371be8f8c2a2a3c691a01fd5e211eR394-R395) instead of in the top level describe like you had before?

agrare:
@shashank-gowda can you add a PR to update the API documentation to add these new endpoints? https://github.com/ManageIQ/manageiq-documentation/tree/master/api

agrare (approved):

agrare:
@shashank-gowda have you added these endpoints to the API documentation?

PR #1314, opened 2026-03-25
Sort provider subclasses alphabetically

agrare:
This list is shown to the user to select a provider type and it is extremely confusing when it is not sorted alphabetically.

Before:
image

After:
image

### DIFF
--- a/app/controllers/api/providers_controller.rb
+++ b/app/controllers/api/providers_controller.rb
@@ -154,7 +154,7 @@ module Api
         po.merge(ems.ems_type => ems.options_description)
       end
 
-      supported_providers = supported_types_for_create.map do |klass|
+      supported_providers = supported_types_for_create.sort_by(&:description).map do |klass|
         if klass.supports?(:regions)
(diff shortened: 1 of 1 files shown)


agrare:
@Fryguy  should the UI be the one doing this sorting?  Don't we need to take the user's locale into consideration for this?

Fryguy:
That's a good question - I'm not sure if provider names are translated. However, it's possible they _could_ be so I agree with you that we should do it client side.

Fryguy:
Since we sort the regions here also, I'll merge anyway.

Fryguy:
Backported to `tal` in commit f113e064948bbef049316be556deb42ca4f6fb85.

```text
commit f113e064948bbef049316be556deb42ca4f6fb85
Date:   Wed Mar 25 09:13:01 2026 -0400

    Merge pull request #1314 from agrare/sort_provider_subclasses_by_description
    
    Sort provider subclasses alphabetically
    
    (cherry picked from commit b0f32fd9e5603440b86852d1e4aaadef879087c5)
```

PR #1319, opened 2026-04-29
Add Paths and Parameters to OpenAPI spec generation

agrare:


.github/workflows/openapi-generate.yml
```diff
@@ -0,0 +1,82 @@
...
+    - cron: '0 0 * * 0'
+  workflow_dispatch:
+    # Allow manual trigger
+  push:
+    branches:
+      - master
+    paths:
+      - 'config/api.yml'
+      - 'lib/manageiq/api/open_api/**'
+      - 'app/controllers/api/**'
```
Fryguy: Formatting for consistency across our repos.


```suggestion
  push:
    branches:
    - master
    paths:
    - app/controllers/api/**
    - config/api.yml
    - lib/manageiq/api/open_api/**
  schedule:
  - cron: 0 0 * * 0
  workflow_dispatch:
```

.github/workflows/openapi-generate.yml
```diff
@@ -0,0 +1,82 @@
...
+permissions:
```
Fryguy, line 17: Include concurrency


```suggestion
concurrency:
  group: "${{ github.workflow }}-${{ github.ref }}"
  cancel-in-progress: true
permissions:
```

.github/workflows/openapi-generate.yml
```diff
@@ -0,0 +1,82 @@
...
+  contents: read
+jobs:
+  generate-openapi:
```
Fryguy: Minor, but should this be openapi-generate (matches the file name)?

.github/workflows/openapi-generate.yml
```diff
@@ -0,0 +1,82 @@
...
+    services:
+      postgres:
+        image: manageiq/postgresql:13
+        env:
+          POSTGRESQL_USER: root
+          POSTGRESQL_PASSWORD: smartvm
+          POSTGRESQL_DATABASE: vmdb_test
+        options: "--health-cmd pg_isready --health-interval 2s --health-timeout 5s --health-retries 5"
+        ports:
+          - 5432:5432
```
Fryguy:
```suggestion
        ports:
        - 5432:5432
```

.github/workflows/openapi-generate.yml
```diff
@@ -0,0 +1,82 @@
...
+    env:
+      PGHOST: localhost
+      PGPASSWORD: smartvm
+      RAILS_ENV: test
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v4
```
Fryguy:
```suggestion
      uses: actions/checkout@v6
```

.github/workflows/openapi-generate.yml
```diff
@@ -0,0 +1,82 @@
...
+    - name: Set up Ruby
```
Fryguy, line 41: Do we need before_install?


```suggestion
    - name: Set up system
      run: bin/before_install
    - name: Set up Ruby
```

.github/workflows/openapi-generate.yml
```diff
@@ -0,0 +1,82 @@
...
+        bundler-cache: true
+      timeout-minutes: 30
+    - name: Prepare tests
+      run: bin/setup
+    - name: Generate OpenAPI specification
+      run: bundle exec rake manageiq:api:openapi_generate
+    - name: Create Pull Request
+      uses: peter-evans/create-pull-request@v8
+      with:
+        token: ${{ secrets.GITHUB_TOKEN }}
```
Fryguy, line 54: Reminder to self that I need to associate this token with this repo after merge.

.github/workflows/openapi-generate.yml
```diff
@@ -0,0 +1,82 @@
...
+          This PR updates the OpenAPI specification (`config/openapi.json`) based on the current API configuration.
+          
+          ### Changes
+          - Regenerated OpenAPI spec from `config/api.yml`
+          - Updated schemas, paths, and operations
+          
+          ### Generated by
+          - Workflow: `${{ github.workflow }}`
+          - Run: `${{ github.run_number }}`
+          - Triggered by: `${{ github.event_name }}`
```
Fryguy: This is a bit wordy - In the [locale file](https://github.com/ManageIQ/manageiq/blob/master/.github/workflows/locale_update_all.yaml#L74C104-L74C215) I used this instead, which is a direct link

```
${{ github.workflow_ref }}\n${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}
```

So I can see a reference back to the Action run that generated it, but maybe at the top

.github/workflows/openapi-generate.yml
```diff
@@ -0,0 +1,82 @@
...
+          
+          ### Review Checklist
+          - [ ] Verify new/modified endpoints are documented correctly
+          - [ ] Check parameter definitions are accurate
+          - [ ] Ensure response schemas match actual API responses
+          - [ ] Validate security schemes are properly defined
+          
+          ---
+          *This PR was automatically generated by GitHub Actions*
```
Fryguy, line 78: This part is kind of cool - wondering if we should add that to the other ones, perhaps along with a reference back to the action that generated it.  Maybe the "generated by" here can _be_ the link.

.github/workflows/openapi-generate.yml
```diff
@@ -0,0 +1,82 @@
...
+        branch: openapi_generate
+        delete-branch: true
+        push-to-fork: miq-bot/kubevirt-sdk-ruby
```
Fryguy: 😉

lib/manageiq/api/open_api/generator.rb
```diff
@@ -3,6 +3,10 @@ module Api
     module OpenApi
       class Generator
         require 'json'
+        require_relative 'parameter_builder'
+        require_relative 'schema_builder'
+        require_relative 'operation_builder'
+        require_relative 'path_builder'
```
Fryguy: Can we move these to the top of the file?  They are loaded at load time anyway, and are kind of hidden here.

.github/workflows/openapi-generate.yml
Fryguy: Second reminder, I also need to fork this repo on the bot account.

lib/manageiq/api/open_api/operation_builder.rb
```diff
@@ -0,0 +1,257 @@
...
+        end
+
+        def self.build_resource_show_operation(collection_name, collection, model_schema_name)
+          {
+            "summary"     => "Get a #{collection[:description]&.singularize || collection_name.to_s.singularize}",
+            "description" => "Returns a single #{collection[:description]&.singularize || collection_name.to_s.singularize} by ID",
+            "operationId" => "get_#{collection_name.to_s.singularize}",
+            "tags"        => [collection[:description] || collection_name.to_s.titleize],
+            "parameters"  => [
+              {"$ref" => "#{PARAMETERS_PATH}/resourceId"}
```
Fryguy: I wonder if we should use something like File.join here. I'm fine merging as is, but I could see minor double-// mistakes showing up in the future.

lib/manageiq/api/open_api/schema_builder.rb
```diff
@@ -0,0 +1,274 @@
...
+                      "$ref" => "#{SCHEMAS_PATH}/Action"
+                    }
+                  }
+                }
+              }
+            ]
+          }
+        end
+
+        def self.build_standard_responses
```
Fryguy: Is this just a constant? I don't see anything changing in here (but perhaps it will in future PRs?)

Fryguy: ~Actually I think these are all constants.~ Sorry the first and last are - the ones in the middle aren't ESCROLLTOOFAST haha

lib/manageiq/api/open_api/operation_builder.rb
```diff
@@ -0,0 +1,257 @@
...
+              "content"  => {
+                "application/json" => {
+                  "schema" => {
+                    "$ref" => "#{SCHEMAS_PATH}/#{model_schema_name}"
+                  }
+                }
+              }
+            },
+            "responses"   => {
+              "201" => {
```
Fryguy: Do we actually return a 201?

lib/manageiq/api/open_api/operation_builder.rb
```diff
@@ -0,0 +1,257 @@
...
+          {
+            "summary"     => "Delete a #{collection[:description]&.singularize || collection_name.to_s.singularize}",
+            "description" => "Deletes an existing #{collection[:description]&.singularize || collection_name.to_s.singularize}",
+            "operationId" => "delete_#{collection_name.to_s.singularize}",
+            "tags"        => [collection[:description] || collection_name.to_s.titleize],
+            "parameters"  => [
+              {"$ref" => "#{PARAMETERS_PATH}/resourceId"}
+            ],
+            "responses"   => {
+              "204" => SchemaBuilder.build_standard_responses["204"],
```
Fryguy: Same here - do we issue a 204?

Fryguy:
Nice it dropped from +120k to +90k - good savings

lib/manageiq/api/open_api/operation_builder.rb
agrare: Removed :+1:

lib/manageiq/api/open_api/operation_builder.rb
agrare: Dropped

PR #1320, opened 2026-05-08
[WIP] Add support for collection attributes (attributes=vms.name)

### LINKED ISSUE #871: Add support for getting attributes on has_many associations (&attributes=vms.name)
Because we can't ask for a single attribute on a plural association and if we don't have an exposed virtual attribute, I think this would require us to bring back the whole collection proxy with all of their attributes (all vms with all of their real attributes) in order to get just one or two.

This doesn't work:
```
/api/providers?expand=resources&attributes=name,vms.name
```

Results in calling `name` on the collection proxy, resulting in the name of the collection, not an array of names for each Vm.

```
    "resources": [
        {
            "href": "http://localhost:3000/api/providers/4",
            "name": "REDACTED1",
            "id": "4",
            "vms": {
                "name": "Vm"
            }
        },
        {
            "href": "http://localhost:3000/api/providers/3",
            "name": "REDACTED2",
            "id": "3",
            "vms": {
                "name": "Vm"
            }
        },

```

Meanwhile, singular associations work:

```
/api/vms?expand=resources&attributes=name,hardware.cpu_sockets
```

```
    "resources": [
        {
            "href": "http://localhost:3000/api/vms/485",
            "name": "REDACTED3",
            "id": "485",
            "hardware": {
                "cpu_sockets": 1
            }
        },
        {
            "href": "http://localhost:3000/api/vms/338",
            "name": "REDACTED4",
            "id": "338",
            "hardware": {
                "cpu_sockets": 4
            }
        },
```

Related to #805

jrafanie:
Enables requesting specific attributes on has_many associations using dot notation.

Before: attributes=vms.name returned {"vms": {"name": "Vm"}}
After: Returns array of VMs with requested attributes plus href and id

Uses eager loading to prevent N+1 queries.

Fixes #871

### DIFF
--- a/app/controllers/api/base_controller/renderer.rb
+++ b/app/controllers/api/base_controller/renderer.rb
@@ -313,12 +313,60 @@ module Api
 
       def fetch_indirect_virtual_attribute(_type, resource, base, attr, object_hash)
         query_related_objects(base, resource, object_hash)
-        return unless attr_accessible?(object_hash[base], attr)
-        value  = virtual_attribute_search(object_hash[base], attr)
+        related_obj = object_hash[base]
+
+        if collection_association?(related_obj) && @req.collection_attributes_for(base).any?
+          fetch_collection_attributes(base, related_obj)
+        else
+          fetch_standard_virtual_attribute(related_obj, base, attr)
+        end
+      end
+
+      def fetch_standard_virtual_attribute(related_obj, base, attr)
+        return unless attr_accessible?(related_obj, attr)
+
+        value  = virtual_attribute_search(related_obj, attr)
         result = {attr => normalize_attr(attr, value)}
-        # set nil vtype above to "#{type}/#{resource.id}/#{base.tr('.', '/')}/#{attr}" to support id normalization
         base.split(".").reverse_each { |level| result = {level => result} }
         [value, result]
+      end
+
+      def fetch_collection_attributes(association_path, collection)
+        collection_type = get_collection_type(collection)
+        requested_attrs = @req.collection_attributes_for(association_path)
+        return [nil, {}] if collection_type.nil? || requested_attrs.empty?
+
+        # TODO: Only physical attributes supported. Virtual attributes on associations (vms.v_total_snapshots)
+        # would cause N+1 queries - one per associated record even with eager loading. Consider adding
+        # aggregated virtuals to primary model instead (e.g., provider.v_total_vms_snapshots).
+        attrs_to_render = (requested_attrs + ['id']).uniq
+        items = collection.map do |item|
+          normalize_hash(collection_type, item, {:render_attributes => attrs_to_render})
+        end
+        association_name = association_path.split(".").last
+        result = {association_name => items}
+        [items, result]
+      end
+
+      # Check for ActiveRecord collections (CollectionProxy, Relation) without triggering queries
+      # Both have .loaded? method, plain Arrays don't
+      def collection_association?(obj)
+        obj.respond_to?(:loaded?)
+      end
+
+      def get_collection_type(collection)
+        klass = if collection_association?(collection)
+                  collection.klass
+                elsif collection.kind_of?(Array) && collection.first
+                  collection.first.class
+                end
+
+        return nil unless klass
+
+        # Use base_model to handle STI classes
+        base_klass = klass.respond_to?(:base_model) ? klass.base_model : klass
+        collection_name = collection_config.name_for_klass(base_klass)
+        collection_name&.to_sym
       end
 
       #
@@ -606,7 +654,25 @@ module Api
       end
 
       def determine_include_for_find(klass)
-        attrs = virtual_attributes_for(klass) do |type, attr_name, attr_base|
+        attrs = determine_include_for_find_vattrs(klass)
+        collections = determine_include_for_find_collections(klass)
+
+        all_includes = (attrs || []) + collections
+
+        if all_includes.any?
+          all_includes.each_with_object({}) do |key, include_for_find|
+            if (virtual_includes = klass.virtual_includes(key))
+              ActiveRecord::Base.merge_includes(include_for_find, virtual_includes)
+            else
+              nested = include_for_find
+              key.to_s.split(".").each { |k| nested = nested[k] ||= {} }
+            end
+          end
+        end
+      end
+
+      def determine_include_for_find_vattrs(klass)
+        virtual_attributes_for(klass) do |type, attr_name, attr_base|
           if klass.virtual_includes(attr_name) && !klass.attribute_supported_by_sql?(attr_name) && attr_base.blank?
             attr_name
           else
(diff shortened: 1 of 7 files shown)


app/controllers/api/base_controller/renderer.rb
```diff
@@ -311,16 +311,62 @@ def add_custom_action_hrefs(value)
...
       end
 
+      def fetch_collection_attributes(_type, _resource, association_path, collection)
+        requested_attrs = @req.collection_attributes_for(association_path)
+
+        return [nil, {}] if requested_attrs.empty?
+
+        collection_type = get_collection_type(collection)
+
+        return [nil, {}] unless collection_type
```
Fryguy: I think this can be trimmed up


```suggestion
        requested_attrs = @req.collection_attributes_for(association_path)
        collection_type = get_collection_type(collection)
        return [nil, {}] if requested_attrs.empty? || collection_type.nil?
```

app/controllers/api/base_controller/renderer.rb
```diff
@@ -311,16 +311,62 @@ def add_custom_action_hrefs(value)
...
+        association_name = association_path.split(".").last
+        result = {association_name => items}
+        [items, result]
+      end
+
+      # Check for ActiveRecord collections (CollectionProxy, Relation) without triggering queries
+      # Both have .klass method, plain Arrays don't
+      def collection_association?(obj)
+        obj.respond_to?(:klass)
+      end
```
Fryguy: This is interesting - I feel like this could be done a better way or there should be a helper method elsewhere that already does this.

lib/api/request_adapter.rb
```diff
@@ -43,6 +43,40 @@ def attributes
       @attributes ||= @params['attributes'].to_s.split(',')
     end
 
+    # Groups attributes by association path for collection attribute requests
+    # Example: "name,vms.name,vms.cpu_count" =>
+    #   { direct: [:name], associations: { "vms" => [:name, :cpu_count] } }
```
Fryguy: I presume this is only 1 level deep? (I'm fine with that - I'm assuming this will be expanded to do N-levels deep later).

Fryguy: Oh duh - it's in the comment right below - comment might be better up here instead.

app/controllers/api/base_controller/renderer.rb
jrafanie:
```ruby
      def collection_association?(obj)
        obj.respond_to?(:loaded?)
      end
```

This seems more intention revealing and uses a more public API... What do you think? 

Note, an inline `obj.respond_to?(:loaded?)`: 

```ruby
if related_obj.respond_to?(:loaded?) && @req.collection_attributes_for(base).any?
```

seems worse than:

```ruby
if collection_association?(related_obj) && @req.collection_attributes_for(base).any?
```

app/controllers/api/base_controller/renderer.rb
jrafanie: should be done, ptal

app/controllers/api/base_controller/renderer.rb
jrafanie: should be done, ptal

lib/api/request_adapter.rb
jrafanie: should be done, ptal

app/controllers/api/base_controller/renderer.rb
Fryguy: Yeah, I was more surprised that we didn't have this method already implemented. I agree on the helper method being more intention revealing.

app/controllers/api/base_controller/renderer.rb
```diff
@@ -311,16 +311,64 @@ def add_custom_action_hrefs(value)
...
-        value  = virtual_attribute_search(object_hash[base], attr)
+        related_obj = object_hash[base]
+
+        # Special handling for collection associations with sub-attributes (e.g., vms.name, vms.vendor)
+        # Returns array of resources with requested attributes instead of single attribute value
+        if collection_association?(related_obj) && @req.collection_attributes_for(base).any?
+          return fetch_collection_attributes(type, resource, base, related_obj)
+        end
+
+        # Standard virtual attribute handling for single attributes (e.g., hardware.host.name)
```
jrafanie: Note, this method should possibly be renamed as we are now fetching collection_associations or indirect virtual attributes.  The comments make this path clearer but it might make sense to make these method calls for clearer intent.

app/controllers/api/base_controller/renderer.rb
```diff
@@ -311,16 +311,64 @@ def add_custom_action_hrefs(value)
...
       end
 
+      def fetch_collection_attributes(_type, _resource, association_path, collection)
+        requested_attrs = @req.collection_attributes_for(association_path)
+        collection_type = get_collection_type(collection)
+        return [nil, {}] if requested_attrs.empty? || collection_type.nil?
+
+        # TODO: Virtual attributes not yet supported - only physical attributes (database columns)
+        # Consider: Support virtual attributes on associations (vms.v_total_snapshots) or expose
+        # aggregated virtual attributes on primary collection where they can be properly eager loaded?
```
jrafanie: This is a big one... this adds only vms.name type support, it doesn't add  virtual column/attributes on the has many association (so not, vms.v_total_snapshots)

It makes me think we might have to weigh the pros/cons of exposing virtual attributes at the primary collection side (which may not make sense), or adding support for pulling back virtual attribute/columns from these has many associations.

jrafanie: > This is a big one... this adds only vms.name type support, it doesn't add virtual column/attributes on the has many association (so not, vms.v_total_snapshots)
> 
> It makes me think we might have to weigh the pros/cons of exposing virtual attributes at the primary collection side (which may not make sense), or adding support for pulling back virtual attribute/columns from these has many associations.

in other words, I don't know that things like 

```
GET /providers?attributes=vms.v_total_snapshots...
```

would even make sense.

Or even worse, a virtual column on providers themselves:

```
GET /providers?attributes=v_total_snapshots...
```

app/controllers/api/base_controller/renderer.rb
```diff
@@ -311,16 +311,64 @@ def add_custom_action_hrefs(value)
...
+      def fetch_indirect_virtual_attribute(type, resource, base, attr, object_hash)
         query_related_objects(base, resource, object_hash)
-        return unless attr_accessible?(object_hash[base], attr)
-        value  = virtual_attribute_search(object_hash[base], attr)
+        related_obj = object_hash[base]
+
+        if collection_association?(related_obj) && @req.collection_attributes_for(base).any?
+          fetch_collection_attributes(type, resource, base, related_obj)
+        else
+          fetch_standard_virtual_attribute(related_obj, base, attr)
```
jrafanie: I split out the logic into two methods for the conditionals so hopefully it's easier to follow

app/controllers/api/base_controller/renderer.rb
```diff
@@ -606,7 +654,25 @@ def attr_base_uses_rbac?(attr_base)
       end
 
       def determine_include_for_find(klass)
-        attrs = virtual_attributes_for(klass) do |type, attr_name, attr_base|
+        attrs = determine_include_for_find_vattrs(klass)
+        collections = determine_include_for_find_collections(klass)
```
jrafanie: This method added the collections to the list of things we need to build the includes for.  Since we now have virtuals and collections doing similar things, each was added or moved to methods that return what needs to be included and then loops over them and builds the includes.

Fryguy:
I think this got conflicted by #1324

PR #1322, opened 2026-05-13
Preload nested RBAC associations to reduce N+1 queries

### LINKED ISSUE #1321: N+1 query when an RBAC eligible association is included in a primary collection's attributes
@Fryguy found an N+1 query when an RBAC eligible association is included in a primary collection's attributes:

```
GET /api/vms?expand=resources&attributes=name,host,ext_management_system,
  ems_cluster,storage,operating_system,snapshots,tags,...
```

The log showed and N + 1 pattern - one query per VM for each RBAC association:
```
  SELECT "hosts".* FROM "hosts" WHERE "hosts"."id" = $1 LIMIT $2
  SELECT "ext_management_systems".* FROM "ext_management_systems" WHERE ...
  SELECT "ems_clusters".* FROM "ems_clusters" WHERE ...
```
  (repeated for each of 108 VMs)

jrafanie:
Removed RBAC check in `determine_include_for_find` to allow eager loading
of nested RBAC associations (hardware.host.name, ext_management_system.name,
storage.name, ems_cluster.name). Previously these were skipped, causing a
query for each VM. Now they're eager-loaded and RBAC re-filters in bulk.

For example, in a specific database setup with 108 vms returned, and the
following API request:

http://localhost:3000/api/vms?expand=resources&attributes=name,hardware.host.name,ext_management_system.name,storage.name

The query count looks like this:

Before:
Completed 200 OK in 4530ms (Views: 0.3ms | ActiveRecord: 1401.9ms (2207 queries, 1152 cached) | GC: 479.2ms)

After:
Completed 200 OK in 3168ms (Views: 0.3ms | ActiveRecord: 541.5ms (787 queries, 763 cached) | GC: 461.6ms)

Reduction of queries from 2207 to 787 (64%)

Fixes #1321

### DIFF
--- a/app/controllers/api/base_controller/renderer.rb
+++ b/app/controllers/api/base_controller/renderer.rb
@@ -611,7 +611,7 @@ module Api
           if klass.virtual_includes(attr_name) && !klass.attribute_supported_by_sql?(attr_name) && attr_base.blank?
             attr_name
           # Case 2: Direct association (e.g., "snapshots", "storage", "ems_cluster")
-          # Not dot notation, check if the requested attribute is a direct association and include it
+          # Eager load associations (RBAC participating or not) to reduce N+1 queries.
           elsif attr_base.blank?
             reflection = klass.reflect_on_association(attr_name.to_sym)
             if reflection && [:has_many, :has_one, :has_and_belongs_to_many, :belongs_to].include?(reflection.macro)
@@ -620,11 +620,10 @@ module Api
               next
             end
           # Case 3: Nested attribute (e.g., "hardware.host.name")
-          # Include the base path for eager loading with specific exceptions
+          # Eager load nested associations (RBAC participating or not) to reduce N+1 queries.
+          # Exception: custom virtual_attribute_accessor (handled via accessor).
           else
             next if virtual_attribute_accessor(type, attr_name)
-            next if attr_base_uses_rbac?(attr_base)
-
             attr_base
           end
         end
--- a/spec/requests/vms_spec.rb
+++ b/spec/requests/vms_spec.rb
@@ -464,6 +464,43 @@ describe "Vms API" do
         get api_vms_url, :params => {:expand => "resources", :attributes => "host,ext_management_system,ems_cluster,storage"}
       end.to make_database_queries(:count => be <= 15, :matching => query_match)
     end
+
+    context "with restricted user" do
+      let(:tenant)         { FactoryBot.create(:tenant) }
+      let(:role)           { FactoryBot.create(:miq_user_role) }
+      let(:group)          { FactoryBot.create(:miq_group, :tenant => tenant, :miq_user_role => role) }
+      let(:other_group)    { FactoryBot.create(:miq_group) }
+      let!(:restricted_vm) { FactoryBot.create(:vm_vmware, :host => host, :ems_id => ems.id, :miq_group => other_group) }
+
+      before do
+        vm.update(:miq_group => group)
+        @user.update(:miq_groups => [group])
+        @role = role
+      end
+
+      it "applies RBAC to expanded associations" do
+        api_basic_authorize action_identifier(:vms, :read, :resource_actions, :get)
+
+        get api_vms_url, :params => {
+          :expand     => "resources",
+          :attributes => "name,host.name,ext_management_system.name"
+        }
+
+        resources = response.parsed_body.fetch("resources")
+        resource_ids = resources.map { |r| r["id"] }
+
+        expect(resource_ids).to include(vm.id.to_s)
+        expect(resource_ids).not_to include(restricted_vm.id.to_s)
+
+        allowed_resource = resources.detect { |resource| resource["id"] == vm.id.to_s }
+
+        expect(allowed_resource).to include(
+          "name"                  => vm.name,
+          "host"                  => {"name" => vm.host.name},
+          "ext_management_system" => {"name" => vm.ext_management_system.name}
+        )
+      end
+    end
   end
 
   context 'Vm edit' do
--- a/spec/requests/vms_spec.rb
+++ b/spec/requests/vms_spec.rb
@@ -300,15 +300,15 @@ describe "Vms API" do
     end
   end
 
-  # Query count expectations for eager loading:
+  # Query count expectations:
   # - has_many associations: 6 queries
-  #   * 3 pagination queries (count filtered results, get page of IDs, recount for subquery)
-  #   * 1 total count query (for collection metadata)
-  #   * 2 data queries (get distinct IDs, fetch VMs with associations eagerly loaded)
+  #   * 3 pagination queries
+  #   * 1 total count query
+  #   * 2 data queries
   # - belongs_to associations: 4 queries
-  #   * 2 pagination queries (count filtered results, get page of IDs)
-  #   * 1 total count query (for collection metadata)
-  #   * 1 data query (fetch VMs with association eagerly loaded via LEFT OUTER JOIN)
+  #   * 2 pagination queries
+  #   * 1 total count query
+  #   * 1 data query
   context "eager loads requested direct associations" do
     let!(:vm_with_associations) do
       FactoryBot.create(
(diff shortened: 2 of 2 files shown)


app/controllers/api/base_controller/renderer.rb
```diff
@@ -612,7 +612,11 @@ def determine_include_for_find(klass)
           else
             next if attr_base.blank?
             next if virtual_attribute_accessor(type, attr_name)
-            next if attr_base_uses_rbac?(attr_base)
+            # Removed RBAC check to allow preloading of RBAC-participating associations.
```
Fryguy: I realize this PR is still WIP, but I'm not a fan of these types of comments because they don't make sense to future readers and read more like a commit message. If we do need a comment here it should probably read something like

```
# NOTE: We intentionally do not do an RBAC check here as it prevents preloading of RBAC-participating
# associations leading to N+1 queries. Core RBAC may still re-filter these associations when accessed
```

This way it shows to future readers that we don't want them to include RBAC here, or if they choose to they have to take the N+1 into consideration.

Fryguy:
If there isn't already one, please also add a spec somewhere that shows that RBAC is actually being applied on these associations in the end.

app/controllers/api/base_controller/renderer.rb
jrafanie: Oh yeah, this is absolutely going away.. it's explaining the why for when I come back to it to review it.

app/controllers/api/base_controller/renderer.rb
```diff
@@ -612,7 +612,11 @@ def determine_include_for_find(klass)
           else
             next if attr_base.blank?
             next if virtual_attribute_accessor(type, attr_name)
-            next if attr_base_uses_rbac?(attr_base)
```
jrafanie, line 626: @Fryguy @agrare This section will conflict with https://github.com/ManageIQ/manageiq-api/pull/1324/ so once that's in, I can rebase and get this one in.

In the meantime, I'll update this one to use the same pattern for testing preload/eager loading via query counts as was done in that PR

jrafanie: (obviously, this comment will just go way though and I'll update the new comment that will land from the other PR to mention the specific exceptions that will remain)

jrafanie:
ok, this should be ready for re-review since #1324 is now merged and introduced some new comments explaining the `determine_include_for_find` method.

Fryguy:
is there a spec that shows RBAC post-filtering the VMs? If not, I think we need that and show that it's not incurring more database queries.

spec/requests/vms_spec.rb
```diff
@@ -383,12 +382,17 @@ def query_match_regexp(*tables)
       api_basic_authorize action_identifier(:vms, :read, :resource_actions, :get)
       query_match = query_match_regexp("vms", "hardwares", "hosts")
 
+      # After fix: 9 queries (3 base + 6 RBAC re-filtering for 6 VMs)
+      # Before fix: was 21 queries (N+1 for each host access without preloading)
```
Fryguy: This wording is weird to me - I don't like comments that talk about a fix - that can really just live in the git commit body and shown via the spec itself.



```suggestion
      # Query count expectations: 9 queries (3 base + 6 RBAC re-filtering for 6 VMs)
```

spec/requests/vms_spec.rb
```diff
@@ -383,12 +382,17 @@ def query_match_regexp(*tables)
...
+      # Note: RBAC associations are preloaded (visible in LEFT OUTER JOIN) but
+      # ManageIQ core RBAC re-filters them when accessed, causing additional queries.
+      # This is still better than the original N+1 (21 queries).
```
Fryguy: And this line should go away too.
```suggestion
```

spec/requests/vms_spec.rb
```diff
@@ -415,6 +419,60 @@ def query_match_regexp(*tables)
     end
   end
 
+  # Query count expectations for eager loading RBAC associations:
+  # - RBAC associations (host, ems, cluster, storage) are preloaded via LEFT OUTER JOINs
+  # - ManageIQ core RBAC re-filters associations when accessed, causing additional queries
+  # - Query count: ~15 queries (base queries + RBAC re-filtering)
+  # - Without preloading: Would be ~20+ queries per VM (N+1 pattern)
```
Fryguy: Again, we don't need this "before this fix" type comments - I don't find them helpful to future readers (at least not in the code - fine in the git commit message)

jrafanie: Thanks, yes.  I cleaned up some of the comments but should make this more concise too.

jrafanie:
Still working on a better rbac test...

jrafanie:
> If there isn't already one, please also add a spec somewhere that shows that RBAC is actually being applied on these associations in the end.

I added a test to verify a restricted vm is not returned and the restricted vm's host name and ems name are also not returned.

app/controllers/api/base_controller/renderer.rb
jrafanie: done

spec/requests/vms_spec.rb
```diff
@@ -415,6 +418,96 @@ def query_match_regexp(*tables)
...
+        get api_vms_url, :params => {
+          :expand     => "resources",
+          :attributes => "name,host.name,ext_management_system.name"
+        }
+
+        resources = response.parsed_body.fetch("resources")
+        resource_ids = resources.map { |r| r["id"] }
+
+        expect(resource_ids).to include(vm.id.to_s)
+        expect(resource_ids).not_to include(restricted_vm.id.to_s)
```
jrafanie, line 493: verification of only permitted vms in the response

spec/requests/vms_spec.rb
```diff
@@ -415,6 +418,96 @@ def query_match_regexp(*tables)
...
+
+        allowed_resource = resources.detect { |resource| resource["id"] == vm.id.to_s }
+
+        expect(allowed_resource).to include(
+          "name"                  => vm.name,
+          "host"                  => {"name" => vm.host.name},
+          "ext_management_system" => {"name" => vm.ext_management_system.name}
```
jrafanie, line 500: verification of only associations from permitted vms are included

spec/requests/vms_spec.rb
jrafanie: should be fixed now

spec/requests/vms_spec.rb
jrafanie: don

jrafanie: or done

spec/requests/vms_spec.rb
```diff
@@ -300,15 +300,15 @@ def query_match_regexp(*tables)
...
+  #   * 3 pagination queries
+  #   * 1 total count query
+  #   * 2 data queries
   # - belongs_to associations: 4 queries
-  #   * 2 pagination queries (count filtered results, get page of IDs)
-  #   * 1 total count query (for collection metadata)
-  #   * 1 data query (fetch VMs with association eagerly loaded via LEFT OUTER JOIN)
+  #   * 2 pagination queries
+  #   * 1 total count query
+  #   * 1 data query
```
jrafanie, line 311: Since I was cleaning up query count comments, I made these a bit less wordy

jrafanie:
ok, I think this is ready for review again @Fryguy

Fryguy (approved):

Fryguy:
Backported to `tal` in commit e2b13dd235cfc2d018172992db4a2f1f49dd634d.

```text
commit e2b13dd235cfc2d018172992db4a2f1f49dd634d
Date:   Fri May 29 14:56:19 2026 -0400

    Merge pull request #1322 from jrafanie/fix-n-plus1-for-rbac-supporting-associations-as-attributes
    
    Preload nested RBAC associations to reduce N+1 queries
    
    (cherry picked from commit 76fd9101361cdf965375a4a77320b58b3c10e29e)
```

PR #1324, opened 2026-05-18
Include direct associations in eager loading when requested

jrafanie:
When associations like snapshots, tags, storages, etc. are explicitly requested in the attributes parameter, they should be eager-loaded to prevent N+1 queries.

Previously, only associations with virtual_includes defined or nested associations (e.g., hardware.disks) were included. Direct associations were skipped, causing N+1 queries.

This change checks if a requested direct attribute is an ActiveRecord association and includes it in the eager loading if so.

For example, in a specific database setup with 1000 vms returned, and the following API request:

http://localhost:3000/api/vms?expand=resources&attributes=snapshots,tags,storages

The query count looks like this:

Before:
Completed 200 OK in 6631ms (Views: 0.3ms | ActiveRecord: 2005.1ms (3017 queries, 0 cached) | GC: 578.6ms)

After:
Completed 200 OK in 5972ms (Views: 0.3ms | ActiveRecord: 1684.1ms (1019 queries, 1 cached) | GC: 577.6ms)

It still can be an enormous payload because we're asking for N number of things with a large number of default attributes for 1000 vms, but this will ensure we eager load it because we asked for it.

### DIFF
--- a/app/controllers/api/base_controller/renderer.rb
+++ b/app/controllers/api/base_controller/renderer.rb
@@ -607,10 +607,21 @@ module Api
 
       def determine_include_for_find(klass)
         attrs = virtual_attributes_for(klass) do |type, attr_name, attr_base|
+          # Case 1: Direct virtual attribute (e.g., "ram_size")(Not association.column format (dot notation), therefore attr_base is blank)
           if klass.virtual_includes(attr_name) && !klass.attribute_supported_by_sql?(attr_name) && attr_base.blank?
             attr_name
+          # Case 2: Direct association (e.g., "snapshots", "storage", "ems_cluster")
+          # Not dot notation, check if the requested attribute is a direct association and include it
+          elsif attr_base.blank?
+            reflection = klass.reflect_on_association(attr_name.to_sym)
+            if reflection && [:has_many, :has_one, :has_and_belongs_to_many, :belongs_to].include?(reflection.macro)
+              attr_name
+            else
+              next
+            end
+          # Case 3: Nested attribute (e.g., "hardware.host.name")
+          # Include the base path for eager loading with specific exceptions
           else
-            next if attr_base.blank?
             next if virtual_attribute_accessor(type, attr_name)
             next if attr_base_uses_rbac?(attr_base)
 
--- a/spec/requests/vms_spec.rb
+++ b/spec/requests/vms_spec.rb
@@ -300,6 +300,74 @@ describe "Vms API" do
     end
   end
 
+  # Query count expectations for eager loading:
+  # - has_many associations: 6 queries
+  #   * 3 pagination queries (count filtered results, get page of IDs, recount for subquery)
+  #   * 1 total count query (for collection metadata)
+  #   * 2 data queries (get distinct IDs, fetch VMs with associations eagerly loaded)
+  # - belongs_to associations: 4 queries
+  #   * 2 pagination queries (count filtered results, get page of IDs)
+  #   * 1 total count query (for collection metadata)
+  #   * 1 data query (fetch VMs with association eagerly loaded via LEFT OUTER JOIN)
+  context "eager loads requested direct associations" do
+    let!(:vm_with_associations) do
+      FactoryBot.create(
+        :vm_vmware,
+        :host                  => host,
+        :ext_management_system => ems,
+        :ems_cluster           => FactoryBot.create(:ems_cluster, :ext_management_system => ems),
+        :storage               => FactoryBot.create(:storage),
+        :tags                  => [FactoryBot.create(:tag)],
+        :snapshots             => [FactoryBot.create(:snapshot)]
+      )
+    end
+
+    it "snapshots" do
+      api_basic_authorize action_identifier(:vms, :read, :resource_actions, :get)
+      query_match = query_match_regexp("vms", "snapshots")
+
+      expect do
+        get api_vms_url, :params => {:expand => "resources", :attributes => "snapshots", :filter => ["id=#{vm_with_associations.id}"]}
+      end.to make_database_queries(:count => 6, :matching => query_match) # 6 queries for has_many (see comment above)
+    end
+
+    it "storage" do
+      api_basic_authorize action_identifier(:vms, :read, :resource_actions, :get)
+      query_match = query_match_regexp("vms", "storages")
+
+      expect do
+        get api_vms_url, :params => {:expand => "resources", :attributes => "storage", :filter => ["id=#{vm_with_associations.id}"]}
+      end.to make_database_queries(:count => 4, :matching => query_match) # 4 queries for belongs_to (see comment above)
+    end
+
+    it "tags" do
+      api_basic_authorize action_identifier(:vms, :read, :resource_actions, :get)
+      query_match = query_match_regexp("vms", "taggings", "tags")
+
+      expect do
+        get api_vms_url, :params => {:expand => "resources", :attributes => "tags", :filter => ["id=#{vm_with_associations.id}"]}
+      end.to make_database_queries(:count => 6, :matching => query_match) # 6 queries for has_many (see comment above)
+    end
+
+    it "ems_cluster" do
+      api_basic_authorize action_identifier(:vms, :read, :resource_actions, :get)
+      query_match = query_match_regexp("vms", "ems_clusters")
+
+      expect do
+        get api_vms_url, :params => {:expand => "resources", :attributes => "ems_cluster", :filter => ["id=#{vm_with_associations.id}"]}
+      end.to make_database_queries(:count => 4, :matching => query_match) # 4 queries for belongs_to (see comment above)
+    end
+
+    it "multiple associations" do
+      api_basic_authorize action_identifier(:vms, :read, :resource_actions, :get)
+      query_match = query_match_regexp("vms", "snapshots", "storages", "taggings", "tags")
+
+      expect do
+        get api_vms_url, :params => {:expand => "resources", :attributes => "snapshots,storage,tags", :filter => ["id=#{vm_with_associations.id}"]}
+      end.to make_database_queries(:count => 6, :matching => query_match) # 6 queries when mixing has_many and belongs_to (see comment above)
+    end
+  end
+
   context "Vm index with nested indirect virtual attribute that participates in Rbac ('hardware.host.name')" do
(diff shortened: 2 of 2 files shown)


app/controllers/api/base_controller/renderer.rb
```diff
@@ -609,8 +609,15 @@ def determine_include_for_find(klass)
         attrs = virtual_attributes_for(klass) do |type, attr_name, attr_base|
           if klass.virtual_includes(attr_name) && !klass.attribute_supported_by_sql?(attr_name) && attr_base.blank?
             attr_name
+          elsif attr_base.blank?
+            # Include direct associations when explicitly requested in attributes
+            reflection = klass.reflect_on_association(attr_name.to_sym)
+            if reflection && [:has_many, :has_one, :has_and_belongs_to_many, :belongs_to].include?(reflection.macro)
+              attr_name
```
jrafanie, line 618: for "hardware.cpu_sockets", attr_base is cpu_sockets.

For "snapshosts", attr_base is "".  This check is really "is this a non-dot notation association"... or a direct association..., and if so, snag this association name as we need to eager load it.

If desirable, I can make a method out of that logic.

spec/requests/vms_spec.rb
```diff
@@ -298,6 +319,95 @@ def query_match_regexp(*tables)
...
     end
+
+  end
+
+  context "eager loads requested direct associations" do
+    let(:vm_with_associations) { create_vm_with_associations }
+
+    before do
+      vm_with_associations
+    end
```
agrare: Could this be

```suggestion
    let!(:vm_with_associations) { create_vm_with_associations }
```

spec/requests/vms_spec.rb
```diff
@@ -42,6 +42,27 @@ def query_match_regexp(*tables)
...
+    queries.count
+  end
+
+  def create_vm_with_associations
+    vm_record = FactoryBot.create(:vm_vmware, :host => host, :ems_id => ems.id)
+    FactoryBot.create(:snapshot, :vm_or_template => vm_record)
+    vm_record.update(:storage => FactoryBot.create(:storage))
+    vm_record.tags << FactoryBot.create(:tag)
+    vm_record.update(:ems_cluster => FactoryBot.create(:ems_cluster))
+    vm_record
```
agrare: I think most of these can be done in the initial `FactoryBot.create` call and avoid extra sql updates

```suggestion
    FactoryBot.create(:vm_vmware, :host => host, :ems_id => ems.id, :storage => FactoryBot.create(:storage), :tags => [FactoryBot.create(:tag)], :ems_cluster => FactoryBot.create(:ems_cluster)).tap do |vm_record|
      FactoryBot.create(:snapshot, :vm_or_template => vm_record)
    end
```

agrare: Actually snapshots will work this way too:

```suggestion
    FactoryBot.create(
      :vm_vmware,
      :host => host,
      :ext_management_system => ems,
      :ems_cluster => FactoryBot.create(:ems_cluster, :ext_management_system => ems),
      :storage => FactoryBot.create(:storage, :ext_management_system => ems),
      :tags => [FactoryBot.create(:tag)],
      :snapshots => [FactoryBot.create(:snapshot)])
```

app/controllers/api/base_controller/renderer.rb
agrare: > for "hardware.cpu_sockets", attr_base is cpu_sockets.

That's actually surprising, based on the names I would have expected `attr_base` to be `hardware` and `attr_name` to be `cpu_sockets`

spec/requests/vms_spec.rb
```diff
@@ -298,6 +319,95 @@ def query_match_regexp(*tables)
...
+  end
+
+  context "eager loads requested direct associations" do
+    let(:vm_with_associations) { create_vm_with_associations }
+
+    before do
+      vm_with_associations
+    end
+
+    it "snapshots" do
```
jrafanie, line 325: I'm debating dropping some of these extra tests... but considering this was broken, we should probably have a regression test to ensure this doesn't break again.

spec/requests/vms_spec.rb
```diff
@@ -42,6 +42,27 @@ def query_match_regexp(*tables)
...
+    queries = []
+    subscriber = ActiveSupport::Notifications.subscribe("sql.active_record") do |_name, _start, _finish, _id, payload|
+      queries << payload[:sql] if payload[:sql] =~ pattern
+    end
+
+    yield
+
+    ActiveSupport::Notifications.unsubscribe(subscriber)
+    queries.count
+  end
```
Fryguy: We have the db_query_matchers gem which I think gives you the ability to do this so I don't think you need a custom method.

spec/requests/vms_spec.rb
```diff
@@ -298,6 +319,95 @@ def query_match_regexp(*tables)
...
+        get api_vms_url, :params => {:expand => "resources", :attributes => "snapshots", :filter => ["id=#{vm_with_associations.id}"]}
+      end
+
+      vm2 = create_vm_with_associations
+
+      query_count_after = count_queries_matching(query_match_regexp("vms", "snapshots")) do
+        get api_vms_url, :params => {:expand => "resources", :attributes => "snapshots", :filter => ["id=#{vm_with_associations.id},#{vm2.id}"]}
+      end
+
+      expect(query_count_after).to eq(query_count_before)
```
Fryguy: re: db_query_matchers, you can either use `make_database_queries(:count => X, :matching => pattern)`, or there is a query count object in there (works similar to this method).

Also if we know the number of queries, we should declare that up front. I think it's better to hardcode into the test the number of expected queries than a before/after match (because if one reintroduces the N+1 in both places, then the test still passes, which we don't want)

app/controllers/api/base_controller/renderer.rb
jrafanie:
```ruby
      def split_virtual_attribute(attr)
        attr_parts = attr_split(attr)
        return [attr_parts.first, ""] if attr_parts.length == 1
        [attr_parts.last, attr_parts[0..-2].join(".")]
      end
```

all of the callers look like:

```ruby
attr_name, attr_base = split_virtual_attribute(vattr)
```

spec/requests/vms_spec.rb
jrafanie: Thanks, yeah, I'll see how easy it is to pull that in... Part of it was to discuss if others felt these tests are worthwhile doing...  I think so... if you agree, we should try reusing existing mechanisms to do the query counting.

spec/requests/vms_spec.rb
jrafanie: ah interesting, the random 3, 4, 5 numbers I had previously seemed pretty hard to reason about, but the problem you bring up might be worse than something being not easy to understand quickly.

app/controllers/api/base_controller/renderer.rb
agrare: yeah not important to this PR I just would have expected that to be the inverse

spec/requests/vms_spec.rb
Fryguy: Yes, I think the tests are worthwhile to show that we are properly eager loading associations.  Might make more sense as core renderer specs as opposed to in vms_spec, but vms is one of our biggest and most important endpoints, so there's that.

spec/requests/vms_spec.rb
Fryguy: Some where else, I think keenan wrote comments like `3 queries (one for the base, one for snapshots, one for whatevers)`, just to give a rough idea where the magic comes from

app/controllers/api/base_controller/renderer.rb
jrafanie: Yup, tell me about it

spec/requests/vms_spec.rb
jrafanie: Yes, I like that much better especially if we're only creating what we need to create once and verify the hardcoded query count, see below, it's much simpler to do this once and drop the before entirely.

app/controllers/api/base_controller/renderer.rb
jrafanie: > for "hardware.cpu_sockets", attr_base is cpu_sockets.

Sorry, I misspoke because you're right, that doesn't look right.

Indeed, for hardware.cpu_sockets, attr_base is hardware, and attr_name is cpu_sockets.

app/controllers/api/base_controller/renderer.rb
```diff
@@ -607,10 +607,21 @@ def attr_base_uses_rbac?(attr_base)
...
+          # Not dot notation, check if the requested attribute is a direct association and include it
+          elsif attr_base.blank?
+            reflection = klass.reflect_on_association(attr_name.to_sym)
+            if reflection && [:has_many, :has_one, :has_and_belongs_to_many, :belongs_to].include?(reflection.macro)
+              attr_name
+            else
+              next
+            end
+          # Case 3: Nested attribute (e.g., "hardware.host.name")
+          # Include the base path for eager loading with specific exceptions
```
jrafanie, line 623: I added some comments to hopefully help navigate what this method is doing...

jrafanie:
ok, @agrare @Fryguy, I think I addressed your helpful comments. Based on the feedback, I also added some explanation to the code to better help understand the code change and the context around it.

app/controllers/api/base_controller/renderer.rb
agrare: These are very helpful 👍

agrare (approved):

app/controllers/api/base_controller/renderer.rb
jrafanie: yeah, my confusion and your question made me add it for future me and you 😉

Fryguy (approved):

Fryguy:
Backported to `tal` in commit ca20ec7af9eb603bc389eea96ae251f01ce8e6f7.

```text
commit ca20ec7af9eb603bc389eea96ae251f01ce8e6f7
Date:   Tue May 19 11:53:35 2026 -0400

    Merge pull request #1324 from jrafanie/fix-direct-association-includes
    
    Include direct associations in eager loading when requested
    
    (cherry picked from commit aab2492b841fc1e4781f195040eab550dca3efcd)
```

PR #1327, opened 2026-06-18
Add Configured Systems assignment prefix for tags

ohhAndy:
This PR adds configured systems to the hard coded Assignment Prefix array, to allow tagging for configured systems. 

### DIFF
--- a/app/controllers/api/mixins/chargeback_assignment.rb
+++ b/app/controllers/api/mixins/chargeback_assignment.rb
@@ -11,7 +11,7 @@ module Api
       }.freeze
 
       ALLOWED_TAG_PREFIXES = {
-        'Compute' => %w[vm container_image],
+        'Compute' => %w[vm container_image configured_system],
         'Storage' => %w[storage]
       }.freeze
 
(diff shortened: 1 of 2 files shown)


app/controllers/api/mixins/chargeback_assignment.rb
```diff
@@ -11,7 +11,7 @@ module ChargebackAssignment
       }.freeze
 
       ALLOWED_TAG_PREFIXES = {
-        'Compute' => %w[vm container_image],
+        'Compute' => %w[vm container_image configured_system],
```
jrafanie, line 14: @agrare this seems really simple... do you know if there was a reason this wasn't added for configured system?   Was it just an oversight?

agrare: Yeah I didn't know this existed, so I'm guessing when people added configured systems they didn't know it existed either

jrafanie (approved):
LGTM.  I don't see any problem with adding this. My only hesitation is why it was excluded and perhaps it was just an oversight.

PR #1329, opened 2026-06-22
Fix the N+1 problem in chargeback assignments controller

ohhAndy:
## Summary

Fixes severe performance issue in the `GET /api/chargebacks?expand=resources&attributes=assigned_to` endpoint by implementing eager loading of nested associations. Response time reduced from 2-10+ seconds to under 500ms.

## Problem

The chargebacks API endpoint with `attributes=assigned_to` was experiencing:

- Response time: 2-10+ seconds for a 1.2KB response
- Query count: 40+ database queries for just 3 chargeback rates

## Root Cause

N+1 query problem in the `assigments_to_result()` method. When processing assignments, the code accessed:

- `tag.classification.name`
- `tag.classification.description`
- `tag.category.name`
- `label.resource_id`

Each access triggered individual database queries for every assignment, resulting in:

- 1 query for chargeback rates
- N queries for tags
- N queries for classifications
- N queries for categories
- N queries for label resources

## Solution

Added eager loading using `MiqPreloader.preload()` in the `assigments_to_result()` method:

1. **Added `preload_tag_associations()`** - Batch loads all tag classifications and categories before processing
2. **Added `preload_label_associations()`** - Batch loads all label resources (ContainerImages) before processing
3. **Updated `assigments_to_result()`** - Calls both preload methods before mapping assignments

This reduces 40+ queries to ~25 queries by loading all nested associations in 2-3 batch queries instead of N individual queries.

### DIFF
--- a/app/controllers/api/mixins/chargeback_assignment.rb
+++ b/app/controllers/api/mixins/chargeback_assignment.rb
@@ -235,9 +235,29 @@ module Api
       def assigments_to_result(compute_assignments, assignment_keys = [:cb_rate])
         return [] if compute_assignments.empty?
 
+        # Preload tag associations to avoid N+1 queries
+        preload_tag_associations(compute_assignments)
+        preload_label_associations(compute_assignments)
+
         key = (compute_assignments.first.keys - assignment_keys).first
 
         compute_assignments.map { |x| result_assignment(x, key, assignment_keys == [:cb_rate]) }
+      end
+
+      def preload_tag_associations(assignments)
+        # Extract all tags from assignments
+        tags = assignments.filter_map { |a| a[:tag]&.first&.tag }
+        return if tags.empty?
+
+        MiqPreloader.preload(tags, [:classification, :category])
+      end
+
+      def preload_label_associations(assignments)
+        # Extract all labels (CustomAttributes) from assignments
+        labels = assignments.filter_map { |a| a[:label]&.first }
+        return if labels.empty?
+
+        MiqPreloader.preload(labels, :resource)
       end
 
       def fetch_rates_from_params(params_assignments)
--- a/spec/requests/chargebacks_spec.rb
+++ b/spec/requests/chargebacks_spec.rb
@@ -6,6 +6,10 @@ RSpec.describe "chargebacks API" do
     FactoryBot.create(:chargeback_tier, :chargeback_rate_detail_id => chargeback_rate_detail.id, :start => 0, :finish => Float::INFINITY, :fixed_rate => 0.0, :variable_rate => 0.0)
   end
 
+  def query_match_regexp(*tables)
+    /SELECT.*FROM\s"(?:#{tables.flatten.join("|")})"/m
+  end
+
   def convert_to_response_hash(resource)
     resource.attributes.map do |attribute_name, attribute_value|
       attribute_value = attribute_value.to_s if attribute_value.present? && (attribute_name.include?("id") || attribute_value == Float::INFINITY)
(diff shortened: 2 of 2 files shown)


Fryguy:
@jrafanie Please review.
@ohhAndy Please fix the rubocop issues as per the bot comment.

app/controllers/api/mixins/chargeback_assignment.rb
```diff
@@ -235,11 +235,33 @@ def result_assignment(record, key, with_rate)
...
 
         compute_assignments.map { |x| result_assignment(x, key, assignment_keys == [:cb_rate]) }
       end
 
+      def preload_tag_associations(assignments)
+        # Extract all tags from assignments
+        tags = assignments.map { |a| a[:tag]&.first&.tag }.compact
+        return if tags.empty?
+
+        # Use MiqPreloader to preload associations on already-loaded objects (consistent with NotificationsController)
```
Fryguy: This comment is not needed as it's redundant (MiqPreloader's purpose is to preload associations on already-loaded objects, so no need to say it), and mentioning NotificationsController is not necessary.

app/controllers/api/mixins/chargeback_assignment.rb
```diff
@@ -235,11 +235,33 @@ def result_assignment(record, key, with_rate)
...
+        MiqPreloader.preload(tags, [:classification, :category])
+      end
+
+      def preload_label_associations(assignments)
+        # Extract all labels (CustomAttributes) from assignments
+        labels = assignments.map { |a| a[:label]&.first }.compact
+        return if labels.empty?
+
+        # Use MiqPreloader to preload associations on already-loaded objects (consistent with NotificationsController)
```
Fryguy: Ditto

spec/requests/chargebacks_spec.rb
```diff
@@ -90,6 +90,36 @@ def convert_to_response_hash(resource)
...
+      ChargebackRate.set_assignments(:compute, [temp])
+    end
+
+    api_basic_authorize collection_action_identifier(:chargebacks, :read, :get)
+
+    # Count queries - should use eager loading to avoid N+1
+    query_count = 0
+    query_counter = lambda { |_name, _started, _finished, _unique_id, payload|
+      query_count += 1 unless payload[:name] == "SCHEMA" || payload[:sql] =~ /^(BEGIN|COMMIT|SAVEPOINT|RELEASE)/
+    }
```
Fryguy: There is a query counting helper, so that should be used instead of this.

Fryguy: @ohhAndy  Here's an example from another test (coincidentally checking N+1 queries), that uses the count helpers: https://github.com/ManageIQ/manageiq-api/blob/e7ef45021177b701f1e76f9be5b93ea75debfc9e/spec/requests/vms_spec.rb#L112

ohhAndy: thank you, Ill adjust it now

spec/requests/chargebacks_spec.rb
```diff
@@ -90,6 +94,30 @@ def convert_to_response_hash(resource)
...
+    api_basic_authorize collection_action_identifier(:chargebacks, :read, :get)
+
+    query_match = query_match_regexp("chargeback_rates", "chargeback_rate_detail_currencies", "chargeback_rate_details", "chargeable_fields", "chargeback_rate_detail_measures", "chargeback_tiers", "chargeback_rate_detail_measure_units", "custom_attributes", "classifications", "tags")
+
+    # With eager loading, query count should be significantly reduced
+    # Without eager loading, it would be 40+ queries (base queries + 3 rates * N associations each)
+    # With eager loading, we expect around 25-30 queries (base system queries + optimized association loading)
+    expect do
+      get api_chargebacks_url, :params => {:expand => 'resources', :attributes => 'assigned_to'}
+    end.to make_database_queries(:count => be < 35, :matching => query_match)
```
Fryguy, line 115: From a test perspective, I think we should know exactly how many queries this will be. Might be better to be exact? @jrafanie Thoughts?

spec/requests/chargebacks_spec.rb
```diff
@@ -6,6 +6,10 @@
     FactoryBot.create(:chargeback_tier, :chargeback_rate_detail_id => chargeback_rate_detail.id, :start => 0, :finish => Float::INFINITY, :fixed_rate => 0.0, :variable_rate => 0.0)
   end
 
+  def query_match_regexp(*tables)
+    /SELECT.*FROM\s"(?:#{tables.flatten.join("|")})"/m
```
jrafanie, line 10: Note, with this, we now have it in 3 places... I wonder if we should have this in a helper for the test.

https://github.com/ManageIQ/manageiq-api/blob/e7ef45021177b701f1e76f9be5b93ea75debfc9e/spec/requests/notifications_spec.rb#L9-L11

https://github.com/ManageIQ/manageiq-api/blob/e7ef45021177b701f1e76f9be5b93ea75debfc9e/spec/requests/vms_spec.rb#L41-L43

jrafanie: Note, we can do this later, but wanted to mention it.

spec/requests/chargebacks_spec.rb
jrafanie, line 115: Note, we might want to see if we can create less things so this number is more explainable.  Maybe 1 or 2 chargeback rates instead of 3.  If there is some setup in the before that we don't need, we can move this group of tests to a separate context with more minimal setup in the before.

I say this because I have no idea why there's 35 even if we preload the associations now.

spec/requests/chargebacks_spec.rb
Fryguy: Probably spec/support helper

PR #1331, opened 2026-06-25
[WIP] Expose request_logs as a read-only subcollection of all request types

jrafanie:
Request logs were already being tracked per-request in the database but were
not exposed through the API. This changes makes them accessible without
introducing new RBAC complexity - if a user can view the request, they can
view its logs.

The subcollection reuses the miq_request_show identifier so no new product
features need to be defined or granted. The same pattern is applied across
all four request collections (requests, service_requests, provision_requests,
and automation_requests) for consistency.

### DIFF
--- a/config/api.yml
+++ b/config/api.yml
@@ -345,6 +345,7 @@
     :klass: AutomationRequest
     :subcollections:
     - :request_tasks
+    - :request_logs
     :collection_actions:
       :get:
       - :name: read
@@ -2989,6 +2990,7 @@
     :klass: MiqProvisionRequest
     :subcollections:
     - :request_tasks
+    - :request_logs
     :collection_actions:
       :get:
       - :name: read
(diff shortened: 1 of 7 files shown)


config/api.yml
```diff
@@ -345,6 +345,7 @@
     :klass: AutomationRequest
     :subcollections:
     - :request_tasks
+    - :request_logs
```
Fryguy, line 348: Alphabetical.

🏃💨

jrafanie: OH, we added @Fryguy as a test


```
  1) API configuration (config/api.yml) collections is sorted a-z
     Failure/Error: expect(actual).to eq(expected)

       expected: [:accounts, :actions, :alert_actions, :alert_definition_profiles, :alert_definitions, :alerts, :auth,...mplates, :tenant_groups, :tenants, :time_profiles, :users, :vms, :volume_mappings, :widgets, :zones]
            got: [:accounts, :actions, :alert_actions, :alert_definition_profiles, :alert_definitions, :alerts, :auth,...mplates, :tenant_groups, :tenants, :time_profiles, :users, :vms, :volume_mappings, :widgets, :zones]

       (compared using ==)

       Diff:

       @@ -106,8 +106,8 @@
         :rates,
         :regions,
         :reports,
       - :request_logs,
         :request_tasks,
       + :request_logs,
         :requests,
         :resource_actions,
         :resource_pools,
     # ./spec/lib/api/api_config_spec.rb:10:in `block (3 levels) in <top (required)>'
```

config/api.yml
```diff
@@ -3275,7 +3277,20 @@
...
+  :request_logs:
+    :description: Request Logs
+    :options:
+    - :subcollection
+    :verbs: *g
+    :klass: RequestLog
+    :subcollection_actions:
+      :get:
+      - :name: read
+        :identifier: miq_request_show
```
jrafanie, line 3289: I need to double check what the classic UI and (if the service UI has it, which it might not) uses in order to determine who can see request logs... the current logic is to show request logs for people who can see the request.

PR #1332, opened 2026-06-26
[WIP] Add export action to service_dialogs API

jrafanie:
SUPER WIP... I just looked to see if we have service dialog import/export via API and apparently import exists so I wanted to see what it would require to do export.  If needed, we can use this mechanism.

Exposes POST /api/service_dialogs/:id with action=export and bulk POST /api/service_dialogs with action=export and resources=[...].

Uses DialogSerializer to produce the same hash structure that the existing import (create_resource) already accepts, making export/import round-trippable through the API without transformation.

The export action reuses the miq_ae_customization_explorer RBAC identifier, consistent with read and copy.

### DIFF
--- a/config/api.yml
+++ b/config/api.yml
@@ -3741,6 +3741,8 @@
         :identifier: dialog_edit_editor
       - :name: copy
         :identifier: dialog_copy_editor
+      - :name: export
+        :identifier: miq_ae_customization_explorer
       - :name: template_service_dialog
         :identifier: dialog_new_editor
     :resource_actions:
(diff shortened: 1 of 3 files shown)


config/api.yml
```diff
@@ -3741,6 +3741,8 @@
         :identifier: dialog_edit_editor
       - :name: copy
         :identifier: dialog_copy_editor
+      - :name: export
+        :identifier: miq_ae_customization_explorer
```
jrafanie, line 3745: note, this is the same we use in the UI in miq_ae_customization_controller.rb

```ruby
def export_service_dialogs
  assert_privileges('miq_ae_customization_explorer') # feature like miq_ae_customization_service_dialog_import_export is missing
...
```

Fryguy: Interesting - wondering if now is the time to add the feature?

PR #1334, opened 2026-07-17
Fix security headers spec broken by Rack 3 lowercase header requirement caused by secure_headers upgrade

jrafanie:
We updated secure_headers from 3.x to 7.x:
https://github.com/ManageIQ/manageiq/pull/23912

Rack 3 requires all response header names to be lowercase. secure_headers complied by lowercasing its HEADER_NAME constants (x-frame-options, content-security-policy, etc.) so it no longer conflicts with Rack::Lint.

The old assertions compared title-case keys (X-Frame-Options) against the actual headers hash and always missed, producing a wall of diff noise.

Switch to individual per-header eq assertions using response.headers[] which does case-insensitive lookup via Rack's HeaderHash, so the test works regardless of the casing used by secure_headers or Rails. Also lowercase the key in the content_security_policy_for helper to match.

Related to:
https://www.github.com/github/secure_headers/pull/533
https://www.github.com/github/secure_headers/pull/551

### DIFF
--- a/spec/requests/headers_spec.rb
+++ b/spec/requests/headers_spec.rb
@@ -71,14 +71,11 @@ RSpec.describe "Headers" do
 
       get(api_entrypoint_url)
 
-      expected = {
-        "X-Content-Type-Options"            => "nosniff",
-        "X-Download-Options"                => "noopen",
-        "X-Frame-Options"                   => "SAMEORIGIN",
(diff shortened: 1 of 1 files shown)


jrafanie:
This should make master 💚 after https://github.com/ManageIQ/manageiq/pull/23912 was merged

PR #1337, opened 2026-07-27
Add YAML response support for settings subcollection

ohhAndy:
Add YAML response for settings subcollection, to allow ui-classic to pull settings hash for advanced settings editing. 

@miq-bot add-reviewer @Fryguy
@miq-bot assign @Fryguy

### DIFF
--- a/app/controllers/api/base_controller.rb
+++ b/app/controllers/api/base_controller.rb
@@ -136,8 +136,13 @@ module Api
 
     def validate_response_format
       accept = request.headers["Accept"]
-      return if accept.blank? || accept.include?("json") || accept.include?("*/*")
+      return if accept.blank? || accept.include?("json") || accept.include?("*/*") || accept.include?("yaml")
+
       raise UnsupportedMediaTypeError, "Invalid Response Format #{accept} requested"
+    end
+
+    def yaml_request?
+      request.headers["Accept"].to_s.include?("yaml")
     end
 
     def set_access_control_headers
--- a/app/controllers/api/subcollections/settings.rb
+++ b/app/controllers/api/subcollections/settings.rb
@@ -21,7 +21,10 @@ module Api
           return
         end
 
-        render_resource :settings, resource_settings(resource)
+        settings = resource_settings(resource)
+        return render(:plain => settings.to_yaml, :content_type => "application/yaml") if yaml_request?
+
+        render_resource(:settings, settings)
       end
 
       def settings_query_resource(resource)
--- a/spec/requests/servers_spec.rb
+++ b/spec/requests/servers_spec.rb
@@ -172,6 +172,15 @@ RSpec.describe "Servers" do
     let(:original_timeout) { server.settings_for_resource[:api][:authentication_timeout] }
     let(:super_admin) { FactoryBot.create(:user, :role => 'super_administrator', :userid => 'alice', :password => 'alicepassword') }
 
+    it "returns settings as YAML when Accept: application/yaml is requested" do
+      api_basic_authorize(:ops_settings)
+
+      get(api_server_settings_url(nil, server), :headers => {"Accept" => "application/yaml"})
+
+      expect(response).to have_http_status(:ok)
+      expect(response.content_type).to include("application/yaml")
+      expect { YAML.safe_load(response.body) }.not_to raise_error
+    end
 
     it "shows the settings to an authenticated user with the proper role" do
       api_basic_authorize(:ops_settings)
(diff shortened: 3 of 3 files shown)


app/controllers/api/base_controller.rb
```diff
@@ -136,10 +136,15 @@ def set_gettext_locale
...
     def validate_response_format
       accept = request.headers["Accept"]
-      return if accept.blank? || accept.include?("json") || accept.include?("*/*")
+      return if accept.blank? || accept.include?("json") || accept.include?("*/*") || accept.include?("yaml")
+
       raise UnsupportedMediaTypeError, "Invalid Response Format #{accept} requested"
     end
 
+    def yaml_request?
+      request.headers["Accept"].to_s.include?("yaml")
```
Fryguy, line 145: I believe this is handled through `request.format` or `request.formats` - Rails takes care of parsing the Accept header and setting the request format.

app/controllers/api/base_controller.rb
```diff
@@ -136,10 +136,15 @@ def set_gettext_locale
 
     def validate_response_format
       accept = request.headers["Accept"]
-      return if accept.blank? || accept.include?("json") || accept.include?("*/*")
+      return if accept.blank? || accept.include?("json") || accept.include?("*/*") || accept.include?("yaml")
```
Fryguy, line 139: Strange - this should probably be changed to also look at request.format / request.formats for both the json and yaml legs.

ohhAndy: I think this the current approach is intended. request.format doesn't work since MimeTypes doesn't register application/yaml. It currently only registers application/x-yaml for some reason. A workaround would probably cause more problems then solve them.

spec/requests/servers_spec.rb
```diff
@@ -172,6 +172,15 @@
...
     let(:super_admin) { FactoryBot.create(:user, :role => 'super_administrator', :userid => 'alice', :password => 'alicepassword') }
 
+    it "returns settings as YAML when Accept: application/yaml is requested" do
+      api_basic_authorize(:ops_settings)
+
+      get(api_server_settings_url(nil, server), :headers => {"Accept" => "application/yaml"})
+
+      expect(response).to have_http_status(:ok)
+      expect(response.content_type).to include("application/yaml")
+      expect { YAML.safe_load(response.body) }.not_to raise_error
```
Fryguy, line 182: Interestingly, this test is not sufficient. JSON is a subset of YAML, so you can pass almost any JSON string to YAML.safe_load and it will always work.

```ruby
> json = {:a => 1, :b => [1, 2]}.to_json
=> "{\"a\":1,\"b\":[1,2]}"
> YAML.safe_load(json)
=> {"a"=>1, "b"=>[1, 2]}
```

So, if the code is wrong and ends up returning JSON, this test will still pass.

I think adding a second test that shows it's _not_ valid JSON would be helpful.


```suggestion
      expect { YAML.safe_load(response.body) }.not_to raise_error
      expect { JSON.parse(response.body) }.to raise_error(JSON::ParserError)
```

app/controllers/api/base_controller.rb
Fryguy: Oh that's interesting - ok then we can leave it.

app/controllers/api/subcollections/settings.rb
```diff
@@ -21,7 +21,10 @@ def settings
           return
         end
 
-        render_resource :settings, resource_settings(resource)
+        settings = resource_settings(resource)
+        return render(:plain => settings.to_yaml, :content_type => "application/yaml") if yaml_request?
```
Fryguy, line 25: Does bypassing render_resource lose anything for us here? I think we are losing the outer REST structure if we do this? It's hard to tell, because I thought this particular subcollection worked differently than others.

Fryguy:
Do we accept the YAML on the POST side as well? Or is that for a followup PR?

app/controllers/api/subcollections/settings.rb
ohhAndy: What do you mean by the outer REST structure?

ohhAndy:
> Do we accept the YAML on the POST side as well? Or is that for a followup PR?

I dont think I explicitly added it for POST, but it works with PATCH which is all I need for the advanced settings.

app/controllers/api/subcollections/settings.rb
Fryguy: like paging and a top-level "resource" key - I can't remember what settings looks like when it renders.

PR #1340, opened 2026-08-12
Clear filter_expression in group API when null is passed

ohhAndy:
The Issue:
When updating a group entitlement from an expression back to tag filters, sending "filter_expression": null was ignored by the API because if filter_expression evaluated to false. The existing filter_expression remained in the database, colliding with the new tag filters and causing a validation error.

The Fix:
The API now checks if filter_expression key was present in the request (key?("filter_expression")) and explicitly sets filter_expression: nil on the entitlement record, allowing expressions to be cleared when switching to tag filters.

@miq-bot add-label bug
@miq-bot assign @jrafanie


app/controllers/api/groups_controller.rb
```diff
@@ -59,13 +59,15 @@ def parse_set_tenant(data)
...
+      filters_present           = data.key?("filters")
+      filter_expression_present = data.key?("filter_expression")
+      filters                   = data.delete("filters")
+      filter_expression         = data.delete("filter_expression")
+      filter_expression         = filter_expression["exp"] if filter_expression && filter_expression["exp"]
+      if filters_present || filter_expression_present
         entitlements = {"id" => entitlement_id}
         entitlements["filters"]           = filters.stringify_keys.to_h          if filters
-        entitlements["filter_expression"] = MiqExpression.new(filter_expression) if filter_expression
+        entitlements["filter_expression"] = filter_expression ? MiqExpression.new(filter_expression) : nil
```
jrafanie: Can you have filters, filter expression, or both?  It's not obvious to me if they're supposed to be mutually exclusive or not.

ohhAndy: the managed filters and filter expression are mutually exclusive, but belongsto filters are seperate

Fryguy: if there were mutual exclusions they should be handled at the model level, not at the API level.

ohhAndy: When I was testing for the UI, the API would return an error saying that I can't set managed filters and filter expressions together. So the API already handles the mutual exclusion. The issue is, if I originally filter by expression for tags and try switching to filter by regular tags (managed tags) it just doesn't properly save it.

jrafanie: Sure, I was more asking because of the conditionals as it's unclear without that model knowledge here.  I don't have a better suggestion.

spec/requests/groups_spec.rb
```diff
@@ -261,6 +261,45 @@
...
+      api_basic_authorize collection_action_identifier(:groups, :edit)
+
+      filters = {"managed" => [["/managed/area/1"]]}
+
+      post(api_group_url(nil, group1), :params => gen_request(:edit, "filters" => filters, "filter_expression" => nil))
+
+      expect(response).to have_http_status(:ok)
+      group1.reload
+      expect(group1.entitlement.filter_expression).to be_nil
+      expect(group1.entitlement.filters).to eq(filters)
```
jrafanie, line 380: do we have adequate test coverage for each of the supported ways to provide filters, expressions, etc.?  Do we test for the API response for more common invalid payloads such as both managed filters and filter expressions?

ohhAndy: I added more testing coverage, to cover the different cases.

spec/requests/groups_spec.rb
```diff
@@ -261,6 +316,150 @@
...
+      existing_filters = {"managed" => [["/managed/area/1"]], "belongsto" => ["/managed/infra/1"]}
+      group1.create_entitlement(:filters => existing_filters)
+      api_basic_authorize collection_action_identifier(:groups, :edit)
+
+      post(api_group_url(nil, group1), :params => gen_request(:edit, "description" => "renamed"))
+
+      expect(response).to have_http_status(:ok)
+      group1.reload
+      expect(group1.entitlement.filters).to eq(existing_filters)
+      expect(group1.entitlement.filter_expression).to be_nil
```
jrafanie: Should we create the entitlement with a filter_expression?  We created it with a filter but not a filter expression so we're not really verifying the filter_expression was left untouched.

jrafanie (approved):
LGTM

